Issue #50, opened 2025-07-05
Transmitting the parity shards when an object doesn't need to be fixed


weisd:
### Discussed in https://github.com/rustfs/rustfs/discussions/46

<div type='discussions-op-text'>

<sup>Originally posted by **zzhpro** July  4, 2025</sup>
In the implementation of `GetObject`, it seems that parity shards of an object are transmitted to the coordinator regardless of whether the object need to be fixed. This wastes network bandwidth and can harm cluster performance. Is there a reason for doing this or simply am I wrong about the behavior?</div>

PR #322, opened 2025-07-31
perf: avoid transmitting parity shards when the object is good

zzhpro:
## Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [x] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues

#46 
#50 

## Summary of Changes

When executing `GetObject`, transmitting the data shards only when the object is fine. Read the parity shards only when some of the data shards are broken.
I've added some unit tests and tested this implementation on a three-nodes RustFS setup, the object is available when one of the nodes is down or one part file is lost

## Checklist
- [x] Code is formatted with `cargo fmt --all`
- [x] Passed `cargo clippy --all-targets --all-features -- -D warnings`
- [x] Passed `cargo check --all-targets`

## Impact
- [ ] Other impact:

## Additional Notes

---

### DIFF
--- a/crates/ecstore/src/erasure_coding/decode.rs
+++ b/crates/ecstore/src/erasure_coding/decode.rs
@@ -116,35 +116,24 @@ where
                 }
             }
 
-            let mut running_readers = self.data_shards;
-            let mut next_reader_index = self.data_shards;
+            let mut success = 0;
             while let Some((i, result)) = sets.next().await {
-                running_readers -= 1;
                 match result {
                     Ok(v) => {
                         shards[i] = Some(v);
+                        success += 1;
                     }
                     Err(e) => {
                         errs[i] = Some(e);
 
-                        // The number of shards we already got
-                        let fine_shards_count = shards.iter().filter(|s| s.is_some()).count();
-                        // The number of shards we need to decode the block
-                        let need_shards = self.data_shards - fine_shards_count;
-                        // The number of shards we haven't try to read yet
-                        let untriggered_readers = self.total_shards - next_reader_index;
-
-                        if need_shards > running_readers + untriggered_readers {
-                            // We're doomed, just wait for the running readers to finish
-                            continue;
-                        }
-
                         if let Some(future) = fut_iter.next() {
                             sets.push(future);
-                            running_readers += 1;
-                            next_reader_index += 1;
                         }
                     }
+                }
+
+                if success >= self.data_shards {
+                    break;
                 }
             }
         }


weisd (approved):

shiroleeee:
You can try using `futures::stream::FuturesUnordered` to simplify concurrency control. For example, in the following case, we put all async blocks for reader data reading into a Vec, then control the concurrency level by limiting the number of pushes to FuturesUnordered. When errors occur, you can quickly push an additional async block for execution without waiting for all tasks in the current batch to complete before proceeding to the next round.

```rust
        use futures::stream::{FuturesUnordered, StreamExt};

        let mut futures = Vec::with_capacity(num_readers);
        for (i, reader) in self.readers.iter_mut().enumerate() {
            if let Some(r) = reader {
                futures.push(async move {
                    let mut buf = vec![0u8; shard_size];
                    match r.read(&mut buf).await {
                        Ok(n) => {
                            buf.truncate(n);
                            (i, Ok(buf))
                        }
                        Err(e) => (i, Err(Error::from(e))),
                    }
                });
            }
        }
        if futures.len() > self.data_shards {
            let mut sets = FuturesUnordered::new();
            let mut fns_iter = futures.into_iter();
            for _ in 0..self.data_shards {
                if let Some(f) = fns_iter.next() {
                    sets.push(f)
                }
            }

            while let Some((i, result)) = sets.next().await {
                match result {
                    Ok(v) => {}
                    Err(e) => {
                        // error!("Error reading shard {}: {}", i, e);
                        if let Some(f) = fns_iter.next() {
                            sets.push(f)
                        }
                    }
                }
            }
        }
```

zzhpro:
@shiroleeee  thanks for your advice, I'll improve this.

shiroleeee:
You've done great work! However, I think the logic for determining whether sufficient data shards are obtained could be further simplified. For example, as shown below, we can simply verify if the number of successful returns matches the total count of data shards.

```rust
            let mut success = 0;
            while let Some((i, result)) = sets.next().await {
                match result {
                    Ok(v) => {
                        shards[i] = Some(v);
                        success += 1;
                    }
                    Err(e) => {
                        errs[i] = Some(e);
                        if let Some(future) = fut_iter.next() {
                            sets.push(future);
                        }
                        continue;
                    }
                }

                if success == self.data_shards {
                    break;
                }
            }
```

zzhpro:
@shiroleeee  You're right. Your implementation is actually equivalent to mine while being simpler. I've applied the changes in the new commit.

shiroleeee (approved):

Issue #581, opened 2025-09-24
It may be a bug in multi-part upload


loverustfs:
Here is a summary of the four bugs related to the S3 SDK multipart update:

https://github.com/rustfs/rustfs/issues/580
https://github.com/rustfs/rustfs/issues/541
https://github.com/rustfs/rustfs/issues/540
https://github.com/rustfs/rustfs/issues/464

Issue #580: Users are encountering an "InvalidRequest: invalid content type" error when uploading files via the AWS S3 API. The issue seems to be specific to RustFS, as uploads work with AWS Cloud Storage and Cloudflare R2. This is a recurring bug.

Issue #541: A user is unable to upload files using rustfs-macos-aarch64-v1.0.0-alpha.56.zip, receiving an "InvalidPart" error. The problem is resolved by downgrading to a previous version (1.0.0-alpha.50).

Issue #540: An error occurs when uploading .ipa files to RustFS storage using the AWS S3 standard protocol. The user receives "NotFound" (404) and "InvalidPart" errors, as well as a 501 "Not Implemented" error for PutBucketCors.

Issue #464: A user is experiencing an "InvalidRequest: missing required multipart fields" error when using form-data for POST requests, even though the same request works with Minio. Another user reported a similar issue with uploading .ipa files and getting an "invalid content type" error, suggesting a problem with the application/octet-stream MIME type.

Nugine:
Upstream project `s3s` has received this bug report. For `rustfs` members, you can try to reproduce the bugs with `aws-sdk-rust` in [s3s-e2e](https://github.com/Nugine/s3s/tree/main/crates/s3s-e2e). So we can do regression testing continuously.

loverustfs:
https://github.com/rustfs/rustfs/issues/625

Nugine:
Upstream bug fixed
+ https://github.com/s3s-project/s3s/issues/326
+ https://github.com/rustfs/rustfs/pull/626

PR #626, opened 2025-10-08
fix: CompleteMultipartUpload encryption

Nugine:
Related
+ #581


Nugine:
@guojidan Please check whether it works correctly.

Issue #887, opened 2025-11-18
RUSTFS_SERVER_DOMAINS breaks console bucket administration


RogerSik:
RUSTFS_SERVER_DOMAINS is needed to set for working vhost style. But when I set it breaks the administration of buckets in console

Working/tested console features:

- Login
- Access Keys

Not working console features:

- Bucket List
- Bucket Creation

<img width="1652" height="531" alt="Image" src="https://github.com/user-attachments/assets/ff1fd27b-5a18-444d-aab2-d1d373aefcbe" />

<img width="1639" height="803" alt="Image" src="https://github.com/user-attachments/assets/2f344d52-7a46-4aa2-a68d-0e8ea15f685a" />

.env
```yaml
RUSTFS_ACCESS_KEY="REDACTED"
RUSTFS_SECRET_KEY="REDACTED"
RUSTFS_CONSOLE_ENABLE=true

RUSTFS_SERVER_DOMAINS="rustfs.example.org"
RUSTFS_ADDRESS=0.0.0.0:9000
RUSTFS_CONSOLE_ADDRESS=0.0.0.0:9001
RUSTFS_EXTERNAL_ADDRESS=:9000

RUSTFS_CONSOLE_CORS_ALLOWED_ORIGINS="https://console.rustfs.example.org"
```

compose.yaml
```yaml
services:
  rustfs:
    # https://hub.docker.com/r/rustfs/rustfs
    image: rustfs/rustfs:1.0.0-alpha.67
    hostname: unraid
    container_name: rustfs
    env_file: .env
    ports:
      - '9000:9000' # s3 api
      - '9001:9001' # management ocnsole
    volumes:
      - /mnt/cache/rustfs:/data
    tmpfs:
      - /logs:size=1G
    restart: unless-stopped
```

rustfs is reverse proxied by (external) traefik v3.6.0

RogerSik:
I have rustfs/rustfs:1.0.0-alpha.67 running. I don't know why but if I upgrade to alpha.68 it breaks also the console login itself. Invalid credentials are my error. So witched back to alpha.67

loverustfs:
We will review the code and deployment. We will get back to you once we have updates.

loverustfs:
[...]

PR #967, opened 2025-12-03
fix: handle reverse proxy forwarded headers correctly in console config

### LINKED ISSUE #966: RustFS uses internal ports in console config.json
**Describe the bug**

I have set up both the API and the console behind an Nginx reverse proxy, following the provided examples.

When inspecting the `config.json` loaded by the console, I can see that the protocol and hostname are correctly taken from the forwarded headers for the baseURL.

The problem is that the port still remains the internally published port, taken [from here](https://github.com/rustfs/rustfs/blob/main/rustfs/src/server/http.rs#L218).

An additional problem is that the S3 endpoint hostname is also taken from the forwarded header, even though the API is published under a different hostname.

This is my docker setup:

```yaml
services:
  rustfs:
    image: rustfs/rustfs:latest
    restart: always
    ports:
      - 127.0.0.1:13122:13122
      - 127.0.0.1:13123:13123
    environment:
      - RUSTFS_VOLUMES=/data
      - RUSTFS_ADDRESS=:13122
      - RUSTFS_CONSOLE_ADDRESS=:13123
      - RUSTFS_CONSOLE_ENABLE=true
      - RUSTFS_SERVER_DOMAINS=rustfs.example.com
      - RUSTFS_CORS_ALLOWED_ORIGINS=https://myapp.example.com
      - RUSTFS_CONSOLE_CORS_ALLOWED_ORIGINS=https://console.rustfs.example.com
      - RUSTFS_ACCESS_KEY=…
      - RUSTFS_SECRET_KEY=…
      - RUSTFS_LOG_LEVEL=info
    healthcheck:
      test: [ "CMD", "sh", "-c", "curl -f http://localhost:13122/health && curl -f http://localhost:13123/rustfs/console/health" ]
      interval: 30s
      timeout: 10s
      retries: 3
    volumes:
      - ./data:/data
```

DASPRiD:
Fixes #966

## Problem

When RustFS is deployed behind a reverse proxy (like Nginx):

1. The console's `config.json` included internal ports (e.g., `:13123`) instead of public ports
2. The S3 endpoint used the console's hostname instead of allowing a separate API hostname
3. Standard forwarded headers were not properly handled

### 1. Fixed Port Extraction Logic

- Added support for `X-Forwarded-Host` header (takes priority over `Host` header)
- Added support for `X-Forwarded-Port` header
- Implemented smart port inference: defaults to 443 for HTTPS and 80 for HTTP
- URLs now omit the port when it's the default for the scheme

### 2. Added Independent S3 Endpoint Configuration

- New environment variable: `RUSTFS_API_PUBLIC_ENDPOINT`
- Allows separate configuration for API and Console domains
- Falls back to console URL if not set (backward compatible)

### 3. Updated Documentation

- Added comprehensive documentation for `RUSTFS_API_PUBLIC_ENDPOINT`
- Updated reverse proxy configuration examples

## Example Usage

```yaml
services:
  rustfs:
    image: rustfs/rustfs:latest
    environment:
      - RUSTFS_API_PUBLIC_ENDPOINT=https://rustfs.example.com
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,23 @@ make help-docker                      # Show all Docker-related commands
 
 Follow the instructions in the [Helm Chart README](https://charts.rustfs.com/) to install RustFS on a Kubernetes cluster.
 
+### 5\. Nix Flake (Option 5)
+
+If you have [Nix with flakes enabled](https://nixos.wiki/wiki/Flakes#Enable_flakes):
+
+```bash
+# Run directly without installing
+nix run github:rustfs/rustfs
+
+# Build the binary
+nix build github:rustfs/rustfs
+./result/bin/rustfs --help
+
+# Or from a local checkout
+nix build
+nix run
+```
+
 -----
 
 ### Accessing RustFS
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,8 +2,7 @@
 
 ## Supported Versions
 
-Use this section to tell people about which versions of your project are
-currently being supported with security updates.
+Security updates are provided for the latest released version of this project.
 
 | Version | Supported          |
 | ------- | ------------------ |
@@ -11,8 +10,10 @@ currently being supported with security updates.
 
 ## Reporting a Vulnerability
 
-Use this section to tell people how to report a vulnerability.
+Please report security vulnerabilities **privately** via GitHub Security Advisories:
 
-Tell them where to go, how often they can expect to get an update on a
-reported vulnerability, what to expect if the vulnerability is accepted or
-declined, etc.
+https://github.com/rustfs/rustfs/security/advisories/new
+
+Do **not** open a public issue for security-sensitive bugs.
+
+You can expect an initial response within a reasonable timeframe. Further updates will be provided as the report is triaged.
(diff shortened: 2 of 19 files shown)


DASPRiD:
I'm not certain, but I think this should also address #887

NightSling:
Please merge this asap, @loverustfs!

loverustfs:
> Please merge this asap, @loverustfs！

We're reviewing the code to trace the context of the bug and assess the impact of the fix. Thanks for the contribution and the suggestion!

loverustfs:
The bug was likely caused by the three automated builds we ran yesterday. It seems DockerHub picked up the artifact from the third build.

We're going to release alpha.72 first.

@DASPRiD Great solution. We just need to evaluate the scope of its impact before merging.

https://github.com/rustfs/rustfs/actions/runs/19895021399

loverustfs:
@houseme  @weisd i think this is a good solution to nginx reverse proxy.

loverustfs:
Hey @DASPRiD ,

The config.json file will be removed in the next version.

Thank you for your contribution.

Issue #1001, opened 2025-12-05
Cluster Behaves Differently to kill vs. Abrupt Power-Off


Pandas886:
## Context
We have a cluster consisting of 4 machines, each running the rustfs service with 4 disks. This setup forms a single, distributed storage cluster.

## Expected Behavior
The cluster should maintain availability and allow normal file uploads from the application even when one node becomes unavailable, as long as the quorum and data redundancy requirements are met.

## Actual Behavior
- Case 1: Terminating the rustfs process (kill)
  - Action: We terminate the rustfs process on one of the machines.
  - Result: The application continues to upload files normally. This behavior is as expected.
- Case 2: Abruptly powering off a machine  
  - Action: We abruptly power off one of the machines (e.g., by cutting the power).
  - Result: The application cannot upload files normally. Furthermore, the Console Web UI for all rustfs nodes becomes unresponsive or appears to be "hung"/"frozen". The entire cluster seems to become unresponsive.

## Problem Summary
The cluster exhibits different behaviors depending on how a node is taken offline:
- It recovers correctly from a graceful (or signal-based) process termination (kill).
- It fails to recover and becomes unresponsive when a node is abruptly powered off.

This suggests a potential issue in how the cluster handles the sudden, ungraceful failure of a node, possibly related to:
-   Network partition detection and handling.
-   Quorum calculation and re-election processes.
-   Session or connection state management when a node vanishes without sending "goodbye" messages.
-   Timeouts for detecting a truly dead node.

## Steps to Reproduce
1.  Deploy a 4-node rustfs cluster (4 machines, each with 4 disks).
2.  Verify the cluster is healthy and the application can upload files.
3.  Abruptly power off one of the machines (e.g., by cutting the power).
4.  Observe that the application can no longer upload files.
5.  Attempt to access the Console Web UI of any rustfs node. It will be unresponsive.

## Additional Information
-   Cluster Size: 4 nodes.
-   Disks per Node: 4.
-   Failure Mode: The key difference is between a process-level kill and a hardware-level power-off.

loverustfs:
Which version do you use?

loverustfs:
Thank you very much for such a professional test, and we will fix it.

Pandas886:
I suspect it's related to `rustfs/crates/protos/src/lib.rs `. Some timeout detection would be needed to more quickly identify the issue. For example:  
```
  let channel = match channel {
        Some(channel) => channel,
        None => {
[...]

PR #1025, opened 2025-12-06
fix: detect dead nodes via HTTP/2 keepalives (Issue #1001)

Jitterx69:
### Resolution for Issue #1001

This PR resolves the issue where the cluster becomes unresponsive after a node experiences an abrupt power-off.

### The Fix
Configures the internal gRPC client (`crates/protos/src/lib.rs`) to use active Application-Layer Heartbeats:
- **HTTP/2 PING Interval**: 5 seconds
- **Timeout**: 3 seconds
- **TCP Keepalive**: 10 seconds (Backup)

### Verification
- Surviving nodes now detect dead peers in ~8 seconds.
- `cargo check`, `cargo test`, and `cargo clippy` passed.

### DIFF
--- a/crates/ecstore/src/erasure_coding/encode.rs
+++ b/crates/ecstore/src/erasure_coding/encode.rs
@@ -149,6 +149,12 @@ impl Erasure {
                         break;
                     }
                     Err(e) if e.kind() == std::io::ErrorKind::UnexpectedEof => {
+                        // Check if the inner error is a checksum mismatch - if so, propagate it
+                        if let Some(inner) = e.get_ref() {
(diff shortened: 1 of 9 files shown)


weisd:
Uploading and downloading don’t use tonic rpc. Instead, they use HttpReader from crates/rio. Maybe we can work in this direction.

Jitterx69:
> Uploading and downloading don’t use tonic rpc. Instead, they use HttpReader from crates/rio. Maybe we can work in this direction.

Indeed, talk more about this. So that I can help fixing that.

Issue #1021, opened 2025-12-06
support copy on write and viewing file history by datetime


Busboy3129:
**Is your feature request related to a problem? Please describe.**
support copy on write and viewing file history by datetime

**Describe the solution you'd like**
 copy on write for a bucket

**Describe alternatives you've considered**

**Additional context**

loverustfs:
Hi  @Busboy3129 ,

Can we use bucket replication?

loverustfs:
Thank you for the suggestion. Currently, we have no plans to implement this feature.

PR #1045, opened 2025-12-07
feat: Add copy-on-write and point-in-time file history

tennisleng:
Implements support for point-in-time object retrieval and file history viewing (#1021).


tennisleng:
Closing in favor of split PRs: #1047, #1048

PR #1106, opened 2025-12-10
helm: Use service.type from Values

simonswine:
I would like to not expose the service externally. (So no ingress and no NodePort)


majinghe:
@simonswine Thanks for your contribution. 

/LGTM

/cc @loverustfs @houseme

PR #1107, opened 2025-12-10
helm: Allow configuration of volumes per replica

simonswine:
This is a lot more flexible than forcing the replica count to be either
4 or 16.

I have verified there are no other breaking changes:

```
$ helm template rustfs helm/rustfs > before.txt
$ helm template rustfs helm/rustfs > after.txt
$ diff -u before.txt after.txt
--- before.txt  2025-12-10 18:24:05
+++ after.txt   2025-12-10 18:24:11
@@ -128,17 +128,15 @@
             - sh
             - -c
             - |
-              if [ "$REPLICA_COUNT" -eq 4 ]; then
-                for i in $(seq 0 $(($REPLICA_COUNT - 1))); do
-                  mkdir -p /data/rustfs$i
-                done;
-              elif [ "$REPLICA_COUNT" -eq 16 ]; then
-                mkdir -p /data
-              fi
-
+              mkdir -p /data/rustfs0
+              mkdir -p /data/rustfs1
+              mkdir -p /data/rustfs2
+              mkdir -p /data/rustfs3
               chown -R 10001:10001 /data
               chown -R 10001:10001 /logs
           volumeMounts:
+            - name: logs
+              mountPath: /logs
             - name: data-rustfs-0
               mountPath: /data/rustfs0
             - name: data-rustfs-1
@@ -147,8 +145,6 @@
               mountPath: /data/rustfs2
             - name: data-rustfs-3
               mountPath: /data/rustfs3
-            - name: logs
-              mountPath: /logs
       containers:
         - name: rustfs
           image: "rustfs/rustfs:latest"
@@ -165,9 +161,6 @@
               name: endpoint
             - containerPort: 9001
               name: console
-          env:
-            - name: REPLICA_COUNT
-              value: "4"
           envFrom:
             - configMapRef:
                 name: rustfs-config
```

### DIFF
--- a/rustfs/src/storage/access.rs
+++ b/rustfs/src/storage/access.rs
@@ -25,7 +25,7 @@ use s3s::{S3Error, S3ErrorCode, S3Request, S3Result, dto::*, s3_error};
 use std::collections::HashMap;
 
 #[allow(dead_code)]
-#[derive(Default, Clone)]
+#[derive(Default, Clone, Debug)]
 pub(crate) struct ReqInfo {
     pub cred: Option<rustfs_credentials::Credentials>,
     pub is_owner: bool,
@@ -38,6 +38,7 @@ pub(crate) struct ReqInfo {
 /// Authorizes the request based on the action and credentials.
 pub async fn authorize_request<T>(req: &mut S3Request<T>, action: Action) -> S3Result<()> {
     let remote_addr = req.extensions.get::<RemoteAddr>().map(|a| a.0);
+
     let req_info = req.extensions.get_mut::<ReqInfo>().expect("ReqInfo not found");
 
     if let Some(cred) = &req_info.cred {
@@ -426,8 +427,12 @@ impl S3Access for FS {
     /// Checks whether the DeleteObjects request has accesses to the resources.
     ///
     /// This method returns `Ok(())` by default.
-    async fn delete_objects(&self, _req: &mut S3Request<DeleteObjectsInput>) -> S3Result<()> {
-        Ok(())
+    async fn delete_objects(&self, req: &mut S3Request<DeleteObjectsInput>) -> S3Result<()> {
+        let req_info = req.extensions.get_mut::<ReqInfo>().expect("ReqInfo not found");
+        req_info.bucket = Some(req.input.bucket.clone());
+        req_info.object = None;
+        req_info.version_id = None;
+        authorize_request(req, Action::S3Action(S3Action::DeleteObjectAction)).await
     }
 
     /// Checks whether the DeletePublicAccessBlock request has accesses to the resources.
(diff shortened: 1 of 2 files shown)


majinghe:
@simonswine I reviewed the change which has different with our Rust Mode Setting. For helm installation, RustFS provides two mode

- Standalone mode: one pod with one pvc, namely SNSD(single node single disk mode)
- Distributed mode: MNMD(multiple node and multiple disk)
    - 4 pods with each pod has 4 pvc, totally 16 pvc
    - 16 pod with each pod has only 1 pvc, totally 16 pvc, so the below code
    
     ```
             {{- if gt (int .Values.volumesPerReplica) 1 }}
            {{- range $i := until (int .Values.volumesPerReplica) }}
            - name: data-rustfs-{{ $i }}
              mountPath: /data/rustfs{{ $i }}
            {{- end }}
            {{- else }}
            - name: data
              mountPath: /data
            {{- end }}
    ```

    means if the replicas is `16` and the `pvc` will also be `16`, and totally pvc will be 16*16=256, that's not our exception.

simonswine:
I see what you are saying, I can fix this with a bit more defaulting, but what I am confused with is, what if I wanna run 8 replicas with two volumes  (or any other number than is not 1-1, 4-4, 16-4 [replica-volumes]), is that be bad for performance or just right out broken?

majinghe:
@simonswine Good question, actually, 8 replicas are also fine, even other replicas( for example, 32 64) are also fine. As you know, MNMD mode has several node and disk combination, but helm chart can not cover all the combinations, so we just support two main combinations( 4 replicas and 16 replicas) which is used frequently. If the current helm chart can not satisfy  the user's requirements, user should modify the helm chart locally, but we will not accept the change.

majinghe:
@simonswine Hi, could y please fix the conflict?

majinghe:
@simonswine Could y pls help to handle the conflict?

houseme:
@majinghe

majinghe:
> @simonswine Good question, actually, 8 replicas are also fine, even other replicas( for example, 32 64) are also fine. As you know, MNMD mode has several node and disk combination, but helm chart can not cover all the combinations, so we just support two main combinations( 4 replicas and 16 replicas) which is used frequently. If the current helm chart can not satisfy the user's requirements, user should modify the helm chart locally, but we will not accept the change.

@simonswine Could y pls change the code according to this logic?

@houseme is our administration, he fix the conflicts, but i think the change code don't satisfy our requirement. If there is no any update, we will close PR due to it hangs on a long time. Thx.

majinghe:
@simonswine Due to the PR, there hasn't been any update for more than 3 weeks. We will close this PR. If you want to continue contributing with this PR, you can reopen it or create a new one. Thx again for you contributio.

PR #1166, opened 2025-12-16
feat: Replace LRU cache with Moka async cache in policy variables

yxrxy:
https://github.com/rustfs/rustfs/issues/1167
## Type of Change
- [x] Performance Improvement


houseme (approved):
LGTM

Issue #1187, opened 2025-12-18
pre-commit when rust is not installed


cod3mas0n:
**Describe the bug**
Pre Commit issue when rust is not installed.

**To Reproduce**
Remove all rust, cargo, rustc, and rust-related's then run `make pre-commit`

**Expected behavior**

```bash
🔧 Formatting code...
cargo fmt --all
error: no such command: `fmt`

help: a command with a similar name exists: `fix`

help: view all installed commands with `cargo --list`
help: find a package to install `fmt` with `cargo search cargo-fmt`
make: *** [Makefile:16: fmt] Error 101
```

**Platform**
 - OS: Linux
 - Distro: Fedora 43

**Additional context**

`cargo fmt` comes from `rustfmt` so it has to be installed before doing `make pre-commit`

`cargo help`

```
Rust's package manager

Usage: cargo [OPTIONS] [COMMAND]
       cargo [OPTIONS] -Zscript <MANIFEST_RS> [ARGS]...

Options:
  -V, --version                  Print version info and exit
      --list                     List installed commands
      --explain <CODE>           Provide a detailed explanation of a rustc error message
  -v, --verbose...               Use verbose output (-vv very verbose/build.rs output)
  -q, --quiet                    Do not print cargo log messages
      --color <WHEN>             Coloring [possible values: auto, always, never]
  -C <DIRECTORY>                 Change to DIRECTORY before doing anything (nightly-only)
      --locked                   Assert that `Cargo.lock` will remain unchanged
      --offline                  Run without accessing the network
      --frozen                   Equivalent to specifying both --locked and --offline
      --config <KEY=VALUE|PATH>  Override a configuration value
  -Z <FLAG>                      Unstable (nightly-only) flags to Cargo, see 'cargo -Z help' for details
  -h, --help                     Print help

Commands:
    build, b    Compile the current package
    check, c    Analyze the current package and report errors, but don't build object files
    clean       Remove the target directory
    doc, d      Build this package's and its dependencies' documentation
    new         Create a new cargo package
    init        Create a new cargo package in an existing directory
    add         Add dependencies to a manifest file
    remove      Remove dependencies from a manifest file
    run, r      Run a binary or example of the local package
    test, t     Run the tests
    bench       Run the benchmarks
    update      Update dependencies listed in Cargo.lock
    search      Search registry for crates
    publish     Package and upload this package to the registry
    install     Install a Rust binary
    uninstall   Uninstall a Rust binary
    ...         See all commands with --list

[...]

PR #1188, opened 2025-12-18
Add: rust installation in Makefile

cod3mas0n:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [x] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues

pre-commit when rust is not installed #1187

## Summary of Changes

- Add `install-rust` target in Makefile.
- Add `install-rust` target as a dependency for `pre-commit` target.

## Checklist

---


cod3mas0n:
This PR will be changed according to the discussion in https://github.com/rustfs/rustfs/issues/1187#issuecomment-3670675345

> Placing the check at the top level of the Makefile means it will execute for every command, even those that don't require Rust (e.g., make help, make clean, or Docker-based targets). This would block users who only want to interact with non-Rust parts of the project.

cod3mas0n:
Hi @loverustfs 

Here is the updated changes as we discussed in https://github.com/rustfs/rustfs/issues/1187#issuecomment-3677754928

## Dependency checks via pattern targets


- `check-<dep-name>`
  Performs a fatal check for a required dependency.
  If `<dep-name>` is not found in PATH, Make exits with an error.

- `warn-<dep-name>`
  Performs a `non-fatal` check for an optional dependency.
  If `<dep-name>` is missing, a warning is printed but execution continues. 
  Put the `test-deps` for future uses .

Grouped dependency targets

Dependencies are grouped by purpose for convenience:

`core-deps` → required core tooling

`fmt-deps` → required formatting tools

`test-deps` → optional test tooling

`pre-commit-deps` → required git hooks tooling

loverustfs:
Hi @cod3mas0n,

Thanks for the contribution! I really like the new check-% and warn-% mechanism; it makes the build requirements much clearer and the error messages more friendly.

However, I have a few suggestions regarding the workflow logic before we merge:

1. Decoupling pre-commit from setup-hooks Currently, the pre-commit target depends on setup-hooks:


```

pre-commit: setup-hooks fmt clippy check test
```


This means every time a developer just wants to run the check suite locally (by typing make pre-commit), it attempts to re-install/configure the git hooks. It is better to keep the "execution of checks" separate from the "installation of environment tools".

2. Hard dependency on the pre-commit framework You added pre-commit install --all in setup-hooks and a check-pre-commit dependency.

The original Makefile used a simple shell script (chmod +x .git/hooks/pre-commit).

This change implies a hard dependency on the external [pre-commit](https://pre-commit.com/) framework (Python). If a user doesn't have this tool installed, make setup-hooks or make pre-commit will fail hard.

## Suggested Changes:

Could you please modify the pre-commit target to remove setup-hooks from its dependencies? It should look like this:


```
.PHONY: pre-commit
# Remove setup-hooks from dependencies here so we don't install hooks every time we run checks
pre-commit: fmt clippy check test
	@echo "✅ All pre-commit checks passed!"
```
Also, please confirm if we want to enforce the external pre-commit tool usage. If so, check-pre-commit is fine, but if we want to keep it optional (or stick to the local script), we should remove pre-commit install --all.

Thanks!

cod3mas0n:
Yes, I can do that 

But if a user does a straightforward clone of the repo, `make setup-hooks` would fail

```
chmod: cannot access '.git/hooks/pre-commit': No such file or directory
```

It checks for `pre-commit` too in the `pre-commit-deps` target via `check-pre-commit` so user should install it.

Being optional is good, and it can be removed from pre-commit, although there are useful pre-commit-hooks out there, and I do suggest using them, and even may be with those hooks, the `pre-commit run --all` can be added.

Awesome hooks Like :

- Security
  - https://github.com/trufflesecurity/trufflehog/blob/main/.pre-commit-hooks.yaml
  - https://github.com/Yelp/detect-secrets/blob/master/.pre-commit-hooks.yaml
- Dockerfile
  - https://github.com/hadolint/hadolint/blob/master/.pre-commit-hooks.yaml

```yaml
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v5.0.0
    hooks:
      - id: check-merge-conflict
      - id: end-of-file-fixer
      - id: trailing-whitespace
      - id: check-added-large-files
      - id: check-executables-have-shebangs
      - id: check-json
      - id: check-symlinks
      - id: detect-private-key

```

`shellcheck`

```yaml
  - repo: https://github.com/jumanjihouse/pre-commit-hooks
    rev: 3.0.0
    hooks:
      - id: shellcheck
        args: [--severity, "error"]
        exclude: "^.git"
        files: "\\.sh$"

```

Please consider these hooks.

Thanks.

loverustfs:
Thanks for the quick feedback. You are absolutely right—since git hooks aren't version-controlled, a fresh clone won't have the .git/hooks/pre-commit file, causing chmod to fail.

However, I'd prefer to avoid introducing a hard dependency on the external Python pre-commit tool for the core build process, as we want to keep the dev environment setup as lightweight as possible for Rust developers.

Proposal: Instead of using pre-commit install, could we simply manage the hook script within the repo (e.g., creating a file at scripts/git-hooks/pre-commit.sh)?

Then, the setup-hooks target in the Makefile would look something like this:
```

setup-hooks:
	@echo "🔧 Setting up git hooks..."
	# Copy or symlink the script from repo to .git/hooks
	cp scripts/git-hooks/pre-commit.sh .git/hooks/pre-commit
	chmod +x .git/hooks/pre-commit
	@echo "✅ Git hooks setup complete!"
```

This way, we fix the "file not found" error on fresh clones without requiring users to install the external pre-commit binary.

What do you think?

cod3mas0n:
That makes sense, and I agree that keeping the Rust dev setup lightweight is important.

That said, I’d still strongly prefer using `pre-commit` for hook management.

While a repo-managed shell hook avoids the Python dependency, it also pushes us toward maintaining ad-hoc shell logic over time. In practice, `.pre-commit-config.yaml` is significantly easier to reason about, review, and extend than a growing shell script—especially as checks evolve.

A few concrete benefits of `pre-commit` here:

- Hooks are declarative and versioned (`.pre-commit-config.yaml`)
- Adding/removing checks is trivial and doesn’t require shell scripting
- Hooks are reproducible and consistent across contributors and CI
- The `pre-commit` binary is a single tool, widely packaged across major distros `apt`, `dnf`, `yum`, `pacman, etc ...

We can also keep it non-blocking for contributors by making hook installation opt-in (`pre-commit install`) and ensuring CI runs the same checks independently.

From a long-term maintenance perspective, I believe `pre-commit` will be simpler and safer than maintaining custom shell hooks ourselves.

Given that development environments naturally accumulate tooling over time, I think using a standard, declarative solution like pre-commit is a fair trade-off here.

loverustfs:
Okay, I'll talk to the person in charge of the Makefile again to see if pre-commit is acceptable.

cod3mas0n:
Hi There @loverustfs 
Have any updates?

loverustfs:
@houseme @majinghe 

Could you please evaluate whether it is necessary to introduce a hard dependency on Python pre-commit tools?

cod3mas0n:
It’s not a hard dependency in the sense of affecting the build or runtime at all. The core build remains completely independent — builds can always be done via direct `make` / `cargo` targets, and we can even provide explicit targets that run without `pre-commit`.

The `pre-commit` tool is strictly scoped to pre-commit checks only. Its role is to provide a consistent, declarative, and maintainable way to run those checks before commits are created.

While it’s technically possible to replicate this with custom shell hooks, at that point we’re effectively re-implementing hook orchestration ourselves. For pre-commit checks specifically, using `pre-commit` is the simplest and most robust option, and avoids long-term maintenance of bespoke shell logic.

Contributors who don’t want to install it are not blocked — they can still build, test, and contribute normally, with CI enforcing the same checks. Given that scope and trade-off, I believe using `pre-commit` is justified and appropriate here.

houseme:
> @houseme @majinghe
> 
> Could you please evaluate whether it is necessary to introduce a hard dependency on Python pre-commit tools?

Personal suggestion is:

1. **Regarding the issue of introducing Python dependencies**: We understand your concerns, introducing Python dependencies indeed increases the requirement for contributors to install Python, which may lead to more issues. Therefore, we will avoid introducing unnecessary Python dependencies to simplify developers' environment configuration.

2. **Regarding error prompts and automatic installation**: You pointed out that when users do not have `cargo` and `rustfmt` installed, clear error prompts should be provided instead of automatic installation. My personal view is: automatic installation of tools goes beyond the scope of the project and may lead to unforeseen issues. Therefore, we will ensure that when users are missing `cargo` or `rustfmt`, clear error prompts are provided to guide users to manually install these dependencies.

cod3mas0n:
Thank You @houseme

Was trying to use and benefit from pre-commit, altho its a binary, its widely used tool to do pre-commit (even in CI) checks, trufflehog and detect-secret for security scans, Hadolint for Dockerfile checks, etc ...

The changes for setup-hooks and pre-commit are reverted now.
It can be merged; changes are just for the `rustfmt` and `cargo` installation checks.

loverustfs:
Hey @cod3mas0n ,
Thanks for the great work! Merged. 🚀

PR #1385, opened 2026-01-05
Default to helm chart version for docker image and not latest

mkrueger92:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Summary of Changes

This PR changes the default image version to the one defined in the helm chart. The version "latest" is a bad choice for a stable production environment.

## Checklist

## Impact
- [ ] Other impact:

## Additional Notes

Leaving the tag blank does a fallback does a fallback:

`image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"`
---


helm/rustfs/values.yaml
```diff
@@ -11,7 +11,7 @@ image:
   # This sets the pull policy for images.
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "latest"
+  tag: ""
```
majinghe, line 14: Currently, RustFS is alpha, we always want user to user latest version, since the latest version has fixed some old bugs and more new features. So, i think, the `latest` tag is better for now. How do y think about this?

mkrueger92: I agree with your goal that users should use the latest version. But since the Helm chart is released along with the RustFS docker image, it feels strange that a version X of the Helm chart does not necessarily install version X of the Docker image. The Helm chart improves too. Therefore, I would suggest to link the release of the Helm chart to the version of the Docker image.

Besides the statement above, there is another aspect: The release pipeline of the Helm chart replaces the version in the values.yaml and the Chart.yaml. The latest tag is not present in the released Helm chart anyway. At best, this is redundant.

For the above reasons I prefer to have the released version in the Chart.yaml and the tag left blank (for users to override the default) in the values.yaml.

@majinghe: Thank you for challenging me and my thoughts. If you prefer it to keep it as it is right now, then feel free to reject the PR.

majinghe: @mkrueger92 Sounds good, will merge this PR.

majinghe:
@mkrueger92  Thx for your contribution.

/LGTM

/cc @houseme @loverustfs

Issue #1381, opened 2026-01-04
OpenBSD: unable to compile


jan-schreib:
**Describe the bug**
Compilation on OpenBSD fails

**To Reproduce**
1. Use OpenBSD
2. Clone the repo
3. Compile with `cargo build`

**Expected behavior**
Compilation succeeds.

**Desktop (please complete the following information):**
OpenBSD hostname 7.8 GENERIC.MP#192 amd64

**Additional context**
Another issue is sysctl-rs - it currently also does not work on OpenBSD.
I opened an [additional PR](https://github.com/johalun/sysctl-rs/pull/73). If the changes get merged the compilation on OpenBSD only fails because of
a TCPKeepalive issue.

houseme:
Hello, regarding the improvements to removing `sysctl` and the reasons and benefits behind it, I hope to get a more detailed explanation from the authors of the issues and PRs. Specifically, I would like to understand why such changes were made and the actual benefits these changes bring to users of the OpenBSD platform. Looking forward to your reply. 

@jan-schreib

jan-schreib:
@houseme 

One reason for the removal was the single use of the library in total in the project. I don't think a single use, especially one that breaks build for an operating system, is justified. The benefits of outsourcing this functionality are minimal and are a wrapper above another wrapper.

My changes use a widely used library directly, while sysctl-rs is (according to github and their release cycle) less used.

So the benefits I see:
1. less dependencies
2. support for more operating systems (since `sysctl` is everywhere, while `sysctlbyname` and `sysctlnametomib` are not)

It introduces a single unsafe call which is otherwise just extracted into it's own library - so no real loss there, imho.

Making rustfs work with sysctl-rs is possible (see my upstream PR) but needs very invasive changes since it's not tested nor supported on e.g. OpenBSD, yet.

jan-schreib:
With all the fixes now applied (#1358 #1360 #1382 #1396) I can finally report that rustfs not compiles and runs (without webinterface) on OpenBSD! 

@loverustfs Feel free to close this issue as soon as you see fit.

houseme:
[...]

PR #1396, opened 2026-01-05
Remove the sysctl crate and use libc's sysctl call interface

jan-schreib:
## Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [x] Refactor
- [x] Other: Dependency reduction

## Related Issues
#1381 
## Summary of Changes
The crate sysctl-rs is used only in http.rs, importing a whole crate for this seems a bit like overkill.
I added an unsafe block to execute the necessary sysctl commands directly. The old way always returned DEFAULT_BACKLOG for at least OpenBSD. On FreeBSD the "kern.ipc.somaxconn" is deprecated and only kept for backward compatibily reasons. The libc crate does not use the new variant "kern.ipc.soacceptqueue".
MacOS and NetBSD share the same sysctl name and behaved like expected.

The new code uses the correct sysctl names, checks the return value of sysctl and returns the default value of 1024 in case an error happened. If everything worked the operating systems returned value is used. Libc is now available in the rustfs workspace.

## Checklist

## Impact
- [ ] Other impact:

## Additional Notes
This 25 line function removes the sysctl-rs dependency and also makes it usable on OpenBSD without a pretty invasive patch to sysctl-rs.

Compiled and tested on FreeBSD, OpenBSD and MacOS.

The sysctl-rs library used the same libc functions that need to be wrapped in unsafe. I don't think its too dangerous to rely on libc in this regard.

---

### DIFF
--- a/rustfs/Cargo.toml
+++ b/rustfs/Cargo.toml
@@ -125,7 +125,7 @@ url = { workspace = true }
 urlencoding = { workspace = true }
 uuid = { workspace = true }
 zip = { workspace = true }
-libc = { workspace = true}
+libc = { workspace = true }
 
 # Observability and Metrics
 metrics = { workspace = true }
--- a/rustfs/src/server/http.rs
+++ b/rustfs/src/server/http.rs
@@ -750,10 +750,10 @@ fn get_conn_queue_len() -> i32 {
     const DEFAULT_BACKLOG: i32 = 1024;
 
     #[cfg(target_os = "openbsd")]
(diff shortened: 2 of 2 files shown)


houseme:
We look forward to you providing a more elegant and secure implementation solution. Thank you for your continued efforts, and we await your update.

jan-schreib:
> We look forward to you providing a more elegant and secure implementation solution. Thank you for your continued efforts, and we await your update.

What do you mean with a more secure implementation? Do you have something in mind?

PR #1498, opened 2026-01-13
In the PVC definition, skip the storageClassName attr if null/empty

arthurdarcet:
## Type of Change
- [ ] New Feature
- [*] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Summary of Changes
Without this, setting the class name to null/empty would:
- at first install, it create a pvc with no storage class, and the cluster will use the default class (which is what i want)
- but at the next upgrade, it will try to apply `storageClassName=""` again, and that will conflict with the `storageClassName=my-default-class` that was added by the storage operator ; and the upgrade will fail

## Checklist
- [*] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines

## Impact
- [ ] Other impact:

### DIFF
--- a/crates/ecstore/src/store.rs
+++ b/crates/ecstore/src/store.rs
@@ -1825,16 +1825,16 @@ impl StorageAPI for ECStore {
             if self.is_suspended(pool.pool_idx).await {
                 continue;
             }
-            match pool
+            return match pool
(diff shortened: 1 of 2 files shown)


majinghe:
@arthurdarcet Thanks for your contribution.

/LGTM

/cc @houseme @loverustfs

Issue #1505, opened 2026-01-14
RustFS client possible bug


river-001:
https://github.com/rustfs/rustfs/blob/68c5c0b834a2656ad93a1101c560ba8e58d2030f/crates/ecstore/src/client/api_put_object_common.rs#L93
This line of code may need to add the ceil function:
part_size_flt = (part_size_flt / configured_part_size as f64).ceil() * configured_part_size as f64;

But I'm not sure, hope it helps

loverustfs:
Hello @river-001 ,


Thank you for your feedback and suggestions! I understand the issue you raised regarding the calculation of `part_size_flt` and will take your suggestions into serious consideration. I will check the relevant code as soon as possible and try to add the `ceil` function to address this problem. If further assistance or discussion is needed, please let me know anytime. Looking forward to your reply, thank you!

river-001:
OK, I'll give it a try.

bytehalo-infra:
https://github.com/rustfs/rustfs/pull/1532
This is my merge, please take a look when you have time.

loverustfs:
Hi @mengyu-sxyz ,

Thank you for your contribution.

Copilot provided some suggested modifications:

https://github.com/rustfs/rustfs/pull/1532

bytehalo-infra:
Okay, I have optimized the code based on the AI's suggestions.

bytehalo-infra:
Solved
https://github.com/rustfs/rustfs/pull/1532

loverustfs:
Ok, i'll close this issue.

Thank you very much.

PR #1532, opened 2026-01-16
fix: improve part size calculation in optimal_part_info function

mengyu-sxyz:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues

#1505

## Bug Description

**Issue 1: Ineffective Floating-Point Calculation (Line 93)**

The expression `(part_size_flt / configured_part_size as f64) * configured_part_size as f64` performs floating-point division followed by multiplication, which mathematically equals the original value without any rounding effect. The original intent was to round up `part_size_flt` to the nearest multiple of `MIN_PART_SIZE`, but this implementation fails to achieve that goal.

**Issue 2: Integer Division Precision Loss (Line 92)**

The calculation `object_size / MAX_PARTS_COUNT` uses integer division, which truncates the fractional part. This can result in a part size that is too small, potentially causing the actual number of parts to exceed `MAX_PARTS_COUNT` when dividing the total object size.

**Issue 3: Missing Minimum Value Protection**

When `object_size / MAX_PARTS_COUNT` is less than `MIN_PART_SIZE`, the code fails to enforce the minimum part size constraint. This could lead to part sizes below the allowed minimum threshold of 16 MiB (or 5 MiB absolute minimum).

## Summary

These bugs could cause the multipart upload logic to:
- Generate part sizes that don't align with expected boundaries
- Create more parts than the maximum allowed (10,000 parts)
- Use part sizes smaller than the minimum required size

The fix addresses all three issues by using proper floating-point arithmetic with ceiling operations and explicit minimum value enforcement.

## Checklist
- [x] Documentation updated (if needed)
- [x] CI/CD passed (if applicable)

## Impact
- [ ] Other impact:

## Additional Notes

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7148,6 +7148,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "futures",
+ "http 1.4.0",
  "rustfs-common",
  "rustfs-config",
  "rustfs-ecstore",
--- a/crates/heal/Cargo.toml
+++ b/crates/heal/Cargo.toml
@@ -49,3 +49,4 @@ serial_test = { workspace = true }
 tracing-subscriber = { workspace = true }
 tempfile = { workspace = true }
 walkdir = { workspace = true }
+http = { workspace = true }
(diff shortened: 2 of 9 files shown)


LeonWang0735:
LGTM

PR #1594, opened 2026-01-23
Fix/correctly handle compression

LeonWang0735:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues

RELATE #1668

## Summary of Changes

When an object is compressed, compression and actual-size metadata were only set in the metadata HashMap but not synchronized to opts.user_defined. This caused compression metadata to be lost when saving object metadata, leading to:
- is_compressed() failing to identify compressed objects
- is_compressed_ok() failing to identify compressed objects which cause get_object unable to decompress to the original file

This fix ensures compression metadata is properly synchronized to opts.user_defined so it can be correctly persisted and retrieved.
## Checklist

### DIFF
--- a/.github/s3tests/s3tests.conf
+++ b/.github/s3tests/s3tests.conf
@@ -120,7 +120,11 @@ secret_key = ${S3_ALT_SECRET_KEY}
 display_name = RustFS Tenant Tester
 
 # tenant user_id
-user_id = rustfstenant
+# Note: Using same user_id as main to avoid teardown failures.
+# RustFS does not currently support multi-tenancy, so the tenant client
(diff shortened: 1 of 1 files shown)


overtrue:
Suggestion: cache `CompressionAlgorithm::default()` in a local variable to avoid calling it twice and keep the value consistent if the default ever changes. e.g. `let algorithm = CompressionAlgorithm::default();` then use `algorithm.to_string()` and `CompressReader::new(hrd, algorithm)`.

PR #1617, opened 2026-01-26
Fix:s3 compatibility

LeonWang0735:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues

RELATE #1668

## Summary of Changes

- Fix object tag ordering: Sort tags by key alphabetically in decode_tags to match AWS S3 behavior. Fixes test_put_obj_with_tags which expects  tags in alphabetical order by Key. Allow to set the value of a tag to an empty string, but cannot set it to a null value. Reference：https://docs.aws.amazon.com/zh_cn/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions

- Fix authentication error code: Return InvalidAccessKeyId (403) instead of UnauthorizedAccess (401) when access key is not found, matching AWS S3 behavior. Fixes test_list_buckets_invalid_auth.
Reference:
https://docs.aws.amazon.com/zh_cn/redshift/latest/dg/s3serviceexception-error.html

- Fix metadata filtering: Filter internal metadata case-insensitively to prevent exposing x-rustfs-internal-* fields regardless of casing. Fixes test_object_copy_versioning_multipart_upload which expects only user-defined metadata in response.
- Fix CopySource parsing for special characters: Parse ?versionId query parameter before URL decoding to correctly handle object keys containing special characters like '?'. Fixes test_multipart_copy_special_names which tests multipart copy with keys like '?versionId'.
- Allow ExpiresIn=0 in presigned URLs: Remove filter that rejected ExpiresIn=0, allowing it to be parsed and correctly return 403 Forbidden when expired. Fixes test_object_raw_get_x_amz_expires_out_range_zero which expects 403 instead of 400 for expired presigned URLs.

fix s3-test:
test_put_obj_with_tags 
test_list_buckets_invalid_auth
test_object_copy_versioning_multipart_upload
test_bucket_policy_allow_notprincipal
s3s-project/s3s#462
test_multipart_copy_special_names (s3s)
test_object_copy_versioned_url_encoding (s3s)
test_object_raw_get_x_amz_expires_out_range_zero (s3s)

## Checklist

### DIFF
--- a/crates/ecstore/src/bucket/tagging/mod.rs
+++ b/crates/ecstore/src/bucket/tagging/mod.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use s3s::dto::Tag;
+use std::cmp::Ordering;
 use std::collections::HashMap;
 use url::form_urlencoded;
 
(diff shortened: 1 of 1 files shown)


houseme:
It is recommended to associate the relevant s3 document addresses

Issue #1641, opened 2026-01-28
Don't work bucket replication


justc0de-it:
**Describe the bug**
When I try start bucket replication get error

```
Connection error for bucket: warp-test, access key: a05AhNoov4 quahC2daSahth, error: failed
to dispatch bucket exists for bucket: warp-test
error:DispatchFailure { source: ConnectorError {
kind: lo, source:
hyper_util::client::legacy::Error(Connect,
ConnectError &quot; dns error&quot;, Custom {
kind: Uncategorized, error: &quot; failed to lookup address information: Name does not resolve&quot; })), connection: Unknown }}
```

**To Reproduce**
1. Go to bucket replication setting
2. Add rules
3. Push Add

**Expected behavior**
From server via CLI domain is resolved. 

**Screenshots**
<img width="306" height="217" alt="Image" src="https://github.com/user-attachments/assets/d76099e2-554a-430d-b77c-104cb5d8d0c4" />

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Safari
 - Version 26.3

**Additional context**
RustFS run as systemd-service on Ubuntu 24.04

loverustfs:
Hello, 

What are your steps & version?

justc0de-it:
1. Created an account, policy, and token for replication.
2. Created a test bucket on both servers with versioning enabled.
3. Go to the bucket settings menu and select Replication.
4. Fill in the fields as shown in the screenshot.
<img width="898" height="987" alt="Image" src="https://github.com/user-attachments/assets/2db5ad20-9c87-4db1-888a-2ff0b89affdb" />
5. After clicking the Add button, I get the error shown in the screenshot.

<img width="383" height="241" alt="Image" src="https://github.com/user-attachments/assets/c7fc0846-26ee-46c8-a5f5-ee299ae34f5d" />

6. RustFS Version

```
root@s3-prod:~# rustfs --version
rustfs 1.0.0-alpha.79
build time   : 2026-01-07 13:08:14 +00:00
build profile: release
build os     : linux-x86_64
rust version : rustc 1.92.0 (ded5c06cf 2025-12-08)
rust channel : stable-x86_64-unknown-linux-gnu
git branch   : 
git commit   : a95e549430851f3c34ee4cc8dd7be60aa200e8c4
git tag      : 1.0.0-alpha.79
git status   :
```

7. RustFS Settings

```
RUSTFS_ACCESS_KEY=USER
[...]

PR #1646, opened 2026-01-28
feat: strip protocol from replication target endpoint before adding it back based on secure

evanofslack:
Strip protocol from replication target endpoint before later adding it back based on `target.secure`. 

## Type of Change
- [x] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
#1641

## Summary of Changes
It is not clear to users that replication target endpoint cannot include protocol (`http://` or `https://`). This tries to strip the protocol off the endpoint if it exists, before it is later added back based on `target.secure`. With this change, users can provide a protocol in their replication target endpoint and it will resolve correctly. 

This is not a required change for feature to work, but would result in less errors and confusion for users. 

## Checklist
- [x] Documentation updated (if needed)
- [x] CI/CD passed (if applicable)

## Impact
- [ ] Other impact:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -735,9 +735,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-s3"
-version = "1.120.0"
+version = "1.121.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "06673901e961f20fa8d7da907da48f7ad6c1b383e3726c22bd418900f015abe1"
+checksum = "61948728b681f88a1e49b9500469cf9e36575a424e745e2c5a651a42386e7d9c"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -868,9 +868,9 @@ dependencies = [
 
 [[package]]
 name = "aws-smithy-async"
-version = "1.2.7"
+version = "1.2.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9ee19095c7c4dda59f1697d028ce704c24b2d33c6718790c7f1d5a3015b4107c"
+checksum = "9330762ee48c6cecfad2cb37b1506c16c8e858c90638eda2b1a7272b56f88bd5"
 dependencies = [
  "futures-util",
  "pin-project-lite",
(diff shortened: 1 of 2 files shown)


loverustfs:
Hello @evanofslack ,

Thank you for your contribution.

Copilot provides some suggestions for modification.

evanofslack:
> Hello @evanofslack ,
> 
> Thank you for your contribution.
> 
> Copilot provides some suggestions for modification.

@loverustfs 

I see, thank you. 

I feel that the current implementation, where the endpoint must not contain the protocol can cause confusion for users. But I also feel that the change I provided has some downsides as copilot has pointed out. Potentially there needs to be validation/sterilization happening right when the endpoint is received as part of the replication config. 

While informing the user of the expected format might be a good first step, I think that parsing and validating the replication config and returning an error on `put` would be the best possible solution. 

Please let me know your thoughts, happy to address this differently but feel it could be worth discussing first.

loverustfs:
> > Hello @evanofslack ,
> > Thank you for your contribution.
> > Copilot provides some suggestions for modification.
> 
> @loverustfs
> 
> I see, thank you.
> 
> I feel that the current implementation, where the endpoint must not contain the protocol can cause confusion for users. But I also feel that the change I provided has some downsides as copilot has pointed out. Potentially there needs to be validation/sterilization happening right when the endpoint is received as part of the replication config.
> 
> While informing the user of the expected format might be a good first step, I think that parsing and validating the replication config and returning an error on `put` would be the best possible solution.
> 
> Please let me know your thoughts, happy to address this differently but feel it could be worth discussing first.

Our plan is to implement filtering and alerts on the front end.

loverustfs:
Of course, your implementation is fantastic!

crates/ecstore/src/bucket/bucket_target_sys.rs
```diff
@@ -629,10 +629,17 @@ impl BucketTargetSys {
             .provider_name("bucket_target_sys")
             .build();
 
+        // Strip the protocol so it can be added back based on `target.secure`
+        let stripped = target
+            .endpoint
+            .strip_prefix("https://")
```
houseme, line 637: `https`=>`RUSTFS_HTTPS_PREFIX`

`http`=>`RUSTFS_HTTP_PREFIX`

evanofslack:
Closing due to other solutions

PR #1999, opened 2026-02-27
fix(crypto): patch polyval SIGILL on x86_64 CPUs without AVX2

wittjeff:
## Type of Change

- [x] Bug Fix

## Summary

RustFS crashes with `SIGILL` (illegal instruction) on x86_64 CPUs that support PCLMULQDQ but **not** AVX2, such as Intel Pentium Gold, Celeron, and some Atom processors. The crash occurs in the AES-GCM encryption path during normal operation (e.g., writing objects).

### Root Cause

The `polyval` 0.7.0-rc.9 crate's `avx2.rs` backend has a bug in its runtime CPU feature detection:

```rust
// avx2.rs line 34 — only checks for pclmulqdq
cpufeatures::new!(clmul, "pclmulqdq");
```

But the functions it guards require **both** PCLMULQDQ and AVX2:

```rust
#[target_feature(enable = "avx2", enable = "pclmulqdq")]
pub(super) unsafe fn expand_key(h: &[u8; 16]) -> ExpandedKey { ... }
```

On CPUs with PCLMULQDQ but without AVX2, the runtime check passes, the AVX2-annotated functions are called, and the process crashes on VEX-encoded instructions like `vmovdqu`.

### Fix

The fix adds `"avx2"` to the runtime feature detection:

```rust
cpufeatures::new!(clmul, "pclmulqdq", "avx2");
```

This ensures the intrinsics path is only used when **both** features are available. CPUs without AVX2 now correctly fall back to polyval's pure-Rust software implementation.

The patched `polyval` crate is vendored under `patches/` with a `[patch.crates-io]` override in the workspace `Cargo.toml`. This patch can be removed once the upstream fix lands and a new polyval version is released.

### Upstream

- Upstream issue: RustCrypto/universal-hashes#315
- Upstream fix PR: RustCrypto/universal-hashes#316

### Affected CPUs

Any x86_64 CPU with PCLMULQDQ but without AVX2, including:
- Intel Pentium Gold (G5400, G5420, etc.)
- Intel Celeron (G4900, G4930, etc.)
- Some Intel Atom processors
- Some older AMD processors

### Call stack at crash

```
rustfs_crypto::encdec::encrypt
  → aes_gcm::AesGcm::encrypt
    → ghash → polyval
      → polyval::backend::intrinsics::intrinsics_impl::expand_key
        → vmovdqu (%rsi),%xmm0    ← SIGILL
```

## Test Plan

- [x] Verified on Intel Pentium Gold G5420 (PCLMULQDQ: yes, AVX2: no)
- [x] `cargo check` passes with the patch applied
- [ ] CI tests pass


heihutu:
Thank you for your contribution. Is this related to `RUSTFLAGS=-C target-cpu`? Currently, we are using `RUSTFLAGS=-C target-cpu=native`, but the upcoming new version will be compiled using `RUSTFLAGS=-C target-cpu=x86-64-v2` to provide better compatibility. @wittjeff

heihutu:
The latest version of the `Polyval` crate, `0.7.0`, has been released.

heihutu:
> The latest version of the `Polyval` crate, `0.7.0`, has been released.

Thank you for your contribution. We are now using the `Polyval` version 0.7.0 released upstream.

PR #2007, opened 2026-02-28
fix(s3): reject SSE-C with partial headers per S3 spec

overtrue:
## Type of Change
- [x] Bug Fix

## Related Issues
S3 compatibility fix per S3_COMPAT_WORKFLOW.

## Summary of Changes

RustFS previously accepted PutObject requests with partial SSE-C headers (e.g. `x-amz-server-side-encryption-customer-algorithm` without key or MD5). Per the S3 API spec, when any SSE-C header is present, all three (algorithm, customer key, customer key MD5) must be provided. Similarly, mixing SSE-S3/SSE-KMS with SSE-C headers must be rejected with 400 InvalidArgument.

**Root cause:** When clients inject SSE-C headers via boto3 `before-call` events, the S3 layer may not populate them into `PutObjectInput`. Validation in `sse_encryption()` was never reached because the input had no SSE-C params.

**Fix:**
1. Add `extract_ssec_params_from_headers()` to read SSE-C params from request headers as fallback
2. Merge header params with input in `execute_put_object` (input takes precedence)
3. Call `validate_sse_headers_for_write()` early, before encryption logic
4. Add unit tests for extraction and validation

**s3-tests now passing:**
- `test_encryption_sse_c_no_key` — expects 400 when algorithm sent without key
- `test_encryption_sse_c_no_md5` — expects 400 when algorithm+key sent without MD5
- `test_put_obj_enc_conflict_c_s3` — expects 400 InvalidArgument when SSE-S3 and SSE-C headers conflict

## Checklist
- [x] I have read and followed the CONTRIBUTING.md guidelines

## Impact
- [ ] Other impact: None

Made with [Cursor](https://cursor.com)

### DIFF
--- a/rustfs/src/storage/sse.rs
+++ b/rustfs/src/storage/sse.rs
@@ -344,20 +344,14 @@ fn sse_invalid_argument(message: &str) -> ApiError {
 }
 
 /// SSE-C parameters extracted from headers (algorithm, key, key MD5).
-pub(crate) type SsecParamsFromHeaders = (
-    Option<SSECustomerAlgorithm>,
-    Option<SSECustomerKey>,
-    Option<SSECustomerKeyMD5>,
-);
+pub(crate) type SsecParamsFromHeaders = (Option<SSECustomerAlgorithm>, Option<SSECustomerKey>, Option<SSECustomerKeyMD5>);
 
 /// Extract SSE-C parameters from request headers.
 /// Used as fallback when the S3 layer does not populate them in the input struct.
 ///
 /// Returns an error if an SSE-C header is present but cannot be parsed as valid UTF-8,
 /// ensuring malformed headers do not bypass validation.
-pub(crate) fn extract_ssec_params_from_headers(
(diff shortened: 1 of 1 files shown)


overtrue:
**Addressing review comments (6600454a):**

1. **Invalid UTF-8 handling**: `extract_ssec_params_from_headers` now returns `Result` and rejects non-UTF-8 header values with `InvalidArgument`, as suggested.

2. **Additional regression tests**:
   - `test_extract_ssec_params_from_headers_rejects_invalid_utf8` — invalid UTF-8 in header
   - `test_validate_sse_headers_for_write_rejects_algorithm_and_key_without_md5` — algo+key without MD5
   - `test_validate_sse_headers_for_write_rejects_ssec_with_managed_sse` — SSE-C + SSE-S3 conflict

3. **Input precedence**: The merge uses `Option::or()` (input `.or`(header)), which guarantees input values take precedence when both are present. This is by construction.

PR #2050, opened 2026-03-02
fix(iam): remove incorrect trailing slash from OIDC issuer URL

### LINKED ISSUE #2049: OIDC discovery fails: normalize_config_url adds incorrect trailing slash to issuer URL
## Description

`normalize_config_url` in `crates/iam/src/oidc.rs` unconditionally appends a trailing slash to the issuer URL:

```rust
if !issuer.ends_with('/') {
    issuer.push('/');
}
```

This was introduced in commit 40692f18 (`fix(iam): address PR 1875 review issues for OIDC STS flows (#1969)`) with the comment:

> *"Ensure trailing slash for correct URL joining in the crate"*

However, the `openidconnect` crate does **not** use the issuer URL for URL joining — it performs a **strict string comparison** between the provided issuer and the `issuer` field returned in the OIDC discovery document. Most OIDC providers return issuer URLs **without** a trailing slash, so the appended slash causes an issuer mismatch error.

## Reproduction

1. Configure RustFS with a Keycloak OIDC provider:
   ```
   RUSTFS_IDENTITY_OPENID_CONFIG_URL=http://keycloak.example.com/realms/myrealm
   ```
2. Keycloak's discovery document returns:
   ```json
   { "issuer": "http://keycloak.example.com/realms/myrealm" }
   ```
3. `normalize_config_url` transforms the URL to `http://keycloak.example.com/realms/myrealm/`
4. The `openidconnect` crate's `discover_async` compares `http://keycloak.example.com/realms/myrealm/` (with slash) against `http://keycloak.example.com/realms/myrealm` (without slash) and fails with an issuer mismatch error.

## Expected Behavior

`normalize_config_url` should return the issuer URL without adding a trailing slash, preserving compatibility with the issuer string returned by the provider's discovery document.

## Affected Providers

Any OIDC provider that does not include a trailing slash in the issuer URL, including:
- Keycloak
- Auth0
- Okta
- Google

## Fix

Remove the trailing slash logic from `normalize_config_url`. A PR with the fix and updated tests is available.

RafaelHerrero:
## Summary

- Removes the trailing slash appended by `normalize_config_url` in `crates/iam/src/oidc.rs`, which causes OIDC discovery to fail with an issuer mismatch error against providers that don't include a trailing slash in their issuer URL (Keycloak, Auth0, Okta, etc.)
- Updates test expectations to match the corrected behavior

Fixes #2049

## Context

The trailing slash was introduced in 40692f18 with the comment *"Ensure trailing slash for correct URL joining in the crate"*, but the `openidconnect` crate performs strict string equality for issuer validation — it does not use the issuer URL for URL joining.

## Changes

```diff
-    if !issuer.ends_with('/') {
-        issuer.push('/');
-    }
```

## Test plan

- [x] `cargo test --package rustfs-iam -- oidc::tests` — all 20 tests pass
- [ ] Integration test with Keycloak OIDC provider


tscibilia:
This causes Authentik to fail, as it uses a trailing slash.
Can we configure this based on the `Configuration Issuer` instead?

Issue #2048, opened 2026-03-02
console ui show 403 when get object added throw mc


shinan88:
docker version v1.0.0-alpha.84, follow the instruction from installation doc. very simple.

mc can add object and downlaod object without problem.

rustfs console ui can upload file and preview file without problem.

but rustfs console ui can not preview file added through mc. from browser network can see 403 for head request for the object

same as mc, mc can not download file add from console ui. got access denied error

shinan88:
test the linux quick install, all default config. same issue. file added through python sdk can be seen in console ui, but can not get file. 403 error. And python sdk can not download file that added through console ui.

loverustfs:
Please wait for the new version to be released to fix the issue.

loverustfs:
Use Alpha.85 https://github.com/rustfs/rustfs/releases/tag/1.0.0-alpha.85

tiny-pangolin:
the issue no longer occurs for me on alpha 85 thanks for the rapid fix

Issue #2047, opened 2026-03-02
Buckets missing after update to alpha.84


janreppien:
**Describe the bug**
Bucket not found after update to alpha.84. Web interface and API show error.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to '`rustfs/console/browser`'
2. Click on any bucket
3. See error

or

1. Call `$ aws s3 sync ./folder s3://$S3_BUCKET/folder/ --endpoint-url "$AWS_ENDPOINT_URL" --only-show-errors`
2. See error

**Expected behaviour**
Bucket contents show in browser.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Zen
 - Version: 1.18.9b (Firefox 147.0.4) (aarch64)

**Additional context**
Error disappeared after downgrading back to alpha.83

Section of logs in log dir:

Errors triggered by web interface:

```
{"timestamp":"2026-03-02T06:59:59.82056461Z","level":"WARN","fields":{"message":"Failed to propagate tracing context: `SpanDisabled`"},"target":"rustfs::server::http","filename":"rustfs/src/server/http.rs","line_number":652,"threadName":"rustfs-worker","threadId":"ThreadId(5)"}

{"timestamp":"2026-03-02T06:59:42.525909844Z","level":"ERROR","fields":{"message":"op returns error","op":"HeadBucket","err":"S3Error(Inner { code: AccessDenied, message: Some(\"Access Denied\"), request_id: None, status_code: None, source: None, headers: None })"},"target":"s3s::ops","filename":"/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/s3s-0.13.0/src/ops/mod.rs","line_number":247,"threadName":"rustfs-worker","threadId":"ThreadId(4)"}
```

Errors triggered by manually calling `$ aws s3 sync ./folder s3://$S3_BUCKET/folder/ --endpoint-url "$AWS_ENDPOINT_URL" --only-show-errors`: 

Returns:
```
fatal error: An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
```

Logs on server:

```
[...]

Issue #2036, opened 2026-03-01
Custom policy ignored after upgrade to Alpha 84


Morethanevil:
**Describe the bug**
I have set a custom bucket policy for Mastodon, anonymous read shall be allowed, so profile pictures and media can be displayed.

This is the policy from the [Mastodon docs](https://docs.joinmastodon.org/admin/optional/object-storage/#minio), it works fine with Alpha 83:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "*"
                ]
            },
            "Action": [
                "s3:GetObject"
            ],
            "Resource": [
                "arn:aws:s3:::mastodon/*"
            ]
        }
    ]
}

```

**To Reproduce**
1. Set this custom policy in Alpha 83
2. Navigate through Mastodon
3. Upgrade to Alpha 84
4. See no media anymore

**Expected behavior**
Expect media to be displayed properly, here with Alpha 83 (blurred by me):

<img width="662" height="486" alt="Image" src="https://github.com/user-attachments/assets/ff330744-ed22-4e06-95e1-ade9f3a692fa" />

**Actual behavior with Alpha 84**

<img width="678" height="490" alt="Image" src="https://github.com/user-attachments/assets/fdd6589d-1980-4547-8f9b-9e9e3848d33d" />

**Additions**
The error-log shows nothing, all the logs are in the normal rustfs-log:

```bash
{"timestamp":"2026-03-01T21:12:46.347196405+01:00","level":"ERROR","fields":{"message":"op returns error","op":"GetObject","err":"S3Error(Inner { code: AccessDenied, message: Some(\"Access Denied\"), request_id: None, status_code: None, source: None, headers: None })"},"target":"s3s::ops","filename":"/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/s3s-0.13.0/src/ops/mod.rs","line_number":247,"threadName":"rustfs-worker","threadId":"ThreadId(25)"}
{"timestamp":"2026-03-01T21:12:46.34912322+01:00","level":"WARN","fields":{"message":"Failed to propagate tracing context: `SpanDisabled`"},"target":"rustfs::server::http","filename":"rustfs/src/server/http.rs","line_number":652,"threadName":"rustfs-worker","threadId":"ThreadId(39)"}
[...]

PR #2053, opened 2026-03-02
fix: restore default CORS fallback and STS object ACL ownership

overtrue:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [x] Refactor
- [ ] Other:

## Related Issues
- #2048
- #2047
- #2036

## Summary of Changes
This PR addresses two regressions introduced during recent refactoring around CORS/auth behavior.

1. Restored correct CORS decision flow for S3 paths:
- If bucket CORS exists, enforce bucket CORS result.
- If bucket CORS exists but does not match, return `403`.
- If bucket CORS does not exist, fall back to global/default CORS (default `*`).

2. Fixed STS object ACL ownership for `PutObject`:
- Object ACL owner is now derived from stable parent identity (`parent_user`) for temporary/service-account credentials.
- This avoids `HEAD/GET 403` caused by ACL owner being bound to rotating STS access keys.

3. Added/updated regression coverage for:
- CORS preflight/response behavior in S3 paths.
- STS ACL owner resolution logic.
- Telemetry env-filter test stability under environment overrides.

## Checklist
- [ ] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines

## Impact
- [x] Other impact:
  Existing objects written with transient STS owner IDs may still need ACL repair or object rewrite to normalize ownership metadata.

## Additional Notes
Validation executed locally:
- `make pre-commit`
- `DEPLOY_MODE=build S3_PORT=9000 S3_HOST=127.0.0.1 scripts/s3-tests/run.sh`

S3 test result in this environment:
- `313 passed, 516 deselected`


overtrue:
Addressed all review comments in commit `929c1741`.

What changed:
1. `rustfs/src/storage/ecfs_extend.rs`
- Added explicit contract docs for `apply_cors_headers`: `None` vs `Some(empty)` vs `Some(non-empty)`.
- Set `Vary: Origin` whenever `Access-Control-Allow-Origin` is reflected (non-`*`), not only wildcard+credentialed path.

2. `rustfs/src/storage/access.rs`
- Replaced per-request `Vec` allocation in `check_acl_access` with a fixed stack array (`[&str; 2]`) + length slice to avoid heap allocation on hot auth path.

3. `crates/obs/src/telemetry/filter.rs`
- Updated noisy-crate suppression decision to consider effective config (including `RUST_LOG`) instead of only `logger_level`.
- Added helper test coverage for `RUST_LOG=debug` behavior.

Validation:
- `make pre-commit` passed locally.

Issue #2041, opened 2026-03-02
InternalError: Io error: No KMS key -- v1.0.0-alpha.84+ fails PUT/GET (KMS error), v1.0.0-alpha.83 works fine


harshadkcon:
### Environment

- RustFS version (working): v1.0.0-alpha.83
- RustFS version (failing): v1.0.0-alpha.84+ (latest image at time of testing)
- Deployment: Docker
- Backend storage: Local filesystem
- Client: AWS CLI / Ansible (boto3 1.37.34 / botocore 1.37.34)

### Description

After upgrading from v1.0.0-alpha.83 to a newer image (alpha.84+), the following issues occur:

1. PUT operations fail with:

   InternalError: Io error: No KMS key available for managed server-side encryption

2. Existing objects are visible in the GUI (bucket listing works),
   but preview / download fails with:
   
   "cannot fetch"

Rolling back to v1.0.0-alpha.83 immediately restores:

- Successful PUT operations
- Successful GET / preview / download of existing objects

No configuration changes were made during upgrade.

### Observed Behaviour

- Bucket metadata loads correctly.
- Object listing works.
- Object data retrieval fails.
- PUT fails with KMS-related error even though KMS is not configured or used in v1.0.0-alpha.83.

### Expected Behaviour

- Upgrade should not introduce mandatory KMS dependency.
- If KMS is required, it should be:
  - Optional
  - Explicitly configurable
  - Clearly documented
  - Backward compatible

### Impact

This regression:

- Breaks object uploads
- Breaks reads of existing data
- Makes upgrade unsafe without full staging validation

### Questions

1. Was managed server-side encryption made mandatory in newer versions?
2. Is KMS configuration now required by default?
3. Is there a migration step required between alpha.83 and alpha.84+?

Happy to provide logs if needed.

reatang:
We apologize for any instability in certain features due to extensive refactoring in this version, I will reproduce the problem and handle it carefully.

loverustfs:
Use Alpha.85 https://github.com/rustfs/rustfs/releases/tag/1.0.0-alpha.85

devyn:
I'm still having this issue with 1.0.0-alpha.85 despite not using KMS 

```
[...]

PR #2054, opened 2026-03-02
fix(sse): allow PUT/GET without KMS when no SSE or bucket default 

reatang:
## Type of Change

- [x] Bug Fix

## Related Issues

#2041


rustfs/src/storage/sse.rs
```diff
@@ -1028,9 +1033,11 @@ async fn apply_managed_encryption_material(
         }
     }
 
-    let kms_key_to_use = kms_key_candidate
-        .clone()
-        .ok_or_else(|| ApiError::from(StorageError::other("No KMS key available for managed server-side encryption")))?;
+    let kms_key_to_use = kms_key_candidate.clone().ok_or_else(|| {
```
87081420rjgg-code, line 1036: 🖕

87081420rjgg-code (commented):
🖕

PR #2072, opened 2026-03-03
fix(heal):heal failed replication via must_replicate instead of check replicate_delete

### LINKED ISSUE #1683: RustFS Active-Active Replication Delete Object Issue
## Problem Description

In RustFS active-active replication, when an object is deleted from the main site, the deletion does not behave correctly. The object is not removed from either the main site or the DR site initially. After several retry attempts, the object may eventually be deleted from the main site, but it incorrectly persists on the DR site.

## Expected Behavior
- Object deleted from main site should be removed from the main site immediately
- Deletion should replicate to the DR site
- Object should be removed from all sites in the cluster

## Actual Behavior
- Object is not deleted from the main site where the deletion was initiated
- Object is not deleted from the DR site either
- After multiple retry attempts, the object may eventually be deleted from the main site
- The deleted object incorrectly persists on the DR (replica) site instead of being removed

### LINKED ISSUE #2071: Bucket level replication recovery when destination is not available
**Describe the bug**
There is no recovery and retry for failed replication events, such as caused by network issue or destination server availability.

**To Reproduce**
1. Setup bucket level replication between RustFS A and B.
2. Verify that replication works
3. Shutdown RustFS B instance.
4. Add `file_while_rustfs_b_is_offline` to RustFS A instance.
5. Startup RustFS  B instance.
6. Add `file_after_recovery` to RustFS A instance
7. Verify that on RustFS B, `file_after_recovery`  is present but not `file_while_rustfs_b_is_offline`

`A` has 3 files:

Image

`B` has 2 files:

Image

**Expected behaviour**

`file_while_rustfs_b_is_offline` should be retried and be replicated.

**Additional context**

Version: `v1.0.0-alpha.85`

LeonWang0735:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues

FIX #2071 
FIX #1683 
## Summary of Changes

heal failed replication via must_replicate instead of check replicate_delete
- Before: The condition was oi.delete_marker || !oi.replication_status.is_empty(). For a normal object that failed replication, replication_status is Failed (non-empty), so the code always called check_replicate_delete, which is for delete markers/version deletes. That produced a dsc that did not replicate_any(), so the scanner never re-queued the object for replication heal.
- After: The condition is oi.delete_marker || (!oi.replication_status.is_empty() && oi.replication_status != ReplicationStatusType::Failed). Objects with Failed status now take the else branch and use must_replicate, so heal correctly gets replication targets and the scanner can re-queue them for retry.
- Result: Failed replication events (e.g. due to destination being offline) are correctly healed by the background scanner instead of being skipped forever.

trigger delete replication for Pending delete marker in single delete
- Before: When a user deletes an object without a version id (creating a delete marker) on the source in active-active replication, the storage layer sets the new delete marker’s replication state to Pending via set_delete_replication_state. Previously only Replica or version_purge_status == Pending triggered replication, so the delete marker was not replicated until the scanner ran.
- After: Add || obj_info.replication_status == ReplicationStatusType::Pending to the existing condition so that delete markers with Pending replication status are replicated immediately, consistent with batch delete (execute_delete_objects) behavior.
- Result: Delete markers created on the source (no version id) are now replicated to the target in real time instead of only after the next scan.
## Checklist

### DIFF
--- a/crates/ecstore/src/data_usage.rs
+++ b/crates/ecstore/src/data_usage.rs
@@ -15,12 +15,8 @@
 pub mod local_snapshot;
 
 use crate::{
-    bucket::metadata_sys::get_replication_config,
-    config::com::read_config,
-    disk::DiskAPI,
-    error::Error,
-    store::ECStore,
-    store_api::{BucketOperations, ListOperations},
+    bucket::metadata_sys::get_replication_config, config::com::read_config, disk::DiskAPI, error::Error, store::ECStore,
+    store_api::ListOperations,
 };
 pub use local_snapshot::{
     DATA_USAGE_DIR, DATA_USAGE_STATE_DIR, LOCAL_USAGE_SNAPSHOT_VERSION, LocalUsageSnapshot, LocalUsageSnapshotMeta,
@@ -38,7 +34,7 @@ use std::{
 };
 use tokio::fs;
 use tokio::sync::RwLock;
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, info};
 
 // Data usage storage constants
 pub const DATA_USAGE_ROOT: &str = SLASH_SEPARATOR;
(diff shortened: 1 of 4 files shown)


LeonWang0735:
@codex review

PR #2112, opened 2026-03-10
fix: resolve silent failure in MQTT bucket event notifications

### LINKED ISSUE #2111: Bucket event notifications (MQTT) silently fail -- events never delivered
## Summary

When configuring RustFS with MQTT-based bucket event notifications (via `RUSTFS_NOTIFY_MQTT_*` environment variables), object upload events are silently broken in RustFS. No events arrive at the MQTT broker, and no errors are surfaced to the user.

## Steps to reproduce

1. Start RustFS with MQTT notify env vars set:
   ```
   RUSTFS_NOTIFY_MQTT_ENABLE_primary=on
   RUSTFS_NOTIFY_MQTT_BROKER_primary=tcp://<broker>:1883
   RUSTFS_NOTIFY_MQTT_TOPIC_primary=rustfs/events
   RUSTFS_NOTIFY_MQTT_QOS_primary=1
   ```
2. Create a bucket and register a bucket event rule (e.g. via `mc event add` or the S3 `PutBucketNotificationConfiguration` API)
3. Upload an object to the bucket
4. Observe: no MQTT message is published, no error logged

## Root cause

Two bugs in the notification initialisation path combine to prevent events from being delivered:

1. **`rustfs/src/server/event.rs` -- spurious early-return guard**
   `init_event_notifier()` contains a guard that checks whether the MQTT *and* webhook subsystems are present in config before calling `initialize()`. The condition uses `||` (OR), so configuring only MQTT (without webhook, which is the normal case) always triggers the early return and `initialize()` is never called. The guard is also redundant -- `initialize()` handles missing config gracefully on its own.

2. **`crates/notify/src/registry.rs` -- hard error on object store timing**
   During target initialisation, the registry attempts to persist config to the object store. If the store isn't available at that exact moment it returns a hard `Err(ServerNotInitialized)`, which prevents the global `NOTIFICATION_SYSTEM` from being set. Even if targets were created successfully in memory, they are discarded and no events can be delivered.

There is also a third environmental issue: the default MQTT queue directory (`/opt/rustfs/events`) does not exist inside a standard container, causing `MQTTTarget::new()` to fail silently. Setting `RUSTFS_NOTIFY_MQTT_QUEUE_DIR_primary=` (empty string) disables the disk queue and avoids this.

## Fix

- Remove the guard in `server/event.rs` -- call `initialize()` unconditionally
- Downgrade the `None` object-store case in `registry.rs` to a warning and return `Ok(successful_targets)` -- targets active in memory are sufficient for event delivery
- Document that `RUSTFS_NOTIFY_MQTT_QUEUE_DIR_*` should be set to empty unless a writable queue directory is explicitly provisioned

A PR with these changes will follow.

simon-escapecode:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
Closes #2111

## Summary of Changes

Bucket event notifications (MQTT and webhook) silently fail because two bugs in the notification initialisation path prevent events from being delivered.

**Fix 1 -- Remove spurious guard in `rustfs/src/server/event.rs`**

`init_event_notifier()` contains a guard that checks whether the MQTT *and* webhook subsystems are present in config before calling `initialize()`. The condition uses `||` (OR), so configuring only MQTT (the normal case) always triggers an early return and `initialize()` is never called. The guard is redundant because `initialize()` handles missing config gracefully on its own.

**Fix 2 -- Soften hard error in `crates/notify/src/registry.rs`**

During target initialisation, the registry attempts to persist config to the object store. If the store is not available at that exact moment it returns a hard `Err(ServerNotInitialized)`, which prevents the global `NOTIFICATION_SYSTEM` from being set. Even if targets were created successfully in memory, they are discarded. This change downgrades the `None` case to a warning and returns `Ok(successful_targets)`.

### Verification

```bash
# Start RustFS with MQTT notify env vars
RUSTFS_NOTIFY_MQTT_ENABLE_primary=on
RUSTFS_NOTIFY_MQTT_BROKER_primary=tcp://<broker>:1883
RUSTFS_NOTIFY_MQTT_TOPIC_primary=rustfs/events
RUSTFS_NOTIFY_MQTT_QOS_primary=1
RUSTFS_NOTIFY_MQTT_QUEUE_DIR_primary=

# Register bucket event rule and upload an object
# Verify MQTT message is received at the broker
```

## Checklist

## Impact
- [ ] Other impact:

## Additional Notes

There is also a third environmental issue worth documenting: the default MQTT queue directory (`/opt/rustfs/events`) does not exist inside a standard container, causing `MQTTTarget::new()` to fail silently. Setting `RUSTFS_NOTIFY_MQTT_QUEUE_DIR_primary=` (empty string) disables the disk queue and avoids this.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6544,9 +6544,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.13"
+version = "0.11.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1906b49b0c3bc04b5fe5d86a77925ae6524a19b816ae38ce1e426255f1d8a31"
+checksum = "434b42fec591c96ef50e21e886936e66d3cc3f737104fdb9b737c40ffb94c098"
 dependencies = [
  "aws-lc-rs",
  "bytes",


rustfs/src/server/event.rs
```diff
@@ -58,31 +57,7 @@ pub(crate) async fn init_event_notifier() {
         }
     };
 
-    info!(
```
houseme, line 61: ？

simon-escapecode: The guard uses `||` (OR), so if either MQTT or webhook is absent from config, it returns early and `initialize()` is never called. Since most deployments only configure one subsystem, this effectively blocks all notifications. The guard is also unnecessary -- `initialize()` already handles missing subsystems gracefully internally -- so the whole block needs to be removed.

houseme (approved):

rustfs/src/server/event.rs
simon-escapecode: Good catch -- the info log should not have been removed. Restored it in the latest push. Only the guard block is removed now, the log line is kept.

crates/notify/src/registry.rs
```diff
@@ -293,10 +293,17 @@ impl TargetRegistry {
                 }
             }
 
-            let Some(store) = rustfs_ecstore::global::new_object_layer_fn() else {
-                return Err(TargetError::ServerNotInitialized(
-                    "Failed to save target configuration: server storage not initialized".to_string(),
-                ));
+            let store = match rustfs_ecstore::global::new_object_layer_fn() {
```
houseme, line 296: Thanks for your contribution, if the 'store' fetch fails, and other operations related to 'store' also fail, does the subsequent operation make sense?

simon-escapecode: Apologies for only just noticing this comment.

Yes - the targets are already created and initialised by this point (lines 234-247). The object store is only used here to persist the notification configuration so it survives a restart. Event delivery itself (MQTT publish, webhook POST etc.) does not depend on the store. Downgrading to a warning means targets still work in memory for the current process lifetime; only config persistence is deferred.

PR #2125, opened 2026-03-10
fix(server): init event notifier when partial notify configured

overtrue:
## Type of Change
- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
- https://github.com/rustfs/backlog/issues/585

## Summary of Changes
- Fix event notifier bootstrap condition to initialize when either MQTT or webhook is configured.
- Avoid incorrectly skipping notifier initialization when one of the two subsystems is absent.

## Checklist
- [ ] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines

## Impact
- [ ] Other impact: startup behavior for partial notify configs is corrected.

## Additional Notes
- Added tests are covered in a dedicated `notify` path where this initialization condition is exercised.

### DIFF
--- a/rustfs/src/server/event.rs
+++ b/rustfs/src/server/event.rs
@@ -62,14 +62,15 @@ pub(crate) async fn init_event_notifier() {
         target: "rustfs::main::init_event_notifier",
         "Global server configuration loaded successfully"
     );
-    // 2. Check if the notify subsystem exists in the configuration, and skip initialization if it doesn't
-    if server_config
+    // 2. Check if at least one notify subsystem exists in the configuration, and skip only when neither is present.
+    let mqtt_configured = server_config
         .get_value(rustfs_config::notify::NOTIFY_MQTT_SUB_SYS, DEFAULT_DELIMITER)
-        .is_none()
-        || server_config
-            .get_value(rustfs_config::notify::NOTIFY_WEBHOOK_SUB_SYS, DEFAULT_DELIMITER)
-            .is_none()
-    {
+        .is_some();
+    let webhook_configured = server_config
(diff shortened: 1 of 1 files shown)


simon-escapecode:
This fix is incomplete and introduces a new failure mode.

The original guard used `||` (if *either* subsystem is missing, bail out), which was clearly wrong. This PR changed it to `&&` (if *both* are missing, bail out), which fixes the single-subsystem case but the guard itself is still unnecessary and fragile:

1. **`initialize()` already handles missing subsystems gracefully.** `create_targets_from_config()` iterates all registered factories. For each factory, it resolves config via `config.0.get(&section_name).unwrap_or_default()` — if no config exists for a target type, the instance ID set is empty and zero tasks are created. The loop simply moves to the next factory. No error, no early return.

2. **The guard will break again when new target types are added.** Kafka, AMQP, Redis, and other notification targets are all supported by the registry. Any deployment using *only* one of those (without MQTT or webhook) will hit the same early return and silently fail — exactly the same class of bug as the original `||` issue.

3. **The `DEFAULT_DELIMITER` import is only needed for this guard.** Removing the guard also removes the import, reducing coupling.

PR #2112 removes the guard entirely, which is the correct fix. It also addresses a second failure point in the notification registry (`ServerNotInitialized` hard error downgraded to a warning when the object store is unavailable at init time) that this PR does not cover.

simon-escapecode:
The complete fix is in PR #2112 (now merged): https://github.com/rustfs/rustfs/pull/2112/files

Issue #1874, opened 2026-02-19
Anonymous access via bucket policy for specific objects — is it supported?


metoant:
## Environment
- RustFS version: rustfs/rustfs:latest
- Client: AWS SDK for Go v2

## Question
I'm trying to allow anonymous public access to specific objects in a bucket 
while keeping others private, without using presigned URLs.

I've tried two approaches and neither works:

### Approach 1 — Canned ACL
Setting `ACL: types.ObjectCannedACLPublicRead` during `PutObject`.
Upload succeeds but anonymous GET returns 403.

### Approach 2 — Tag-based bucket policy
Setting `input.Tagging = "public-read=true"` during `PutObject`, with this policy:

{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {"AWS": "*"},
      "Action": ["s3:GetObject"],
      "Resource": ["arn:aws:s3:::oblax/*"],
      "Condition": {
        "StringEquals": {
          "s3:ExistingObjectTag/public-read": "true"
        }
      }
    }
  ]
}

Anonymous GET still returns 403 AccessDenied.

## What does work
A bucket policy with no conditions allows anonymous access to everything — 
but I need per-object control.

## Questions
1. Are object-level canned ACLs (`public-read`) supported?
2. Are tag-based policy conditions (`s3:ExistingObjectTag`) supported?

GatewayJ:
The issue has been confirmed; coding will expedite the resolution.Both methods will be supported.Thank you for your feedback. I will keep you informed as soon as there is progress.

metoant:
thank you.

loverustfs:
Fixed.

Theoooooo:
I can confirm this issue not fixed
I'm using the v86 alpha version and i still can't use the condition as stated above

<img width="337" height="87" alt="Image" src="https://github.com/user-attachments/assets/4daf24e1-5bb4-415c-808a-19ed33bf5a84" />

<img width="384" height="71" alt="Image" src="https://github.com/user-attachments/assets/a632949a-3c28-4012-837a-d74c69febd87" />

I'll test the v90 later today and see if it was shipped after the v86 version

GatewayJ:
> I can confirm this issue not fixed I'm using the v86 alpha version and i still can't use the condition as stated above
[...]

PR #2315, opened 2026-03-28
feat(iam,admin): prepared IAM auth, ExistingObjectTag, admin permission checks

GatewayJ:
## Type of Change
- [x] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [x] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
Related to #1874 (fixes anonymous and authenticated access when bucket or IAM policies use `s3:ExistingObjectTag` / `ExistingObjectTag/...` conditions; object-level canned ACL behavior is unchanged and out of scope for this PR).

## Summary of Changes
This PR implements **prepared IAM authorization** and **object-tag-aware policy evaluation** so conditions that depend on existing object tags are evaluated with the correct tag map.

- **`crates/iam`**: Add `prepare_auth` / `eval_prepared` (`PreparedIamAuth`) so merged identity policies and STS/session policies are prepared once, `s3:ExistingObjectTag` usage is detected from policy documents, and evaluation reuses that context. Replace the old STS entry path with `prepare_sts_auth` + `eval_prepared`; service accounts use `prepare_service_account_auth` + `eval_prepared`. `is_allowed` delegates to this pipeline.
- **`crates/policy`**: Add helpers to detect `ExistingObjectTag` / `s3:ExistingObjectTag` condition keys in IAM and bucket policies (including structured checks so arbitrary substrings in values do not spuriously trigger tag loads). Add regression tests for condition-key detection.
- **`rustfs/src/storage/access.rs`**: In `authorize_request`, merge object tags into condition values when IAM preparation or the bucket policy indicates they are needed; cache tags per request `(bucket, object, version_id)`; parse bucket policy JSON when possible to decide whether tag conditions apply (malformed JSON with the marker still conservatively loads tags). Remove per-handler prefetch of tag conditions in favor of this centralized path.
- **`crates/e2e_test`**: Add STS `AssumeRole` and bucket/IAM policy E2E coverage for `ExistingObjectTag`; extend common STS helpers and teardown.
- **Admin**: Document why canned policies with an empty `Version` are rejected (`rustfs/src/admin/handlers/policies.rs`); minor `rustfs/src/admin/auth.rs` doc cleanup. Policy-crate tests support admin permission semantics.

## Checklist

## Impact
- [ ] Other impact:

N/A — no intentional API or configuration breaking changes. Authorization may allow previously denied requests when policies reference existing object tags and tags were not loaded before.

## Additional Notes
**Verification (local):**

```bash
make pre-commit
cargo test -p rustfs-policy
cargo test -p rustfs-iam
cargo test -p rustfs --lib -- object_tag
# Optional: e2e (requires awscurl / environment per e2e_test)
cargo test -p e2e_test -- existing_object_tag
```

**Documentation:** N/A for end-user docs; only an explanatory comment was added in the admin canned-policy handler.

**Scope note:** Issue #1874 also asked about object canned ACL `public-read` for anonymous access. This PR does not change canned ACL handling; it addresses tag-based bucket and IAM policies using `s3:ExistingObjectTag`.

---

### DIFF
--- a/rustfs/src/storage/access.rs
+++ b/rustfs/src/storage/access.rs
@@ -25,7 +25,10 @@ use rustfs_ecstore::new_object_layer_fn;
 use rustfs_ecstore::store_api::BucketOperations;
 use rustfs_iam::error::Error as IamError;
 use rustfs_policy::policy::action::{Action, S3Action};
-use rustfs_policy::policy::{Args, BucketPolicy, BucketPolicyArgs, bucket_policy_uses_existing_object_tag_conditions};
+use rustfs_policy::policy::{
+    Args, BucketPolicy, BucketPolicyArgs, bucket_policy_needs_existing_object_tag_for_args,
+    bucket_policy_uses_existing_object_tag_conditions,
+};
 use rustfs_utils::http::AMZ_OBJECT_LOCK_BYPASS_GOVERNANCE;
 use s3s::access::{S3Access, S3AccessContext};
 use s3s::{S3Error, S3ErrorCode, S3Request, S3Result, dto::*, s3_error};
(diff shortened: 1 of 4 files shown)


rustfs/src/storage/access.rs
```diff
@@ -134,102 +180,131 @@ pub async fn authorize_request<T>(req: &mut S3Request<T>, action: Action) -> S3R
...
+            object: object.as_str(),
+            claims,
+            deny_only: false,
+        };
+        let prepared = iam_store.prepare_auth(&action_args).await;
+        let needs_tag_from_iam = prepared.needs_existing_object_tag;
+
+        let needs_tag_from_bucket =
+            !bucket.is_empty() && !object.is_empty() && bucket_policy_uses_existing_object_tag(&bucket).await;
+        let needs_tag = needs_tag_from_iam || needs_tag_from_bucket;
```
overtrue: Because `prepared.needs_existing_object_tag` is derived from the whole merged IAM/session policy, `needs_tag` becomes true even when the current `(action, bucket, object)` cannot match any `ExistingObjectTag` condition. That means an unrelated tag-conditioned statement elsewhere in the policy set now forces `get_object_tags()` on every object request, and any transient failure in that read is turned into `AccessDenied` below. Can we narrow this hint to statements relevant to the current request (or defer the fail-closed fetch until evaluation actually needs an `ExistingObjectTag/*` key)?

GatewayJ: @overtrue  tag loads now use policy_needs_existing_object_tag_for_args / bucket_policy_needs_existing_object_tag_for_args, so we only fetch when a statement with ExistingObjectTag conditions actually reaches evaluation for the current request args (wired through PreparedIamAuth and authorize_request). That should avoid unrelated statements forcing get_object_tags and broad AccessDenied on transient read failures.

PR #2358, opened 2026-03-31
fix(admin): percent-decode group name in DELETE /v3/group/{group}

### LINKED ISSUE #2343: Failed to delete the user group created by the console
**Describe the bug**
1、Failed to delete the user group created by the console in version 90
2. If a Chinese name is used when creating a key in the console, the creation will fail; it will succeed if the name is entirely in English. In previous old versions, creating a key with a Chinese name was allowed.

Image

GatewayJ:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
Closes #2343

## Summary of Changes
`DELETE /rustfs/admin/v3/group/{group}` received the `{group}` path segment as percent-encoded UTF-8 (e.g. `%E5%93%88...`) from the raw URI path / matchit, while IAM and `GET /v3/group?group=...` use decoded names. Deleting groups with non-ASCII names therefore failed with `NoSuchKey` / `NoSuchGroup`. The handler now percent-decodes the path parameter (same approach as tier admin routes) before calling IAM and site-replication hooks.

## Checklist

## Impact
- [ ] Other impact:

## Additional Notes
**Verification:** `cargo check -p rustfs`

**Manual check:** Create a user group with a non-ASCII name in the console, then delete it; expect `200` instead of `404` with `NoSuchKey`.

`make pre-commit` was not run in this session (per contributor request); CI should cover fmt/clippy/tests.


GatewayJ:

PR #2393, opened 2026-04-04
fix: skip missing groups in policy_db_get instead of aborting

### LINKED ISSUE #2392: fix: policy_db_get aborts on first missing OIDC group, discarding all resolved policies
### Summary

`policy_db_get` in `crates/iam/src/manager.rs` uses the `?` operator when iterating over OIDC groups, causing the entire policy resolution to abort on the first non-existent group. Any policies already resolved from earlier groups in the list are discarded.

### Environment

- RustFS version: 1.0.0-alpha.90
- Deployment: 4-node distributed cluster with OIDC via Dex (backed by Entra ID)

### Steps to Reproduce

1. Configure OIDC with `RUSTFS_IDENTITY_OPENID_GROUPS_CLAIM=groups`
2. Create IAM groups for 2 of the user's Entra ID groups (e.g., `consoleAdmin` and `readwrite`)
3. The user's JWT contains 19 Entra ID group UUIDs, but only 2 have corresponding IAM groups
4. Login via SSO

### Expected Behavior

The 2 matching groups should resolve to `consoleAdmin` and `readwrite` policies. The 17 non-existent groups should be silently skipped.

### Actual Behavior

`policy_db_get` iterates over all 19 groups. When it encounters the first non-existent group (alphabetically before the 2 valid ones), `policy_db_get_internal` returns `Err(NoSuchGroup)`. The `?` operator on line 758 propagates this error, causing `policy_db_get` to return `Err` — discarding any policies already accumulated.

Back in `prepare_sts_auth` (line 975), `.unwrap_or_default()` converts this to an empty `Vec<String>`. The fallback to the JWT `policy` claim then runs, but the claim contains group UUIDs (not policy names like `consoleAdmin`), so they don't match any `DEFAULT_POLICIES`. Result: user gets **DENY** with an empty policy.

### Root Cause

`crates/iam/src/manager.rs` line 758:

```rust
let (gp, _) = self.policy_db_get_internal(group, true, present).await?;
```

The `?` should be replaced with a `match` or `unwrap_or_default()` that skips missing groups:

```rust
let (gp, _) = match self.policy_db_get_internal(group, true, present).await {
    Ok(result) => result,
    Err(_) => continue,
};
```

### Workaround

Create IAM groups (without policies) for **all** Entra ID groups that may appear in the JWT. This prevents the `NoSuchGroup` error. Only the groups that should have access need a policy attached.

kharkevich:
## Summary

- Fix `policy_db_get` in `crates/iam/src/manager.rs` to skip non-existent IAM groups instead of aborting the entire policy resolution
- Only skip `NoSuchGroup` errors using the existing `is_err_no_such_group()` helper; propagate all other errors (storage, backend)
- Add regression test covering mixed existing/non-existent groups

## Problem

When iterating over OIDC groups from the JWT in `policy_db_get`, calling `policy_db_get_internal` for a non-existent IAM group returns `Err(NoSuchGroup)`. The `?` operator on line 758 propagates this error, causing the function to abort and discard any policies already resolved from earlier groups.

In OIDC setups where the JWT contains many groups (e.g., 19 Entra ID groups via Dex) but only a subset have corresponding IAM groups with policies attached, the first missing group aborts resolution. The caller (`prepare_sts_auth`) then falls back to the JWT `policy` claim, which contains group UUIDs that don't match any `DEFAULT_POLICIES` — resulting in a full DENY.

## Fix

Replace the `?` with a `match` that `continue`s only on `NoSuchGroup`, allowing the loop to skip missing groups and accumulate policies from groups that do exist. Other errors (storage failures, etc.) are still propagated.

```rust
// Before
let (gp, _) = self.policy_db_get_internal(group, true, present).await?;

// After
let (gp, _) = match self.policy_db_get_internal(group, true, present).await {
    Ok(result) => result,
    Err(err) if is_err_no_such_group(&err) => continue,
    Err(err) => return Err(err),
};
```

## Test plan

- [x] `cargo test -p rustfs-iam` — all 106 tests pass
- [x] Regression test `test_policy_db_get_skips_nonexistent_groups` added
- [x] Deployed to 4-node cluster with 19 Entra ID groups, only 2 mapped — verified correct policy resolution

Fixes #2392

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -410,7 +410,7 @@ dependencies = [
  "arrow-schema",
  "chrono",
  "half",
- "indexmap 2.13.0",
+ "indexmap 2.13.1",
  "itoa",
  "lexical-core",
  "memchr",
@@ -687,9 +687,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-sys"
-version = "0.39.0"
+version = "0.39.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fa7e52a4c5c547c741610a2c6f123f3881e409b714cd27e6798ef020c514f0a"
+checksum = "83a25cf98105baa966497416dbd42565ce3a8cf8dbfd59803ec9ad46f3126399"
 dependencies = [
  "cc",
  "cmake",
@@ -1227,16 +1227,16 @@ dependencies = [
 
 [[package]]
 name = "blake3"
-version = "1.8.3"
+version = "1.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2468ef7d57b3fb7e16b576e8377cdbde2320c60e1491e961d11da40fc4f02a2d"
+checksum = "4d2d5991425dfd0785aed03aedcf0b321d61975c9b5b3689c774a2610ae0b51e"
 dependencies = [
  "arrayref",
  "arrayvec",
  "cc",
  "cfg-if",
  "constant_time_eq",
- "cpufeatures 0.2.17",
+ "cpufeatures 0.3.0",
 ]
 
 [[package]]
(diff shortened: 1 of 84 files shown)


kharkevich:

kharkevich:
@GatewayJ could you please run CI again, right now all should be green

Issue #2366, opened 2026-04-02
Access Keys keep disappearing (JWT issue?)


rsxdalv:
**Describe the bug**
Access Keys / Service accounts disappear (get revoked/become unencryptable) from the UI.

**To Reproduce**
1. mc admin user svcacct add {....}
2. Check UI to see the accounts
3. 20-40 minutes later they are gone.

**Expected behavior**
1. Service accounts do not randomly disappear from UI.
2. They obey the expiration time.
3. They show a more visible error/hint in more accessible logs.

**Screenshots**
<img width="1338" height="679" alt="Image" src="https://github.com/user-attachments/assets/3c0154d1-c4fc-4f25-af4d-07c001c29dcd" />

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Brave

**Additional context**
According to Claude: 
> Root cause (confirmed): RustFS alpha rotates its internal JWT sealing key every ~20 minutes. Service accounts (svcacct) are stored as signed JWTs — they become unreadable after each rotation. The KMS key we set up last session only applies to object encryption, not to this IAM JWT mechanism. There's no config flag to disable the rotation.

More AI diagnostics in regards to KMS key not preventing a extract_jwt failure:
> The errors show extract_jwt_claims failed timestamps starting at 2026-04-02T00:20:18 - that's about 9 hours after RustFS started at 2026-04-01T15:14:18. The KMS key is still present at /run/rustfs-kms/iam-master-key.

> The denki credentials (9FW0RQSIKI7A500A04AQ and JJZY30AIQ5Y62CRB7CET) started failing at 00:20:18 - they were created around 23:55 or so probably, but no... wait. Looking at the context, setup-s3-bucket.sh was run and created those credentials.

> So the credentials appeared in the extract_jwt failures about 5 hours after RustFS started, and then pe-reader and pe-upload started failing at 00:30 with the promptecho accounts that had been working during that 26-minute window. This suggests the KMS approach isn't actually preventing JWT key rotation - the key file exists at the right path, so RustFS is loading it, but it's still rotating its internal sealing key anyway. Let me look more closely at when those extract_jwt failures first appeared.

[...]

PR #2410, opened 2026-04-07
fix(iam): keep service account JWT expiry consistent

weisd:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [x] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
#2366

## Summary of Changes
- Stop adding a default JWT `exp` when a service account is created without an explicit expiration.
- Keep the persisted service account expiration and the JWT `exp` claim in sync when the expiration is updated.
- Allow non-expiring service accounts to load and authenticate without requiring an `exp` claim, while keeping stricter expiration handling for expiring tokens.
- Add regression tests for service account creation and expiration updates.

## Checklist

## Impact
- [ ] Other impact:

## Additional Notes
Verification:
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- `make pre-commit`

### DIFF
--- a/crates/iam/src/manager.rs
+++ b/crates/iam/src/manager.rs
@@ -687,6 +687,8 @@ where
             cr.description = opts.description;
         }
 
+        let token_without_expiration = cr.expiration.is_none();
+
         if opts.expiration.is_some() {
             // TODO: check expiration
             cr.expiration = opts.expiration;
@@ -702,7 +704,7 @@ where
             }
         }
 
-        let mut m: HashMap<String, Value> = if cr.expiration.is_none() {
+        let mut m: HashMap<String, Value> = if token_without_expiration {
             get_claims_from_token_with_secret_allow_missing_exp(&cr.session_token, &current_secret_key)?
         } else {
             get_claims_from_token_with_secret(&cr.session_token, &current_secret_key)?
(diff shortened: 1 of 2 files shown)


robertoberto:
The fix looks good but there's no regression test. If someone refactors the JWT path later, this could silently reappear. I'd like to open a follow-up PR adding unit tests in crates/iam/. Two cases: (1) new_service_account with expiration: None produces a JWT that extract_jwt_claims accepts (no exp required), and (2) a JWT with a past exp is correctly rejected. Both are pure unit tests, no sleep needed. They'd run in the existing make test step. Would that be welcome? @overtrue

overtrue:
> The fix looks good but there's no regression test. If someone refactors the JWT path later, this could silently reappear. I'd like to open a follow-up PR adding unit tests in crates/iam/. Two cases: (1) new_service_account with expiration: None produces a JWT that extract_jwt_claims accepts (no exp required), and (2) a JWT with a past exp is correctly rejected. Both are pure unit tests, no sleep needed. They'd run in the existing make test step. Would that be welcome? @overtrue

Thank you — you’re always welcome to build this with us. fell free to make a PR.

PR #2448, opened 2026-04-09
fix(server): warn on default credentials with console enabled

marshawcoco:
## Summary
- move the default-credentials warning out of the non-console startup branch
- emit the warning whenever RustFS starts with the built-in access key and secret key
- keep startup logging behavior unchanged otherwise

## Test plan
- [x] Run `cargo check -p rustfs --manifest-path C:/Users/guoziyu/rustfs/Cargo.toml`


rustfs/src/server/http.rs
```diff
@@ -210,19 +210,20 @@ pub async fn start_http_server(
...
-                "Detected default credentials '{}:{}', we recommend that you change these values with 'RUSTFS_ACCESS_KEY' and 'RUSTFS_SECRET_KEY' environment variables",
-                rustfs_credentials::DEFAULT_ACCESS_KEY,
-                rustfs_credentials::DEFAULT_SECRET_KEY
-            );
-        }
         info!(target: "rustfs::main::startup","For more information, visit https://rustfs.com/docs/");
         info!(target: "rustfs::main::startup", "To enable the console, restart the server with --console-enable and a valid --console-address.");
     }
 
+    if rustfs_credentials::DEFAULT_ACCESS_KEY.eq(&config.access_key)
```
houseme: Does this warning message need to be printed again?

marshawcoco:

rustfs/src/server/http.rs
marshawcoco: No, it is still printed only once. Previously this warning was only emitted in the `console_enable == false` branch. I moved it out of that branch so it is also emitted when the console is enabled, but it is not duplicated.

houseme: In `main.rs`, `start_http_server` will run twice. Originally, it only printed when running on port `:9000`, but now it prints on both `:9000` and `:9001`.

marshawcoco: You're right. `start_http_server()` runs once for the S3/API listener and once for the console listener, so keeping the warning there caused it to be logged twice when the console was enabled. I've moved the default-credentials warning to the main startup path in `main.rs` so it is emitted once per process start, and added a small regression test for the credential check.

houseme:
Please do `cargo fmt --all` for formatting, thanks!

PR #2493, opened 2026-04-11
fix(storage): fix critical bug in range calculation

zeapo:
Hi,

While using rustfs locally as an s3 storage for my iceberg warehouse, I noticed that sometimes my requests using duckdb failed. Duckdb and other tools read the last 8 bytes of parquet files, as this format stores a metadata footer (end of the file). 

While debugging the issue, I noticed that doing a Range: -8 http returns no data; a 206 Partial Content response with correct headers but still an empty body.

For example, requesting the last 8 bytes of a 3,095,910-byte file:
```
  Request:  Range: bytes=-8

  Response: HTTP/1.1 206 Partial Content
            content-length: 8
            content-range: bytes 3095902-3095909/3095910
            accept-ranges: bytes

            <empty — 0 bytes transferred>
```

The Content-Range and Content-Length headers were computed correctly (the start offset math was right), but the response body was never written. Absolute ranges worked fine for the same file:

```
  Request:  Range: bytes=0-7    → 206, 8 bytes body 
  Request:  Range: bytes=-8     → 206, 0 bytes body 
```

The file size has to be larger than 3MB (at least with smaller files i was unable to reproduce).
 

## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
None

## Summary of Changes

The bug was when calling `create_bitrot_reader`, the caller passed the absolute end offset `till_offset` rather than the length as expected by the function. 

## Checklist
- [-] Documentation updated (if needed)
- [-] CI/CD passed (if applicable)

## Impact
- [ ] Other impact:

## Additional Notes
None

---

### DIFF
--- a/crates/e2e_test/src/range_request_test.rs
+++ b/crates/e2e_test/src/range_request_test.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! End-to-end regression test for invalid GET object ranges.
+//! End-to-end regression tests for invalid and suffix GET object ranges.
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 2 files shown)


zeapo:

houseme (approved):

PR #2499, opened 2026-04-12
fix(s3): preserve equals in signed object paths

### LINKED ISSUE #2477: S3 SignatureDoesNotMatch (403) instead of NoSuchKey (404) for non-existent objects with '=' in path
## Description

When requesting a non-existent object whose key ends with `=` (e.g., `sitemap.xmlage=`), RustFS returns `403 SignatureDoesNotMatch` instead of the expected `404 NoSuchKey`.

The AWS SigV4 signature is correctly computed — the same canonical URI is used for both signing and the HTTP request. The `=` character is valid in S3 object keys and in URI paths per RFC 3986 (it is a `sub-delims` character).

## Steps to Reproduce

1. Sign a `GET` request for an object key containing `=` at the end, e.g.:
   ```
   GET /bucket/path/sitemap.xmlage= HTTP/1.1
   ```
2. Use standard AWS Signature Version 4 signing (path-style, `PercentEncodingMode::Single`, `UriPathNormalizationMode::Disabled`).
3. Send the request to RustFS.

**Expected:** `404 NoSuchKey` (the object does not exist)
**Actual:** `403 SignatureDoesNotMatch`

## Analysis

It appears that RustFS's signature verification normalizes or reinterprets the `=` character in the object key path differently from the canonical request computation. This causes a mismatch between the signature the client computed and the signature RustFS expects, even though both should be operating on the same URI.

Per the [AWS S3 Signature Version 4 specification](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html), the canonical URI should use the request path as-is (with appropriate percent-encoding), and `=` does not require encoding in the path component.

## Environment

- RustFS version: latest (self-hosted)
- Client: custom Rust client using `aws-sigv4` crate (v1.4.2) with `PercentEncodingMode::Single` and `UriPathNormalizationMode::Disabled`
- Request method: `GET`
- Path style: path-based (not virtual-hosted)

## Impact

Low — this only affects requests for non-existent objects with unusual characters in keys. However, it violates the S3 API contract: signature validation should not depend on object existence, and valid path characters should not cause signature mismatches.

jlon:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other: N/A

## Related Issues
- Fixes #2477

## Summary of Changes
- vendor the pinned `s3s` dependency used by RustFS
- preserve `=` in SigV4 canonical object paths so signed GET requests do not fail signature verification for valid S3 object keys
- add an end-to-end regression test covering signed GET on a missing object whose key ends with `=`

## Checklist

## Impact
- [ ] Other impact: N/A

## Additional Notes
Verification commands:
- `cargo fmt --all --check`
- `cargo test -p e2e_test test_signed_get_nonexistent_object_with_trailing_equals_returns_nosuchkey -- --nocapture`

---


jlon:
Superseded by rustfs/s3s#10. The final RustFS-side PR will only bump the s3s revision and keep the regression test after the upstream fix is merged.

PR #2515, opened 2026-04-13
fix: remove dead replace() call in gen_secret_key (credentials.rs)

### LINKED ISSUE #2514: bug: gen_secret_key dead-code replace on URL-safe base64 output in credentials.rs
### Summary

`gen_secret_key` in `crates/credentials/src/credentials.rs` encodes random bytes with `URL_SAFE_NO_PAD` (which uses `-` and `_` as the 62nd and 63rd alphabet characters instead of `+` and slash), and then calls `.replace("/", "+")`. Because URL-safe base64 **never produces a slash character**, this replace is a **dead code path** that never executes — it is a latent bug that suggests the wrong alphabet was assumed.

### Affected file
`crates/credentials/src/credentials.rs` — `gen_secret_key`

### Buggy code

```rust
pub fn gen_secret_key(length: usize) -> std::io::Result<String> {
    use base64_simd::URL_SAFE_NO_PAD;
    // ...
    let encoded = URL_SAFE_NO_PAD.encode_to_string(&key);
    let key_str = encoded.replace("/", "+");  // dead code: URL_SAFE never emits '/'
    Ok(key_str)
}
```

### Root cause
The code appears to have been written assuming standard base64 (which uses `+` and `/`) but then switched to the URL-safe variant (which uses `-` and `_`) without removing the now-redundant replace call. The net effect is:

1. The `.replace("/", "+")` call wastes CPU on every secret-key generation.
2. More importantly, it signals a **design confusion**: if the intent was to produce alphanumeric-plus-sign-only keys (replacing slash with plus), that goal is silently not achieved since URL-safe encoding still produces `-` and `_` characters.

### Fix
Remove the dead replace call. The output of `URL_SAFE_NO_PAD` never contains `/`, so the replace is a no-op:

```rust
pub fn gen_secret_key(length: usize) -> std::io::Result<String> {
    use base64_simd::URL_SAFE_NO_PAD;

    if length < 8 {
        return Err(Error::other("secret key length is too short"));
    }
    let mut rng = rand::rng();
    let mut key = vec![0u8; URL_SAFE_NO_PAD.estimated_decoded_length(length)];
    rng.fill_bytes(&mut key);
    // URL_SAFE_NO_PAD uses '-' and '_' — no '/' ever appears, so no replace needed
    Ok(URL_SAFE_NO_PAD.encode_to_string(&key))
}
```

### Impact
- Dead code executed on every credential generation path.
- Misleading intent: reviewers may incorrectly assume the output is alphanumeric-only.
- If the alphabet were ever changed back to standard base64, the replace would still be semantically wrong.

kuishou68:
Closes #2514

## Problem
`gen_secret_key` in `crates/credentials/src/credentials.rs` encoded random bytes with `URL_SAFE_NO_PAD` then called `.replace("/", "+")`. URL-safe base64 uses `-` and `_` instead of `+` and `/`, so the slash character never appears in the output. The replace was a dead no-op executed on every credential generation.

## Fix
Removed the dead `.replace("/", "+")` call and return the encoded string directly. Added a comment explaining why no replace is needed.

## Changed file
- `crates/credentials/src/credentials.rs` — `gen_secret_key`


houseme:
@kuishou68 Thank you for your contribution. Please explain why the entire document was deleted.

kuishou68:

kuishou68:
@houseme Thank you for reviewing and for the question — I apologize for the confusing commit history.

Here is what happened:

**The first commit in this PR accidentally submitted `credentials.rs` as a 0-byte empty file** (git empty-blob SHA `e69de29`). This was a tooling error during the automated patch-application process — the file content was not properly included in the API payload when the initial commit was created. That is why the first commit shows `−531` deletions.

This was caught immediately, and **a follow-up commit (`e909297`) in the same PR restored the full file from upstream `main`** and then applied the actual targeted fix on top of it (`+532` additions — the original 531 lines + 1 new line).

The **net diff of this PR is exactly 3 additions and 2 deletions** — only:
1. Removing the dead `.replace("/", "+")` call (which was a no-op since `URL_SAFE_NO_PAD` never produces `'/'`).
2. Adding a short comment explaining why.

The file is fully intact. No content was intentionally deleted. Again, I apologize for the messy intermediate commit — it was an unintended side-effect of the tooling used to create the PR. The final state of the file is correct and complete.

weisd (approved):

PR #2653, opened 2026-04-23
fix(admin): harden STS and KMS authorization checks

marshawcoco:
## Summary

- return explicit `UNIMPLEMENTED` gRPC statuses for unfinished node RPC handlers instead of panicking
- add policy parsing support for `sts:AssumeRole` and enforce it in the AssumeRole path
- use existing KMS-specific admin actions for KMS key create/status operations


reatang (approved):
KMS permissions are reasonably allocated.

Issue #1618, opened 2026-01-26
Users disappear from the Console UI and only shown for a short time after they've been active


BalintCsonka:
**Using:**
rustfs: 1.0.0-alpha.81
rc: v0.1.3

**Describe the bug**
I have created several service users for my workloads to use. After these users are idle for a while (~1 minute) they are not shown on the UI, nor can be listed with `rc admin user ls`.

Once I run a workload that uses any of these users, the user is once again visible on the console UI for a while, then disappears again.

This makes managing user settings on the UI nearly impossible and leaves no way to determine all added users.

**To Reproduce**
1. Create user with the console UI and/or the rc admin user add cli.
2. Wait ~1 minute
3. See the user disappear
4. Run a workload with the user
5. See the user appear for some time

**Expected behavior**
Users are visible all time.

**Additional context**
I seem to have started to experience this once I added an additional user along with the first x users already added from the console UI, still clarifying wether they could be connected.

loverustfs:
We'll fix it, plz wait for our comfirmation.

cxymds:
Possibly related issue #1260

LoganZ2:
@BalintCsonka I tested this in the Web UI, but I can’t reproduce the issue so far. Could you please share the exact steps you follow when this happens?
Also, just to clarify: service users are not expected to be shown directly in the main user list.
They are usually visible under each user’s menu instead.

mwullink:
i have the same issue, where i only see subset of created users in the UI and also using rc util.
the credentials for the "invisible" users do seem to work.

had this on version v80 up to v94 at the moment,  in the logging is see entries about users but i don't know what this means.
for example i see these log lines for the invisible user "maarten"


```
[...]

PR #2705, opened 2026-04-27
fix(ecstore): log walk failures in IAM listing path

GatewayJ:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
Relates to #1618

## Summary of Changes
- Added error logging when IAM config listing launches `store.walk(...)` in a detached task and the walk call returns an error.
- Added per-task error logging after `join_all(futures)` in `walk_internal` so failures from set-level listing tasks are no longer silent.
- Kept behavior unchanged (observability-only change): errors are logged, and current return-path semantics remain the same.

## Checklist

## Impact
- [x] Other impact:
  - Improves troubleshooting signal for intermittent/partial IAM listing failures.

## Additional Notes
- Verification commands run:
  - `cargo fmt --all --check`
- `make pre-commit` was not run in this iteration per requester instruction.
- N/A


crates/ecstore/src/store_list_objects.rs
```diff
@@ -829,6 +829,7 @@ impl ECStore {
         let (merge_tx, mut merge_rx) = mpsc::channel::<MetaCacheEntry>(100);
 
         let bucket = bucket.to_owned();
+        let bucket_for_merge = bucket.clone();
```
houseme: `bucket_clone` ?

GatewayJ: Updated in 6d51eba6: renamed to `bucket_clone` for consistency with nearby naming patterns. Thanks for the suggestion.

houseme (approved):

PR #2722, opened 2026-04-28
fix(webdav): decode URL-encoded filenames in path parsing

giter:
## 🐛 Problem

When uploading files via WebDAV with special characters in filenames (e.g., spaces, Chinese characters), the filenames were stored with URL encoding preserved.

## ✅ Solution

Added URL decoding in `parse_path` using `percent_encoding::percent_decode_str`.

### Changes
- **Cargo.toml**: Added `percent-encoding` to webdav feature dependencies
- **driver.rs**: Decode URL-encoded path strings before processing

### Before
```
Client request: /bucket/hello%20world.txt
Stored key:     hello%20world.txt  ❌
```

### After
```
Client request: /bucket/hello%20world.txt
Stored key:     hello world.txt    ✅

### DIFF
--- a/crates/protocols/src/webdav/driver.rs
+++ b/crates/protocols/src/webdav/driver.rs
@@ -21,6 +21,7 @@ use dav_server::fs::{
     DavDirEntry, DavFile, DavFileSystem, DavMetaData, FsError, FsFuture, FsResult, FsStream, OpenOptions, ReadDirMeta,
 };
 use futures_util::{FutureExt, StreamExt, stream};
+use percent_encoding::percent_decode_str;
 use rustfs_utils::path;
 use s3s::dto::*;
 use std::fmt::Debug;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by giter** [comment]

File: crates/protocols/src/webdav/driver.rs [hunk 1]
```diff
@@ -432,6 +433,9 @@ where
     /// Parse WebDAV path to bucket and object key
     fn parse_path(&self, path: &DavPath) -> Result<(String, Option<String>), FsError> {
         let path_str = path.as_url_string();
+        let decoded_path = percent_decode_str(&path_str)
```
houseme, line 436: Unused `decoded_path`

houseme (approved):

giter (commented):
missed,  resolved now

PR #2739, opened 2026-04-29
fix(admin): allow site replication removal with offline peers

cxymds:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
N/A

## Summary of Changes
- Persist site replication removal locally before notifying peer sites.
- Treat peer remove notification failures as non-blocking and surface them in `errorDetail`.
- Add regression coverage for remove-all when peer notification returns an error.

## Checklist
- [x] Documentation updated (if needed) - N/A, no documentation change required
- [ ] CI/CD passed (if applicable) - pending GitHub CI

## Impact
- [x] Other impact: site replication removal can complete locally even when peer notification fails; peer failures are reported in `errorDetail`.

## Additional Notes
In the affected deployment, an unavailable peer may surface as a `403 Forbidden` response with `InvalidAccessKeyId` from the peer remove request. The fix treats that response the same as other peer notification failures: local removal succeeds and the peer failure is reported to the caller.

Verification:
- `cargo test -p rustfs --lib admin::handlers::site_replication::tests::`
- `cargo fmt --all --check`
- `git diff --check`
- `make pre-commit`

`make pre-commit` reported that `cargo-nextest` was unavailable and fell back to `cargo test`; the fallback test run and subsequent checks passed.

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -83,6 +83,7 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use std::sync::OnceLock;
 use std::time::{Duration, Instant};
 use time::OffsetDateTime;
+use tracing::warn;
 use url::{Url, form_urlencoded};
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


cxymds:
Addressed the peer error detail review in fc33f5a8. Peer remove failures are now summarized before being returned or logged: each detail is trimmed and capped at 256 characters with a truncation marker. Added regression coverage for long peer response bodies.

loverustfs (approved):

PR #2765, opened 2026-05-01
fix: handle empty multipart list-parts

### LINKED ISSUE #2760: ListParts returns 500 for a new multipart upload with no parts
Hi,

I'm using RustFS as the S3 backend for Harbor, and I hit a plain S3 compatibility issue before Harbor can even upload the layer data.

`ListParts` on a freshly-created multipart upload with no uploaded parts returns HTTP 500 from RustFS:

```text
Storage resources are insufficient for the read operation
```

Expected behaviour: return 200 with an empty parts list, like S3/MinIO. Harbor / Docker Distribution calls this when it resumes an upload, so the 500 breaks `docker push`.

Environment:

```text
RustFS image: rustfs/rustfs:1.0.0-beta.1
Harbor: 2.14.2
Registry: v2.8.3-23-g0c62ec3e
Docker client: 29.4.1
S3 mode: path style
Bucket: registry
```

Repro:

```bash
export AWS_ACCESS_KEY_ID=...
export AWS_SECRET_ACCESS_KEY=...
export AWS_DEFAULT_REGION=ca-east-1
EP=https://s3-001.domain.com
KEY="debug/empty-listparts-$(date -u +%s)/data"

UPLOAD_ID="$(aws --endpoint-url "$EP" s3api create-multipart-upload \
  --bucket registry \
  --key "$KEY" \
  --query UploadId \
  --output text)"

aws --endpoint-url "$EP" s3api list-parts \
  --bucket registry \
  --key "$KEY" \
  --upload-id "$UPLOAD_ID"
```

If I upload a part first, `ListParts` works. The failure is specifically the empty upload case.

I know #815/#816 were closed and #979 says it fixed a related `ListParts` problem, but this still happens here with `1.0.0-beta.1`. My temporary workaround is a tiny proxy that converts this exact empty-ListParts 500 into an empty 200 response, and Harbor gets past this stage with that.

overtrue:
## Related Issues
Fixes #2760.

## Summary of Changes
- treat missing multipart part directories as an empty upload state when no parts have been uploaded yet
- avoid collapsing the fresh-upload `ListParts` path into a read-quorum failure when the only observed result is `FileNotFound` plus ignored offline-disk errors
- add regression coverage for the empty-upload directory case in `set_disk::multipart`

## Verification
- `cargo test -p rustfs-ecstore collect_list_parts_results_ -- --nocapture`
- `cargo test -p rustfs-ecstore`
- reproduced locally before the fix with a fresh multipart upload followed by `ListParts` returning HTTP 500 / `InternalError`
- reproduced locally after the fix with the rebuilt server returning HTTP 200 and an empty parts list for the same flow
- `cargo fmt --all`
- `cargo fmt --all --check`
- `make pre-commit`

## Impact
`ListParts` now matches S3/MinIO behavior for freshly-created multipart uploads that have no uploaded parts yet, which unblocks clients like Harbor / Docker Distribution that probe empty uploads during resume.

## Additional Notes
The fix is intentionally narrow to the multipart part-directory collection path. Existing early quorum-failure behavior is preserved for cases that do not involve the empty-upload `FileNotFound` state.

### DIFF
--- a/crates/ecstore/src/set_disk/multipart.rs
+++ b/crates/ecstore/src/set_disk/multipart.rs
@@ -49,9 +49,24 @@ where
             Err(_) => {}
         }
 
-        if successful_responses + pending < read_quorum {
+        if successful_responses + pending < read_quorum && !errs.iter().any(|err| matches!(err, Some(DiskError::FileNotFound))) {
             return Err(DiskError::ErasureReadQuorum);
         }
+    }
+
+    if successful_responses < read_quorum {
+        let saw_file_not_found = errs.iter().any(|err| matches!(err, Some(DiskError::FileNotFound)));
+        let only_missing_or_ignored = errs.iter().all(|err| match err {
+            Some(DiskError::FileNotFound) => true,
+            Some(err) => OBJECT_OP_IGNORED_ERRS.contains(err),
+            None => false,
+        });
+
+        if successful_responses == 0 && saw_file_not_found && only_missing_or_ignored {
+            return Err(DiskError::FileNotFound);
+        }
+
+        return Err(DiskError::ErasureReadQuorum);
     }
 
     Ok((errs, object_parts))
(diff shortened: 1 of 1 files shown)


overtrue:
Addressed the Codex review point.

I tightened the early-exit guard so we only defer fail-fast behavior while the empty-upload `FileNotFound` fallback is still genuinely possible (no successful responses, and all observed results are `FileNotFound` or ignored offline-disk errors). If a non-ignored error shows up, we now return `ErasureReadQuorum` immediately again instead of waiting for the slow tail.

Added a regression test covering the mixed `FileNotFound` + non-ignored error case, and re-ran:
- `cargo test -p rustfs-ecstore collect_list_parts_results_ -- --nocapture`
- `cargo fmt --all`
- `cargo fmt --all --check`
- `make pre-commit`

PR #2769, opened 2026-05-01
security: same-origin console CORS, fail-closed helm creds, deny.toml, sample-config hardening

skandragon:
## ⚠️ Breaking changes

Two changes in this PR are behavior-changing for existing users. Please flag if either should be split into its own PR or paired with a chart-major / app-minor version bump.

1. **Helm chart: `helm install` / `helm upgrade` with no overrides will start failing rendering.** The well-known `rustfsadmin` / `rustfsadmin` defaults are no longer rendered silently. See the upgrade-path section below for the three escape hatches.
2. **Console CORS default flipped from `*` to same-origin only.** Anyone running the console cross-origin and relying on the prior wildcard default will see CORS preflights start failing. See the upgrade-path section below.

## Summary

A bundle of low-risk public-hardening changes that came out of an external security review. None of these address an active in-the-wild exploit, and none are critical-class issues; the truly impactful findings have been routed through GitHub Private Vulnerability Reporting separately. This PR addresses the public-safe items.

### Helm chart

Pick one of three escape hatches (the new `fail` message lists them):

- Set explicit non-default creds: `--set secret.rustfs.access_key=... --set secret.rustfs.secret_key=...`
- Point at a Secret you control: `--set secret.existingSecret=<name>`
- Opt back into the well-known dev creds (local dev only): `--set secret.allowInsecureDefaults=true`

This likely warrants a chart `Chart.yaml` version bump (major, per chart semver). Happy to add that to this PR or split the helm change out — whichever you prefer.

### Console CORS

Restore old behavior with `RUSTFS_CONSOLE_CORS_ALLOWED_ORIGINS=*`, or supply an explicit allow-list (recommended).

## Changes

- **Console CORS default**: `RUSTFS_CONSOLE_CORS_ALLOWED_ORIGINS` now defaults to `""` (same-origin only) instead of `"*"`. Operators opt in to wildcard or an explicit allow-list when the console is served under a different origin. Regression tests cover default same-origin behavior and the explicit `*` opt-in. (`crates/config/src/constants/console.rs`, `rustfs/src/admin/console.rs`)
- **NGINX sample**: the commented HTTPS server block in `.docker/nginx/nginx.conf` now sets `ssl_protocols TLSv1.2 TLSv1.3` plus modern ciphers and a session cache, so copy-paste deployments inherit safe defaults.
- **`deny.toml`**: codifies what was previously implicit policy for `cargo deny check` (advisories, sources, bans, licenses). Documents the accepted exceptions (`instant`, `paste` unmaintained transitive; `rsa` Marvin sidechannel tracked separately) with owner + review-month tags.
- **Helm chart fail-closed credentials**: `secret.rustfs.access_key` / `secret_key` now default to empty. `helm template` / `helm install` fails with a security-explanatory message unless one of: explicit non-default creds are supplied, `secret.existingSecret` names a Secret you control, or `secret.allowInsecureDefaults: true` is set. The four code paths are covered by `scripts/test_helm_templates.sh`.
- **`docker-compose.yml`**: split the S3 vs console CORS guidance — the S3 listener still reflects any Origin even when the env var is unset, so the comment now flags that an explicit allow-list is the only constraint.
- **Test isolation**: a few env-var-mutating tests (`ENV_HEALTH_ENDPOINT_ENABLE`, the new console-CORS tests) had a pre-existing race when run in parallel. Added `#[serial]` to the affected tests in `rustfs/src/admin/console.rs` and `rustfs/src/admin/route_registration_test.rs` to fix it.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo test -p rustfs --lib admin::console`
- [x] `cargo test -p rustfs --lib admin::route_registration`
- [x] `cargo deny check` (advisories, sources, bans, licenses)
- [x] `bash scripts/test_helm_templates.sh` — covers RollingUpdate / Recreate strategy, default fail-closed render, well-known-creds fail, `allowInsecureDefaults=true` success, `existingSecret` short-circuit
- [ ] Maintainer: confirm `cargo deny` exception comments reflect intended owner/review cadence
- [ ] Maintainer: decide whether the helm change wants a `Chart.yaml` major bump in this PR or a follow-up

## Notes for reviewers

- I'm happy to split this into smaller PRs if any of the changes would be easier to review in isolation — in particular the helm fail-closed change is the natural one to split out if you'd rather release/ship it on its own cadence.
- If you'd prefer a soft warning over a hard `fail` for the helm credential check, that's a one-line change.


skandragon:

skandragon:
Thanks for the careful review @copilot — all three comments addressed in `ce53d665`:

1. **Helm partial-default (`hasDefaultKey`)**: changed `$isDefault` (both must equal `rustfsadmin`) to `$hasDefaultKey` using `or`, so a single key matching the well-known default fails rendering even when the other is overridden.
2. **Helm partial-empty auto-fill**: split into `$bothEmpty` and `$oneEmpty`. Partial-empty now emits a dedicated `fail` message regardless of `allowInsecureDefaults`, so a missing key can never silently fall back to `rustfsadmin`. Auto-fill only fires when both keys are empty AND `allowInsecureDefaults=true`.
3. **CORS wildcard whitespace**: `origins_str.trim() == "*"` so `" * "` (common when templated via env vars) is treated as wildcard.

`scripts/test_helm_templates.sh` now covers the partial-default and partial-empty paths; `rustfs/src/admin/console.rs` has a `whitespace_padded_wildcard_console_cors_allows_any_origin` regression test.

loverustfs (approved):

majinghe (approved):

PR #2771, opened 2026-05-01
fix(rpc): use map-encoded msgpack for all internode RPC responses

### LINKED ISSUE #2780: v1.0.0-beta.1: storage_info RPC msgpack deserialization error in multi-pool configurations
## Environment

- **RustFS version**: `1.0.0-beta.1` (image: `rustfs/rustfs:1.0.0-beta.1`)
- **Helm chart**: `rustfs/rustfs` v0.1.0
- **Deployment**: Kubernetes StatefulSet, 4 replicas, distributed mode
- **Volume configuration**: 4 space-separated pool expressions:
  ```
  http://rustfs-{0...3}.rustfs-headless.rustfs.svc.cluster.local:9000/data/rustfs0
  http://rustfs-{0...3}.rustfs-headless.rustfs.svc.cluster.local:9000/data/rustfs1
  http://rustfs-{0...3}.rustfs-headless.rustfs.svc.cluster.local:9000/data/rustfs2
  http://rustfs-{0...3}.rustfs-headless.rustfs.svc.cluster.local:9000/data/rustfs3
  ```
- **Topology**: 4 pools × 1 erasure set × 4 drives per set

## Symptom

All pods report `0/1 Ready` indefinitely despite being fully operational. The `/health/ready` endpoint permanently returns HTTP 503:

```json
{
  "details": {
    "iam": {"ready": false, "status": "disconnected"},
    "storage": {"ready": false, "status": "disconnected"}
  },
  "ready": false,
  "service": "rustfs-endpoint",
  "status": "degraded",
  "timestamp": "2026-04-30T16:25:19.852031722Z[Etc/Unknown]",
  "version": "1.0.0-beta.1"
}
```

Kubernetes readiness probes fail continuously, preventing pods from entering Ready state and receiving load-balanced traffic.

## Log Evidence

All pods show repeated warnings on every peer communication attempt (every ~5 seconds):

```
WARN rustfs-worker ThreadId(15) storage_info:handle_storage_info: rustfs_ecstore::notification_sys: crates/ecstore/src/notification_sys.rs:206: peer 172.20.1.99:9000 storage_info failed: Io error: wrong msgpack marker FixArray(1)
WARN rustfs-worker ThreadId(12) storage_info:handle_storage_info: rustfs_ecstore::notification_sys: crates/ecstore/src/notification_sys.rs:206: peer 172.20.1.103:9000 storage_info failed: Io error: wrong msgpack marker FixArray(1)
WARN rustfs-worker ThreadId(17) storage_info:handle_storage_info: rustfs_ecstore::notification_sys: crates/ecstore/src/notification_sys.rs:206: peer 172.20.2.139:9000 storage_info failed: Io error: wrong msgpack marker FixArray(1)
```

The error occurs **bidirectionally on all peer pairs** — every pod sees the error when communicating with every other pod.

## Root Cause

The `notification_sys::storage_info` RPC handler (at `crates/ecstore/src/notification_sys.rs:206`) appears to have a msgpack serialization/deserialization bug when running **multi-pool configurations**.

All peer-to-peer `storage_info` exchanges fail with `wrong msgpack marker FixArray(1)`, which prevents:
- The IAM subsystem from reporting as "ready"
- The storage subsystem from reporting as "ready"
- The `/health/ready` endpoint from returning HTTP 200

The peers can communicate over TCP (DNS resolution works, connections succeed), but the msgpack response payload cannot be deserialized.

## Impact

**What's Broken:**
- `/health/ready` endpoint permanently returns 503
- Kubernetes readiness probes fail
- Pods never reach `1/1 Ready` state
- Load balancer health checks fail

**What Still Works:**
- S3 API operations (GET, PUT, DELETE, LIST all function correctly)
- Erasure coding and data operations
- Distributed lock acquisition
- Data scanning and rebalancing
- Console on port 9001

The cluster is **fully functional for S3 operations** — only the health reporting subsystem is broken.

## Steps to Reproduce

1. Deploy rustfs v1.0.0-beta.1 in Kubernetes with a multi-pool configuration (4 space-separated volume expressions)
2. Wait for pods to start
3. Check pod readiness: `kubectl get pods -n rustfs`
   - Observe: All pods stuck at `0/1 Running`
4. Check health endpoint: `kubectl exec rustfs-0 -n rustfs -- wget -qO- http://localhost:9000/health/ready`
   - Observe: HTTP 503 with IAM/storage "disconnected"
5. Check logs: `kubectl logs rustfs-0 -n rustfs | grep storage_info`
   - Observe: Repeated msgpack marker errors

## Workaround

Replace the HTTP readiness/liveness probes with TCP probes. Since the issue is only in the health endpoint (not the actual S3 service), checking that port 9000 is listening is sufficient:

```yaml
readinessProbe:
  tcpSocket:
    port: 9000
  initialDelaySeconds: 30
  periodSeconds: 5

livenessProbe:
  tcpSocket:
    port: 9000
  initialDelaySeconds: 60
  periodSeconds: 10
```

This allows pods to become Ready and receive traffic normally.

## Expected Behavior

After cluster quorum is established, `/health/ready` should return HTTP 200 with:
```json
{
  "details": {
    "iam": {"ready": true, "status": "connected"},
    "storage": {"ready": true, "status": "connected"}
  },
  "ready": true
}
```

## Additional Context

- Single-pool configurations (single volume expression with `{0...N}` expansion) may not be affected
- The issue appears specific to the space-separated multi-pool volume format
- All pods run the same image, so this is not a version skew issue
- The `format.json` on disk is consistent across all pods

kharkevich:
Fixes #2780

## Summary

All server-side RPC handlers in `health.rs` and `metrics.rs` serialized response structs with `Serializer::new()`, which produces **positional array-encoded msgpack**. The client-side deserializers in `peer_rest_client.rs` use `Deserializer::new()`, which expects **map-encoded msgpack** with named fields.

This mismatch causes `wrong msgpack marker FixArray(1)` errors on `local_storage_info` calls, preventing the erasure store from initializing and leaving **all nodes permanently stuck in `degraded/disconnected` state** in distributed deployments.

`handle_server_info` was the only handler that already used `.with_struct_map()` correctly.

## Root Cause

```rust
// health.rs — broken (array encoding)
info.serialize(&mut Serializer::new(&mut buf))

// health.rs — working (map encoding, only handle_server_info had this)
info.serialize(&mut Serializer::new(&mut buf).with_struct_map())
```

`rmp_serde::Serializer::new()` encodes structs as msgpack arrays (positional). `Deserializer::new()` on the client expects msgpack maps (named fields). The `StorageInfo` struct with its nested `BackendInfo` (containing `HashMap` fields, `#[serde(rename)]` and `skip_serializing_if` attributes) is particularly sensitive to this mismatch.

## Changes

- **`rustfs/src/storage/rpc/mod.rs`**: Added `encode_msgpack_map()` helper that enforces `.with_struct_map()` in one place, so the encoding choice cannot be accidentally omitted in future handlers.
- **`rustfs/src/storage/rpc/health.rs`**: Migrated all 11 handlers to use `encode_msgpack_map()` instead of inlining `Serializer` construction (`handle_get_proc_info`, `handle_get_mem_info`, `handle_get_sys_errors`, `handle_get_sys_config`, `handle_get_se_linux_info`, `handle_get_os_info`, `handle_get_partitions`, `handle_get_net_info`, `handle_get_cpus`, `handle_server_info`, `handle_local_storage_info`).
- **`rustfs/src/storage/rpc/metrics.rs`**: Migrated `handle_get_metrics` to use `encode_msgpack_map()`.
- **`crates/ecstore/src/rpc/peer_rest_client.rs`**: Fixed `get_se_linux_info()` to deserialize as `SysServices` (plural wrapper with `node_common` + `services`) instead of `SysService` (singular with `name` + `status`), aligning the client contract with what the server actually sends.
- **`crates/ecstore/src/notification_sys.rs`**: Updated `get_sys_services()` return type from `Vec<SysService>` to `Vec<SysServices>` to match the corrected client contract.
- **`rustfs/src/storage/rpc/node_service.rs`**: Added 12 round-trip regression tests that mirror the real `PeerRestClient` deserialization path — each test calls the server handler, then deserializes the response bytes into the exact type the client uses. Also removed unused `Serializer`/`Serialize` imports.

## How to Reproduce

1. Deploy a 4-node distributed RustFS cluster with `1.0.0-beta.1`
2. All nodes start, bind ports, resolve hostnames
3. Internode `local_storage_info` RPC fails with: `peer X storage_info failed: Io error: wrong msgpack marker FixArray(1)`
4. ECStore never initializes — all nodes report `storage: disconnected, iam: disconnected`
5. Cluster is completely non-functional

## Test Plan

- [x] Round-trip regression tests for all 12 RPC endpoints (ProcInfo, MemInfo, SysErrors, SysConfig, SysServices, OsInfo, Partitions, NetInfo, Cpus, ServerProperties, RealtimeMetrics, StorageInfo)
- [x] `test_get_se_linux_info_round_trip` validates the SysServices contract fix
- [x] `test_get_metrics_round_trip` validates metrics response deserialization
- [x] `test_server_info_round_trip` validates server_info still works after migration to shared helper
- [ ] Deploy 4-node cluster with this fix — verify `health/ready` returns `ok` / `ready: true`


rustfs/src/storage/rpc/health.rs
```diff
@@ -105,59 +102,56 @@ impl NodeService {
     ) -> Result<Response<GetSeLinuxInfoResponse>, Status> {
         let addr = get_global_local_node_name().await;
         let info = get_sys_services(&addr);
```
overtrue, line 104: This handler is still using a different payload shape from the client contract.

`get_sys_services(&addr)` returns the wrapper type `SysServices` (`node_common + services`), and this PR now serializes that wrapper with map encoding. However, `PeerRestClient::get_se_linux_info()` deserializes `response.sys_services` as a singular `SysService`. After this PR, the marker mismatch is fixed, but this endpoint can still fail at decode time because the payload map contains `node_common` / `services` rather than `name` / `status`.

Could we align the server/client contract here before merging? The field name and server helper suggest the response should probably be decoded as the plural `SysServices` shape, but either direction is fine as long as both sides agree.

It would also be useful to add a regression test that mirrors the real client path: call `handle_get_se_linux_info` / `get_se_linux_info`, then deserialize `response.sys_services` into the exact type used by `PeerRestClient`. The existing tests only assert that the response bytes are non-empty, so they would not catch this remaining incompatibility.

overtrue (commented):
I found one remaining RPC contract mismatch in the touched health handler.

kharkevich:
should be better now

Issue #2608, opened 2026-04-20
[Feature Request] Community Collaboration: Extend Native Event Notification Target Support


houseme:
## 🎯 Background & Objectives

Dear community members,

Thank you for your continued interest in the RustFS project. We have received feedback from multiple users requesting native support for additional message brokers as event notification targets—such as **NATS**, **Kafka**, **AMQP**, **Redis**, and **PostgreSQL**—to enable more flexible integration of file system events (e.g., `ObjectCreated`, `ObjectRemoved`) into existing event-driven architectures.

Currently, RustFS's `crates/targets` module has established an extensible event notification framework based on the `Target` Trait, with reference implementations for `Webhook` and `MQTT`. However, due to resource constraints within the core team, we are unable to deliver native support for all requested endpoints in the short term.

We firmly believe that the vitality of an open-source project stems from community collaboration. Therefore, we warmly invite developers and enthusiasts of all experience levels to join us in building this feature. Whether you are a seasoned Rust expert or contributing to open source for the first time, your effort will bring tangible value to RustFS.

**Core Objectives of This Issue**:  
✅ Clarify the technical approach and implementation specifications for extending target endpoints  
✅ Provide clear development guidance and reference resources for community contributors  
✅ Establish an efficient collaboration workflow to ensure code quality and maintainability

---

## 🔍 Current Architecture Reference: `Target` Trait Design

To help contributors get started efficiently, here is the core interface definition of the event target system (excerpted from [`crates/targets/src/target/mod.rs`](https://github.com/rustfs/rustfs/blob/main/crates/targets/src/target/mod.rs)):

```rust
#[async_trait]
pub trait Target<E>: Send + Sync + 'static
where
    E: Send + Sync + 'static + Clone + Serialize + DeserializeOwned,
{
    fn id(&self) -> TargetID;
    fn name(&self) -> String { ... }
    async fn is_active(&self) -> Result<bool, TargetError>;
    async fn save(&self, event: Arc<EntityTarget<E>>) -> Result<(), TargetError>;
    async fn send_raw_from_store(&self, key: Key, body: Vec<u8>, meta: QueuedPayloadMeta) -> Result<(), TargetError>;
    async fn send_from_store(&self, key: Key) -> Result<(), TargetError>;
    async fn close(&self) -> Result<(), TargetError>;
    fn store(&self) -> Option<&(dyn Store<QueuedPayload, Error = StoreError, Key = Key> + Send + Sync)>;
[...]

PR #2779, opened 2026-05-03
[WIP] feat(targets): add PostgreSQL notify target

### LINKED ISSUE #2629: Feature: Impl of PostgreSQL event notification Target
## Summary
Proposing the implementation of a PostgreSQL event notification target so that RustFS events can be persisted into a relational database — useful for audit trails, compliance logging, and integration with existing business databases already running on Postgres.

## Parent/Related Issue 
- #2608

jaehanbyun:
## Related Issues

- Fixes #2629 
- Refs #2608 (parent — Community Collaboration: Extend Native Event Notification Target Support)

## Summary of Changes

Adds PostgreSQL as the 6th built-in notify target. Implements Target<E> end-to-end (config keys → ecstore default KVS → target args/builder → PostgresTarget impl → connectivity probe → notify factory/registry/integration → admin descriptor + validator) following the pattern established by NATS/Pulsar in [#2618](https://github.com/jaehanbyun/rustfs/issues/2618). 

## Verification

```
make pre-commit                                     
cargo fmt --all --check                              
cargo clippy --all-targets --all-features -- -D warnings  
cargo nextest run --all --exclude e2e_test         
cargo test --all --doc                              
```

## Impact

Additive only. No existing behavior changes.

- The pre-existing #[allow(dead_code)] on NOTIFY_POSTGRES_SUB_SYS is removed and the constant is now active.
- `notify_target_descriptors()` length grows from 5 to 6.
- `notification_target_specs()` length grows from 5 to 6.
- New `RUSTFS_NOTIFY_POSTGRES_*` env vars are inert until enable=on.
- Workspace gains five direct dependencies (`tokio-postgres`, `deadpool-postgres`, `tokio-postgres-rustls`, `rustls-pemfile`, `rustls-native-certs`); `rustls-native-certs` was already a transitive dependency, so the build graph grows by four crates.
- No public API removed or renamed.
- No migration required.

## Additional Notes

---


jaehanbyun:

crates/config/src/constants/targets.rs
```diff
@@ -74,6 +74,21 @@ pub const PULSAR_TLS_HOSTNAME_VERIFICATION: &str = "tls_hostname_verification";
...
+pub const POSTGRES_USER: &str = "user";
+pub const POSTGRES_PASSWORD: &str = "password";
+pub const POSTGRES_DATABASE: &str = "database";
+pub const POSTGRES_SCHEMA: &str = "schema";
+pub const POSTGRES_TABLE: &str = "table";
+pub const POSTGRES_FORMAT: &str = "format";
+pub const POSTGRES_TLS_REQUIRED: &str = "tls_required";
+pub const POSTGRES_TLS_CA: &str = "tls_ca";
+pub const POSTGRES_TLS_CLIENT_CERT: &str = "tls_client_cert";
+pub const POSTGRES_TLS_CLIENT_KEY: &str = "tls_client_key";
```
JaySon-Huang, line 88: Hi @jaehanbyun,

I'm also implementing a target for event notifications in `rustfs` 🤝. During my research on similar object storage implementations (like MinIO), I found that using a single **DSN (Data Source Name) string** instead of separate `host`, `port`, `user`, etc., fields offers better consistency and reduces the migration cost for users in the **MySQL** event notification target.

For reference, MinIO's official documentation for **PostgreSQL** notifications also recommends using a **Connection String** rather than split configuration items:
🔗 [MinIO PostgreSQL Notifications Docs](https://docs.min.io/aistor/administration/bucket-notifications/publish-events-to-postgresql)

Would you be open to refactoring the PostgreSQL target to use a single `connection_string` as well? I think it would make the API design across different SQL databases more uniform.

houseme:
Thanks for the solid PostgreSQL target implementation. I did a deep pass focused on runtime reliability, replay semantics, and operator UX. Below is my adopt/reject recommendation list.

## Recommended to adopt (high priority)

1. Error classification for replay/backoff (`map_pool_error` / `map_pg_error`)
- Why: transient connectivity failures can currently fall into `Network`/`Request`-like permanent paths, while notify replay typically backs off only on `NotConnected`/`Timeout`.
- Risk: queued events may get stuck or retry in the wrong mode during brief PostgreSQL outages.
- Suggested minimal fix: map connection checkout failures, closed-connection errors, and transport-level transient failures to `TargetError::NotConnected` (and explicit timeout cases to `TargetError::Timeout`). Keep deterministic config/permission/schema faults as permanent errors.

2. Admin probe requires `SELECT` privilege (`SELECT ... LIMIT 0`)
- Why: this rejects least-privilege roles that are valid for delivery (write-focused roles), especially in append-only scenarios.
- Risk: valid deployments cannot be configured.
- Suggested minimal fix: at least document the `SELECT` requirement clearly in current behavior; follow up with format-aware probe logic so `access` mode can support write-only role models where feasible.

3. `event_name` persistence format consistency (access mode)
- Why: writing debug-style names (for example `ObjectCreatedPut`) instead of canonical S3 names (for example `s3:ObjectCreated:Put`) makes SQL-side filtering inconsistent with payload and existing event filters.
- Risk: downstream consumers cannot reliably query by standard event names.
- Suggested minimal fix: persist canonical S3 event names in `event_name`.

## Recommended to adopt (medium priority)

4. Validator parity for TLS path checks (`validate_postgres_config`)
- Why: target construction validates absolute TLS paths, but validator path does not fully enforce the same constraints.
- Risk: validation may pass but runtime/init fails later.
- Suggested minimal fix: add absolute-path checks for `tls_ca`, `tls_client_cert`, and `tls_client_key` in validator path as well.

5. Replay-idempotency test coverage around `send_raw_from_store`
- Why: idempotent replay depends on stable event-id derivation from store key in that method.
- Risk: future refactors can silently break replay semantics.
- Suggested minimal fix: add a direct test that exercises `send_raw_from_store` with stable key behavior.

## Lower-priority but useful

6. Regression tests for admin/config mapping
- Add one PostgreSQL case in `notification_target_specs()` tests.
- Add one PostgreSQL encode/decode descriptor regression in `crates/ecstore/src/config/com.rs` tests.

## Not recommended for this PR scope

7. Refactor to a single DSN-style PostgreSQL config model
- This is a broader API-shape change, not a correctness fix for the current PR.
- Suggest handling in a follow-up issue/PR to keep this change set focused.

PR #2828, opened 2026-05-06
fix(rpc): add issue 2815 regression and docker validation

### LINKED ISSUE #2815: [beta.1 regression] handle_local_storage_info missing map encoding — all pods stuck 0/1 Ready in distributed mode, /health/ready returns 503
## Describe the bug

In `v1.0.0-beta.1`, all nodes in a distributed cluster are permanently `0/1 Ready` in Kubernetes. `/health/ready` returns 503 with `"storage": "disconnected"` and `"iam": "disconnected"` on every node, despite S3 data operations working correctly.

## Root cause

The `handle_local_storage_info` handler in `rustfs/src/storage/rpc/health.rs` was not updated to use `encode_msgpack_map` when the rest of the file was migrated. In the `1.0.0-beta.1` git tag it still uses the old tuple-mode serializer:

\`\`\`rust
// 1.0.0-beta.1 (broken)
let mut buf = Vec::new();
if let Err(err) = info.serialize(&mut Serializer::new(&mut buf)) { ... }
\`\`\`

Every other handler in the same file — `handle_server_info`, `handle_get_cpus`, `handle_get_mem_info`, etc. — uses `encode_msgpack_map`. The `v1.0.0-beta.1` git tag (with the `v` prefix) has the fix applied, but **no Docker image has been published for that tag**. Docker Hub only carries `rustfs/rustfs:1.0.0-beta.1`, which maps to the unfixed git tag.

With tuple encoding, `BackendDisks(pub HashMap<String, usize>)` (a newtype struct) serializes as `FixArray(1)`. The receiving peer node fails to deserialize it as a map:

```
WARN peer 10.x.x.x:9000 storage_info failed: Io error: wrong msgpack marker FixArray(1)
     at crates/ecstore/src/notification_sys.rs:206
```

The health readiness check (PR #2662) marks all peer disks as offline when `storage_info` fails, dropping below write quorum → `storage: disconnected` → 503.

This is the same class of issue as #2617 (fixed for `server_info` in PR #2621) — `handle_local_storage_info` was simply missed when the other handlers were updated.

## To Reproduce

1. Deploy `rustfs/rustfs:1.0.0-beta.1` in distributed mode (4+ nodes, e.g. Helm chart `0.1.0`)
2. All pods remain `0/1 Ready`
3. Check logs on any node:
   ```
   WARN peer 10.x.x.x:9000 storage_info failed: Io error: wrong msgpack marker FixArray(1)
        at crates/ecstore/src/notification_sys.rs:206
   ```
4. `curl http://localhost:9000/health/ready` → HTTP 503, `{"storage":{"status":"disconnected"}, "iam":{"status":"disconnected"}, ...}`

## Workaround (for current `1.0.0-beta.1` image users)

Switch the Kubernetes readiness probe from `/health/ready` → `/health`. The liveness endpoint returns HTTP 200 and the data path is unaffected.

## Request

Please either:
- Publish a `rustfs/rustfs:v1.0.0-beta.1` Docker image (or overwrite `1.0.0-beta.1`) matching the `v1.0.0-beta.1` git tag which already contains the fix, **or**
- Release beta.2

The fix is already present in the `v1.0.0-beta.1` git tag and `main`.

## Environment

- Docker image: `rustfs/rustfs:1.0.0-beta.1` (Docker Hub, pushed Apr 29)
- Helm chart: `rustfs-0.1.0`
- Mode: Distributed, 4 nodes, Kubernetes (k3s)

houseme:
## Related Issues
Fixes #2815

## Summary of Changes
- add focused MsgPack regression tests for `StorageInfo` so the internode map-encoding contract behind issue #2815 stays covered
- add `.docker/test/issues-2815/` local verification assets for a 4-node distributed source-built cluster
- move the issue-specific Docker verification flow into the existing `.docker/test/` layout used for repro assets
- add runtime `curl` to `Dockerfile.source` so container healthchecks can execute during local Docker verification
- harden `Dockerfile.source` local multi-arch build behavior by using `TARGETARCH` and an isolated Docker target dir
- configure local distributed Docker verification with explicit non-default `RUSTFS_SECRET_KEY` and `RUSTFS_RPC_SECRET` so internode RPC auth does not fail closed on the default secret

## Verification
- `cargo test -p rustfs storage_info_map_encoding_round_trip_matches_issue_2815_contract --lib`
- `cargo test -p rustfs storage_info_tuple_encoding_uses_array_marker_that_issue_2815_fixed --lib`
- `docker compose -f .docker/test/issues-2815/docker-compose.yml config`
- local 4-node Docker cluster reached `healthy` on all nodes
- `/health/ready` returned `200` for the local cluster after setting explicit non-default RPC/auth secrets
- current local cluster logs no longer show `storage_info failed`, `FixArray(1)`, `No valid auth token`, or `erasure read quorum`

## Impact
- Adds a durable regression guard for the issue #2815 MsgPack contract.
- Adds local-only Docker verification assets for distributed repro and follow-up debugging.
- Improves `Dockerfile.source` for local source-image builds on arm64/amd64 hosts.
- Documents and configures the non-default RPC secret requirement for internode auth in local distributed Docker validation.

## Additional Notes
- `docs/` is ignored on this branch, so the durable tracked Docker verification assets live under `.docker/test/issues-2815/`.
- Current code behavior is that internode RPC authentication must not resolve to the default secret `rustfsadmin`. If `RUSTFS_RPC_SECRET` is unset, the RPC signer falls back to `RUSTFS_SECRET_KEY`, so at least one of them must be set to a non-default shared secret for distributed RPC signing.


houseme:
Local Docker verification for issue #2815 is now green.

Key follow-up finding:
- current code rejects using the default secret `rustfsadmin` for internode RPC authentication
- if `RUSTFS_RPC_SECRET` is unset, RPC auth falls back to `RUSTFS_SECRET_KEY`
- therefore distributed local Docker validation must provide at least one non-default shared secret for RPC signing

The issue-2815 local verification assets now set:
- `RUSTFS_SECRET_KEY=issue-2815-secret`
- `RUSTFS_RPC_SECRET=issue-2815-rpc-secret`

With that in place, the 4-node local cluster reached healthy state and `/health/ready` returned `200`.

The current local logs no longer show:
- `storage_info failed`
- `wrong msgpack marker FixArray(1)`
- `No valid auth token`
- `erasure read quorum`

houseme:
Adopted. I strengthened the `storage_info_map_encoding_round_trip_matches_issue_2815_contract` test to assert that the top-level MsgPack payload starts with a map marker in addition to the round-trip field checks, so it now guards the encoding shape more directly.

majinghe (approved):

loverustfs (approved):

PR #2829, opened 2026-05-06
fix(replication): handle version ID format mismatch with AWS S3

ajax-bakun-n:
## Summary of Changes

1. RustFS uses UUID version IDs; AWS S3 uses alphanumeric strings. A versioned `HeadObject` to an AWS target returns HTTP 400 (root caller) or HTTP 403 (IAM user) with no response body, which previously aborted replication with a spurious error.
2. Add a `is_version_id_mismatch` helper that detects this condition (explicit `InvalidArgument` code, or bare 400/403 when the SDK has no response body to parse).
3. On mismatch, fall back to a versionId-less `HeadObject` via `head_object_fallback` and compare ETags — skip PUT if content already matches, otherwise proceed with PUT as normal.
4. HTTP 400/403 version-ID rejections are excluded from proxy failure metrics.

## Verification

Deployed from branch - replication works. Additionally added tests.

## Additional Notes

---


crates/ecstore/src/bucket/replication/replication_resyncer.rs
```diff
@@ -2536,16 +2571,31 @@ impl ReplicateObjectInfoExt for ReplicateObjectInfo {
...
-                        rinfo.error = Some(e.to_string());
-                        warn!("replication head_object failed bucket:{} arn:{} error:{}", bucket, tgt_client.arn, e);
-                        return rinfo;
+                if e.as_service_error().is_some_and(|se| se.is_not_found()) || has_raw_status(&e, 404) {
+                    // Object not on target yet → fall through to PUT.
+                } else if is_version_id_format_mismatch(&e) {
+                    // Version-ID format mismatch: retry without versionId and compare ETags.
+                    match head_object_fallback(&bucket, &tgt_client, &object).await {
+                        Ok(Some(oi)) => {
+                            replication_action = get_replication_action(&object_info, &oi, self.op_type);
```
overtrue: This fallback still passes the versionId-less HeadObject result into get_replication_action(), which compares the RustFS UUID with oi2.version_id below. For an AWS versioned bucket, an unversioned HeadObject can still return AWS' latest version ID, so equal content will keep producing ReplicationAction::All instead of skipping the PUT as described. Please either ignore version_id for this fallback comparison or use a dedicated ETag/size comparison for the mismatch path.

ajax-bakun-n: @overtrue, thank you for review! Added suggested changes

PR #2838, opened 2026-05-06
fix(storage): add RUSTFS_NETWORK_MOUNT_MODE for CIFS/NFS backends

### LINKED ISSUE #2790: [BUG] Scanner hangs indefinitely on network mount (CIFS/Storagebox) – regression from alpha.99 to beta.1
## Environment
- **RustFS Version:** v1.0.0-beta.1 (regression from 1.0.0-alpha.99)
- **Deployment:** Docker / docker-compose, single node
- **Storage Backend:** Hetzner Storagebox mounted via CIFS
- **OS:** Linux (Hetzner Cloud Server)
- **Bucket:** ~452,000 objects / 232 MiB (many small JSON files, ~530 bytes avg)

> **Note:** Searched existing issues – closest matches are 
> #1760, #2217, #2635 but none describe this specific 
> network mount regression in beta.1.

## Describe the Bug
After upgrading from **1.0.0-alpha.99 to v1.0.0-beta.1**, RustFS 
starts hanging indefinitely after running stably for several hours 
or days. All worker threads become blocked on `readmetadata` 
operations (30–60 seconds per single file), causing all S3 API 
requests to time out completely.

This is a **regression** – alpha.99 handled the exact same setup 
correctly and recovered automatically from transient network 
timeouts via temporary disk-offline → auto-recovery cycle.

The issue is **random/intermittent**: RustFS works fine for hours 
or even days, then suddenly freezes. This matches the NS scanner 
cycle – it works fine between scan cycles, then hangs when a new 
scan cycle starts and hits the slow CIFS mount.

## Steps to Reproduce
1. Mount a Hetzner Storagebox via CIFS as the RustFS data volume
2. Deploy RustFS v1.0.0-beta.1 via docker-compose
3. Have 400,000+ objects in a bucket (many small files)
4. Run for several hours/days
5. Observe all worker threads blocking indefinitely on `readmetadata`

## Expected Behavior
RustFS should handle transient CIFS/network timeouts gracefully 
and recover automatically, as it did in alpha.99.

## Actual Behavior
Worker threads block with `time.busy` values of 30–60 seconds 
per single `readmetadata` call. The NS scanner never completes 
its cycle. The directory size calculation times out after ~2500 
files. All S3 API requests hang until container is manually 
restarted.

## Relevant Log Output

WARN Directory size calculation timeout after 2561 files,
elapsed 6.113920736s, timeout 5.285969596s

WARN Failed to get size for directory data/rustfs
Custom kind TimedOut, error: Timeout after 2561 files

WARN Local disk operation timed out
endpoint: data/rustfs, op: walkdir, timeoutms: 5000

INFO close, time.busy=886s, time.idle=...
target: rustfs_ecstore::disk::local
span: getDiskInfo

INFO close time.busy=33.0s time.idle=5.31ms
target: rustfs_ecstore::disk::local
span: readmetadata
path: guilds/.../users/....jsonxl.meta

INFO walk dir err IoCustom kind Other,
error: disk operation timeout after 5s
WARN listpathraw fallback disk is none

## Root Cause Analysis
PR #2714 (`fix(storage): avoid faulting local drives on transient 
timeouts`) changed timeout handling behavior in beta.1. On local 
NVMe storage this fix is correct and beneficial. However on 
network mounts (CIFS/NFS), where individual metadata reads 
regularly take 30–60 seconds due to network latency, this causes 
the scanner to **retry indefinitely** instead of the previous 
behavior:

- **alpha.99 behavior:** timeout → disk marked temporarily offline 
  → auto-recovery after a few seconds → continues normally
- **beta.1 behavior:** timeout → treated as transient → infinite 
  retry → all worker threads blocked → full hang

## Workaround
Downgrade to `1.0.0-alpha.99` restores stable behavior immediately.

## Related Issues
- #1760 – Scanner stuck indefinitely on IO error files
- #2217 – S3 requests hang indefinitely under concurrent load
- #2620 – Scanner hangs indefinitely on same object path
- #2635 – Increase directory size calculation timeout

Mist7k:
Fixes #2790

When using network-mounted storage (CIFS/NFS), the drive operation timeouts introduced in beta.1 (5s defaults) are too aggressive for backends where individual metadata reads can take 30-60s due to network latency. This causes the scanner to repeatedly timeout, trigger the drive health state machine, and eventually block all worker threads.

This PR adds `RUSTFS_NETWORK_MOUNT_MODE=true` which:
- Raises all drive operation timeouts to 60s (metadata, walkdir, list_dir, disk_info, stall)
- Skips the drive health state machine on timeouts (no Suspect/Offline transitions)

This matches the behavior that was stable on network mounts prior to the per-operation timeout split.

**Usage:**
```
RUSTFS_NETWORK_MOUNT_MODE=true
```

Individual timeout overrides (`RUSTFS_DRIVE_METADATA_TIMEOUT_SECS` etc.) still take precedence when explicitly set.

**Testing:**
- Unit tests for env parsing, timeout override, and state machine bypass


Mist7k:

crates/ecstore/src/disk/disk_store.rs
```diff
@@ -67,35 +70,54 @@ fn get_drive_timeout_duration(env_key: &str, default_secs: u64) -> Duration {
     )
 }
 
+fn network_mount_mode() -> bool {
```
houseme: It’s sufficient to assign values accordingly based on the different environmental configurations. This approach reduces the need for unnecessary checks and also eliminates the need to introduce new environmental variables.

crates/ecstore/src/disk/disk_store.rs
```diff
@@ -67,35 +70,54 @@ fn get_drive_timeout_duration(env_key: &str, default_secs: u64) -> Duration {
...
+    rustfs_utils::get_env_bool(rustfs_config::ENV_NETWORK_MOUNT_MODE, rustfs_config::DEFAULT_NETWORK_MOUNT_MODE)
+}
+
 pub fn get_drive_metadata_timeout() -> Duration {
+    if network_mount_mode() {
```
houseme: It is recommended to reduce duplicate code.

loverustfs:
I still do not recommend implementing this feature. We will continue to explain this to users in the future.

loverustfs:
The risk is too high.

loverustfs (commented):
Merging is not recommended.

Mist7k:
A simple "Use at your Own Risk" is enough

Mist7k:
This flag is opt-in, disabled by default, and explicitly intended for 
operators who have already chosen to run RustFS on network-backed storage 
— a setup that worked fine prior to beta.1. 

If the concern is data integrity: the flag only raises timeouts and skips 
health state transitions on timeout — it does not disable error handling 
or checksums. Happy to add additional warnings to the docs if that helps.

loverustfs:
Hello @houseme @overtrue @weisd @majinghe ,

Do you have any suggestions?

houseme:
> Hello @houseme @overtrue @weisd @majinghe ,
> 
> Do you have any suggestions?

## Request changes (not recommended to merge)

Thank you for the detailed reproduction and thoughtful proposal tied to #2790. The issue is real, but this PR is not aligned with RustFS’s product boundary and failure model.

### 1) Product-boundary conflict (Blocker)
RustFS is positioned for high-performance, predictable-latency object storage on direct-attached local POSIX storage (typically local SSD/NVMe), not CIFS/NFS/GPFS network filesystems.  
In issue #2790, maintainers explicitly stated on 2026-05-07 that this direction is too risky and not recommended.  
Introducing `RUSTFS_NETWORK_MOUNT_MODE` as a first-class env plus Helm value creates an implicit “official support path” signal that conflicts with that boundary.

### 2) Claimed behavior vs actual behavior mismatch (Blocker)
The PR description says timeout handling will skip the drive health state machine.  
In code, only the operation-timeout path in `track_disk_health_with_op` is bypassed. Active health checks and recovery probes still call `mark_failure(...)` and can transition runtime state.  
So the behavior is not consistently “state-machine bypass on timeouts,” which makes operator expectations inaccurate.

### 3) Failure model is weakened in exchange for longer stalls (Blocker)
This mode raises multiple critical operation timeouts to 60s and suppresses timeout-based fault transitions on selected paths.  
That delays real fault detection, increases tail latency, and extends blocked windows—opposite to RustFS’s core operational goals (performance + explicit fault boundaries).  
Even as opt-in, once exposed in official chart paths, support burden and user expectations increase.

### 4) Config contract/documentation is incomplete (Needs fix)
`crates/config/README.md` currently documents existing drive timeout knobs but does not establish a complete contract for this new mode (precedence, risk, applicability, unsupported status).  
Without consensus on support boundaries, this should not be promoted via Helm surface area first.

## Acceptable alternative path
1. Close this PR (do not merge as a product feature).
2. Implement the guardrail already discussed in #2790: detect and warn (or fail fast) on network-mounted backends at startup.
3. For users who still run unsupported setups, recommend explicit tuning via existing timeout env vars only (clearly marked unsupported / at your own risk), instead of adding a backend mode that changes health-state semantics.

Your investigation quality is excellent and valuable; the fix strategy just needs to remain consistent with RustFS’s support policy and reliability model.

Mist7k:
Thanks for the detailed review, @houseme

Regarding point 3 — the suggestion to use existing per-operation timeout env vars (RUSTFS_DRIVE_WALKDIR_TIMEOUT_SECS, RUSTFS_DRIVE_METADATA_TIMEOUT_SECS, etc.) instead of RUSTFS_NETWORK_MOUNT_MODE — I want to clarify why that approach doesn't fully address the problem.

Longer timeouts alone don't fix the failure model.

In my logs, operations are completing in 122s, 886s, even 974s on a CIFS/Storagebox backend. Bumping the timeout vars to 60s would simply mean those operations time out later — but they still time out, and mark_failure(...) still gets called from health checks and recovery probes on other paths, regardless of the per-op timeout setting.

The root issue is that RustFS's health state machine was designed around predictable local NVMe latency. On a network backend, high latency is expected and normal, not a signal of disk failure. No timeout value change addresses that semantic mismatch — the drive still gets marked unhealthy and the instance degrades.

RUSTFS_NETWORK_MOUNT_MODE=true is intentionally a single ergonomic wrapper that:

Sets all relevant timeouts in one place (no partial misconfiguration risk across 6+ vars)

Disables state-machine fault transitions for network-latency-induced stalls

That said, I'm open to discussing the boundary concern further. If the issue is the Helm surface area, I could move this to an explicit "unsupported/use-at-your-own-risk" section of the chart.

houseme:
You can try compiling the package from the main branch, as there have been updates to the disk health detection logic. @Mist7k

Mist7k:
Thanks for the thorough review, @houseme.

Addressed all points in the latest push:

**Point 2 (behavior mismatch):** Fixed. `network_mount_mode()` now consistently bypasses `mark_failure()` in all three paths: `track_disk_health_with_op` (operation timeouts), `monitor_disk_writable` (active health checks), and `monitor_disk_status` (recovery probes). The state machine is fully bypassed when the flag is set.

**Point 3 (failure model):** This is intentional and opt-in. On a network backend, high latency is expected and normal — not a signal of disk failure. The 60s timeout is a floor, not a ceiling: explicit per-operation overrides still take precedence (fixed in previous commit). Operators who set this flag explicitly acknowledge they are outside the supported configuration.

**Point 4 (documentation):** Happy to add a doc comment and README entry if the team prefers. I kept it minimal to avoid implying official support.

**Regarding the alternative path:**
- Recommending existing timeout env vars alone doesn't solve the problem — even with 120s timeouts, `mark_failure()` still fires on every timeout and transitions the drive through Suspect → Offline. The state machine is the root issue, not the timeout values.
- Removing the Helm surface area is fine with me if that's the concern. The env var works without it.

The flag is opt-in, off by default, zero impact on the standard local-storage path. It simply stops RustFS from actively breaking on a setup that previously worked.

Mist7k:
Rebased on current main and squashed into a single commit. Resolves merge conflicts with the recent disk health detection changes.

Mist7k:
Rebased on current main and refactored to build on top of #2996.

Instead of introducing a parallel `network_mount_mode()` check in the timeout path, this now reuses the `TimeoutHealthAction::IgnoreFailure` mechanism that was just merged. When `RUSTFS_NETWORK_MOUNT_MODE=true`, `read_metadata`, `list_dir`, and `disk_info` are routed through `IgnoreFailure` — same pattern #2996 applied to `walk_dir`.

Diff is significantly smaller as a result. The flag is still opt-in and off by default; standard local-storage path is unchanged.

Drive timeout overrides remain (60s default in network mount mode, explicit `RUSTFS_DRIVE_*_TIMEOUT_SECS` take precedence).

cxymds:
Based on my judgment, the current 5-second setting is somewhat aggressive. However, the changes in the PR cannot fundamentally solve this problem, and optimization needs to be done from other aspects.

loverustfs:
@Mist7k Many developers do not agree to merge this PR.
So we'll close this.

Mist7k:
@loverustfs @cxymds @houseme 

I'd like to respectfully push back on this decision.

The argument that this PR "cannot fundamentally solve the problem" is technically correct — but it misses the point. There are currently no system conflicts introduced by these changes. The PR does not break anything for users on local storage, and it does not change any default behavior.

For users on network-mounted storage, this is currently the best available fix. Not a perfect one — but the only one that actually prevents the scanner from triggering mark_failure() on disks that are perfectly healthy. The existing env variables don't cover this path. The recent commits on main don't cover this path. Nothing else currently does.

Closing a PR that works, introduces no regressions, and fills a gap that no other fix currently addresses — while there is no concrete alternative on the roadmap — is a decision I genuinely don't understand.

Even as a temporary measure until a more comprehensive solution is found, this PR has value. "Not the perfect fix" and "should be closed" are two very different things.

I'm happy to incorporate any improvements, revised behavior, or new approaches into this PR if the team has a better direction in mind. But simply closing it without any argumentation about what could be done better, or what an alternative solution would look like, is unprofessional. Affected users deserve more than a closed PR and no path forward.

loverustfs:
> @loverustfs @cxymds @houseme 
> 
> I'd like to respectfully push back on this decision.
> 
> The argument that this PR "cannot fundamentally solve the problem" is technically correct — but it misses the point. There are currently no system conflicts introduced by these changes. The PR does not break anything for users on local storage, and it does not change any default behavior.
> 
> For users on network-mounted storage, this is currently the best available fix. Not a perfect one — but the only one that actually prevents the scanner from triggering mark_failure() on disks that are perfectly healthy. The existing env variables don't cover this path. The recent commits on main don't cover this path. Nothing else currently does.
> 
> Closing a PR that works, introduces no regressions, and fills a gap that no other fix currently addresses — while there is no concrete alternative on the roadmap — is a decision I genuinely don't understand.
> 
> Even as a temporary measure until a more comprehensive solution is found, this PR has value. "Not the perfect fix" and "should be closed" are two very different things.
> 
> I'm happy to incorporate any improvements, revised behavior, or new approaches into this PR if the team has a better direction in mind. But simply closing it without any argumentation about what could be done better, or what an alternative solution would look like, is unprofessional. Affected users deserve more than a closed PR and no path forward.

I can especially understand your feelings, and I think you are right to do so.

However, the compatibility with the network file system will make the user's risk very great.

We don't want users to use the network file system.

Mist7k:
@loverustfs 
This PR does not introduce any new risk for network file system users — it restores the behavior that existed in alpha.89 and earlier, which ran stably on network mounts with zero issues reported. No new code paths are added, no new failure modes introduced. The only thing this PR does is prevent a regression that Beta.1 caused. There is no risk to revert.

Beyond that — MinIO, Garage, and other S3-compatible solutions handle network mounts without issues. "Network storage is risky" is not a universal truth, it is a consequence of how Beta.1 changed the scanner behavior.

loverustfs:
We're exploring a more reliable implementation, but unfortunately we can't give a definite answer at this moment. We all agree that the current solution isn't the direction we want to take. Sorry again for the inconvenience, and thank you for your patience.


https://github.com/rustfs/rustfs/issues/2790#issuecomment-4378629635
https://github.com/rustfs/rustfs/issues/2790#issuecomment-4384206055

weisd (approved):

crates/ecstore/src/disk/disk_store.rs
```diff
@@ -60,17 +60,38 @@ lazy_static::lazy_static! {
...
-    Duration::from_secs(rustfs_utils::get_env_u64(
-        rustfs_config::ENV_DRIVE_MAX_TIMEOUT_DURATION,
-        rustfs_config::DEFAULT_DRIVE_MAX_TIMEOUT_DURATION_SECS,
-    ))
+    let explicit = rustfs_utils::get_env_opt_u64(rustfs_config::ENV_DRIVE_MAX_TIMEOUT_DURATION);
+    if let Some(v) = explicit {
+        return Duration::from_secs(v);
+    }
+    if network_mount_mode() {
+        return Duration::from_secs(rustfs_config::NETWORK_MOUNT_DRIVE_TIMEOUT_SECS);
```
houseme: Do you need to compare the sizes?

crates/ecstore/src/disk/disk_store.rs
```diff
@@ -60,17 +60,38 @@ lazy_static::lazy_static! {
...
-        rustfs_utils::get_env_opt_u64_with_aliases(env_key, &[rustfs_config::ENV_DRIVE_MAX_TIMEOUT_DURATION])
-            .unwrap_or(default_secs),
-    )
+    let explicit =
+        rustfs_utils::get_env_opt_u64_with_aliases(env_key, &[rustfs_config::ENV_DRIVE_MAX_TIMEOUT_DURATION]);
+    if let Some(v) = explicit {
+        return Duration::from_secs(v);
+    }
+    if network_mount_mode() {
+        return Duration::from_secs(rustfs_config::NETWORK_MOUNT_DRIVE_TIMEOUT_SECS);
```
houseme: Do you need to compare the sizes?

houseme (changes requested):
If a value has been set for the corresponding key, and `network_mount_mode` is enabled later, the setting may still not take effect. Please verify the situation further and optimize the code accordingly.

Mist7k:
Addressed all review feedback:

**@houseme "Do you need to compare the sizes?":**
The `network_mount_mode()` check acts as a fallback default — explicit per-operation env vars (`RUSTFS_DRIVE_*_TIMEOUT_SECS`) always take precedence. If an operator sets both `RUSTFS_NETWORK_MOUNT_MODE=true` and `RUSTFS_DRIVE_METADATA_TIMEOUT_SECS=120`, the 120 wins. The 60s is only applied when no explicit override is present.

**@houseme "setting may still not take effect":**
This is the intended behavior documented in the PR description. The precedence order is: explicit env override > network_mount_mode default > compiled default. Added a test (`network_mount_mode_explicit_override_takes_precedence`) that verifies this.

**Copilot (High) — active health check probe path:**
Fixed. `get_drive_active_check_timeout()` now also respects `network_mount_mode()` (60s instead of 5s). Additionally, `monitor_disk_writable` and `monitor_disk_status` now bypass `mark_failure()` when network mount mode is enabled, preventing the active probe path from transitioning drives to Suspect/Offline.

**Copilot (Medium) — incomplete test coverage:**
Fixed. `network_mount_mode_overrides_drive_timeouts` now asserts all 7 timeout getters: metadata, list_dir, disk_info, walkdir, walkdir_stall, max_timeout, and active_check.

**Copilot (Low) — Helm placement:**
Fixed. Moved `network_mount_mode` from `config.rustfs.scanner` to `config.rustfs` (top-level), since it affects drive behavior globally.

Mist7k:
@houseme 

please retry

overtrue (changes requested):
Requesting changes because network_mount_mode still suppresses mark_failure for all active health-check and recovery-probe failures, not just timeout-specific failures. That can hide real I/O errors and leave disk state/metrics inaccurate. Please keep non-timeout failures in the health state machine, and make the env-mutating timeout tests hermetic by unsetting every timeout env var they assert against.

PR #2846, opened 2026-05-07
fix(iam): keep error state on initial load failure

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Ensure IAM initialization fails closed when the initial cache load cannot complete after all retry attempts.

  Previously, `init` could mark IAM as ready even after the initial `load` failed on every retry. This change preserves
  the final load error, returns it from initialization, and keeps the IAM state as `Error`.

  A focused regression test was added to verify that repeated initial load failures:
  - return the underlying error
  - `git diff --check`

  Not run due to local memory constraints:
  - `make pre-commit`

  ## Impact

  IAM now fails closed during startup if its initial data load fails after all retries. This prevents the system from
  reporting IAM readiness with an incomplete or unavailable in-memory IAM cache.

  No API, configuration, migration, or documentation changes are expected.

  ## Additional Notes

  `make pre-commit` still needs to pass before this PR should be marked ready for review.


houseme (approved):
This PR is aligned with RustFS security expectations for IAM bootstrap and correctly implements fail-closed behavior.

What I verified:
1. Constructor path now propagates initialization failures instead of panicking in non-test code:
   - `IamCache::new` returns `Result<Arc<Self>>`
   - `init_iam_sys` uses `?` to bubble the error
2. Initial-load retry exhaustion now preserves and returns the final underlying error, while keeping IAM state as `Error` (not `Ready`).
3. Regression coverage is in place for repeated initial-load failure and validates:
   - error is returned
   - readiness remains false
   - state remains `Error`
4. Test-path adjustment is correct (`unwrap` remains test-only), consistent with crate guidance to avoid panic-driven control flow in runtime paths.

Non-blocking optimization suggestions:
1. Consider classifying retryable vs non-retryable bootstrap errors to fail faster on deterministic configuration/format failures.
2. Consider adding a startup metric/log field that explicitly records “initial IAM load attempts exhausted” for easier production diagnosis.
3. Consider a focused integration test at `init_iam_sys` boundary to assert global singleton is not set when bootstrap fails.

Overall: LGTM, behavior is safer and matches RustFS fail-closed IAM expectations.

Issue #2775, opened 2026-05-02
RustFS alpha.94 LIST recursion bug


SyntaxErr0r42:
RustFS alpha.94 has a LIST recursion bug that caps single-prefix enumeration at ~6888 entries (breaks Harbor GC/catalog).

Empirically, `mc ls --recursive rustfs/<bucket>/<prefix>` returns at most ~6888 entries even when the prefix actually contains more. Per-object `mc stat` returns the "missing" objects fine, the data is on disk and readable but LIST drops them.

Discovered while migrating a Harbor registry bucket currently on MinIO Community edition (RELEASE.2025-09-07) with 22933 objects. LIST under another bucket with 13575 objects also returned only 6888. Sub-prefixes returning <10K each enumerated correctly.

**Why this blocks Harbor**: Harbor's garbage collector traverses LIST. Objects invisible to LIST get marked unreferenced and deleted. Catalog API depends on LIST too.

loverustfs:
Please upgrade to the latest version before testing.

gabrialm:
---
name: RustFS — LIST cap blocks Harbor migration (alpha.94 → beta.1)
description: RustFS is not viable as a Harbor S3 backend on lab1. beta.1 (2026-04-29) fixed the alpha.95/.96 FaultyDisk regression (#2601), but a hardcoded ~6,888-key cap on a single recursive LIST persists in beta.1 with the exact same ceiling.
type: reference
originSessionId: 1ff6295f-cf21-47f9-9696-fafd2f7c7ec3
---
## Bug 1 — FaultyDisk regression (#2601) — FIXED in beta.1
On `1.0.0-alpha.95`/`.96` with single-volume `RUSTFS_VOLUMES=/path`, after some sustained write load (~26 GiB into a 22879-object harbor mirror), every read and write started returning HTTP 500 with `"erasure write quorum"` / `"Storage resources are insufficient for the read operation"`. Internal log: `not enough disks to write: [Some(FaultyDisk)]`. RustFS marked the (healthy) disk FaultyDisk on its own.

**Status (2026-05-02)**: re-tested against `1.0.0-beta.1` (build 2026-04-29). The full 39.80 GiB / 23,533-object harbor mirror completed cleanly in 2m02s at 333 MiB/s with no FaultyDisk events and no error-priority journal entries. **FIXED.**

Tracking: https://github.com/rustfs/rustfs/issues/2601.

[...]

Issue #2916, opened 2026-05-11
Streaming file listing cuts out without any error and without returning full list


XANi:
On directories of few thousand files, `minio-client ls --recursive local/srv-dc1-d1-gence-filer1/restic/data` will only return first few hundred files with no notion of any error happening.

that also causes some backup tools like `restic` to trip and fail because they think they got full directory listing, leading to essentially a corrupted repository

It seems to be very erratic, some directories get few hundred, others few thousand files.  For example:
```
rustfs@d1-backup1:~$ /usr/local/bin/minio-client ls --recursive local/srv-dc1-dc1-metrics 
[2026-03-27 19:31:56 CET]   155B STANDARD restic/config
[2026-05-09 15:31:51 CEST]  16MiB STANDARD restic/data/00/0000a6ebee7a5a4ac964e8b0447b98eb6c267efa2f0df193823f18559af3b086
[2026-05-09 15:48:59 CEST]  16MiB STANDARD restic/data/00/00021d6316f804814d2e05575799cbe71e6d5aa7bc40a1195b2455294d9d5312
[2026-04-02 22:05:14 CEST]  16MiB STANDARD restic/data/00/0002525b493a6ee5d77c6b99718c3586b10e6aa637962f8ddbfddc1d87a238fb
...
[2026-03-28 07:43:32 CET]  18MiB STANDARD restic/data/07/07ed3c7d29d48ff769d7b55e4299f4f6d665ef2de9df73179337fcadb6201fd4
[2026-05-09 15:18:21 CEST]  16MiB STANDARD restic/data/07/07ed63c4878f1f024cbdd362d2b0d003331a892c919cc774d61022983267f48c
[2026-04-01 09:10:45 CEST]  16MiB STANDARD restic/data/07/07edd1c98981f4b32f2f82b0bab6b7de3654d58b268becc9241f2f6487c6b24b
[2026-03-28 07:57:11 CET]  16MiB STANDARD restic/data/07/07eeaca7aa6f3b13e47bb66b3e2774330412ee52898794d541180165a31c06f6
[2026-05-05 03:51:27 CEST]  16MiB STANDARD restic/data/07/07effd50c45a3cbe0557533daa7c9eb2c669abc99fa7a48d8f8b5877e76b26a5
rustfs@d1-backup1:~/data/srv-dc1-dc1-metrics/restic/data$ 

```
while directory goes all the way to `/ff/`
```
[...]

PR #2919, opened 2026-05-11
fix(ecstore): preserve list marker set index

marshawcoco:
## Related Issues

  Related to #2916 and #2775.

  ## Summary of Changes

  Fixes list continuation marker encoding so the `s` marker tag preserves `set_idx` instead of incorrectly reusing `pool_idx`.

  Adds a regression test that round-trips a marker with different pool and set indexes to ensure pagination resumes with the original set index.

  ## Verification

  - `cargo fmt --all` - passed
  - `cargo fmt --all --check` - passed
  - `cargo test -p rustfs-ecstore list_path_marker_round_trip_preserves_set_index` - passed
  - `cargo test -p rustfs-ecstore` - failed on existing Windows path/local disk tests unrelated to this marker change

  ## Impact

  No API, configuration, migration, or deployment changes.

  The change improves continuation marker correctness for listing pagination when pool and set indexes differ.

  ## Additional Notes

  `make pre-commit` has not been run yet, so this PR should remain draft or blocked until the full required gate passes.

### DIFF
--- a/crates/ecstore/src/store_list_objects.rs
+++ b/crates/ecstore/src/store_list_objects.rs
@@ -1488,7 +1488,11 @@ mod test {
         };
 
         let marker = opts.encode_marker("photos/2026/image.jpg");
-        assert_eq!(marker, "photos/2026/image.jpg[rustfs_cache:v1,id:list-cache-id,p:3,s:7]");
+        let expected_marker = format!(
+            "photos/2026/image.jpg[rustfs_cache:{},id:list-cache-id,p:3,s:7]",
+            super::MARKER_TAG_VERSION
+        );
+        assert_eq!(marker, expected_marker);
 
         let mut parsed = ListPathOptions {
             marker: Some(marker),
--- a/crates/ecstore/src/store_list_objects.rs
+++ b/crates/ecstore/src/store_list_objects.rs
@@ -213,7 +213,7 @@ impl ListPathOptions {
                 MARKER_TAG_VERSION,
                 id.to_owned(),
                 self.pool_idx.unwrap_or_default(),
-                self.pool_idx.unwrap_or_default(),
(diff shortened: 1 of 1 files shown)


houseme (approved):

XANi:
I've checked it and problem is still here
```
rust version : rustc 1.95.0 (59807616e 2026-04-14)
rust channel : stable-x86_64-unknown-linux-gnu
git branch   : fix/list-marker-set-index
git commit   : e4e8f6dedf539cb1068a46e04b46b98479dd9b01
```
result:
```
rustfs@d1-backup1:~/data/srv-dc1-dc1-metrics/restic/data$ for a in {1..10} ; do /usr/local/bin/minio-client ls --recursive local/srv-dc1-dc1-metrics  |wc -l ; done
4965
5514
5817
6581
7432
7835
8591
9600
9986
```
server have few HDDs in RAID setup

marshawcoco:
Thanks for testing this.@XANi

I think this reproduction is hitting a different listing path from this PR. This PR only fixes the continuation marker round-trip bug where `encode_marker()` wrote `pool_idx` into the `s:` field instead of `set_idx`.

The remaining issue appears to be in recursive directory walking: `LocalDisk::scan_dir()` currently uses `trim_start_matches(current)` when deriving `forward_to`, which can over-strip repeated prefix components such as `engineering/
engineering/...` and skip child directories. That is the issue being addressed separately in SamuraJey/rustfs#1.

So I agree this PR alone does not fully resolve #2775; it fixes one related marker bug, while the repeated-prefix recursion issue needs the separate `scan_dir` fix.

XANi:
@marshawcoco thanks for the info, I'll look into that PR

SamuraJey:
Yeah, i am working on fixing 2775, hope to finish soon

PR #2923, opened 2026-05-11
Fix #2775 recursive list handling in LocalDisk::scan_dir()

### LINKED ISSUE #2775: RustFS alpha.94 LIST recursion bug

SamuraJey:
## Related Issues

  Fixes #2775.

  ## Summary of Changes

  Fixed recursive object listing pagination when a continuation marker points into a path whose component repeats the current prefix, for example engineering/engineering/....

  The bug was caused by deriving the forward_to child component with prefix trimming semantics that could remove repeated path components and skip the next directory during recursive LIST traversal.

  Changes include:

  - Corrected scan_dir continuation forwarding so it strips only the current directory prefix.
  - Added a focused scan_dir unit regression for repeated prefix components.
  - Added a S3 ListObjectsV2 e2e regression test.

  ## Verification

  - cargo fmt --all --check
  - cargo test -p rustfs-ecstore test_scan_dir_forward_to_repeated_prefix_component -- --nocapture
  - cargo test -p e2e_test test_list_objects_v2_repeated_prefix_continuation -- --nocapture
  - make pre-commit

  ## Impact

  Recursive ListObjectsV2 and marker-based listing now return all matching objects across continuation pages when object keys contain repeated prefix components.

  No API, configuration, or deployment changes are expected.

  ## Additional Notes

  PR #2919 fixes a separate marker round-trip issue and does not resolve #2775 by itself. The small e2e regression added here to reproduce case mentioned in issue.

### DIFF
--- a/crates/e2e_test/src/list_objects_v2_pagination_test.rs
+++ b/crates/e2e_test/src/list_objects_v2_pagination_test.rs
@@ -154,9 +154,20 @@ mod tests {
 
         create_bucket(&client, bucket).await.expect("Failed to create bucket");
 
-        let expected_keys: Vec<String> = (0..5)
-            .map(|idx| format!("{prefix}engineering/project-{idx:03}/artifact.txt"))
-            .collect();
+        let expected_keys = vec![
+            format!("{prefix}alpha-000/artifact.txt"),
+            format!("{prefix}engineering/engineering/project-000/artifact.txt"),
+            format!("{prefix}engineering/engineering/project-001/artifact.txt"),
+            format!("{prefix}engineering/project-000/artifact.txt"),
+            format!("{prefix}engineering/project-001/artifact.txt"),
+            format!("{prefix}engineering/project-002/artifact.txt"),
+            format!("{prefix}zulu-000/artifact.txt"),
+        ];
+        let noise_keys = [
+            "different/prefix/prefix/project-000/artifact.txt",
+            "engineering-other/project-000/artifact.txt",
+            "unrelated/engineering/project-000/artifact.txt",
+        ];
 
         for key in &expected_keys {
             client
@@ -168,10 +179,24 @@ mod tests {
                 .await
                 .expect("Failed to put object");
         }
+        for key in noise_keys {
+            client
+                .put_object()
+                .bucket(bucket)
+                .key(key)
+                .body(ByteStream::from_static(b"x"))
+                .send()
+                .await
+                .expect("Failed to put noise object");
+        }
 
         let listed_keys = list_all_objects_v2(&client, bucket, prefix, PAGE_SIZE).await;
         let seen: HashSet<String> = listed_keys.iter().cloned().collect();
 
+        assert_eq!(
+            listed_keys, expected_keys,
+            "Issue #2775 regression: repeated-prefix pagination must return exactly the expected keys in lexicographic order"
+        );
         assert_eq!(
             listed_keys.len(),
             expected_keys.len(),
--- a/crates/e2e_test/src/list_objects_v2_pagination_test.rs
+++ b/crates/e2e_test/src/list_objects_v2_pagination_test.rs
@@ -30,7 +30,6 @@ mod tests {
     use crate::common::{RustFSTestEnvironment, init_logging};
     use aws_sdk_s3::Client;
     use aws_sdk_s3::primitives::ByteStream;
-    use aws_sdk_s3::types::EncodingType;
     use serial_test::serial;
     use std::collections::HashSet;
     use tracing::info;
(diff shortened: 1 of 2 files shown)


SamuraJey:
@XANi Can you please check your test cases against my fix?

XANi:
@SamuraJey no change.

If I drop cache in the middle of the checks  I get smaller number of listed files

```
 for a in {1..10} ; do /usr/local/bin/minio-client ls --recursive local/srv-dc1-dc1-metrics  |wc -l ; done
158
303
470
625
766
924
1207
# about here I was running  echo 3 > /proc/sys/vm/drop_caches
163 
311
484
```

they all also seem to run on timeout as they exit after 5-6 seconds mark at first

```
~/data/srv-dc1-dc1-metrics/restic/data$ for a in {1..10} ; do /usr/bin/time --format='total %es' /usr/local/bin/minio-client ls --recursive local/srv-dc1-dc1-metrics  |wc -l ; done

total 5.98s
134
total 5.06s
273
total 5.11s
430
total 5.13s
583
total 5.13s
736
total 5.19s
886
total 10.23s
1229
total 6.04s
1404
total 6.04s
1598
total 6.09s
1776
```
but after few runs it takes a bit longer
```
~/data/srv-dc1-dc1-metrics/restic/data$ for a in {1..10} ; do /usr/bin/time --format='total %es' /usr/local/bin/minio-client ls --recursive local/srv-dc1-dc1-metrics  |wc -l ; done
total 15.54s
7427
total 11.26s
7768
total 16.13s
8426
total 12.43s
8814
total 16.45s
9459
total 13.04s
9817
total 17.34s
10447
total 13.88s
10822
total 17.48s
11459
total 14.78s
11846
```

SamuraJey:
I read your issue #2916 and tried to repeat the experiment. I have a few ideas, but it seems the error you are encountering isnt caused by the problem I m fixing in this pr. I think that there is a problem somewhere around list_path_raw() in metacache_set.rs, but I'm not sure.

The listing may be hitting the server-side walk_dir timeout and returning a partial successful ListObjectsV2 response.

Could you please collect the following debug info?

1. RustFS server logs around one affected `minio-client ls --recursive` run, especially lines containing:
   - `Local disk operation timed out`
   - `op="walk_dir"` or `op=walk_dir`
   - `list_path_raw: fallback disk is none`
   - `list_path_raw reader peek timed out`
   - `walk dir err`
   - `list_merged err`
   - `gather_results err`

2. The relevant timeout configuration/env values:
   - `RUSTFS_DRIVE_WALKDIR_TIMEOUT_SECS`
   - `RUSTFS_DRIVE_WALKDIR_STALL_TIMEOUT_SECS`
   - `RUSTFS_DRIVE_MAX_TIMEOUT_DURATION`
   - any other `RUSTFS_DRIVE_*TIMEOUT*` vars

3. One `minio-client` debug run for the failing listing:
   ```bash
   /usr/local/bin/minio-client --debug ls --recursive local/srv-dc1-dc1-metrics > /tmp/list.out 2> /tmp/list.debug
   wc -l /tmp/list.out
   grep -E 'list-type=2|HTTP/|Response Time|IsTruncated|NextContinuationToken|<Error>' /tmp/list.debug
   ```
   If the debug output is large, the last few ListObjectsV2 requests/responses are enough.

4. A comparison run with a larger walkdir timeout, if possible. This one is worked for my local environment:
   ```bash
   RUSTFS_DRIVE_WALKDIR_TIMEOUT_SECS=30 rustfs ...
   ```
   Then repeat:
   ```bash
   /usr/bin/time --format='total %es' /usr/local/bin/minio-client ls --recursive local/srv-dc1-dc1-metrics | wc -l
   ```
   We want to know whether the count becomes stable/full when the timeout is increased.

5. Approximate bucket shape:
   - total expected object count if known;
   - whether the bucket contains many empty directories / directory marker objects;
   - whether the listed prefix contains deep nested paths;

We want to confirm that RustFS logs a `walk_dir` timeout around the same time the client receives HTTP 200 with a smaller than expected object count and no further continuation token.

XANi:
> RustFS server logs around one affected minio-client ls --recursive run, especially lines containing:

There are no new entries generated during the test

> The relevant timeout configuration/env values:

None of those are defined so they are at default values. Are they documented somewhere ? Regardless of the bug I probably should increase it for HDD setup.

> One minio-client debug run for the failing listing:
this is log:
[debug.log](https://github.com/user-attachments/files/27627971/debug.log)

this is raw http transaction:

[http.txt](https://github.com/user-attachments/files/27627978/http.txt)

> We want to know whether the count becomes stable/full when the timeout is increased.

first run took a bit over 30s and returned partial with no error
```
total 31.03s
922
```
second and subsequent run appears to read all of them
```
total 690.78s
62635
total 55.25s
62635
total 46.97s
62635
```


>Approximate bucket shape:

>    total expected object count if known;
>    whether the bucket contains many empty directories / directory marker objects;
>    whether the listed prefix contains deep nested paths;

it's 256 directories, 00 to ff containing a bunch of ~16MB chunks each (it's a restic backup, it chunks the files into up to 16MB parts for deduplication purpose). This particular one has 62k files, other big one is around 100k.

It is hosted on RAID6 array of 5 20TB hard drives so not that quick IO

I found the bug when I ran some repo verification checks and one of them basically streamed the list of chunks to compare it with indexes.

The lack of error when some timeout is hit is really problematic here, because the application just assumes the files vanished, removes them from index and marks the backup jobs that used them as corrupted.

SamuraJey:
Okay, i am pretty sure, that your issue is not relevant to this particular PR. But i think we gathered some valuable information about it and probably made easier to fix it

XANi:
Yeah I'll continue any feedback etc. in my original ticket

houseme (approved):

weisd (approved):

PR #2958, opened 2026-05-14
fix(ecstore): use hex sha256 for delete objects

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Fixes the SHA256 payload hash used when signing multi-object delete requests.

  `RequestMetadata.content_sha256_hex` is later written to `X-Amz-Content-Sha256`, so it must contain the lowercase hex
  SHA256 digest expected by SigV4. The multi-object delete path was instead base64-encoding the SHA256 bytes, while
  already using base64 only where appropriate for `Content-Md5`.

  Changes:
  - Added a small helper for lowercase hex SHA256 encoding.
  - Updated `DeleteObjects` request metadata to pass the hex SHA256 body digest.
  - Added a regression test that verifies the generated delete body hash is 64-byte lowercase hex and not base64.

  ## Verification

  - `cargo fmt --all --check`
  - `cargo test -p rustfs-ecstore multi_object_delete_sha256_is_lowercase_hex --lib`
  - `cargo test -p rustfs-ecstore client::api_remove --lib`
  - `git diff --check`

  ## Impact

  Improves SigV4 compatibility for multi-object delete requests. No API, configuration, or migration changes.

  ## Additional Notes
  N/A.


houseme (approved):

houseme (approved):
Thanks for the focused fix.

I verified the change path in `api_remove` and `transition_api`: this correctly switches the DeleteObjects payload hash from base64 to lowercase hex before `X-Amz-Content-Sha256` is signed/sent, which aligns with SigV4 expectations.

LGTM and approved.

Follow-up suggestion (non-blocking): add one regression assertion that validates the request metadata/header path for this DeleteObjects flow, so a future accidental fallback to base64 in assignment wiring is caught directly.

PR #2974, opened 2026-05-15
fix: handle Windows paths in pre-commit tests

### LINKED ISSUE #2963: make pre-commit fails locally on Windows due lint and platform-specific tests
**Describe the bug**
  Running `make pre-commit` on Windows fails on current `main` due unrelated lint and platform-specific test failures. This blocks local validation for otherwise unrelated changes.

  **To Reproduce**
  1. Use a Windows development environment.
  2. Check out the latest `main`.
  3. Run:
     ```bash
     make pre-commit

  4. See failures including:
      - crates/io-core/src/io_profile.rs dead-code under -D warnings
      - Windows path expectation failures where tests expect Unix-style paths such as /tmp/data but receive Windows paths such as C:/tmp/data
      - Rustls CryptoProvider default provider initialization failure in some tests

  **Expected behavior**
  make pre-commit should pass on Windows, or Windows-incompatible checks should be gated/skipped with clear justification so contributors can validate unrelated changes locally.

  **Screenshots**
  N/A

   - OS: Windows
   - Browser: N/A
   - Version: N/A

   - Device: N/A
   - OS: N/A
   - Browser: N/A
   - Version: N/A
  
  **Additional context**
  This was observed while validating an unrelated rustfs-ecstore regression test. The focused verification passed:

  cargo fmt --all --check
  cargo test -p rustfs-ecstore test_get_object_reader_decrypts_managed_sse_range_on_plaintext_semantics --lib
  git diff --check

  The make pre-commit failure appears unrelated to that local change.

marshawcoco:
## Related Issues

  Fixes #2963

  ## Summary of Changes

  - Fix Windows path and URL endpoint parsing differences
  - Update tests that assumed Unix absolute paths, `/dev/null`, filesystem type values, or immediate file size refreshes
  - Gate Unix-only SFTP host key permission checks to Unix platforms
  - Fix hardcoded Unix absolute paths in MySQL and Redis target tests
  - Clean up non-Linux/macOS clippy dead code and unused import issues

  ## Verification

  - `make pre-commit`
  - `cargo clean`

  ## Impact

  Improves Windows compatibility for local pre-commit checks without changing the intended Unix/macOS behavior.

### DIFF
--- a/crates/ecstore/src/disk/endpoint.rs
+++ b/crates/ecstore/src/disk/endpoint.rs
@@ -82,17 +82,22 @@ impl TryFrom<&str> for Endpoint {
                 #[cfg(not(windows))]
                 let path = Path::new(&path).absolutize()?;
 
-                // On windows having a preceding SlashSeparator will cause problems, if the
-                // command line already has C:/<export-folder/ in it. Final resulting
-                // path on windows might become C:/C:/ this will cause problems
-                // of starting rustfs server properly in distributed mode on windows.
-                // As a special case make sure to trim the separator.
                 #[cfg(windows)]
-                let path = Path::new(&path[1..]).absolutize()?;
+                let path = if has_leading_slash_windows_drive(&path) {
+                    // Url::path() exposes file-like Windows paths as `/C:/...`.
+                    // Strip only that synthetic leading slash; plain URL paths
+                    // such as `/export1` must stay URL paths, not become
+                    // relative paths under the current drive.
+                    Path::new(&path[1..]).absolutize()?.to_string_lossy().into_owned()
+                } else {
+                    path
+                };
+                #[cfg(windows)]
+                let path = Path::new(&path);
 
                 debug!("endpoint try_from: path={}", path.display());
 
-                if path.parent().is_none() || Path::new("").eq(&path) {
+                if path.parent().is_none() || path.as_os_str().is_empty() {
                     return Err(Error::other("empty or root path is not supported in URL endpoint"));
                 }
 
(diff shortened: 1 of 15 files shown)


houseme (approved):

houseme:
Completed a focused review for PR #2974.

Conclusion: **one must-fix issue remains before merge**.

Blocking issue:
- `crates/ecstore/src/disk/endpoint.rs` (`has_leading_slash_windows_drive`) currently matches any `/X:*` path.
- This is too broad for URL paths from `Url::path()`. A path like `http://host/C:foo` is a valid URL path segment, but current logic strips the leading slash and absolutizes it as a local Windows path, which mutates remote endpoint semantics.

Suggested fix:
- Restrict the Windows-drive rewrite to `/X:/...` (require a separator after colon, e.g. `bytes[3] == b'/'`), instead of all `/X:*` forms.
- Add regression tests for both:
  - `http://host/C:/data` (should apply Windows-drive handling)
  - `http://host/C:foo` (should stay as URL path semantics, no drive rewrite)

After this adjustment, the rest of the changes look aligned with the Windows pre-commit compatibility goal.

overtrue (changes requested):
Please fix the Windows drive-path predicate before merge. In crates/ecstore/src/disk/endpoint.rs, has_leading_slash_windows_drive currently accepts any /X:* path, so a URL endpoint path like http://host/C:foo can be stripped and absolutized as a local Windows path. Restrict the rewrite to real absolute drive paths such as /C:/data, and add regressions for both /C:/data and /C:foo URL path semantics.

marshawcoco:
> Please fix the Windows drive-path predicate before merge. In crates/ecstore/src/disk/endpoint.rs, has_leading_slash_windows_drive currently accepts any /X:* path, so a URL endpoint path like http://host/C:foo can be stripped and absolutized as a local Windows path. Restrict the rewrite to real absolute drive paths such as /C:/data, and add regressions for both /C:/data and /C:foo URL path semantics.

Thanks for the review. I’ve updated `has_leading_slash_windows_drive` to only match real absolute Windows drive paths like `/C:/data`, instead of all `/X:*` URL paths.

  I also added Windows-only regression coverage for both cases:
  - `http://host/C:/data` applies the drive-path handling
  - `http://host/C:foo` keeps normal URL path semantics

  Verified with:
  - `cargo test -p rustfs-ecstore disk::endpoint --lib`
  - `cargo fmt --check`
  - `git diff --check`

  Could you please take another look?

overtrue (approved):

PR #2982, opened 2026-05-16
feat: Helm chart support extra volumes

CptOfEvilMinions:
## Related Issues
N/A

## Summary of Changes
Adds`extraVolumes` and `extraVolumeMounts` support to the RustFS Helm chart, following the same pattern already used by `extraEnv`. This was needed to support mounting a custom CA bundle so RustFS can trust a self-signed or private OIDC server certificate, without which RustFS cannot validate the OIDC provider and the SSO login option is not presented to users.

- [helm/rustfs/values.yaml](helm/rustfs/values.yaml): Added extraVolumes: [] and extraVolumeMounts: [] default values
- [helm/rustfs/templates/deployment.yaml](helm/rustfs/templates/deployment.yaml): Added {{- with }} blocks to render extra volumes in spec.volumes and extra volume mounts in the container's volumeMounts

## Verification
Rendered the chart locally with `helm template` using a custom `extraVolumes`/`extraVolumeMounts` values override and confirmed the output includes the expected volume and mount entries.
Confirmed default behavior (empty lists) produces no change to the rendered manifest.

## Impact
- User-facing: Users can now mount arbitrary volumes (e.g., a ConfigMap containing a custom CA bundle) without patching the chart, enabling OIDC SSO against servers with private or self-signed TLS certificates.
- Compatibility: Fully backwards-compatible — defaults are empty lists, so existing deployments are unaffected.
- Configuration: Two new optional Helm values: `extraVolumes` and `extraVolumeMounts`.

## Additional Notes
The motivating use case is passing a custom CA bundle into the RustFS container so the OIDC TLS handshake succeeds against an internal identity provider. Without trust for the OIDC server cert, RustFS silently omits the SSO login option. A typical values snippet:

```yaml

extraVolumes:
  - name: rustfs-ca-bundle
    configMap:
      name: ca-bundle
      items:
        - key: ca-certificates.crt
          path: ca.crt

extraVolumeMounts:
  - name: rustfs-ca-bundle
    mountPath: /opt/rustfs-tls
    readOnly: true

...
extraEnv:
  - name: SSL_CERT_FILE
    value: /opt/rustfs-tls/ca.crt
```


CptOfEvilMinions:

majinghe (approved):

majinghe:
@CptOfEvilMinions Thanks for your contribution.

CptOfEvilMinions:
bump

houseme (approved):

PR #2997, opened 2026-05-18
fix(ecstore): correct is_truncated logic in ListObjectsV2 pagination

DemoMacro:
## Related Issues
N/A

## Summary of Changes

- `is_truncated` in `list_objects_generic` only relied on `get_objects.len() > max_keys`, but `get_objects` could be reduced below `max_keys` after `forward_past` + `from_meta_cache_entries_sorted_infos` processing (e.g. `to_fileinfo` failures silently skipping entries), causing premature `is_truncated = false` and data loss in pagination.
- Save the `disk_has_more` signal (`err.is_none()` on list_path result) before `err.take()` consumes it, and use it as a fallback condition for `is_truncated` when `get_objects` count falls below `max_keys`.
- Add `!get_objects.is_empty()` guard to ensure `next_marker` is always valid when `is_truncated = true`, preventing truncated responses without a continuation token.

## Verification
- `cargo check -p rustfs-ecstore` passed

## Impact
Fixes ListObjectsV2 pagination data loss when entries are filtered out during processing. No API or compatibility changes.

## Additional Notes
N/A


DemoMacro:
> 
> Missing signatures:
> 
> * @DemoMacro
> 
> 
> 

houseme:
Request changes: this patch can introduce false-positive pagination truncation. err.is_none() indicates raw metacache limit behavior, not a guaranteed next visible result after post-processing (delimiter collapse or skipped entries). Please adjust truncation detection to visible-result semantics and add regression tests for collapsed-prefix and skipped-entry paths before merge.

overtrue (changes requested):
Requesting changes because the new fallback can produce false-positive truncation for delimiter listings. In crates/ecstore/src/store_list_objects.rs:375, disk_has_more is based on raw metacache entries, but several raw keys can collapse into one visible CommonPrefix in from_meta_cache_entries_sorted_infos. With many keys under the same directory and MaxKeys greater than the visible prefix count, this can return IsTruncated=true and NextContinuationToken set to the same common prefix even though there is no next visible result, causing repeated pages. Please base truncation on visible results and add regression coverage for collapsed-prefix and skipped-entry cases.

houseme:
please fix ci `cargo fmt --all`. @DemoMacro

DemoMacro:
Apologies for the shortcomings in the initial implementation, and for not following the project's `cargo fmt` and clippy requirements (as specified in `CONTRIBUTING.md`) on the first submission. This added unnecessary work during review — I'll make sure to run the full pre-commit checks going forward.

Thank you @overtrue @houseme for the accurate review. Regarding the false-positive truncation caused by raw keys collapsing into CommonPrefixes in delimiter listings, the fix has been completely reworked:

1. Preserved the original `get_objects` truncation logic to maintain correct lexicographic truncation.
2. Added a post-separation check: when `is_truncated=false` but `disk_has_more=true` and visible results exist, set `is_truncated=true` and recompute `next_marker` from visible results.
3. Saved `disk_has_more` signal before `err.take()` consumes it.
4. Applied the same fix to `inner_list_object_versions` which had the identical bug.
5. Added regression tests covering collapsed-prefix and small-page delimiter scenarios.
6. Fixed `cargo fmt` and clippy issues.

Thanks again for the thorough review — it helped catch a real production issue.

overtrue (approved):

overtrue (changes requested):
The delimiter-collapse false-positive still appears to remain. If more than MaxKeys raw entries all collapse into fewer visible CommonPrefixes, the current get_objects/disk_has_more path can still return IsTruncated=true even though there is no next visible result. The new collapsed-prefix test uses only 32 raw keys with MaxKeys=1000, so it never exercises the raw-entry limit case. Please add coverage with raw entries > MaxKeys but visible results < MaxKeys, and base truncation on visible-result semantics.

overtrue (changes requested):
Requesting changes because the current fallback no longer covers the skipped-entry path described by the PR. If from_meta_cache_entries_sorted_infos drops entries, for example due to to_fileinfo failures, and the visible count stays below MaxKeys, disk_has_more is ignored and the response can still return IsTruncated=false even though later valid entries are unreachable. Please either restore a marker-safe way to continue after skipped raw entries or narrow the PR/body and tests to the delimiter-only fix.

DemoMacro:
Thanks @overtrue for the continued review. The latest change distinguishes two scenarios:

- **No delimiter**: get_objects reduction can only come from to_fileinfo failures (skipped entries). Use `disk_has_more && visible > 0` to cover the original skipped-entry bug.
- **With delimiter**: reduction may be from prefix collapse. Use `disk_has_more && visible >= max_keys` to avoid false-positive truncation.

Both functions (list_objects_generic + inner_list_object_versions) are updated with the same logic.

houseme (approved):

PR #2998, opened 2026-05-18
fix: prevent disabled profiling from burning CPU

### LINKED ISSUE #2941: [BUG] beta.2 scanner consumes ~200% CPU with zero disk IO on quiescent 4-drive SNSD — metacache (starshard) hot loop + futex storm
## Summary

On rustfs `1.0.0-beta.2` running as a single-node 4-drive EC:2 deployment, the rustfs process consumes a sustained **~200% CPU (≈2 cores)** with **zero disk IOPS** on all four backing drives, no client write load, and an empty `rustfs.log` / silent docker stdout. CPU profiling with `perf` shows the workers are pinned in a `futex`-storm on top of the `starshard::ShardedHashMap` metacache, with all backing file access satisfied from the kernel page cache. The pattern persists after applying the rustfs#2601 timeout-workaround env-vars (which addressed the FaultyDisk false-trip thrash).

The symptom appears to be a metacache hot loop in beta.2 that doesn't match any single existing open issue, but overlaps with **#2620**, **#2790 / PR #2714**, and the beta.2-era **PR #2724**.

### rustfs

| | |
|---|---|
| rustfs version | `rustfs/rustfs:1.0.0-beta.2` (docker image, built 2026-05-06T13:01:44Z) |
| Container OS | Alpine Linux 3.23.4 |
| Deployment shape | Single-node, single pool, **1 erasure set × 4 drives, EC:2** |
| Drive config | `RUSTFS_ERASURE_SET_DRIVE_COUNT=4`, `RUSTFS_STORAGE_CLASS_STANDARD=EC:2` |
| Scanner | `RUSTFS_SCANNER_SPEED=slow` |
| Observability | `RUSTFS_OBS_ENDPOINT=http://host.docker.internal:4318`, `RUSTFS_OBS_LOGGER_LEVEL=warn` |
| rustfs#2601 workaround applied | `RUSTFS_DRIVE_METADATA_TIMEOUT_SECS=30`, `_DISK_INFO_=30`, `_LIST_DIR_=60`, `_WALKDIR_=120`, `_WALKDIR_STALL_=30`, `_ACTIVE_CHECK_=30`, `RUSTFS_DRIVE_SUSPECT_FAILURE_THRESHOLD=10` |

### Workload scale (per `mc admin info`)

| | |
|---|---|
| Total S3 objects | **13,690** across 7 buckets |
| Total logical data | **12 GiB** used |
| Raw capacity | 1.8 TiB (4 × 932 GiB) |
| Drive utilisation | 2.6% (24 GiB used per drive) |
| Per-drive file count (host xfs) | ~15,300 files / drive (1 data + 1 metadata shard per object, evenly distributed) |
| Inode usage | <1% (39 K used / 488 M total per drive) |
| Tenants | Mimir TSDB blocks, Loki chunks, Tempo blocks — all background retention LIST activity, no client write traffic during the observation window |

This is a small dataset. 13.7K objects on a 4-drive erasure set should be trivial for rustfs to scan. The CPU symptom is not workload-induced — the scanner has very little to walk.

### Host

| | |
|---|---|
| Host OS / kernel | Ubuntu 24.04, Linux 6.8.0-111-generic |
| Hypervisor | QEMU/KVM on Proxmox VE (guest reports SMBIOS 2.8, Q35+ICH9, `pc-q35-10.1`) |
| Guest vCPU | **4 vCPUs**, AMD Ryzen 7 PRO 8845HS (Zen 4 mobile), reported 3.79 GHz, 512 KB L2/core (host has 8C/16T) |
| Guest RAM | **16 GiB**, 13 GiB free during observation (rustfs RSS = 554 MiB, low pressure) |
| Guest swap | 4 GiB, ~20 MiB in use |
| Network | virtio (eth0), 1 Gbps homelab LAN |

### Storage layout

| Aspect | Value |
|---|---|
| Boot disk | `/dev/vda` (96 GiB, virtio-blk, rota=1) — irrelevant to rustfs |
| rustfs data disks | `/dev/sda`, `/dev/sdb`, `/dev/sdc`, `/dev/sdd` (4 × 932 GiB) |
| Transport | **virtio-scsi** (path: `pci0000:00:1e.0/.../virtio1/host0/target0:0:0/0:0:0:N/block/sdX`) |
| QEMU device model | `QEMU HARDDISK` (default SCSI model string; not actually a rotating disk) |
| **Guest-visible rotational flag** | **`rota=0` on all four data drives** — guest treats them as SSDs |
| Block size | 4 KiB DISC-GRAN, 512 B phys/hw sectors |
| Discard granularity | 4 KiB, max 1 GiB per discard |
| I/O scheduler | `[none]` (no kernel scheduler — appropriate for SSD/virtio) |
| Mount options | `xfs rw,noatime,nodiratime,attr2,inode64,logbufs=8,logbsize=32k,noquota` |
| Mount points (host) | `/mnt/rustfs0` ← `/dev/sdd`, `/mnt/rustfs1` ← `/dev/sdc`, `/mnt/rustfs2` ← `/dev/sdb`, `/mnt/rustfs3` ← `/dev/sda` (bind-mounted into the container as `/data/rustfs0..3`) |

### Storage backing (re: NVMe performance model)

I'm aware that rustfs's documented performance model targets NVMe-class storage. The guest-visible block devices for this deployment all advertise `rota=0` (non-rotational) — from rustfs's point of view, it is talking to SSD-class storage with an unscheduled queue. The Proxmox-host backing of the virtio-scsi devices is not introspectable from inside the guest; I can provide the Proxmox storage-pool configuration on request.

Despite the homelab-grade environment, the symptom is not throughput- or latency-bound — it is CPU-bound with zero disk activity (see the perf data below). A workload-mismatch "rustfs is tuned for NVMe" caveat would explain slow I/O, but should not explain a sustained 200% CPU on an idle deployment. The profile shows the workers pinned in user-space metacache logic + futex wakeups, not in any I/O wait path.

## Observed behavior

For >2 hours of continuous observation, confirmed across container restarts:

- **CPU**: rustfs process at 188–315% sustained (`top`, `docker stats`, and `container_cpu_usage_seconds_total` all agree)
- **Disk**: `node_disk_reads_completed_total` and `node_disk_writes_completed_total` on `sda`/`sdb`/`sdc`/`sdd` (the four xfs drives) are flat — **0 iops** for 90+ minutes
- **Memory**: 554 MiB RSS, stable
- **Network**: 343 MB lifetime in 49 min uptime — average ~120 KB/s in+out, dominated by Mimir/Tempo LIST traffic
- **Logs**: `/logs/rustfs.log` is 0 bytes; `docker logs rustfs` produces no new lines over a 30-minute window at level=warn
- **Client traffic**: traefik access log shows ~30 req/min, all from Mimir+Tempo retention/marker LIST operations (`prefix=index/markers/`, `prefix=tempo/markers/`). No write traffic, no heavy reads.

The post-restart trajectory is reproducible: after `docker compose up -d`, CPU drops to ~0% for 25–30 minutes (post-init quiet), then ramps to ~200% and stays there indefinitely.

## CPU profile (perf record, 20s @ 99Hz with stack traces)

```
sudo perf record -F 99 -g -p $(pgrep -o rustfs) -- sleep 20
sudo perf report --children --max-stack 12 --sort=overhead_children,symbol
```

### Top callers / hot paths

| Layer | % CPU (children) | Notes |
|---|---|---|
| `entry_SYSCALL_64_after_hwframe` | **40.04%** | All work routes through syscalls — heavily threaded |
| ↳ `__x64_sys_futex` → `do_futex` | **24.06%** | Thread synchronization storm |
| ↳↳ `futex_wake` → `wake_up_q` → `__pv_queued_spin_lock_slowpath` | 14.40% | Many threads contending wake-up paths |
| ↳↳ `futex_wait` → `schedule` → `__schedule` | 8.09% | Threads parking after no-op cycles |
| ↳ `__x64_sys_open` → `path_openat` → `link_path_walk` | 3.77% | Metadata-file opens — fully served by page cache (no disk IOPS) |

### Top rustfs-internal hot functions (self time)

| % CPU | Symbol | Notes |
|---|---|---|
| **3.90%** | `starshard::ShardedHashMap<K,V,S>::get_or_init_shard` | Metacache shard lookup/init |
| 3.32% | `memcpy` | |
| **2.01%** | `starshard::ShardedHashMap<K,V,S>::remove` | Metacache eviction |
| 1.86% | `_mi_page_malloc_zero` | mimalloc allocation |
| **1.81%** | `tracing_subscriber::filter::env::EnvFilter::enabled` | OTLP filter checks (cost paid even at warn level) |
| 1.72% | `memset` | |
| 1.60% | `_mi_page_free_collect` | mimalloc free |
| **1.49%** | `tracing::__macro_support::__is_enabled` | Per-call macro guard for tracing |

Combined `starshard::ShardedHashMap` operations ≈ **6%** of total CPU (rustfs-internal); combined `tracing` filter overhead ≈ **3.3%**; mimalloc churn ≈ 3.5%.

## Interpretation

The picture is a **futex storm driven by `starshard::ShardedHashMap` churn**: scanner / heal workers continuously look up and insert/remove cache entries; each operation wakes peer workers, which immediately re-park. The cache lookups are backed by `open()` calls on small metadata files (xl.meta) that are fully resident in the kernel page cache — which is why `iostat` and node_exporter show zero physical disk activity despite roughly 7K open syscalls/sec.

`starshard` is the data structure that **PR #2724** ("harden and optimize metacache path") in beta.2 modified, and the symptom is consistent with a refresh-coalescing / cache-validate path that retries without forward progress when a cached entry is treated as stale. The `tracing::__is_enabled` overhead is measurable but secondary — disabling OBS export entirely would only shave ~3% of total CPU.

## What's been ruled out

- **Client load**: ~30 req/min of small LISTs is orders of magnitude too low to drive 200% CPU.
- **Disk faulting / rustfs#2601**: the timeout-workaround env-vars are in place and the FaultyDisk thrash pattern has stopped (no `mc admin scanner info` faulting transitions, no I/O cycling on the drives).
- **OOM / restart loop**: `RestartCount=0`, `OOMKilled=false`; `dmesg` / `journalctl` show no OOM kills.
- **OTLP push back-pressure**: `RUSTFS_OBS_ENDPOINT` reaches an alloy / otel-collector gateway with a healthy file-backed queue; HTTP push latencies stay low.
- **Workload bursts**: traefik access logs show steady ~30 req/min through the entire observation window.
- **Bitrot scanner findings**: an existing low-rate "bitrot shard file size mismatch" tracking issue runs at ~5/min — far too low a rate to drive ~200% CPU.

## Related / overlapping issues

- **#2620** (closed, alpha.96) — "Scanner hangs indefinitely on the same object path; `list_path_raw: revjob err … canceled` + heal-channel `channel closed` busy-retry". Same *shape* (quiescent single-node, scanner pinned, zero per-minute counters). Whether the fix that closed #2620 is still effective in beta.2 against the metacache codepath is worth checking.
- **#2790** (open, beta.1 regression) — bisects scanner hang to **PR #2714** ("avoid faulting local drives on transient timeouts"). That retry-loop logic is still present in beta.2; the env-var workaround tames the FaultyDisk side effect, but the underlying retry topology may be what's feeding the futex storm.
- **PR #2724** (merged in beta.2) — filemeta metacache hardening. Touches the exact `starshard` codepath that's hot in this profile; possible regression vector worth re-reviewing for cache-validate retry without forward progress.

## Reproduction

Minimal setup:

```yaml
# docker-compose.yml (essentials)
services:
  rustfs:
    image: rustfs/rustfs:1.0.0-beta.2
    environment:
      - RUSTFS_VOLUMES=/data
      - RUSTFS_ERASURE_SET_DRIVE_COUNT=4
      - RUSTFS_STORAGE_CLASS_STANDARD=EC:2
      - RUSTFS_SCANNER_SPEED=slow
      - RUSTFS_OBS_LOGGER_LEVEL=warn
      - RUSTFS_DRIVE_METADATA_TIMEOUT_SECS=30
      - RUSTFS_DRIVE_DISK_INFO_TIMEOUT_SECS=30
      - RUSTFS_DRIVE_LIST_DIR_TIMEOUT_SECS=60
      - RUSTFS_DRIVE_WALKDIR_TIMEOUT_SECS=120
      - RUSTFS_DRIVE_WALKDIR_STALL_TIMEOUT_SECS=30
      - RUSTFS_DRIVE_ACTIVE_CHECK_TIMEOUT_SECS=30
      - RUSTFS_DRIVE_SUSPECT_FAILURE_THRESHOLD=10
    volumes:
      - /data/rustfs0:/data/rustfs0
      - /data/rustfs1:/data/rustfs1
      - /data/rustfs2:/data/rustfs2
      - /data/rustfs3:/data/rustfs3
```

Populate the deployment with a modest set of small objects (this reproduces reliably with ~13.7K objects / 12 GiB across 7 buckets — Loki chunks named `<hex>:<hex>:<hex>` ~10 KiB each, plus Mimir TSDB blocks and Tempo trace blocks), then let the deployment idle with only periodic small LIST operations against the bucket (~30 req/min, from Mimir+Tempo retention loops in this case). After ~25–30 min of post-init quiet, CPU climbs to ~200% and stays there indefinitely.

The trigger is the scanner cycle resuming after the post-startup grace period — not the workload itself. The dataset is small: <1% inode usage on each xfs leg, 3% disk usage.

## Attachments

The two perf reports are hosted as a public gist (text size made inlining impractical):

- **Gist**: https://gist.github.com/feicipet/9cf89492339b2cc307fffce7aeb9d72a
  - `perf-report.txt` — self-time view of the top hot functions (90 KB; report depth 30)
  - `perf-report-children.txt` — full call-graph with parents (880 KB; depth 12)
- `rustfs.perf` — original `perf record` binary data (945 KB), retained locally; happy to upload on request for `perf script` / flamegraph generation.

I can supply additional diagnostics on request — a longer profile, a flamegraph SVG, `perf stat` event counters, a `mc admin scanner info` snapshot, or `/proc/$PID/status` and `/proc/$PID/stack` dumps.

weisd:
## Related Issues
Fixes #2941

## Summary of Changes
- Short-circuit allocator profiling allocation and deallocation tracking when profiling is disabled, avoiding live allocation map access on the default disabled path.
- Clear live allocation and stack tracking state when profiling is turned off, so runtime toggles do not retain stale samples.
- Add regression coverage for disabled deallocation tracking and cleanup on disable.

## Verification
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs profiling::allocator --lib`
- `cargo clippy -p rustfs --lib --all-features -- -D warnings`
- `LC_ALL=en_US.UTF-8 LC_CTYPE=en_US.UTF-8 LANG=en_US.UTF-8 make pre-commit`
- `cargo zigbuild -p rustfs --lib --target x86_64-unknown-linux-musl`

## Impact
Default-disabled allocation profiling no longer pays per-deallocation tracking-map overhead on non-GNU builds. No configuration or API changes.

## Additional Notes
An `aarch64-unknown-linux-musl` local check was attempted but blocked by local zig/cc-rs target argument handling before crate code was checked.

---

Thank you for contributing to RustFS! We will review your PR as soon as possible.

### DIFF
--- a/rustfs/src/profiling/allocator.rs
+++ b/rustfs/src/profiling/allocator.rs
@@ -73,6 +73,11 @@ pub fn set_enabled(enabled: bool) {
     let _ = &*STACK_CACHE;
 
     ENABLED.store(enabled, Ordering::Relaxed);
+
+    if !enabled {
+        LIVE_ALLOCATIONS.clear();
+        STACK_CACHE.clear();
+    }
 }
 
 fn should_sample(size: usize) -> bool {
@@ -321,20 +326,26 @@ fn dump_profile_inner(path: &Path) -> Result<(), String> {
 // Helper to handle sampling logic
 #[inline(always)]
 fn handle_alloc_sampling(ptr: *mut u8, size: usize) {
-    if !ptr.is_null() {
-        // Check reentrancy guard BEFORE calling should_sample
-        if !REENTRANCY_GUARD.replace(true) {
-            if should_sample(size) {
-                record_alloc(ptr, size);
-            }
-            REENTRANCY_GUARD.set(false);
+    if ptr.is_null() || !ENABLED.load(Ordering::Relaxed) {
+        return;
+    }
+
+    // Check reentrancy guard BEFORE calling should_sample
+    if !REENTRANCY_GUARD.replace(true) {
+        if should_sample(size) {
+            record_alloc(ptr, size);
         }
+        REENTRANCY_GUARD.set(false);
     }
 }
 
 // Helper to handle dealloc logic
 #[inline(always)]
 fn handle_dealloc_sampling(ptr: *mut u8) {
+    if ptr.is_null() || !ENABLED.load(Ordering::Relaxed) {
+        return;
+    }
+
     if !REENTRANCY_GUARD.replace(true) {
         record_dealloc(ptr);
         REENTRANCY_GUARD.set(false);
@@ -439,6 +450,36 @@ mod tests {
 
         REENTRANCY_GUARD.set(false);
         set_enabled(false);
+    }
+
+    #[test]
+    #[serial]
+    fn test_disabled_dealloc_skips_tracking_map() {
+        set_enabled(false);
+
+        let key = usize::MAX - 1;
+        LIVE_ALLOCATIONS.insert(key, (1, Arc::new(vec![key])));
+
+        handle_dealloc_sampling(key as *mut u8);
+
+        assert!(LIVE_ALLOCATIONS.get(&key).is_some());
+        LIVE_ALLOCATIONS.remove(&key);
+    }
+
+    #[test]
+    #[serial]
+    fn test_disabling_clears_tracking_maps() {
+        set_enabled(true);
+
+        let key = usize::MAX - 2;
+        let stack_key = u64::MAX - 2;
+        LIVE_ALLOCATIONS.insert(key, (1, Arc::new(vec![key])));
+        STACK_CACHE.insert(stack_key, Weak::new());
+
+        set_enabled(false);
+
+        assert!(LIVE_ALLOCATIONS.get(&key).is_none());
+        assert!(STACK_CACHE.get(&stack_key).is_none());
     }
 
     #[test]


weisd:
This PR is superseded by #2983, which removes the non-GNU tracing allocator path and uses the plain allocator on unsupported profiling targets. The remaining conflict is a modify/delete conflict in rustfs/src/profiling/allocator.rs, so there is no patch from this branch to carry forward.

PR #3010, opened 2026-05-19
feat(ecstore): add internode transport boundary and TCP baseline runner

marshawcoco:
## Related Issues
  N/A

  ## Summary of Changes
  - Added a reproducible internode transport TCP baseline runner:
    - `scripts/run_internode_transport_baseline.sh`
    - Exports `summary.csv` and (optionally) `internode_metric_deltas.csv` when `--metrics-url` is provided.
  - Introduced a pluggable internode data-plane transport boundary in ecstore:
    - Added `crates/ecstore/src/rpc/internode_data_transport.rs`
    - Added `InternodeDataTransport` trait and request types for read/write/walk-dir streams.
    - Added default `tcp-http` backend (`TcpHttpInternodeDataTransport`).
    - Added env selector `RUSTFS_INTERNODE_DATA_TRANSPORT` with safe fallback to `tcp-http`.
  - Wired `RemoteDisk` data-plane paths to the transport boundary:
    - `read_file_stream`
    - `create_file`
    - `append_file`
    - `walk_dir`
  - Updated RFC doc with baseline runner entry point:
    - `INTERNODE_DATA_TRANSPORT_RFC.md`

  ## Verification
  - `bash -n scripts/run_internode_transport_baseline.sh`
  - `cargo fmt --all`

  ## Impact
  - No default behavior change: internode data path still uses current TCP/HTTP routes.
  - No control-plane transport change.
  - Adds a stable extension boundary for future RDMA/RoCE/InfiniBand backends while preserving TCP fallback.
  - Adds benchmark execution entry point and machine-readable artifacts for baseline comparison.

  ## Additional Notes
  - This PR intentionally does not implement RDMA backend logic.
  - Scope is limited to baseline tooling + transport abstraction boundary.


overtrue (changes requested):
Please address the unresolved script issues before merge. scripts/run_internode_transport_baseline.sh currently requires rg even though it does not use it, which can make the baseline runner fail on valid environments, and it always emits/advertises internode_metric_deltas.csv even when --metrics-url is not provided, contradicting the documented optional artifact behavior.

marshawcoco:
Addressed the unresolved baseline script issues: removed the unused `rg` requirement from `scripts/run_internode_transport_baseline.sh`, and gated `internode_metric_deltas.csv` creation/append/artifact output on `--metrics-url`. Also addressed Copilot's env constant and repeated env lookup comments by moving the transport env/default constants to `rustfs-config` and caching the resolved transport with `OnceLock`.

overtrue (changes requested):
Blocking: .github/actions/setup/action.yml:47-57 now installs protoc only through apt on Linux, but .github/workflows/build.yml:166-173 and 222-230 still use this shared setup action for macOS and Windows release-build matrix entries. Those runners no longer get arduino/setup-protoc or another protoc install path, so main/tag/manual build releases can fail outside the current PR CI. Please keep a cross-platform protoc setup or add explicit macOS/Windows installation before this lands.

marshawcoco:
Quick correction on my previous note: after resolving merge conflicts with latest main, the setup action now uses `rustfs/setup-protoc@v3.0.1` as the cross-platform protoc installer (including macOS/Windows), while Linux also keeps `apt` system dependencies. So macOS/Windows release matrix jobs still have protoc coverage and we are no longer relying on the previous brew/arduino split.

overtrue (approved):
Looks good.

PR #3049, opened 2026-05-21
fix(helm): add LoadBalancer service type support

dcode:
## Summary

The service template only handled `ClusterIP` and `NodePort` via `if`/`else if` branches. When `service.type=LoadBalancer` is set, no branch matched and the `type` field was omitted from the rendered `Service` manifest entirely — causing Kubernetes to silently fall back to `ClusterIP`.

- Add the missing `LoadBalancer` branch to `service.yaml`
- Add three optional supporting values (all no-ops when unset):
  - `service.loadBalancerIP` — request a specific IP from the LB controller
  - `service.loadBalancerClass` — select a specific LB implementation (e.g. `io.cilium/bgp`, `metallb`)
  - `service.loadBalancerSourceRanges` — restrict allowed client source CIDRs

## Motivation

For bare-metal / on-prem clusters using software load balancers (Cilium LB-IPAM, MetalLB, kube-vip), a `LoadBalancer`-type service is the natural exposure primitive. S3 clients configure an endpoint URL directly rather than relying on hostname-based HTTP routing, so a stable routable IP from the LB pool is often preferable to fronting with an Ingress — especially for workloads doing large multipart uploads or byte-range GETs where an L7 proxy can introduce buffering or timeout issues.

## Test plan

- [x] `helm template` with `service.type=LoadBalancer` renders `type: LoadBalancer` with optional fields populated
- [x] `helm template` with `service.type=ClusterIP` still renders correctly (no regression)
- [x] `helm template` with `service.type=NodePort` still renders correctly (no regression)
- [x] `loadBalancerIP`, `loadBalancerClass`, `loadBalancerSourceRanges` are omitted from output when left empty/unset


houseme:
> 
> Missing signatures:
> 
> * @dcode
> 
> 
> 

@dcode

dcode:

majinghe (approved):

majinghe:
@dcode Thanks for your contribution.

houseme (approved):

PR #3052, opened 2026-05-22
fix(utils): cover sha256 checksum validation

overtrue:
## Related Issues

Related to rustfs/backlog#646

## Summary of Changes

- Add focused coverage for `is_sha256_checksum` to lock its current checksum validation semantics: exactly 64 lowercase hexadecimal characters.
- Remove the stale optimization TODO after validating that the existing implementation is already zero-allocation and direct.
- Add the missing `sha2` dependency to the `crypto` feature so `rustfs-utils --features crypto` compiles independently instead of only when another feature also enables `sha2`.

## Verification

- `PATH="$HOME/.cargo/bin:$PATH" cargo test -p rustfs-utils --features crypto sha256_checksum_ -- --nocapture`
- `PATH="$HOME/.cargo/bin:$PATH" cargo test -p rustfs-utils --features crypto crypto:: -- --nocapture`
- `PATH="$HOME/.cargo/bin:$PATH" cargo fmt --all --check`
- `PATH="$HOME/.cargo/bin:$PATH" cargo clippy -p rustfs-utils --features crypto --tests -- -D warnings`

## Impact

No runtime behavior change for checksum validation. The `crypto` feature now declares the dependency needed by its own SHA-256 helpers. Keeping the existing checksum implementation avoids introducing allocation or a slower validation path.

## Additional Notes

I compared simple replacement strategies locally and kept the existing implementation because it was already at least as fast in the tested cases.


overtrue:
Updated the branch to address the review feedback by completing the `is_sha256_checksum` return docs and explicitly documenting the lowercase-hex requirement.

Verification:
- `cargo test -p rustfs-utils --features crypto sha256_checksum_ -- --nocapture`
- `cargo test -p rustfs-utils --features crypto crypto:: -- --nocapture`
- `cargo fmt --all --check`
- `cargo clippy -p rustfs-utils --features crypto --tests -- -D warnings`

houseme (approved):

PR #3055, opened 2026-05-22
closed: exploratory internode transport simulator spike

marshawcoco:
This was an exploratory spike around the internode transport boundary.

After maintainer feedback, I am narrowing the OSS scope. The follow-up work will focus only on the backend-neutral `InternodeDataTransport` adapter boundary and the existing `tcp-http` path.

This PR is superseded by a smaller OSS-scoped cleanup PR. The replacement PR will avoid simulator code, hardware-specific transport wording, and non-default backend implementation details.


marshawcoco:
Closing this exploratory PR because the OSS scope has been narrowed. I will follow up with a smaller PR focused only on the backend-neutral `InternodeDataTransport` adapter boundary and the default `tcp-http` path. The follow-up PR will not include simulator code or hardware-specific transport plans.

PR #3059, opened 2026-05-22
fix(tls): resolve RUSTFS_TLS_PATH startup regression

### LINKED ISSUE #3053: RustFS Startup Failed: TLS Certificate Directory Not Found
**Describe the bug**

TLS cannot be initialized when `RUSTFS_TLS_PATH` is configured.

Using the same TLS certificate directory and the same startup configuration:

- `1.0.0-beta.2` works as expected
- `1.0.0-beta.3` fails
- `1.0.0-beta.4` fails

**To Reproduce**

1. Prepare a TLS certificate directory.
2. Set `RUSTFS_TLS_PATH` to that directory, for example `/path/to/certs/`.
3. Start RustFS.
4. Observe that TLS initialization fails on `1.0.0-beta.3` and `1.0.0-beta.4`.

Sanitized log:

```text
[FATAL] Server encountered an error and is shutting down:
TLS is explicitly configured via RUSTFS_TLS_PATH/tls_path='/path/to/certs/'
but TLS acceptor initialization failed:
TLS material I/O error: failed to discover TLS certificates under '/path/to/certs/':
The certificate directory does not exist or is not a directory: /path/to/certs/
```

**Expected behavior**

**Screenshots**

Image

Image

**Additional context**

houseme:
## Related Issues
Fixes #3053.

## Summary of Changes
- validate the configured `RUSTFS_TLS_PATH` explicitly before building the server TLS acceptor
- centralize server certificate material resolution into a single decision flow
- preserve multi-cert/SNI behavior when domain-specific certificate directories are present
- return clearer startup errors for missing TLS directories, non-directory paths, missing root cert/key files, and empty TLS directories
- add focused regression coverage for root single-cert, symlinked TLS directories, invalid paths, empty directories, and mixed root/domain certificate layouts

## Verification
- `cargo test -p rustfs --lib tls_material -- --nocapture`
- `make pre-commit`

## Impact
- fixes the TLS startup regression reported in #3053 for explicit `RUSTFS_TLS_PATH` configurations
- preserves existing multi-cert/SNI deployments instead of silently downgrading them to root single-cert handling
- improves operator-facing startup diagnostics for TLS configuration problems

## Additional Notes
- no `docs/` artifacts are included in this PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9252,6 +9252,7 @@ dependencies = [
  "pin-project-lite",
  "pprof-pyroscope-fork",
  "rand 0.10.1",
+ "rcgen",
  "reqwest",
  "rmp-serde",
  "rsa 0.10.0-rc.18",
--- a/rustfs/Cargo.toml
+++ b/rustfs/Cargo.toml
@@ -195,6 +195,7 @@ temp-env = { workspace = true, features = ["async_closure"] }
 tracing-subscriber = { workspace = true }
 opentelemetry_sdk = { workspace = true }
 rsa = { workspace = true }
+rcgen = { workspace = true }
 
 [build-dependencies]
 http.workspace = true
--- a/rustfs/src/server/tls_material.rs
+++ b/rustfs/src/server/tls_material.rs
@@ -602,7 +602,9 @@ pub(crate) fn spawn_reload_loop(tls_path: String, holder: Arc<TlsAcceptorHolder>
                             info!("TLS certificates reloaded successfully");
                             holder.swap(&new_holder);
                         }
-                        Ok(None) => debug!("TLS reload: no server certificates found in directory, skipping"),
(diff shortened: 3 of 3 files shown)


cxymds (approved):

houseme:
Addressed the two Copilot follow-ups and re-ran the full gate.

Updates:
- kept the reload path defensive by handling the `Ok(None)` branch explicitly, but changed the message to reflect that it is now an unexpected state for a configured TLS path instead of a normal "no certificates found" flow
- made the test rustls crypto-provider setup idempotent so it does not panic if another part of the same test binary installs the provider first

Verification:
- `cargo test -p rustfs --lib tls_material -- --nocapture`
- `make pre-commit`

Result:
- `make pre-commit` passes again after the follow-up changes

PR #3085, opened 2026-05-26
fix(heal): restore single disk data during deep heal

marshawcoco:
## Related Issues

  Fixes https://github.com/orgs/rustfs/discussions/2964

  ## Summary of Changes

  Fix admin deep heal for a single-server, four-disk deployment when one disk's bucket data is cleared.

  - Treat all-volume-not-found listing results as an empty listing during object walk so heal can continue using the remaining disks.
  - Prevent recursive bucket heal from passing `remove=true` into the bucket metadata heal step, avoiding deletion of healthy source bucket data before
  object repair.
  - Keep object-level recursive bucket heal using the requested `remove`, `recreate`, and `scanMode` options.
  - Pass admin heal options into erasure-set object heal instead of hard-coding normal scan options.
  - Serialize deep erasure-set page object healing to avoid namespace write-lock contention.
  - Move page concurrency semaphore acquisition inside the queued future to avoid blocking the producer before futures can run.
  - Add focused regression coverage for listing behavior, deep scan concurrency, and recursive bucket heal option handling.

  ## Verification

  - `RUSTFLAGS="--cfg tokio_unstable" cargo build --bin rustfs`
  - `RUSTFLAGS="--cfg tokio_unstable" cargo test -p rustfs-heal --lib recursive_bucket_heal -- --nocapture`
  - `RUSTFLAGS="--cfg tokio_unstable" cargo test -p rustfs-heal --lib erasure_healer -- --nocapture`
  - `RUSTFLAGS="--cfg tokio_unstable" cargo test -p rustfs-ecstore walk_result_from_set_errors -- --nocapture`
  - Manual reproduction: single server with four local disks, uploaded 2000 1 MiB objects, stopped server, removed one disk's bucket data, restarted without
  GET/read-repair, ran admin deep heal, verified the cleared disk restored 2000 `xl.meta` files and 2000 `part.1` files.

  ## Impact

  Admin deep heal can now restore object metadata and parts on a cleared single disk in a four-disk erasure set without relying on read-repair first.

  Deep erasure-set page object healing now prioritizes correctness by running object repairs serially for deep scans.

  ## Additional Notes

  N/A

### DIFF
--- a/crates/ecstore/src/store_list_objects.rs
+++ b/crates/ecstore/src/store_list_objects.rs
@@ -70,7 +70,7 @@ fn ensure_non_empty_listing_disks(bucket: &str, path: &str, disks: &[DiskStore])
 fn walk_result_from_set_errors(errs: &[Option<Error>]) -> Result<()> {
     if is_all_not_found(errs) {
         if is_all_volume_not_found(errs) {
-            warn!(?errs, "treating all-volume-not-found listing result as empty during walk");
+            return Err(StorageError::VolumeNotFound);
         }
 
         return Ok(());
@@ -1660,9 +1660,9 @@ mod test {
     }
 
     #[test]
-    fn walk_result_from_set_errors_treats_all_volume_not_found_as_empty_listing() {
+    fn walk_result_from_set_errors_returns_volume_not_found_when_all_sets_report_missing_volume() {
         walk_result_from_set_errors(&[Some(StorageError::VolumeNotFound), Some(StorageError::VolumeNotFound)])
-            .expect("all volume-not-found set errors should be treated as an empty listing");
+            .expect_err("all volume-not-found set errors should stay fatal for generic walk callers");
     }
 
     #[test]
--- a/crates/ecstore/src/store_list_objects.rs
+++ b/crates/ecstore/src/store_list_objects.rs
@@ -70,7 +70,7 @@ fn ensure_non_empty_listing_disks(bucket: &str, path: &str, disks: &[DiskStore])
 fn walk_result_from_set_errors(errs: &[Option<Error>]) -> Result<()> {
     if is_all_not_found(errs) {
         if is_all_volume_not_found(errs) {
-            return Err(StorageError::VolumeNotFound);
+            warn!(?errs, "treating all-volume-not-found listing result as empty during walk");
         }
 
         return Ok(());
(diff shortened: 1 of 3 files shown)


overtrue (changes requested):
Blocking: crates/ecstore/src/store_list_objects.rs:72 now turns all-VolumeNotFound walks into success for every ECStore::walk caller, but this PR heal listing path uses list_objects_v2 through crates/heal/src/heal/storage.rs:596, not walk. That silently hides missing buckets/config for existing walk callers such as IAM and replication without fixing the deep-heal listing path; please scope the empty-listing behavior to the heal listing path or prove the global walk semantic change is safe.

overtrue:
Thanks for the work here. After comparing this with #3084, I do not think this PR is the better merge candidate in its current form.

My main concern is that this branch still keeps the old `forceStart` behavior in `HealChannelProcessor`: `forceStart=true` still overrides explicit admin options and turns `remove/recreate/updateParity` on. For the reported deep-heal scenario, `forceStart` should only affect queue/admission behavior; it should not reinterpret a request that explicitly says `remove=false`.

#3084 also covers more of the actual repair path: admin option propagation, `forceStart` admission semantics, erasure-set healer options, disk-index error transposition, inline-data part verification, per-copy metadata/checksum handling, and an e2e regression that wipes one disk and verifies admin deep heal rebuilds it. This PR has useful ideas, especially keeping bucket metadata heal non-destructive during recursive bucket heal, and possibly serializing deep object heal to avoid namespace lock contention. I would prefer extracting those pieces into #3084 or a follow-up PR instead of merging this branch as-is.

So my recommendation is: do not merge #3085 directly; use #3084 as the base candidate, then selectively carry over the safer bucket-metadata/object-heal option split if still needed.

marshawcoco:
Updated this branch on top of #3084 / latest main and narrowed the PR to the pieces that are still useful as a follow-up.

Changes kept after the rebase/merge:

- Removed this PR’s global `store_list_objects.rs` walk semantic change; the branch no longer changes ECStore walk/listing behavior relative to main.
- Kept #3084 forceStart semantics: `forceStart` is only admission/queue behavior and does not override explicit admin heal options.
- Kept recursive bucket metadata heal non-destructive by using `remove=false` for the bucket metadata phase while preserving the requested object heal options for recursive object repair.
- Kept deep erasure-set object heal serialized and skips the `object_exists` read-path precheck in deep scan to avoid duplicate read-repair/object-heal lock contention.

Local verification after merging latest main:

```bash
cargo fmt --all -- --check
RUSTFLAGS="--cfg tokio_unstable" cargo test -p rustfs-heal --lib erasure_healer -- --nocapture
RUSTFLAGS="--cfg tokio_unstable" cargo test -p rustfs-heal --lib recursive_bucket_heal -- --nocapture
RUSTFLAGS="--cfg tokio_unstable" cargo test -p rustfs-heal --lib force_start -- --nocapture
RUSTFLAGS="--cfg tokio_unstable" cargo test -p rustfs-ecstore walk_result_from_set_errors -- --nocapture
RUSTFLAGS="--cfg tokio_unstable" cargo build --bin rustfs
RUSTFLAGS="--cfg tokio_unstable" cargo test -p e2e_test heal_erasure_disk_rebuild -- --nocapture
```

houseme (approved):

overtrue (approved):
Looks good.

PR #3090, opened 2026-05-27
fix(lifecycle): make transition worker resize nonblocking

GatewayJ:
## Summary

- Replace the shared transition worker stop channel with per-worker cancellation tokens tracked in a small worker registry.
- Avoid awaiting stop-signal sends while resizing lifecycle transition workers.
- Remove the unused public transition worker entrypoint and restore tests through the registry path.

## Why

The old resize path could block while sending stop signals through a bounded channel, and shared stop signals could leave stale cancellation messages that affect later workers. Tracking each worker directly makes resize deterministic and keeps cancellation scoped to the worker being removed.

## Validation

- `cargo fmt --all`
- `cargo test -p rustfs-ecstore transition_worker_resize_cancels_removed_workers_directly -- --nocapture --test-threads=1`
- `cargo test -p rustfs-ecstore transition_state_init_honors_runtime_configured_worker_count -- --nocapture --test-threads=1`
- `cargo test -p rustfs-ecstore --lib -- --nocapture --test-threads=1`

### DIFF
--- a/crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs
+++ b/crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs
@@ -75,6 +75,8 @@ use time::OffsetDateTime;
 use tokio::select;
 use tokio::sync::mpsc::{Receiver, Sender};
 use tokio::sync::{RwLock, mpsc};
+use tokio::task::JoinHandle;
+use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, warn};
 use uuid::Uuid;
 use xxhash_rust::xxh64;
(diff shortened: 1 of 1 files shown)


houseme (approved):

houseme (approved):
I reviewed the PR body, the full patch, and the surrounding transition-worker flow in crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs.

I did not find a blocking correctness issue in this change.

The move from a shared stop channel to per-worker cancellation tokens makes the downsize path non-blocking and avoids stale stop signals affecting later workers. The updated regression coverage also exercises both runtime-configured worker sizing and direct cancellation of removed workers.

I did not find an explicitly linked issue for this PR in the PR metadata or timeline.

Code review verdict: approve.

CI status note: Build RustFS Debug Binary and Test and Lint were still pending when I checked, so merge readiness still depends on those checks passing.

PR #3094, opened 2026-05-27
fix(iam): avoid stale cache replacement on walk errors

### LINKED ISSUE #3044: Access keys intermittently disappear from console list, but S3 API still works
## Description

I created an access key from the RustFS console. After that, I tested the access key through the S3 API, and both file upload and download worked correctly.

However, I noticed a strange issue in the RustFS console:

After refreshing the access key list in the console, the access key sometimes disappears from the list. In most cases, the access key is not visible in the console. Occasionally, after refreshing many times, it appears again, but this happens very rarely, roughly once in dozens of refresh attempts.

The confusing part is that even when the access key is not visible in the console, the same access key still works normally through the S3 API. I can still upload and download objects successfully.

## Environment

- RustFS version: beta.4
- Deployment mode: distributed cluster
- Number of nodes: 4 cloud servers
- Disks per node: 4 disks
- Reverse proxy: Nginx

## Nginx / Access Test

I tested both of the following Nginx configurations:

1. Proxying only to a single RustFS node
2. Load balancing across all 4 RustFS nodes

The issue exists in both cases.

I also tested with an incognito/private browser window, but the behavior did not improve.

## What I checked

I checked the RustFS data directory on every node and every mounted disk.

The access key exists on disk under the following path:

`/data/rustfs/disk2/.rustfs.sys/config/iam/service-accounts`

I was able to find the service account/access key data on the disks.

So it seems that:

- The access key was created successfully
- The access key exists in the backend storage
- The access key is valid for S3 API operations
- But the console access key list does not consistently show it

## Actual Behavior

The access key list in the RustFS console intermittently does not show the created access key.

Most of the time, the access key is missing from the console list. Occasionally, it appears after many refresh attempts.

## Expected Behavior

The RustFS console should consistently show the created access key as long as the access key exists and is valid.

## Reproduction Steps

1. Deploy a RustFS distributed cluster with 4 nodes and 4 disks per node.
2. Open the RustFS console.
3. Create a new access key from the console.
4. Use the generated access key to test S3 API upload/download.
5. Confirm that S3 API upload/download works.
6. Refresh the access key list in the console.
7. Observe that the access key sometimes disappears from the list.
8. Refresh many times.
9. Occasionally, the same access key may appear again.

## Additional Notes

This does not look like a browser cache issue, because I tested with an incognito/private browser window.

It also does not look like a simple Nginx load balancing issue, because I tested both single-node proxy and 4-node load balancing, and the behavior was the same.

Since the S3 API continues to work and the service account files exist on disk, I suspect this may be related to how the console queries or aggregates IAM/service account data in distributed mode.

Could you please help confirm whether this is a known issue in beta.4, or if there is any additional debug information I should provide?

Image

Image

GatewayJ:
## Related Issues

Fixes #3044.

## Summary of Changes

- Propagate IAM config walk failures back to `list_all_iamconfig_items` instead of only logging them.
- Stop `load_all` from replacing service-account-related IAM caches with partial reload results after a walk failure.
- Avoid replacing `users`, `sts_accounts`, and `sts_policies` when any of those caches changed during the full reload window, preserving newer notification-driven updates.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-iam --lib`

## Impact

When IAM config walking fails during a full reload, RustFS now keeps the existing IAM cache instead of replacing it with incomplete data. This avoids intermittent missing service accounts in admin list responses while preserving normal successful reload behavior. No API, configuration, or data format changes are expected.

## Additional Notes

The cache replacement guard follows the same safety goal as MinIO's reload timestamp check, adapted to RustFS's `ArcSwap` cache structure.

### DIFF
--- a/crates/iam/src/store/object.rs
+++ b/crates/iam/src/store/object.rs
@@ -362,6 +362,7 @@ impl ObjectStore {
         let (tx, mut rx) = mpsc::channel::<ObjectInfoOrErr>(100);
 
         let path = prefix.to_owned();
+        let sender_on_error = sender.clone();
         tokio::spawn(async move {
             if let Err(err) = store
                 .walk(ctx.clone(), Self::BUCKET_NAME, &path, tx, WalkOptions::default())
@@ -378,6 +379,12 @@ impl ObjectStore {
                     error = %err,
                     "system path walk failed"
                 );
+                let _ = sender_on_error
+                    .send(StringOrErr {
+                        item: None,
+                        err: Some(err.into()),
+                    })
+                    .await;
             }
         });
 
@@ -1018,6 +1025,9 @@ impl Store for ObjectStore {
     }
 
     async fn load_all(&self, cache: &Cache) -> Result<()> {
+        let users_snapshot = cache.users.load();
+        let sts_accounts_snapshot = cache.sts_accounts.load();
+        let sts_policies_snapshot = cache.sts_policies.load();
         let listed_config_items = self.list_all_iamconfig_items().await?;
 
         let mut policy_docs_cache = CacheEntity::new(get_default_policyes());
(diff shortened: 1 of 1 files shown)


houseme (changes requested):
Thanks for working on the #3044 fix. I found one blocking issue in the new reload guard.

`crates/iam/src/store/object.rs:1072,1128,1172,1190,1262`

The new pointer check only protects the final `users` / `sts_accounts` / `sts_policies` stores, but `policy_docs`, `groups`, `user_policies`, and `group_policies` are still overwritten earlier from the full-reload snapshot. A concurrent notification can update `users` + `user_policies` (or `groups`) after the snapshots are taken and before those earlier stores run. In that case the guard skips the final user stores, but the earlier stale stores have already landed, leaving the cache set internally inconsistent.

A concrete example is `user_notification_handler(UserType::Reg|Svc)`: it refreshes both `users` and `user_policies`. If that notification fires before `load_all()` reaches `cache.user_policies.store(...)`, the full reload can roll `user_policies` back to the stale snapshot while keeping the newer `users` entry because the final pointer guard bails out.

Please move the cache writes behind a single consistency guard, or extend the snapshot/guard logic to every cache that `load_all()` mutates, so a concurrent notification cannot partially roll back the reload.

houseme (approved):
Re-read the latest change set on commit `0059e32861633f097f01033c33c91544d1625d5c`.

The previous blocking issue is addressed: `load_all()` now snapshots all IAM caches that it mutates and commits them behind a single guard, so concurrent notification-driven cache updates no longer allow a partial rollback from the full reload path.

I did not find a new blocking issue in the updated diff.

PR #3098, opened 2026-05-27
fix: retry namespace lock quorum contention

overtrue:
## Summary
- retry recoverable namespace lock quorum misses within the configured acquire timeout
- treat only client/task failures as unrecoverable quorum loss, not ordinary lock contention responses
- add a clustered concurrent overwrite regression test for namespace lock quorum errors

## Validation
- cargo fmt --all --check
- cargo test -p rustfs-lock namespace_lock -- --nocapture
- cargo build -p rustfs
- RUST_LOG=warn cargo test -p e2e_test namespace_lock_quorum_test::test_concurrent_cluster_overwrites_do_not_fail_namespace_lock_quorum -- --nocapture

### DIFF
--- a/crates/e2e_test/src/lib.rs
+++ b/crates/e2e_test/src/lib.rs
@@ -142,4 +142,7 @@ mod replication_extension_test;
 #[cfg(test)]
 mod snowball_auto_extract_test;
 
+#[cfg(test)]
+mod namespace_lock_quorum_test;
+
 pub mod tls_gen;
--- a/crates/e2e_test/src/namespace_lock_quorum_test.rs
+++ b/crates/e2e_test/src/namespace_lock_quorum_test.rs
@@ -1,0 +1,119 @@
+// Copyright 2024 RustFS Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::common::RustFSTestClusterEnvironment;
+use aws_sdk_s3::Client;
+use aws_sdk_s3::error::SdkError;
+use bytes::Bytes;
+use serial_test::serial;
+use std::sync::Arc;
+use tokio::sync::Barrier;
+use tracing::{info, warn};
+
+const BUCKET: &str = "namespace-lock-quorum-bucket";
+const KEY: &str = "thumb/79/concurrent-overwrite.jpg";
+
+type TestResult = Result<(), Box<dyn std::error::Error + Send + Sync>>;
+
+async fn put_object(client: Client, payload: Vec<u8>, writer_id: usize) -> Result<(), String> {
+    client
+        .put_object()
+        .bucket(BUCKET)
+        .key(KEY)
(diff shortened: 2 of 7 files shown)


houseme (changes requested):
I found one blocking correctness issue in `crates/lock/src/distributed_lock.rs`. The new `hard_failures` tracking is not enough yet because the generic early-exit at `individual_locks.len() + pending.len() < required_quorum` still runs before the unrecoverable branch. In a 4-node/quorum-3 case, two `Remote lock RPC failed: ...` responses plus one ordinary miss will hit that earlier path and return `"Failed to acquire quorum"`, so `acquire_guard()` classifies it as retryable quorum contention and keeps retrying until timeout instead of surfacing `LockError::QuorumNotReached` immediately. Please preserve the hard-failure classification in that first impossible-to-reach-quorum path (or merge the two checks), and add coverage for remote RPC failure because the new healthy-cluster overwrite test does not exercise this degraded-cluster case.

overtrue:
Updated the PR to address the latest review feedback:

- Remote lock RPC timeout responses are now classified as hard remote-lock failures, so unreachable/timed-out peers contribute to quorum-loss detection consistently.
- Impossible quorum attempts now fail fast even when no individual locks have been acquired yet, instead of waiting for delayed pending tasks.
- Added focused regression coverage for both cases.

Validation:
- `cargo test -p rustfs-lock distributed_lock::tests`
- `cargo clippy -p rustfs-lock --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`
- Continuous Integration run 26543544014 passed on head `8f78b6bf8b79f0bf3acce1b0182db65a747bcdfb`.

houseme (changes requested):
I re-reviewed the latest head `0d4ab1921b78a391191b48c1e0079e6c3f8c27a8` and there is still a blocking correctness issue in `crates/lock/src/distributed_lock.rs`.

The new retry loop in `acquire_lock_quorum_with_retry()` still returns too early for transient contention/timeout responses because `pending_cleanup_spawned` is treated as a hard stop condition. In the current code, if an attempt becomes impossible before all spawned tasks finish, `acquire_lock_quorum_once()` returns a retryable failure with `pending_cleanup_spawned = true`, and `acquire_lock_quorum_with_retry()` immediately returns that result instead of retrying.

That breaks the intended transient-timeout recovery path. I reproduced it on the latest PR head with:

`cargo test -p rustfs-lock namespace::tests -- --nocapture`

which fails in `test_namespace_lock_distributed_retries_transient_acquire_timeout` with:

`transient timeout should be retried before the acquire budget expires: Timeout { resource: "bucket/object-flaky-acquire@latest", timeout: 1s }`

The relevant flow is around:
- `acquire_lock_quorum_with_retry()` early-return condition (`pending_cleanup_spawned`)
- `acquire_lock_quorum_once()` returning retryable contention after spawning pending cleanup for the remaining tasks

Please keep cleanup of outstanding tasks, but do not let `pending_cleanup_spawned` suppress retries for retryable contention/timeout results. Right now the PR’s own regression test proves the retry behavior is still broken.

houseme (approved):
Re-reviewed the latest head `9d4147ea834d24e5296f9548a8e1016c28564772`.

The earlier blocking correctness issues around remote RPC hard-failure classification and premature retry termination are addressed now. I also re-ran the targeted `rustfs-lock` tests and the lock-only clippy gate locally.

I left only two minor Copilot nits open for possible follow-up around logging noise/verbosity; they are non-blocking from my perspective.

houseme:
Re-checked the latest head `9d4147ea834d24e5296f9548a8e1016c28564772` against all existing review feedback.

The earlier blocking correctness issues are addressed now, and all review threads have been resolved. The remaining suggestions were non-blocking follow-up items rather than merge blockers.

PR #3099, opened 2026-05-27
fix(ecstore): offload erasure encoding from async workers

### LINKED ISSUE #3097: Distributed lock RPC timeout under high-concurrency PutObject due to tokio runtime thread starvation
## Summary

Under high-concurrency `PutObject` workloads (e.g., Apache Paimon writing multiple manifest/manifest-list files in a single snapshot commit), all distributed lock RPCs time out simultaneously, causing cascading failures including heal task failures and 500 errors returned to S3 clients.

## Root Cause Analysis

After investigating the source code, the root cause appears to be **tokio runtime thread starvation**:

1. **EC encoding is synchronous CPU-intensive computation** running inside async tasks on the tokio worker thread pool (`crates/ecstore/src/erasure_coding/erasure.rs`)
2. **Lock RPCs and data writes share the same tokio runtime** (`rustfs/src/server/runtime.rs`, worker threads = physical core count)
3. Under high-concurrency writes, EC encoding saturates all worker threads → lock RPC futures cannot be polled → 5s timeout fires
4. After timeout, `evict_failed_connection` clears the cached gRPC channel → reconnection storm → further degradation

## Observed Symptoms

```
WARN "Evicting cached remote lock connection after RPC failure" op="lock" reason="RPC timed out after 5s"
WARN "Failed to acquire lock on client, error: Lock acquisition timeout"
WARN "acquire_lock_quorum error: Failed to acquire quorum: 0/5 required"
ERROR "Heal operation failed: ... Quorum not reached: required 5, achieved 0"
```

All nodes timeout simultaneously — this is consistent with the **requesting node's runtime being starved** (futures never get scheduled), rather than remote nodes being unresponsive.

## Contributing Factors

- **HDD storage**: slow random IOPS causes `spawn_blocking` thread pool saturation, back-pressuring worker threads
- **Shared port for gRPC and S3 API** (`HybridService` in `server/hybrid.rs`): data traffic and lock RPCs share the same TCP connection, though this is secondary to the runtime starvation issue
- **No retry mechanism**: unlike MinIO's lock implementation which retries with exponential backoff (250ms to 5s) within a 10-minute window, RustFS fails immediately after a single 5s timeout
- **Every PutObject acquires a distributed write lock** (`set_disk.rs:963`, before `rename_data`), even without HTTP preconditions — this amplifies the lock RPC volume

## Suggested Fixes (by priority)

1. **Move EC encoding to `spawn_blocking` or a dedicated thread pool** — this is the root fix. CPU-intensive Reed-Solomon computation should not run on tokio worker threads.

2. **Add retry with backoff for lock acquisition** — a single timeout should not be terminal. Exponential backoff (e.g., 250ms → 5s, total budget configurable) would prevent cascading failures.

3. **Consider removing distributed lock for unconditional PutObject** — MinIO's approach: unconditional overwrites rely on erasure coding quorum for consistency, no distributed lock needed. Only conditional writes (If-None-Match, etc.) require locking.

4. **Separate gRPC and S3 API onto different ports** — reduces interference between data traffic and lock RPCs, though this alone won't fix runtime starvation.

## Use Case: Apache Paimon on RustFS

Apache Paimon (data lakehouse format) performs the following during a single snapshot commit:
- Write 3-5 manifest files (PutObject, small files ~KB)
- Write 1 manifest-list file (PutObject)
- Write 1 snapshot file (via rename = CopyObject + DeleteObject)

This burst of ~5-10 concurrent S3 operations per commit triggers the lock contention described above.

## Environment

- RustFS distributed cluster (multiple nodes, erasure coding)
- HDD storage
- Host network mode (Docker networking ruled out)
- `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT` at default (5s)

## Reproduction

1. Deploy a multi-node RustFS cluster with HDD
2. Run Apache Paimon (or similar workload) that issues 5-10 concurrent PutObject requests per commit batch
3. Observe lock RPC timeouts in logs when write concurrency is sustained

marshawcoco:
## Related Issues
Fixes #3097.

## Summary of Changes
- Offload erasure block encoding in the async write pipeline to Tokio's blocking pool.
- Apply the same scheduling behavior to the async erasure callback helper.
- Keep the existing shard encoding logic and write quorum flow unchanged.

## Verification
- `cargo fmt --all`
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-ecstore erasure_coding::encode::tests::encode_shutdowns_writers_after_small_shards`
- `cargo test -p rustfs-ecstore erasure_coding::erasure::tests::test_encode_stream_callback_async_channel_decode`

## Impact
High-concurrency PutObject workloads spend EC encoding CPU time outside Tokio worker threads, reducing the chance that lock RPC and other async tasks are starved by synchronous Reed-Solomon work.

## Additional Notes
This is intentionally scoped to the EC encoding starvation path. Lock retry behavior and PutObject lock semantics are left unchanged.

### DIFF
--- a/crates/ecstore/src/erasure_coding/encode.rs
+++ b/crates/ecstore/src/erasure_coding/encode.rs
@@ -227,10 +227,15 @@ impl Erasure {
                     Ok(n) if n > 0 => {
                         total += n;
                         let erasure = self.clone();
-                        let block = buf[..n].to_vec();
-                        let res = tokio::task::spawn_blocking(move || erasure.encode_data(&block))
-                            .await
-                            .map_err(|err| std::io::Error::other(format!("EC encode task failed: {err}")))??;
+                        let encode_buf = std::mem::take(&mut buf);
+                        let (res, returned_buf) = tokio::task::spawn_blocking(move || {
+                            let res = erasure.encode_data(&encode_buf[..n]);
+                            (res, encode_buf)
+                        })
+                        .await
+                        .map_err(|err| std::io::Error::other(format!("EC encode task failed: {err}")))?;
+                        buf = returned_buf;
+                        let res = res?;
                         let queued_bytes = queued_block_bytes(&res);
                         rustfs_io_metrics::add_ec_encode_inflight_bytes(queued_bytes);
                         if let Err(err) = tx.send(res).await {
--- a/crates/ecstore/src/erasure_coding/encode.rs
+++ b/crates/ecstore/src/erasure_coding/encode.rs
@@ -226,7 +226,11 @@ impl Erasure {
                 match rustfs_utils::read_full(&mut reader, &mut buf).await {
                     Ok(n) if n > 0 => {
                         total += n;
-                        let res = self.encode_data(&buf[..n])?;
+                        let erasure = self.clone();
+                        let block = buf[..n].to_vec();
+                        let res = tokio::task::spawn_blocking(move || erasure.encode_data(&block))
+                            .await
+                            .map_err(|err| std::io::Error::other(format!("EC encode task failed: {err}")))??;
                         let queued_bytes = queued_block_bytes(&res);
                         rustfs_io_metrics::add_ec_encode_inflight_bytes(queued_bytes);
                         if let Err(err) = tx.send(res).await {
--- a/crates/ecstore/src/erasure_coding/erasure.rs
+++ b/crates/ecstore/src/erasure_coding/erasure.rs
@@ -587,7 +587,12 @@ impl Erasure {
                 Ok(n) if n > 0 => {
                     warn!("encode_stream_callback_async read n={}", n);
                     total += n;
-                    let res = self.encode_data(&buf[..n]);
+                    let erasure = self.clone();
+                    let block = buf[..n].to_vec();
+                    let res = match tokio::task::spawn_blocking(move || erasure.encode_data(&block)).await {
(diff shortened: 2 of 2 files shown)


marshawcoco:
I also ran a local synthetic runtime-starvation repro to validate the direction of this fix.

  Setup:
  - Tokio runtime with `worker_threads=1`
  - `Erasure::encode` over a 256 MiB payload
  - In parallel, a 10 ms Tokio timer task records the maximum scheduling delay
  - Same temporary test was run against the parent commit (`4648de9e`) and this PR branch

  Results:
  - Before this PR (`4648de9e`): `elapsed_ms=5673 max_timer_delay_ms=128`
  - With this PR: `elapsed_ms=5640 max_timer_delay_ms=33`

  This does not reproduce the full 5s lock RPC timeout from #3097 on my local machine, because the repro uses in-memory writers and does not include HDD latency, multi-node gRPC, connection eviction, or real concurrent S3 traffic. It does reproduce the underlying scheduling symptom: synchronous EC encoding delays Tokio timer/task scheduling, and moving the encoding work to `spawn_blocking` significantly reduces that delay in the same workload.

overtrue (commented):
I did not find a blocking code issue; holding approval until the pending S3 Implemented Tests check completes.

houseme (approved):

PR #3101, opened 2026-05-27
fix(lock): retry transient distributed lock timeouts

marshawcoco:
## Related Issues
  Related to #3097.

  ## Summary of Changes
  - Keep cached remote lock connections when a local lock RPC timeout fires.
  - Continue evicting cached remote lock connections for explicit gRPC status errors.
  - Add bounded retry/backoff for transient distributed lock acquire failures within the caller's acquire timeout budget.
  - Retry only when an acquire attempt obtains zero node locks, avoiding retries while partial lock cleanup may still be in progress.

  ## Verification
  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `git diff --check`
  - `cargo test -p rustfs-ecstore rpc::remote_locker::tests -- --nocapture`
  - `cargo test -p rustfs-lock namespace::tests`

  ## Impact
  Distributed lock acquisition is more tolerant of transient RPC scheduling delays without extending the configured acquire timeout budget. Local lock RPC
  timeouts no longer trigger cached channel eviction, reducing connection churn after temporary runtime pressure.

  ## Additional Notes
  This PR is scoped to lock RPC timeout handling and distributed lock acquire retry behavior. It does not change PutObject lock semantics or quorum
  requirements.

### DIFF
--- a/crates/ecstore/src/rpc/remote_locker.rs
+++ b/crates/ecstore/src/rpc/remote_locker.rs
@@ -101,7 +101,12 @@ impl RemoteClient {
             }
             Err(_) => {
                 let reason = format!("RPC timed out after {:?}", timeout_duration);
-                self.evict_connection(op, &reason).await;
+                warn!(
+                    addr = %self.addr,
+                    op,
+                    reason,
+                    "Remote lock RPC timed out without evicting cached connection"
+                );
                 Err(LockError::timeout(format!("remote lock RPC {op} on {}", self.addr), timeout_duration))
             }
         }
@@ -494,7 +499,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_remote_client_acquire_lock_respects_request_timeout_and_evicts_connection() {
+    async fn test_remote_client_acquire_lock_respects_request_timeout_without_evicting_connection() {
         ensure_test_rpc_secret();
         let (addr, accept_task) = spawn_hanging_listener().await;
         cache_lazy_channel(&addr).await;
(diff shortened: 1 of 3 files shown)


houseme (changes requested):
[P1] `acquire_lock_quorum_with_retry` currently shrinks each acquire attempt to `request.acquire_timeout / 3`, which regresses ordinary lock handoff that should still succeed near the end of the caller's original timeout budget.

In `crates/lock/src/distributed_lock.rs`, `attempt_request.acquire_timeout = Self::lock_acquire_attempt_timeout(...)` means a caller that asked for a 1s lock wait now gives each node only about 333ms per attempt before we back off and retry. That is fine for transient RPC starvation, but it changes the semantics for real lock contention because `LocalClient::acquire_lock` legitimately waits for up to `request.acquire_timeout` for the current holder to release the lock.

I reproduced this on the PR head with a targeted test: owner A acquires the distributed lock, owner B starts `get_write_lock(..., Duration::from_secs(1))`, owner A releases at 950ms, and owner B still times out on this branch. Before this change, the same scenario succeeds because the first acquire attempt is allowed to wait for the full 1s budget.

That means the retry loop is not only masking transient RPC scheduling stalls, it is also turning normal late lock release into a false timeout. Please preserve the original full-budget wait semantics for contention (for example by only slicing the timeout for known RPC-timeout paths, or by letting an individual attempt wait for the remaining budget when the failure mode is lock contention rather than transport starvation).

houseme (approved):
Re-reviewed the latest head `bd5e841f`.

The previous timeout-budget regression in `acquire_lock_quorum_with_retry` is fixed now: the retry path preserves the caller's remaining acquire budget, and the new late-release regression test covers the scenario that was failing before.

I also re-ran the targeted checks below on the latest head:
- `cargo test -p rustfs-lock test_namespace_lock_distributed_retries_transient_acquire_timeout -- --nocapture`
- `cargo test -p rustfs-lock test_namespace_lock_distributed_waits_full_timeout_for_late_release -- --nocapture`
- `cargo test -p rustfs-ecstore rpc::remote_locker::tests -- --nocapture`

No further findings from this review pass.

PR #3105, opened 2026-05-28
fix(iam): serialize IAM cache writes

GatewayJ:
## Related Issues
N/A

## Summary of Changes
- Serialize IAM cache writes with a cache-level write lock while keeping ArcSwap reads lock-free.
- Commit related IAM cache map updates together for full reloads and notifications.
- Fix notification cache edge cases for group reverse memberships and STS policy mappings.
- Add regression tests for group notification membership creation and STS policy mapping updates.

## Verification
- `make pre-commit`
- `cargo fmt --all --check`
- `cargo check -p rustfs-iam`
- `cargo test -p rustfs-iam --lib`

## Impact
IAM cache write paths are serialized to avoid transient inconsistent views during full reloads and cross-node notifications. Read paths remain lock-free via ArcSwap. No configuration, API, or storage format changes.

## Additional Notes
`make pre-commit` used the `cargo test` fallback because `cargo-nextest` is not installed locally.


houseme (approved):
Reviewed the IAM cache refactor and notification-path fixes.

I verified the multi-map cache commit path now publishes one coherent snapshot under `with_write_lock`, and the two targeted regressions are covered by dedicated tests:
- group notification rebuilds reverse memberships for first-time users
- STS policy mapping notifications update `sts_policies`

No new blocking issues found in this patch set. Approving pending CI.

PR #3115, opened 2026-05-29
feat(rio): rio_v2 is compatible with minio for storing data.

reatang:
This update improves `MinIO-compatible` object read/write capabilities under `rio-v2`: 
- Compressed objects use `klauspost/compress/s2` framed streams, recognizing `X-Minio-Internal-compression` and `actual-size`, and supporting part index/range reads; 
- SSE-S3, SSE-KMS, and SSE-C are uniformly switched to the DAREv2 object-key model, using X-Minio-Internal-* to store sealed keys, KMS data keys, and encryption contexts, while retaining historical direct mode read compatibility with x-rustfs-encryption-*, ensuring that compressed, encrypted, and combined objects can be recovered according to MinIO metadata semantics.

Data structure and algorithm compatibility with miniio for disk data persistence.
This design incorporates compatibility features; it will only be triggered when `features=rio-v2`.

[x] Compression Engine
[x] Encryption Engine


crates/rio-v2/tests/minio_generated_fixtures.rs
```diff
@@ -0,0 +1,282 @@
...
+    .unwrap_or_else(|err| panic!("decode {}: {err}", xl_meta.display()))
+}
+
+fn read_manifest(case_id: &str) -> ManifestRecord {
+    read_json(&case_dir(case_id).join("manifest.json"))
+}
+
+fn require_generated_fixture_root() {
+    let root = fixture_root();
+    assert!(
```
overtrue, line 99: CI is failing because non-ignored fixture tests require generated files that are not committed. Please either commit the minio-generated fixtures or mark these tests ignored/skip when fixtures are absent so the default test suite passes.

overtrue (changes requested):
CI is failing because non-ignored fixture tests require generated files that are not committed. Please either commit the minio-generated fixtures or mark these tests ignored/skip when fixtures are absent so the default test suite passes.

overtrue (changes requested):
Blocking: rustfs/src/app/object_usecase.rs:1132-1147 still attaches per-chunk inspect logging on every GetObject stream and passes response_content_length as usize to bytes_stream even though the code already clamps expected. Please gate the chunk logging behind tracing-chunk-debug and pass expected to bytes_stream.

overtrue (approved):
Looks good.

PR #3117, opened 2026-05-29
fix(s3): preserve listing pagination parity

weisd:
## Related Issues
N/A

## Summary of Changes
- Preserve S3 listing response encoding for object keys, common prefixes, markers, prefixes, delimiters, and start-after values when URL encoding is requested.
- Tighten listing pagination around version markers, including explicit null-version markers and raw-vs-encoded next marker comparisons.
- Add a configurable list quorum path and reduce non-versioned listing metadata reads to the requested latest version.
- Harden cached listing marker parsing so object names containing bracket characters are not truncated.
- Add focused regression coverage for V1, V2, version listing, marker parsing, quorum selection, and metadata resolution behavior.

## Verification
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- `git diff --check`
- `make pre-commit`

## Impact
- Improves S3 listing compatibility for encoded responses and versioned pagination.
- Adds `RUSTFS_API_LIST_QUORUM` for list quorum tuning; invalid values fall back to strict behavior.
- No migration or storage format change.

## Additional Notes
- Local verification used the Rust 1.95 toolchain path explicitly because the host Homebrew clippy resolved to 1.94.
- `make pre-commit` was run with `LC_ALL=en_US.UTF-8` because `C.UTF-8` is unavailable on this macOS host.


overtrue (approved):

houseme (approved):
Reviewed the latest PR #3117 diff end-to-end with focus on S3 listing semantics and version pagination. The changes for URL-encoded response parity, null version marker handling, marker-scoped version pagination, and non-versioned metadata resolution are internally consistent and covered by focused regressions. I found no blocking issues in this scope.

PR #3129, opened 2026-05-30
fix(scanner): support PBS subfolder alert threshold

### LINKED ISSUE #3126: Scanner incompatible with Proxmox Backup Server S3 datastore: hardcoded 50,000 subfolder threshold below PBS's required 65,536
**Describe the bug**

RustFS crashes and restarts in a loop when used as a Proxmox Backup Server (PBS) S3 backend. The scanner detects that the `.chunks` directory has more than 50,000 subdirectories and aborts, triggering a process restart. This happens because PBS unconditionally pre-creates exactly 65,536 subdirectories (`0000`–`ffff`) under `.chunks` when initializing any datastore — this is an explicit, documented PBS design requirement. RustFS's hardcoded 50,000 threshold is below that, making the two incompatible by design.

**To Reproduce**

1. Run RustFS 1.0.0-beta.6 via Docker
2. Configure a Proxmox Backup Server 4.x datastore with an S3 backend pointing at RustFS
3. Initialize the PBS datastore — PBS creates all 65,536 subdirectories under `.chunks`
4. RustFS scanner detects the threshold violation, aborts, and restarts in a loop

**Expected behavior**

RustFS should handle buckets with more than 65,536 direct subfolders without crashing. The scanner threshold should be at least 65,538 (PBS's documented minimum) or configurable via environment variable.

**Screenshots**

Relevant log output:

```json
{"level":"WARN","fields":{"message":"scanner detected folder with excessive direct subfolders","root":"/data","folder":"pbs-s3/pbs-s3/.chunks","folders":65389,"threshold":50000},"target":"rustfs_scanner::scanner_folder"}
{"level":"WARN","fields":{"message":"scan_folder: failed to scan child folder pbs-s3/pbs-s3/.chunks: Scanner error: Operation cancelled"},"target":"rustfs_scanner::scanner_folder"}
{"level":"WARN","fields":{"message":"scan_folder: failed to scan child folder pbs-s3/pbs-s3: Scanner error: Operation cancelled"},"target":"rustfs_scanner::scanner_folder"}
{"level":"WARN","fields":{"message":"Audit system not initialized, cannot stop"},"target":"rustfs::server::audit"}
```

This sequence repeats indefinitely — every scanner cycle triggers a restart.

**Additional context**

PBS documents this requirement explicitly at https://pbs.proxmox.com/docs/storage.html:

> "The File Layout requires the file system to support at least 65,538 subdirectories per directory. That number comes from the 2¹⁶ pre-created chunk namespace directories, and the `.` and `..` default directory entries."

The 50,000 threshold is hardcoded in `crates/scanner/src/scanner_folder.rs`. Raising it above 65,538 or exposing it as `RUSTFS_SCANNER_MAX_SUBFOLDERS` would fix the incompatibility.

- **RustFS version:** 1.0.0-beta.6 (`rustfs/rustfs:latest` Docker image)
- **PBS version:** 4.2.1
- **OS:** Debian Linux (ARM64)

marshawcoco:
## Related Issues

  Fixes #3126.

  ## Summary of Changes

  Raise the default scanner excessive-folder alert threshold from `50_000` to `65_538` so the default configuration supports Proxmox
  Backup Server's pre-created `.chunks` namespace layout.

  Add a focused regression test for the default threshold while preserving the existing `RUSTFS_SCANNER_ALERT_EXCESS_FOLDERS`
  environment override behavior.

  ## Verification

  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `cargo test -p rustfs-scanner excessive_folders_threshold`
  - `git diff --check`

  Not run:
  - `make pre-commit` per request.

  ## Impact

  The scanner no longer emits excessive direct-subfolder alerts by default for PBS-compatible chunk directory layouts. Deployments
  can still override the threshold with `RUSTFS_SCANNER_ALERT_EXCESS_FOLDERS`.

  No API, storage format, or migration impact.

  ## Additional Notes

  N/A

### DIFF
--- a/crates/config/src/constants/scanner.rs
+++ b/crates/config/src/constants/scanner.rs
@@ -65,11 +65,12 @@ pub const ENV_SCANNER_ALERT_EXCESS_VERSION_SIZE: &str = "RUSTFS_SCANNER_ALERT_EX
 pub const DEFAULT_SCANNER_ALERT_EXCESS_VERSION_SIZE: u64 = 1024 * 1024 * 1024 * 1024;
 
 /// Environment variable that controls how many subfolders trigger scanner alerts.
-/// - Example: `export RUSTFS_SCANNER_ALERT_EXCESS_FOLDERS=50000`
+/// - Example: `export RUSTFS_SCANNER_ALERT_EXCESS_FOLDERS=65538`
 pub const ENV_SCANNER_ALERT_EXCESS_FOLDERS: &str = "RUSTFS_SCANNER_ALERT_EXCESS_FOLDERS";
 
 /// Default subfolder count that triggers scanner alerts.
-pub const DEFAULT_SCANNER_ALERT_EXCESS_FOLDERS: u64 = 50_000;
+/// Allows Proxmox Backup Server's chunk namespace layout.
+pub const DEFAULT_SCANNER_ALERT_EXCESS_FOLDERS: u64 = 65_538;
 
 /// Environment variable that controls whether the scanner sleeps between operations.
 /// When `true` (default), the scanner throttles itself. When `false`, it runs at full speed.
--- a/crates/scanner/src/scanner_folder.rs
+++ b/crates/scanner/src/scanner_folder.rs
@@ -1666,7 +1666,7 @@ mod tests {
     #[cfg(unix)]
     use std::os::unix::fs::{PermissionsExt, symlink};
     use std::sync::atomic::AtomicBool;
-    use temp_env::with_var;
+    use temp_env::{with_var, with_var_unset};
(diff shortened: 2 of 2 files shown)


houseme (approved):
Reviewed with issue #3126 context. The change raises DEFAULT_SCANNER_ALERT_EXCESS_FOLDERS from 50_000 to 65_538 and updates the env-var docs/example, while preserving the RUSTFS_SCANNER_ALERT_EXCESS_FOLDERS override path. The added regression test covers the default-unset behavior explicitly. This matches the PBS .chunks namespace requirement and I found no blocking regressions in scope.

PR #3141, opened 2026-05-31
fix(rebalance): require target goal before completion

### LINKED ISSUE #3137: Unclear rebalance behavior after adding second pool
**Describe the bug**
After extend my test installation with a second pool starting a rebalance seems do nothing. The new pool is shown and is used. Storing new data is done as well on the second pool. 
```shell
> rc admin pool list root
Pools:
  Pool 0: http://node{1...4}.storage.node:9000/data1
    Decommission: not started
    Usage:        322.05 GiB / 3.64 TiB
  Pool 1: http://node{5...8}.storage.node:9000/data1
    Decommission: not started
    Usage:        74.60 GiB / 3.64 TiB
```
Starting a rebalancing the status shows the following. It finish shortly after start:
```
> rc admin rebalance status root
Rebalance: configured
  ID: 5a461fa0-ca64-4900-9b2d-b68e817c2c2a

Per-pool usage:
  Pool 0: 8.65% used (Completed)
    Progress:   0 B moved, 0 objects, 0 versions
    Remaining:  7 buckets, ETA 0s
    Elapsed:    0s
  Pool 1: 2.00% used (None)
```

**To Reproduce**
1. Setup a four node (one disk each)
2. Fill up with data
3. Add a second pool same configuration (four node, one disk each)
4. Start rebalance

**Expected behavior**
Rebalancing data over both pools.

**Screenshots**
Image

**Additional context**
* Docker Deployment using latest image 1.0.0-beta.6

marshawcoco:
## Related Issues

  Fixes #3137.

  ## Summary of Changes

  - Require a rebalance pool to actually reach the target free-space ratio before marking it completed.
  - Remove the previous fixed tolerance that could complete rebalance before moving any data.
  - Add a regression test for the two-pool imbalance reported in #3137.

  ## Verification

  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `git diff --check`
  - `cargo test -p rustfs-ecstore rebalance_goal --lib`
  - `cargo test -p rustfs-ecstore complete_rebalance_pools_at_goal --lib`
  - `cargo test -p rustfs-ecstore rebalance::rebalance_unit_tests --lib`
  - `cargo test -p rustfs-ecstore --lib`

  ## Impact

  Rebalance may run longer and move data in cases that previously completed early while still below the target balance.

  ## Additional Notes

  N/A

### DIFF
--- a/crates/ecstore/src/rebalance.rs
+++ b/crates/ecstore/src/rebalance.rs
@@ -1002,7 +1002,7 @@ impl ECStore {
                 return true;
             }
 
-            // Mark pool rebalance as done if within 5% of the PercentFreeGoal
+            // Mark pool rebalance as done only after it reaches the PercentFreeGoal.
             let pfi = if pool_stat.init_capacity == 0 {
                 0.0
             } else {
@@ -4138,7 +4138,7 @@ mod rebalance_unit_tests {
     }
 
     #[test]
-    fn test_rebalance_goal_reached_tolerance_and_regression() {
+    fn test_rebalance_goal_reached_requires_target_ratio() {
         let init_free_space = 150_u64;
         let init_capacity = 800_u64;
         let goal = 0.35_f64;
--- a/crates/ecstore/src/rebalance.rs
+++ b/crates/ecstore/src/rebalance.rs
@@ -1032,7 +1032,7 @@ fn rebalance_goal_reached(init_free_space: u64, init_capacity: u64, bytes: u64, 
     }
 
     let pfi = (init_free_space + bytes) as f64 / init_capacity as f64;
-    (pfi - percent_free_goal).abs() <= 0.05 + f64::EPSILON
+    pfi + f64::EPSILON >= percent_free_goal
 }
 
 fn percent_free_ratio(total_free: u64, total_cap: u64) -> f64 {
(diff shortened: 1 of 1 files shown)


houseme:
**Comprehensive Review of PR #3141**

This PR correctly addresses the root cause of #3137. The analysis below covers correctness, edge cases, and test coverage.

**Root Cause Fix**
The `rebalance_goal_reached` function previously used a two-sided ±5% tolerance (`(pfi - goal).abs() <= 0.05`), which allowed a pool to be marked `Completed` even when it had not yet reached the target free-space ratio. In the reported scenario (pool 0 at 8.65% used, target ~94.675% free), pool 0 was within the 5% tolerance window and was immediately marked completed without moving any data. The fix correctly changes this to a one-sided check (`pfi + ε >= goal`), ensuring a pool must actually reach the target before completion.

**Mathematical Verification**
Using the data from #3137:
- Pool 0: free=9135, capacity=10000 → 91.35% free
- Target free ratio: (9135+9800)/20000 = 94.675%
- Old: |91.35% − 94.675%| = 3.325% ≤ 5% → incorrectly marked complete
- New: 91.35% < 94.675% → correctly continues rebalancing

**Edge Cases**
- Pool already above target: `pfi > goal` → correctly marked complete
- Pool exactly at target: `pfi == goal` → correctly marked complete
- Pool just below target: continues rebalancing, correct
- `init_capacity == 0`: returns `false`, excluded from rebalance, correct
- Floating-point boundary: `f64::EPSILON` handles precision correctly

**Test Coverage**
The updated tests correctly reflect the new semantics. The new regression test `test_rebalance_goal_not_reached_for_issue_3137_initial_imbalance` directly uses the usage ratios from the issue report (8.65% / 2.00%), which is good for preventing regressions. The `complete_rebalance_pools_at_goal` test adjustment (bytes 50 → 100) ensures the pool actually reaches the 45% goal under the new logic.

**Minor Observation (non-blocking)**
In `check_if_rebalance_done`, the `pfi` variable (lines 1006-1010) is computed but only used for the log message — the actual goal check is delegated to `rebalance_goal_reached` which recomputes it internally. This is a harmless redundancy but could be cleaned up in a follow-up.

**Verdict: LGTM.** The change is minimal, targeted, and correctly fixes the early-completion bug. The one-sided goal check is the right semantic — a pool should not be considered balanced until it has actually reached the target ratio.

overtrue (approved):

PR #3145, opened 2026-05-31
feat(scanner): add scanner budgets and progress metrics

marshawcoco:
## Related Issues                                                                                              

  N/A

  ## Summary of Changes

  - Preserve regular scanner cycles on single-disk deployments when lifecycle or replication rules are active, while keeping the
  lower-churn default for plain single-disk installs.
  - Add optional scanner concurrency budgets:
    - `RUSTFS_SCANNER_MAX_CONCURRENT_SET_SCANS`
    - `RUSTFS_SCANNER_MAX_CONCURRENT_DISK_SCANS`
  - Add scanner runtime progress metrics for queued/active set scans, queued/active disk bucket scans, and scan permit wait time.
  - Expose scanner `active_paths` through rustfs-obs.
  - Fix scanner report `current_started` so an active scan cycle is not overwritten by the global init time.

  ## Verification

  - `cargo fmt --all --check`
  - `cargo test -p rustfs-common report_ --lib`

  ## Impact

  Backward compatible by default. The new concurrency budget environment variables default to `0`, which preserves the existing
  topology/disk-count based behavior unless explicitly configured.

  Single-disk deployments with active lifecycle or replication rules keep the regular scanner cadence so maintenance work is not
  delayed by the plain single-disk low-churn default.

  Scanner observability is expanded with additional runtime progress metrics.

  ## Additional Notes

  N/A

### DIFF
--- a/crates/common/src/metrics.rs
+++ b/crates/common/src/metrics.rs
@@ -600,13 +600,16 @@ impl Metrics {
     pub async fn report(&self) -> ScannerMetricsReport {
         let mut m = ScannerMetricsReport::default();
 
-        if let Some(cycle) = self.get_cycle().await {
+        let has_cycle = if let Some(cycle) = self.get_cycle().await {
             m.current_cycle = cycle.current;
             m.cycles_completed_at = cycle.cycle_completed;
             m.current_started = cycle.started;
-        }
+            true
+        } else {
+            false
+        };
 
-        if let Some(init_time) = crate::get_global_init_time().await {
+        if !has_cycle && let Some(init_time) = crate::get_global_init_time().await {
             m.current_started = init_time;
         }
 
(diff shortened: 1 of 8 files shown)


houseme:
Approve. Three clean, well-scoped changes. A few notes:

**Single-disk cadence fix** — Good call making `inspection_failed` conservative. If we can't read bucket configs, defaulting to regular cycle is the safe choice. The `needs_regular_cycle()` early-break in the bucket loop is nice — avoids unnecessary config fetches.

**Concurrency budgets** — Standard semaphore + RAII guard pattern, done right. `decrement_atomic_usize` with `fetch_update` + `saturating_sub` is correct. `scanner_concurrency_limit(0, N) => N` preserves existing behavior by default — backward compat is clean.

**`current_started` fix** — This was a real bug. `GLOBAL_INIT_TIME` stomping an active cycle's `started` time would make `last_activity_seconds` report wrong values. Test covers it directly.

**One thing to fix:** `ongoing_buckets` is set from `active_paths.len()`, which counts active scan paths, not buckets. Someone querying this metric expecting "how many buckets have active maintenance work" will get a different answer. Consider renaming to `active_scan_paths` to match the actual semantics.

**Nit:** `report_preserves_current_cycle_started_time` mutates `GLOBAL_INIT_TIME` without `#[serial]`. Low risk since it's write-then-restore, but `#[serial]` would make intent explicit if other tests ever touch that global.

houseme (approved):

PR #3152, opened 2026-06-01
feat(scanner): expose cycle progress metrics

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Expose scanner cycle progress metrics for both the currently running cycle and the last finished cycle.

This adds object, directory, and bucket-drive scan counters to the scanner runtime report and Prometheus metrics. It also derives per-second progress rates from those counters and the current or last cycle duration, so operators can see both scanner work volume and scanner throughput without changing scanner scheduling or walk behavior.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-common scan_cycle_work --lib`

## Impact

Adds scanner observability metrics only. No scanner behavior, scheduling, storage format, or configuration changes.

## Additional Notes

N/A


houseme (approved):

houseme:
**Review**

Clean observability-only PR. The `start/finish_scan_cycle_work` bracketing in `run_data_scanner_cycle` is correct — both success and error paths call finish, so the last cycle counters always get populated. `saturating_sub` for the delta is the right call.

The current cycle report computes work-in-progress by subtracting the stored start snapshot from the live lifetime counters, so operators see real-time progress during a running cycle. Last cycle stores the final snapshot at finish time. Both approaches are correct for their use case.

`scanner_work_rate_per_second` handles edge cases (zero, infinity, NaN) properly — returns 0.0 in all degenerate cases.

One thing worth noting: `current_cycle_*_per_second` is a real-time instantaneous rate (count_so_far / elapsed_so_far), while `last_cycle_*_per_second` is an average rate (total / duration). Both are useful but they mean different things. Not a bug, just something to keep in mind when building dashboards — don't compare them directly.

**CI:** S3 Implemented Tests failed, but this PR is metrics-only with no behavior changes. Likely flaky — worth a re-run to confirm.

LGTM.

PR #3159, opened 2026-06-01
feat(scanner): expand scanner observability metrics

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Expand data scanner observability with additional non-overlapping runtime diagnostics.

This PR records and exports explicit bucket-drive scan starts and failures, scanner throttle/yield configuration, scanner cycle and bitrot-cycle configuration, and per-cycle scanner work counters for yield time, lifecycle actions, heal candidates, replication checks, usage saves, and bucket-drive failures.

It intentionally leaves the existing set-scan backlog gauges and namespace-shape alert counters on their current direct metrics paths instead of re-exporting duplicate scanner collector metrics.

## Verification

- `cargo fmt --all`
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-common scanner --lib`
- `cargo test -p rustfs-scanner scanner --lib`

## Impact

Adds scanner diagnostics and Prometheus metrics only. No scanner scheduling, storage format, API, or configuration behavior changes are expected.

## Additional Notes

This is scoped as a follow-up to the merged scanner observability PRs and avoids duplicating metrics that are already emitted elsewhere.


houseme:
`#[allow(clippy::too_many_arguments)]`

houseme (commented):
Overall this is a solid observability expansion — good metric coverage and thorough test follow-through. Here are the key findings:

### 1. `done_drive()` missing on error path (should fix)

In `scanner_io.rs`, `done_drive()` (i.e. `Metrics::time(Metric::ScanBucketDrive)`) is only called in the `Ok` branch. The `Err` branch skips it entirely. Consequences:
- Failed scans contribute zero latency observations to `ScanBucketDrive`
- `bucket_scans_finished` silently excludes failures, making its semantics ambiguous ("completed successfully" vs "ended")
- The new `BucketDriveFailureGuard` records the failure *count*, but not the latency

**Fix**: call `done_drive()` in both branches so latency observations cover all outcomes.

### 2. ILM double-counting

In `scanner_folder.rs`, every ILM action triggers two independent counters:
- `done_ilm(1)()` → increments `operations[Metric::Ilm]` (the lifetime ILM counter)
- `global_metrics().record_scanner_ilm_action(1)` → increments `scanner_ilm_actions` (a separate `AtomicU64`)

Both represent "total ILM actions" but live in different storage. In `report()`, `ilm_action_total` comes from `lifetime(Metric::Ilm)` while `current_cycle_ilm_actions` comes from `scanner_ilm_actions`. They should always be equal but use independent atomics, so they can diverge.

**Fix**: have `record_scanner_ilm_action` derive its value from `lifetime(Metric::Ilm)` instead of maintaining a separate counter, or drop the separate counter and compute the cycle delta from the `operations` array.

### 3. Yield semantic mixing

`record_scanner_yield` is called in two distinct contexts:
- `scanner_folder.rs` — after `tokio::task::yield_now()` (cooperative task yield)
- `sleeper.rs` — after throttle sleep (intentional rate-limiting)

Both increment the same `Metric::Yield` counter and `scanner_yield_duration_millis`. Operators cannot distinguish "scanner yielded due to CPU contention" from "scanner slept per configured throttle policy".

**Fix**: use separate `Metric` variants (e.g. `Metric::ThrottleSleep` vs `Metric::Yield`) so the two behaviors are independently observable.

### 4. `record_scan_cycle_work` parameter explosion

10 parameters with `#[allow(clippy::too_many_arguments)]`. Since `ScanCycleWorkSnapshot` already exists, the method should accept a single snapshot struct instead of 10 unpacked `u64` values. This eliminates the clippy suppression and makes the call site cleaner.

### 5. Minor: bool → bool → u64 type chain

`throttle_idle_mode_enabled` / `bitrot_cycle_enabled` are `AtomicBool` in `Metrics`, `bool` in `ScannerMetricsReport`, and `u64` in `ScannerStats`. The conversion chain (`u64::from(bool)`) works but is non-obvious. Consider keeping `bool` throughout or using a consistent `u64` representation at the storage layer.

---

All items are non-blocking. The PR is in good shape overall — these are improvements for follow-up.

PR #3163, opened 2026-06-02
fix(ci): stabilize Rust setup action

overtrue:
Related issue(s): CI failure on main and PR checks. Background: GitHub Actions setup failed before Rust checks because self-hosted runners lacked unzip for setup-protoc, and cargo-nextest installation used an install-action ref that failed on the current runner/tool release. Solution: install unzip in the shared Rust setup action and use the supported nextest install-action shorthand. Validation: CI rerun is pending on this PR.


.github/actions/setup/action.yml
```diff
@@ -83,7 +84,7 @@ runs:
       uses: taiki-e/install-action@cargo-zigbuild
 
     - name: Install cargo-nextest
-      uses: taiki-e/install-action@cargo-nextest
+      uses: taiki-e/install-action@nextest
```
houseme: cargo-nextest

houseme (changes requested):
crate name error

houseme (approved):

PR #3166, opened 2026-06-02
feat(scanner): add cycle budget observability

marshawcoco:
## Related Issues

N/A

## Summary of Changes

- Add an optional scanner cycle runtime budget via `RUSTFS_SCANNER_CYCLE_MAX_DURATION_SECS`, with `0` preserving the existing unbounded cycle behavior.
- Record partial scanner cycles when the runtime budget stops a cycle before full completion.
- Expand scanner metrics with cycle budget/config fields, partial cycle counts, bucket-drive scan start/failure signals, and per-cycle work snapshots.
- Split scanner cooperative yield metrics from throttle sleep metrics so operators can distinguish scheduler yields from configured throttling.
- Keep scanner cycle ILM action metrics scanner-only while preserving the global `Metric::Ilm` lifecycle counter.

## Verification

- `cargo fmt --all`
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-scanner scanner_folder --lib`
- `cargo test -p rustfs-scanner budget --lib`
- `cargo test -p rustfs-scanner scanner::tests::test_scanner_cycle_budget_cancels_after_duration --lib`
- `cargo test -p rustfs-scanner data_usage_define::tests::test_retry_save_op_times_out_and_returns_error_after_retries --lib`

## Impact

- Adds a new optional scanner environment variable: `RUSTFS_SCANNER_CYCLE_MAX_DURATION_SECS`.
- Default behavior is unchanged because the default runtime budget is disabled.
- Adds and clarifies scanner Prometheus metrics for cycle budget, partial cycles, throttle sleeps, cooperative yields, and scanner-only lifecycle action counts.

## Additional Notes

Throttle sleep metrics now represent intentional scanner rate limiting, while yield metrics represent cooperative runtime yields.


houseme:
One blocking issue here: `current_cycle_ilm_actions` / `last_cycle_ilm_actions` are no longer scanner-only.

This patch switches scanner cycle accounting to `Metric::Ilm`, but `Metrics::time_ilm()` is also used by non-scanner lifecycle paths like `transition_object()` and `apply_expiry_on_*()`. As a result, background lifecycle work outside the scanner can now bleed into the scanner cycle metrics, which makes the new observability inaccurate.

Previously this was isolated by the dedicated scanner ILM counter in `scanner_folder.rs`. I think we should keep that separation.

houseme (approved):

PR #3171, opened 2026-06-02
fix(admin): prefer explicit TLS for site replication endpoint

### LINKED ISSUE #2723: [Issue/Discussion] Challenges with Site-to-Site Replication: UID mismatches, TLS SAN validation, and Endpoint Protocol
**Description**
I have successfully deployed 2 RustFS in a single-node configuration using Docker Compose with bucket HTTPS replication enabled. While the system is now operational and healthy, I encountered several hurdles that could be improved for better developer experience and security.
I have 2 VMs with rustfs container on each VM.

**1. UID Mismatch and Permission Issues**
By default, the RustFS container runs with UID 10001. In my environment (and many others), the host storage and volumes are aligned with a different UID (e.g., 995).

- The Issue: This caused a Permission Denied (os error 13) at startup, specifically at rustfs/src/main.rs:239, as the binary could not initialize the /logs directory or read the TLS certificates mounted from the host.

- Solution implemented: I had to explicitly override the container user in my docker-compose.yml using user: "995:992" to match my host's storage permissions.

- Feedback: It would be helpful if the documentation highlighted the need for this alignment when using host-mounted volumes.

**2. TLS Healthcheck and SAN Validation**
When enabling TLS, the default healthcheck using 127.0.0.1 fails if the IP is not explicitly included in the Certificate's Subject Alternative Name (SAN).

- Error: curl: (60) SSL: no alternative certificate subject name matches target ipv4 address '127.0.0.1'.

- Workaround: I resolved this by modifying the Docker healthcheck command to bypass strict name validation:

```
healthcheck:
  test: ["CMD", "sh", "-c", "curl -f -k https://127.0.0.1:9000/health && curl -f -k https://127.0.0.1:9001/rustfs/console/health"
```

- Question: Is there a recommended way to perform secure healthchecks internally without adding 127.0.0.1 to production certificates?

**3. Site-to-Site Replication Endpoint & Protocol**
My current replication endpoint is configured using http.
http://<FQDN>:9001

- The Question: Since I have a full TLS stack (Certs and CA) mounted and working for the S3 API and Console, is it possible to switch the replication endpoint to HTTPS?

- I can't enable site replication because my environment uses HTTPS, but the endpoints use HTTP.
  When I try, i have 3 sites : 
Image

  When I try to enable replication on the HTTPS site, I get: 

Image

When I try to enable replication on the HTTP site, I get: 

Image

- Security Concern: How does RustFS handle the remote CA validation for outgoing replication requests? Can we specify a trusted CA for inter-site communication to ensure data is encrypted and verified in transit between nodes?

- Can we change the endpoint to use HTTPS?

- Do you have an up-to-date document?

**Environment Summary**
- Deployment: Docker Compose
- Topology: Single Node
- Image: rustfs:alpha-0.98
- Auth: TLS enabled with custom CA
- bucket replication enable : HTTPS

cxymds:
## Related Issues
Fixes #2723.

## Summary of Changes
- Prefer explicit TLS configuration when inferring the local site replication endpoint.
- Avoid persisting the local site as `http://...` when the node is actually serving HTTPS.
- Add a regression test covering the explicit TLS path over HTTP runtime endpoint case.

## Verification
- `cargo test -p rustfs test_runtime_tls_enabled_prefers_explicit_tls_over_http_runtime_endpoint -- --nocapture`
- `cargo test -p rustfs test_request_endpoint_ -- --nocapture`
- `cargo fmt --all --check`
- Manual 3-node verification: created site replication across three HTTPS nodes and confirmed a bucket created on one site appeared on all three sites.

## Impact
- Fixes three-site replication setups where the local site endpoint could be saved with the wrong protocol.
- No intended API contract change.

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -536,8 +536,12 @@ async fn site_replication_peer_client() -> S3Result<reqwest::Client> {
     built
 }
 
-fn runtime_tls_enabled() -> bool {
-    if let Some(tls_enabled) = get_global_endpoints_opt().and_then(|endpoints| {
+fn runtime_tls_enabled_with(endpoints: Option<&rustfs_ecstore::endpoints::EndpointServerPools>) -> bool {
+    if !rustfs_utils::get_env_str(ENV_RUSTFS_TLS_PATH, DEFAULT_RUSTFS_TLS_PATH).is_empty() {
+        return true;
+    }
+
+    if let Some(tls_enabled) = endpoints.and_then(|endpoints| {
         endpoints
             .as_ref()
             .iter()
@@ -548,7 +552,12 @@ fn runtime_tls_enabled() -> bool {
         return tls_enabled;
     }
 
-    !rustfs_utils::get_env_str(ENV_RUSTFS_TLS_PATH, DEFAULT_RUSTFS_TLS_PATH).is_empty()
+    false
+}
+
+fn runtime_tls_enabled() -> bool {
+    let endpoints = get_global_endpoints_opt();
+    runtime_tls_enabled_with(endpoints.as_ref())
 }
 
 fn query_pairs(uri: &Uri) -> HashMap<String, String> {
@@ -3030,6 +3039,8 @@ mod tests {
     use super::*;
     use http::{HeaderMap, HeaderValue, Uri};
     use rustfs_common::{get_global_outbound_tls_generation, set_global_outbound_tls_generation};
+    use rustfs_ecstore::disk::endpoint::Endpoint;
+    use rustfs_ecstore::endpoints::{EndpointServerPools, Endpoints, PoolEndpoints};
     use serial_test::serial;
     use temp_env::with_var;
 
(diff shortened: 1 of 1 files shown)


houseme (approved):

houseme (commented):
Good fix, directly addresses the root cause of the scheme-drift issue in #2723.

**What changed**: `runtime_tls_enabled()` now checks `RUSTFS_TLS_PATH` *before* falling back to the runtime endpoint URL scheme. Previously, if the endpoint was `http://` but the operator had explicitly configured TLS certs, the function would return `false` and the local site would get persisted as `http://...` — which is exactly the bug the reporter hit.

**Code looks clean**:
- Extracting `runtime_tls_enabled_with(endpoints)` is a nice touch — testable without global state.
- The fallback at the end is now just `false` instead of re-checking the env var, which is correct since it was already checked above.
- `is_empty()` guard on `RUSTFS_TLS_PATH` handles the empty-string case properly.

**Test**: Covers the main regression scenario (explicit TLS path + http:// endpoint → returns true). That's the critical one.

**A few minor observations**:

1. Only one test case added. Could also cover the "no TLS path, http endpoint → false" and "no TLS path, https endpoint → true" paths to verify the priority ordering doesn't break the existing behavior. Not blocking, but would be good for confidence.

2. The call site at line 687 uses `runtime_tls_enabled()` as a scheme fallback when neither `X-Forwarded-Proto` nor the URI scheme is available. With this change, setting `RUSTFS_TLS_PATH` will force `https` in that fallback even when the node is actually behind a reverse proxy terminating TLS and forwarding HTTP. That's probably the *correct* behavior (if you set TLS path, you want HTTPS), but worth being aware of for operators who might have a mixed setup.

3. Issue #2723 also mentions the peer admin `reqwest` client not loading custom CAs for site-to-site RPC — that's a separate problem from the scheme inference and won't be fixed by this PR. The reporter confirmed this is still a blocker after the scheme drift fixes. Just noting so it doesn't get lost.

LGTM, clean and focused fix.

PR #3176, opened 2026-06-02
fix(s3select): validate scan_range protocol and parquet overlap

GatewayJ:
## Summary
- Enforce `scan_range` protocol validity for SelectObjectContent inputs.
- Allow scan-range for CSV, JSON LINES, and Parquet, and keep JSON DOCUMENT rejected.
- Tighten Parquet scan-range behavior to select row groups by overlapping byte spans.
- Reduce `SelectScanRange` external field coupling via constructor/accessors.
- Simplify `TableHandle` abstraction to a single `Arc<dyn TableProvider>`.
- Add regression tests for protocol validation and scan-range behavior.

## Related Issues
- Local fix set for scan-range protocol and boundary correctness.

## Verification
- N/A

## Notes
- No functional changes outside the scan-range/Parquet selection path.


overtrue (changes requested):
Requesting changes: crates/s3select-query/src/dispatcher/parquet_table.rs:151-155 selects Parquet row groups by overlapping byte span, but S3 Select ScanRange processes row groups whose start is within the requested range; a range ending inside one row group can incorrectly include the next row group. Please select by row-group start offset and add a boundary test.

GatewayJ:
Done. Fixed in commit: `parquet_access_plan` now selects row groups by row-group start offset (or fallback to conservative scan if unavailable), not by overlap span. This aligns with S3 Select ScanRange semantics and avoids including the next row group when range ends inside a row group.

Added regressions:
- `access_plan_selects_row_group_by_row_group_start`
- `access_plan_uses_row_group_start_not_column_span_overlap`

Validation run: `cargo test -p rustfs-s3select-query --lib` (2/2 new tests passed). `cargo fmt --all --check` passed.

GatewayJ:
I fixed the two follow-up items:

1) `row_group_start_offset` now only uses `row_group.file_offset` and no longer falls back to column offsets. If missing, access plan keeps conservative full-scan behavior in `access_plan_for_scan_range`.
2) Removed redundant `column_byte_start`/`column_start_offset` path; `ColumnChunkMetaData` import is also removed as no longer needed.

Verification: `cargo test -p rustfs-s3select-query --lib` and `cargo fmt --all --check` both pass.

overtrue (approved):

houseme (approved):

PR #3182, opened 2026-06-03
fix(tier): recover

mbartsch:
## Related Issues

N/A

## Summary of Changes

Fixes three bugs that caused errors when reading or restoring objects tiered to a warm S3-compatible backend.

**Bug 1 — nil UUID sent as `?versionId` (`filemeta/version.rs`)**

When `transitioned-versionID` metadata is absent or contains zero bytes (objects tiered to a non-versioned bucket), the old code called `.unwrap_or_default()` on the UUID parse, silently producing `Uuid::nil()` (`00000000-0000-…`). That nil UUID was then forwarded as `?versionId=00000000-…` in the tier GET request, causing a `NoSuchVersion` error on every read.

Fixed by replacing `unwrap_or_default()` with `.and_then(|v| Uuid::from_slice(v.as_slice()).ok()).filter(|u| !u.is_nil())`, which returns `None` for absent, wrong-length, or nil-UUID bytes. Applied to both `MetaObject::to_fileinfo` and `MetaDeleteMarker::to_fileinfo`.

**Bug 2 — `warm_backend_s3sdk.rs` ignored `rv` and range opts**

The AWS SDK-based warm backend was unconditionally omitting `version_id` and `range` from GET requests, and `version_id` from DELETE requests. Now both are correctly forwarded when non-empty. Error messages are also propagated instead of being swallowed.

**Bug 4 — `copy_object` returns 501 for tiered objects (`set_disk.rs`, `object_usecase.rs`)**

`mc cp --storage-class STANDARD obj obj` on a tiered object always returned 501 (`NotImplemented`) because:
1. `execute_copy_object` set `metadata_only = true` only when source and destination are the same key *and* the object is not tiered — but for tiered objects the guard was not applied, leaving `metadata_only = false`.
2. `set_disk::copy_object` unconditionally rejected `metadata_only = false` with `NotImplemented`.

Fixed across both layers:
- `object_usecase.rs`: extracts and validates `storage_class` from the request; allows self-copy when a storage class change is requested; skips `metadata_only = true` for tiered objects so the full data path is taken; applies the requested storage class to copy opts.
- `set_disk.rs`: for a same-bucket/same-key copy with `metadata_only = false` and a populated `put_object_reader`, calls `put_object` directly — writing the tier data back locally and effectively de-tiering the object.

**How `mc cp --storage-class STANDARD` flows after the fix:**
1. mc sends `PUT /bucket/key` with `x-amz-copy-source`, `x-amz-metadata-directive: REPLACE`, `x-amz-storage-class: STANDARD`
2. `execute_copy_object` → `get_object_reader` fetches data from tier backend → stored in `src_info.put_object_reader`
3. `set_disk::copy_object` detects same-key self-copy with a reader and calls `put_object`
4. New xl.meta written locally with STANDARD storage class, no tier metadata → object de-tiered

**Observability improvements**

Added structured `debug!` / `error!` log lines in `get_transitioned_object_reader` that record tier name, tier object key, version ID, and offset/length before and after the request. Enable with `RUST_LOG=rustfs_ecstore::bucket::lifecycle=debug`.

**Tooling**

Extended `dump_fileinfo` example to print `transition_status`, `transition_tier`, `transitioned_obj`, and `transition_ver_id` from `xl.meta`, making it easier to diagnose tiered objects offline.

---

This has been AI-assisted. I found the bugs and directed the fixes; I verified the logic at each step and confirmed the changes are working in my homelab — objects are now correctly retrieved from a cold storage archive (Cubbit DS3).

## Verification

```bash
# Type-check affected crates
cargo check -p rustfs-ecstore
cargo check -p rustfs-filemeta

# Run crate-level tests
cargo test -p rustfs-ecstore
cargo test -p rustfs-filemeta

# Full pre-PR gate
make pre-commit

# Manual: build and run dump_fileinfo against a tiered object's xl.meta
cargo build -p rustfs-filemeta --example dump_fileinfo
./target/debug/examples/dump_fileinfo /srv/rustfs/data/disk0/<bucket>/<object>/xl.meta
# Confirmed: transition_ver_id: <none> for non-versioned tier, no NoSuchVersion error on GET

# Storage class restore (de-tier):
# mc cp --storage-class STANDARD alias/bucket/tiered-object alias/bucket/tiered-object
# Confirmed: 200 OK, object written back locally with STANDARD class
```

## Impact

- **Behavior**: Objects tiered to non-versioned S3-compatible buckets are now correctly retrieved without sending a nil UUID as `?versionId`. Objects can be de-tiered by copying with `--storage-class STANDARD`. Objects tiered to versioned buckets are unaffected.
- **Compatibility**: No change to xl.meta on-disk format or metadata keys. Existing xl.meta files with empty `transitioned-versionID` values are handled correctly by the new read path.
- **API / deployment / configuration**: No changes required.


mbartsch:

overtrue (changes requested):
Please fix the SDK backend range construction before merge: crates/ecstore/src/tier/warm_backend_s3sdk.rs:159 computes end with unchecked i64 addition, while the existing HTTP backend routes the range through GetObjectOptions::set_range; a large offset/length can overflow or emit an invalid Range header.

mbartsch:
I think I found another related bug, when recovering the Storage Class is not updated, so it still shows the old storage class.
So I might be adding more commits to this PR , as it is related to the same bug, please hold the review until I check this.

Thanks!

mbartsch:
Fixed second bug (copy same object with StorageClass after recover failed. now work as expected)

.config/make/build-docker.mak
```diff
@@ -6,11 +6,11 @@
 .PHONY: build-docker
 build-docker: SOURCE_BUILD_IMAGE_NAME = rustfs-$(BUILD_OS):v1
 build-docker: SOURCE_BUILD_CONTAINER_NAME = rustfs-$(BUILD_OS)-build
-build-docker: BUILD_CMD = /root/.cargo/bin/cargo build --release --bin rustfs --target-dir /root/s3-rustfs/target/$(BUILD_OS)
+build-docker: BUILD_CMD = cargo build --release --bin rustfs --target-dir /root/s3-rustfs/target/$(BUILD_OS)
```
houseme: Before adding it to `PATH`, you’ll encounter an error stating that `Cargo` cannot be found.

mbartsch: Reverted!

houseme (approved):
Thanks for the follow-up here. The overflow issue in the SDK range handling is addressed now, and the added regression coverage around transitioned version IDs makes the tier read path much safer.

The restore/de-tier flow also reads clearly after the latest updates. I don't see any remaining blockers from my side.

overtrue (approved):
Looks good.

PR #3185, opened 2026-06-03
feat(scanner): add scanner budget progress controls

marshawcoco:
## Related Issues

N/A

## Summary of Changes

- Add per-cycle scanner work budgets for objects and directories.
- Preserve partial data-usage cache progress when a scanner cycle stops after reaching a budget.
- Record scanner partial-cycle stop reasons for runtime, object, and directory budgets.
- Expose active scanner path age so operators can identify stalled scan paths.
- Expose scanner set and disk-bucket queue, active, and concurrency state metrics.

## Verification

- `cargo test -p rustfs-scanner scanner_io::tests:: --lib`
- `cargo fmt --all --check`
- `git diff --check`

## Impact

- Adds optional scanner environment variables:
  - `RUSTFS_SCANNER_CYCLE_MAX_OBJECTS`
  - `RUSTFS_SCANNER_CYCLE_MAX_DIRECTORIES`
- Default behavior is unchanged because both new budgets default to `0`, which means unlimited.
- Adds scanner metrics for budget stop reasons, active path age, and scan queue/progress state.

## Additional Notes

This builds on the scanner cycle budget observability work and keeps this PR focused on progress accounting and budget visibility. It also includes the scan-mode, bitrot-cycle, and scanner work-state signals needed to diagnose budgeted partial cycles without mixing in resumable checkpoint logic.

Scheduler backoff, incremental checkpoints, and source-aware lifecycle/replication/heal budgeting are intentionally left for follow-up PRs.


houseme (approved):

houseme (changes requested):
Nice direction overall, but the partial-progress path still looks incomplete.

When a budgeted bucket scan returns `Partial`, the bucket cache is saved locally, but that partial result is not merged back into the set-level cache or published upstream. In practice that can still leave operators looking at stale usage after a partial cycle.

Please wire the partial result into the set aggregation/publish path before merge.

marshawcoco:
Thanks, I updated the partial path so a `ScannerDiskScanOutcome::Partial` no longer only saves the bucket-local cache. After saving the partial bucket cache, it now also sends the flattened bucket root entry through the same set-level aggregation channel used by completed scans. That lets the set cache merge the partial progress and publish an updated usage view upstream during a budgeted partial cycle.
I also factored the bucket-root flattening into a shared helper so the complete and partial paths use the same aggregation shape, and added a regression test for that flattening behavior.

houseme (changes requested):
Thanks for the follow-up. The direction is right, but this still does not quite close the gap.

In the `Partial` path, the code now tries to send the flattened bucket result back into the set-level aggregation flow, but that send happens after checking `ctx_clone.is_cancelled()`. Since hitting the budget cancels the token, this branch will usually break before the partial result is sent.

So the partial bucket cache is still saved locally, but the set-level cache / published usage can still stay stale after a budgeted partial cycle.

houseme (approved):

PR #3201, opened 2026-06-03
feat(table-catalog): add catalog store entry models

marshawcoco:
## Related Issues
  N/A

  ## Summary of Changes
  Adds the initial S3 Tables catalog store boundary for future Iceberg REST Catalog work.

  - Defines stable catalog entry models for table buckets, namespaces, tables, and commit logs.
  - Adds catalog entry state and commit log status enums with stable JSON serialization.
  - Adds a `TableCatalogStore` trait covering the core read/write operations needed by later catalog handlers.
  - Adds unit tests for JSON shape stability, unknown-field rejection, and trait object usability.

  ## Verification
  - `cargo fmt --all --check`

  ## Impact
  No user-facing API changes. This is an internal catalog data model and trait boundary for upcoming S3 Tables catalog implementation.

  ## Additional Notes
  This PR does not add Iceberg REST HTTP routes or change existing S3 object behavior.

### DIFF
--- a/rustfs/src/table_catalog.rs
+++ b/rustfs/src/table_catalog.rs
@@ -20,7 +20,7 @@
 
 #![allow(dead_code)]
 
-use std::fmt;
+use std::{collections::BTreeMap, fmt};
 
 use rustfs_ecstore::bucket::{
     metadata::{BUCKET_TABLE_CONFIG, BUCKET_TABLE_RESERVED_PREFIX},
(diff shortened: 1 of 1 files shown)


marshawcoco:
@codex review

houseme (changes requested):
Requesting changes because the store boundary does not yet capture the core commit semantics this feature will need.

1. The TableCatalogStore trait lacks an atomic commit/update operation. Right now load_table(), append_commit_log(), and table mutation methods are all separate. That leaves a race window for concurrent writers and makes it difficult for a store implementation to guarantee optimistic concurrency correctly.

2. CommitLogEntry models idempotency_key, but the trait does not provide a way to query by it. That likely pushes retry deduplication into handlers or forces another trait change very soon.

The entry structs themselves look fine, but the current trait feels more like provisional CRUD than a stable catalog commit boundary. I would suggest adding an atomic commit API that binds version-token validation, table update, and commit-log persistence together, and also exposing idempotency-key lookup before merging this as the catalog store boundary.

houseme (approved):
Reviewed the latest update. The missing atomic commit boundary and idempotency-key lookup are now covered by `commit_table()` and `get_commit_by_idempotency_key()`, and the trait shape now looks aligned with the intended catalog commit flow. Approving.

PR #3206, opened 2026-06-04
feat(table-catalog): add object-backed catalog store

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Adds the object-backed implementation for the S3 Tables catalog store introduced by the catalog entry model work.

  - Adds `ObjectTableCatalogStore` and a `TableCatalogObjectBackend` abstraction.
  - Adds an `ECStore`-backed catalog object adapter.
  - Stores table bucket, namespace, table, commit log, and idempotency index records behind the `TableCatalogStore` interface.
  - Implements the MVP table commit path with expected token/current metadata validation, metadata object existence checks, conditional table entry updates,
  commit log writes, and idempotency index writes.
  - Hashes untrusted table, commit, and idempotency identifiers before using them in internal object paths.
  - Updates commit log lookup methods to include the table bucket so object-backed storage can resolve the correct catalog backing.
  - Adds focused unit coverage for object path safety, successful commits, stale token conflicts, and catalog store behavior.

  ## Verification

  - `cargo fmt --all --check`

  ## Impact

  This is internal S3 Tables catalog plumbing only. It does not add public REST routes, lifecycle cleanup, maintenance jobs, or user-facing S3 API behavior changes.

  ## Additional Notes

  This store remains an MVP object-backed implementation behind the catalog abstraction. A future strong metadata KV implementation should be able to replace the backing without changing REST handlers.

### DIFF
--- a/rustfs/src/table_catalog.rs
+++ b/rustfs/src/table_catalog.rs
@@ -20,13 +20,26 @@
 
 #![allow(dead_code)]
 
-use std::{collections::BTreeMap, fmt};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fmt,
+    sync::Arc,
+};
 
+use http::HeaderMap;
 use rustfs_ecstore::bucket::{
     metadata::{BUCKET_TABLE_CONFIG, BUCKET_TABLE_RESERVED_PREFIX},
     metadata_sys,
 };
-use serde::{Deserialize, Serialize};
+use rustfs_ecstore::error::StorageError;
+use rustfs_ecstore::{
+    set_disk::get_lock_acquire_timeout,
+    store_api::{HTTPPreconditions, ObjectOptions, PutObjReader, StorageAPI},
+};
+use serde::{Deserialize, Serialize, de::DeserializeOwned};
+use sha2::{Digest, Sha256};
+use tokio::io::AsyncReadExt;
+use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


houseme (changes requested):
Requesting changes because the current `commit_table()` write order can leave the catalog in a partially committed state.

Right now the table entry CAS update is written first, and only after that do we write the commit log object and the idempotency index object. If either later write fails, the method returns an error even though the table has already advanced to the new metadata location and version token.

That means a caller can observe a failed commit while the table state is already committed, and commit history / idempotency lookup can be left inconsistent with the actual table state.

Please make the commit sequence failure-safe. At minimum, the externally visible table update should not happen before the commit can be durably recovered. A staged commit record or another recoverable ordering would address this.

marshawcoco:
Fixed in the updated branch.

  `commit_table()` now writes a staged commit record before the table entry CAS update. If an idempotency key is present, the idempotency index is staged before the table pointer can advance as well.

  After the table CAS succeeds, commit log/idempotency finalization is best-effort because the staged commit record is already durable recovery state. A retry of the same commit can recover from the staged record when the table has already advanced and finalize it as committed.

  I also added regression coverage for:
  - idempotency staging failure leaving the table pointer unchanged
  - post-CAS finalization failure being recoverable by retry

houseme (approved):

PR #3207, opened 2026-06-04
feat(scanner): add partial scan resume hints

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Adds a lightweight resume hint to scanner data usage cache metadata so a later scanner cycle can continue closer to the last completed folder after a partial scan.

  This PR:
  - adds optional `scan_resume_after` metadata to `DataUsageCacheInfo`
  - keeps old JSON and MessagePack cache payloads backward compatible through serde defaults
  - records the last completed folder into partial cache state
  - reorders child folder scans around the resume hint on the next cycle
  - clears the resume hint after a successful full scan
  - adds focused tests for old-cache deserialization, folder ordering, partial resume, and successful hint cleanup

  ## Verification

  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `cargo test -p rustfs-scanner resume`
  - `cargo check -p rustfs-scanner`

  ## Impact

  Improves scanner progress preservation when a cycle is interrupted by budget cancellation. Existing data usage cache files remain readable because the new metadata field defaults to absent.
  ## Additional Notes

  N/A

### DIFF
--- a/crates/scanner/src/data_usage_define.rs
+++ b/crates/scanner/src/data_usage_define.rs
@@ -227,6 +227,8 @@ pub struct DataUsageCacheInfo {
     pub replication: Option<Arc<ReplicationConfig>>,
     #[serde(default)]
     pub failed_objects: HashMap<String, u64>,
+    #[serde(default)]
+    pub scan_resume_after: Option<String>,
 }
 
 /// Data usage cache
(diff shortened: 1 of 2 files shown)


houseme (changes requested):
Requesting changes because the current resume-hint ordering does not actually prioritize the overall next scan position when both new and existing child folders are present.

Right now `scan_resume_after` is applied separately to `new_folders` and `existing_folders`, but the scanner still processes all `new_folders` before any `existing_folders`. If the resume hint points into the existing-child set, a tight directory budget can still be consumed entirely by newly discovered folders before the scanner ever reaches the intended resume area.

That means the new hint can fail to move the scan materially closer to the last completed folder, and in a steady stream of newly created children it can starve the old unfinished subtree for multiple cycles.

Please apply the resume hint to the combined child scan order rather than only within each subgroup, and add a test that mixes new and existing children with a small directory budget.

houseme (approved):

PR #3216, opened 2026-06-04
fix(signer): address post-merge review comments

overtrue:
Address review comments from PR #3150 (merged).

- Restore get_host_addr as best-effort wrapper (String return type preserved)
- Replace bare expect("err") with descriptive messages in unsigned trailer helper
- Simplify aws-chunked header construction

20 signer tests pass.

### DIFF
--- a/crates/signer/src/request_signature_streaming_unsigned_trailer.rs
+++ b/crates/signer/src/request_signature_streaming_unsigned_trailer.rs
@@ -25,7 +25,7 @@ pub fn streaming_unsigned_v4(
 ) -> request::Request<Body> {
     let headers = req.headers_mut();
 
-    let chunked_value = HeaderValue::from_str("aws-chunked").expect("aws-chunked is a valid header value");
+    let chunked_value = HeaderValue::from_static("aws-chunked");
     headers.insert(http::header::TRANSFER_ENCODING, chunked_value);
     if !session_token.is_empty() {
         headers.insert(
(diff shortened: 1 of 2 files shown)


houseme:
## Review

Thanks for following up on the #3150 review. The `expect()` message improvements and `aws-chunked` simplification look good. One blocker and two minor items:

### Blocker: `get_host_addr` regression

`utils.rs:48` — the PR description says "Restore `get_host_addr` as best-effort wrapper", but `unwrap_or_default()` drops the `host` header fallback for relative URIs entirely. A request with `URI=/object` + `Host: bucket.example.com` previously returned `"bucket.example.com"` and now returns `""`.

Since `get_host_addr` is a `pub` function, either:
- Keep the `MissingUriHost` → host-header fallback (matches legacy behavior), or
- Remove/deprecate the function if it's truly unused — don't silently change its contract.

### Minor

1. **`HeaderValue::from_static`** — `"aws-chunked"` is static ASCII; use `HeaderValue::from_static("aws-chunked")` instead of `from_str(...).expect(...)` (zero-copy, infallible).

2. **`streaming_unsigned_v4` panics** — the improved `expect()` messages are great, but the function still panics on bad session tokens or time format failures. Consider returning `Result` in a follow-up.

The first item needs a fix before merge. The rest can be follow-ups.

overtrue:
Fixed the relative-URI Host fallback regression and switched the static aws-chunked header to HeaderValue::from_static in f73594d13.

overtrue:
Fixed in f73594d; verified with cargo fmt --all --check, cargo test -p rustfs-signer, cargo check -p rustfs-signer, and cargo clippy -p rustfs-signer --all-targets --all-features -- -D warnings.

overtrue:
Addressed the new Copilot follow-ups in ca368ca8: legacy host fallback now uses the URI host when the Host header is invalid, and streaming unsigned header insertion is non-panicking with focused tests.

PR #3243, opened 2026-06-06
refactor(embedded): remove single-instance restriction

overtrue:
## Related Issues
Relates to https://github.com/rustfs/backlog/issues/653 (item 8, Phase 5)

## Summary of Changes
Phase 5 of global singleton consolidation. Remove the SERVER_STARTED AtomicBool guard that prevented multiple embedded server instances per process.

### Changes:
- Remove SERVER_STARTED AtomicBool static
- Remove AlreadyStarted error variant from ServerError
- Remove set_global_init_guard closure and its call
- Update module doc comment to reflect multi-instance support

### Multi-instance behavior:
- Multiple RustFSServer instances may now coexist in one process
- Each instance should use a unique port and volume directory
- Global configuration state (region, endpoints, erasure type) is shared — last initialized values take effect

## Verification
```bash
cargo check -p rustfs
cargo clippy -p rustfs -- -D warnings
cargo fmt --all --check
```
- Full workspace compiles
- Clippy clean
- Formatting clean

## Impact
Enables running multiple embedded servers in the same process (e.g., for parallel integration tests). No behavioral change for single-instance usage.

## Additional Notes
The global singletons (OnceLock, LazyLock) are still shared across instances. This is acceptable for the embedded use case where each server uses unique ports/volumes. Full isolation would require migrating all globals to per-instance state (future work).

### DIFF
--- a/rustfs/src/embedded.rs
+++ b/rustfs/src/embedded.rs
@@ -40,12 +40,11 @@
 //! }
 //! ```
 //!
-//! # Multi-instance Support
+//! # Limitations
 //!
-//! Multiple `RustFSServer` instances may exist per process. Each instance
-//! should use a unique port and volume directory. Global configuration state
-//! (region, endpoints, erasure type) is shared across instances — the last
-//! initialized values take effect.
+//! Only **one `RustFSServer`** may exist per process because the underlying
+//! storage engine uses process-global singletons (`OnceLock`). Attempting to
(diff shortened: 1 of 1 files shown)


overtrue:
Closing — review correctly identified that global OnceLock singletons still prevent true multi-instance support. The guard should remain until globals are fully isolated. Reverted changes.

PR #3244, opened 2026-06-06
fix: clean old data dirs on object overwrite

### LINKED ISSUE #3231: Resource leak (disk) on object overwrite on unversioned buckets
[I wrote this issue by hand, and the reproduction was assisted by AI.]

When overwriting objects on unversioned buckets, a data directory is leaked with each overwrite. For some applications, such as index files or s3fs mounts, this can quickly lead to resource exhaustion and performance issues.

I discovered this bug by writing to an append-only log file that was mounted over s3fs. Each write to the file triggered a new PUT to rustfs, leaving the orphaned data dir described above. Eventually, ListObjects started to time out for us, as we unknowingly had 68GB of storage to scan, despite well under 1GB of actual objects.

I think reproducing this requires objects > 128 KiB because below this size, they are inlined in the metadata.

**To Reproduce**
I had Claude create a repro repository at https://github.com/crw2998/rustfs-overwrite-leak-repro/. You should be able to clone this repo and run `./repro.sh` or `uv run repro.py`. The scripts pull the latest rustfs (right now it's 1.0.0-beta.7), start the server, and then make a series of overwriting 508KiB PUTs. After each 100 PUTs it reports the performance of ListObjects. You will see that ListObjects gets slower and slower, even though the logical amount of data that we're storing is not changing. At the end, the script will report the amount of disk space occupied by the bucket.

**Expected behavior**
I would expected these orphaned data directories to be garbage collected after an overwriting PUT.

**Environment**
RustFS 1.0.0-beta.7
Debian Trixie x86_64
also reproduces on my M1 Mac.

I think this issue is a possible root cause of #3191 and maybe #3175

GatewayJ:
## Related Issues
Fixes #3231

## Summary of Changes

When overwriting objects on unversioned buckets, the old data directory was not being cleaned up, causing resource leaks. Each overwrite left behind an orphaned data directory that accumulated over time, leading to storage exhaustion and degraded ListObjects performance.

This fix addresses two issues in the cleanup path:

1. **`reduce_common_data_dir` now skips `None` votes**: Previously, `None` values (from failed disks) participated in the quorum vote. When `None` had the highest count, it could be returned instead of the actual data directory UUID, preventing cleanup. Now only `Some` values are counted, ensuring deterministic results.

2. **Introduce `cleanup_disks` for targeted cleanup**: Previously, `commit_rename_data_dir` was called with all online disks, which could attempt deletion on disks that didn't have the old data directory. Now we compute `cleanup_disks` containing only disks that successfully renamed and reported the same `old_data_dir` as the quorum winner, ensuring cleanup only happens where the old directory actually exists.

The fix applies to both single PUT and multipart commit paths.

## Verification
- `make pre-commit`
- `cargo test -p rustfs-ecstore reduce_common_data_dir`

## Impact
- **User-facing**: Fixes resource leak on object overwrite for unversioned buckets. Applications that frequently overwrite objects (e.g., s3fs mounts, append-only log files) will no longer accumulate orphaned data directories.
- **Compatibility**: No API or configuration changes. Behavior is backward compatible.
- **Performance**: ListObjects performance will no longer degrade due to accumulated orphaned data directories.

## Additional Notes

The fix includes a new test case that validates `reduce_common_data_dir` correctly ignores `None` votes and returns the quorum-winning UUID.

---

### DIFF
--- a/crates/ecstore/src/set_disk.rs
+++ b/crates/ecstore/src/set_disk.rs
@@ -1171,7 +1171,7 @@ impl ObjectIO for SetDisks {
                 object_lock_guard = Some(self.acquire_write_lock_diag("put_object_commit", bucket, object).await?);
             }
 
-            let (online_disks, _, op_old_dir) = Self::rename_data(
+            let (online_disks, _, op_old_dir, cleanup_disks) = Self::rename_data(
                 &shuffle_disks,
                 RUSTFS_META_TMP_BUCKET,
                 tmp_dir.as_str(),
@@ -1183,7 +1183,7 @@ impl ObjectIO for SetDisks {
             .await?;
 
             if let Some(old_dir) = op_old_dir {
-                self.commit_rename_data_dir(&online_disks, bucket, object, &old_dir.to_string(), write_quorum)
+                self.commit_rename_data_dir(&cleanup_disks, bucket, object, &old_dir.to_string(), write_quorum)
                     .await?;
             }
 
@@ -3840,7 +3840,7 @@ impl MultipartOperations for SetDisks {
 
         self.cleanup_multipart_path(&parts).await;
 
-        let (online_disks, versions, op_old_dir) = Self::rename_data(
+        let (online_disks, versions, op_old_dir, cleanup_disks) = Self::rename_data(
             &shuffle_disks,
             RUSTFS_META_MULTIPART_BUCKET,
             &upload_id_path,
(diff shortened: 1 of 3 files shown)


overtrue (approved):

houseme (approved):

KyleLeneau:
This is not working as expected or I am missing something cause I still have runaway disk usage.

PR #3245, opened 2026-06-06
feat(table-catalog): tighten REST load/register compatibility

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Tightens the Iceberg REST catalog MVP compatibility around table registration and loading.

  - Removes the non-standard create-table endpoint from `/config` advertised endpoints while keeping the internal route available.
  - Keeps the MVP commit route unadvertised because it still uses the RustFS CAS request shape rather than the standard Iceberg REST `requirements` plus
  `updates` shape.
  - Updates register-table request parsing to use the Iceberg REST `overwrite` field.
  - Rejects `overwrite=true` explicitly because overwrite registration is not implemented yet.
  - Loads the current Iceberg metadata JSON object from the table metadata location and returns it in load/register/create table responses instead of returning `metadata: null`.
  - Adds tests covering endpoint advertisement, register request shape, and metadata payload propagation through load/register helpers.

  ## Verification

  - `cargo fmt --all && cargo fmt --all --check && cargo test -p rustfs admin::handlers::table_catalog::tests --lib`

  ## Impact

  Iceberg REST clients will no longer discover the MVP create-table endpoint from `/config` while its request body remains non-standard. Load/register responses now require the referenced metadata object to exist, parse as JSON, and have a JSON object at the top level.

  ## Additional Notes

  This does not implement the full standard Iceberg REST create-table or commit-table request shapes. Those remain gated until RustFS can handle schema-driven create requests and standard commit updates safely.

### DIFF
--- a/rustfs/src/admin/handlers/table_catalog.rs
+++ b/rustfs/src/admin/handlers/table_catalog.rs
@@ -38,7 +38,6 @@ const TABLE_CATALOG_ENDPOINTS: &[&str] = &[
     "GET /{warehouse}/namespaces/{namespace}",
     "DELETE /{warehouse}/namespaces/{namespace}",
     "GET /{warehouse}/namespaces/{namespace}/tables",
-    "POST /{warehouse}/namespaces/{namespace}/tables",
     "POST /{warehouse}/namespaces/{namespace}/register",
     "GET /{warehouse}/namespaces/{namespace}/tables/{table}",
     "DELETE /{warehouse}/namespaces/{namespace}/tables/{table}",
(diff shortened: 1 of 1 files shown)


marshawcoco:
@codex review

marshawcoco:
@overtrue   This failure appears to be caused by the CI runner missing `rg`/ripgrep.

overtrue:
> @overtrue This failure appears to be caused by the CI runner missing `rg`/ripgrep.

I'll fix it.

PR #3247, opened 2026-06-06
feat(scanner): expose scan partial source status

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  - Add partial-cycle source attribution to `ScannerMetricsReport`.
  - Expose `last_cycle_partial_source`, `last_cycle_partial_source_code`, and `partial_cycles_by_source`.
  - Attribute object and directory budget stops to the usage scanner source.
  - Keep runtime budget stops source-neutral because they are not owned by one scanner work source.
  - Preserve existing scanner behavior; this is status/accounting only.

  ## Verification

  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `git diff --check`
  - `cargo test -p rustfs-common report_tracks_partial_scan_cycle_without_failed_increment --lib`
  - `cargo test -p rustfs-scanner test_scan_cycle_partial_source_maps_budget_reason --lib`
  - `cargo test -p rustfs-common --lib`
  - `cargo test -p rustfs-scanner scanner::tests --lib`

  ## Impact

  Scanner status consumers can now distinguish why a partial cycle stopped from which scanner source caused the stop. No configuration, storage format, or scheduling behavior changes are introduced.

  ## Additional Notes

  This completes the source-aware scanner status/accounting groundwork before lifecycle, replication, and heal behavior hardening.

### DIFF
--- a/crates/common/src/metrics.rs
+++ b/crates/common/src/metrics.rs
@@ -267,6 +267,31 @@ impl ScannerWorkSource {
         }
     }
 
+    fn code(self) -> u8 {
+        match self {
+            Self::Usage => 1,
+            Self::Lifecycle => 2,
+            Self::BucketReplication => 3,
+            Self::SiteReplication => 4,
+            Self::Heal => 5,
+            Self::Bitrot => 6,
+            Self::Alerts => 7,
+        }
+    }
+
+    fn from_code(code: u8) -> Option<Self> {
+        match code {
+            1 => Some(Self::Usage),
+            2 => Some(Self::Lifecycle),
+            3 => Some(Self::BucketReplication),
+            4 => Some(Self::SiteReplication),
+            5 => Some(Self::Heal),
+            6 => Some(Self::Bitrot),
+            7 => Some(Self::Alerts),
+            _ => None,
+        }
+    }
+
     fn all() -> &'static [Self] {
         &Self::ALL
     }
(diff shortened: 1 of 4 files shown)


marshawcoco:
@codex review

houseme (changes requested):
the direction here looks good, and the partial-source attribution is wired correctly through the dedicated scanner status path.

The remaining gap is that the newly added `last_cycle_partial_source` / `partial_cycles_by_source` fields are only exposed via `/v3/scanner/status`. The aggregated metrics path still maps scanner data through the older `MadminScannerMetrics` shape, so consumers going through `collect_local_metrics` / admin metrics still cannot see the new fields.

Please extend that aggregated metrics output path as well; otherwise the PR only exposes this new status information on one side.

marshawcoco:
Thanks for the review. Fixed: I extended the aggregated metrics path as well, so `MadminScannerMetrics` now carries `last_cycle_partial_source`, `last_cycle_partial_source_code`, and `partial_cycles_by_source`, and `to_madmin_scanner_metrics` maps them from the scanner report. I also added coverage for the mapping and merge behavior.

houseme (approved):
Looks good to me now. The aggregated metrics path has been updated to carry the new partial-source fields as well, so the exposure gap from the previous round is closed.

Approving from the code review side.

PR #3249, opened 2026-06-06
fix(replication): normalize local site endpoint port

### LINKED ISSUE #2723: [Issue/Discussion] Challenges with Site-to-Site Replication: UID mismatches, TLS SAN validation, and Endpoint Protocol

cxymds:
## Related Issues
Fixes #2723.

## Summary of Changes
- normalize the local site replication endpoint to always use the S3 API port instead of reusing the admin/console request port
- keep the existing `request_endpoint()` behavior for non-site-replication callers and scope the port normalization to site replication local peer resolution only
- reject persisted site replication peers that still point at the console port with a clear `InvalidRequest` error instead of letting replication fail later with a signature mismatch
- add regression tests for console-host-header normalization, IPv6 host preservation, and console-port peer rejection

## Verification
- `cargo fmt --all`
- `cargo fmt --all --check`
- `cargo test -p rustfs site_replication`
- `make pre-commit`

## Impact
- fixes a site replication endpoint drift bug where requests entering through `:9001` could be persisted and later reused as data-plane replication targets
- existing invalid site replication state using the console port now fails fast with an explicit error message
- no API contract change outside site replication behavior

## Additional Notes
- this PR intentionally does not auto-migrate existing site replication state; it prevents new bad local peer endpoints from being written and makes existing console-port peers fail with a clearer error

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -20,6 +20,7 @@ use crate::admin::site_replication_identity::{
 };
 use crate::admin::utils::{encode_compatible_admin_payload, read_compatible_admin_body};
 use crate::auth::{check_key_valid, get_session_token};
+use crate::config::get_config_snapshot;
 use crate::error::ApiError;
 use crate::server::{ADMIN_PREFIX, RemoteAddr};
 use base64::Engine;
@@ -30,7 +31,8 @@ use http::{HeaderMap, HeaderValue, Uri};
 use hyper::{Method, StatusCode};
 use matchit::Params;
 use rustfs_config::{
-    DEFAULT_CONSOLE_PORT, DEFAULT_DELIMITER, DEFAULT_RUSTFS_TLS_PATH, ENV_RUSTFS_TLS_PATH, MAX_ADMIN_REQUEST_BODY_SIZE,
+    DEFAULT_CONSOLE_ADDRESS, DEFAULT_DELIMITER, DEFAULT_RUSTFS_TLS_PATH, ENV_RUSTFS_CONSOLE_ADDRESS, ENV_RUSTFS_TLS_PATH,
+    MAX_ADMIN_REQUEST_BODY_SIZE,
 };
 use rustfs_ecstore::bucket::bucket_target_sys::BucketTargetSys;
 use rustfs_ecstore::bucket::metadata::{
@@ -723,11 +725,27 @@ fn request_endpoint(uri: &Uri, headers: &HeaderMap) -> String {
     format!("{scheme}://{host}")
 }
 
+fn runtime_console_port() -> Option<u16> {
+    let console_address = get_config_snapshot()
+        .map(|snapshot| snapshot.console_address.clone())
+        .unwrap_or_else(|| rustfs_utils::get_env_str(ENV_RUSTFS_CONSOLE_ADDRESS, DEFAULT_CONSOLE_ADDRESS));
+
+    let parse_target = if console_address.starts_with(':') {
+        format!("127.0.0.1{console_address}")
+    } else {
+        console_address
+    };
+
+    Url::parse(&format!("http://{parse_target}"))
+        .ok()
+        .and_then(|parsed| parsed.port_or_known_default())
+}
+
 fn site_replication_local_endpoint(uri: &Uri, headers: &HeaderMap) -> String {
     let endpoint = request_endpoint(uri, headers);
     match Url::parse(&endpoint) {
         Ok(mut parsed) => {
-            if parsed.set_port(Some(global_rustfs_port())).is_ok() {
+            if parsed.port_or_known_default() == runtime_console_port() && parsed.set_port(Some(global_rustfs_port())).is_ok() {
                 parsed.to_string().trim_end_matches('/').to_string()
             } else {
                 endpoint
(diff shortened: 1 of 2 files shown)


houseme (approved):

overtrue (changes requested):
Blocking: rustfs/src/admin/handlers/site_replication.rs:798 rejects every peer endpoint whose port is 9001 by comparing against DEFAULT_CONSOLE_PORT, but the S3 API port is configurable; a valid remote configured with --address :9001 will now fail target/config creation with InvalidRequest. Please compare against the actual console listener or only reject the known local console-derived endpoint while allowing peers that use global_rustfs_port().

houseme (changes requested):
Requesting changes for two edge cases before we merge this.

1. `validate_site_replication_peer_endpoint()` hard-codes the console port as `9001`, but `RUSTFS_CONSOLE_ADDRESS` is configurable in this repo. That means custom console ports will slip through, and it could also reject a valid setup if the S3 API is intentionally bound to `9001`.
2. `site_replication_local_endpoint()` now rewrites the local endpoint port to `global_rustfs_port()` unconditionally. This PR is trying to fix console-port drift, but this broader rewrite can also change a valid externally reachable S3 endpoint coming from a load balancer or reverse proxy into an internal backend port that peers cannot reach.

I think the direction is right, and the new tests are helpful. We just need to make both checks depend on the actual runtime console configuration, and only rewrite the port when we know the incoming endpoint is using the console port.

cxymds:
> Requesting changes for two edge cases before we merge this.
> 
> 1. `validate_site_replication_peer_endpoint()` hard-codes the console port as `9001`, but `RUSTFS_CONSOLE_ADDRESS` is configurable in this repo. That means custom console ports will slip through, and it could also reject a valid setup if the S3 API is intentionally bound to `9001`.
> 2. `site_replication_local_endpoint()` now rewrites the local endpoint port to `global_rustfs_port()` unconditionally. This PR is trying to fix console-port drift, but this broader rewrite can also change a valid externally reachable S3 endpoint coming from a load balancer or reverse proxy into an internal backend port that peers cannot reach.
> 
> I think the direction is right, and the new tests are helpful. We just need to make both checks depend on the actual runtime console configuration, and only rewrite the port when we know the incoming endpoint is using the console port.


validate_site_replication_peer_endpoint() no longer hardcodes 9001. It now checks against the runtime RUSTFS_CONSOLE_ADDRESS, so custom console ports are validated correctly.
site_replication_local_endpoint() no longer rewrites the local endpoint port unconditionally to global_rustfs_port(). It now only rewrites when the current request port is confirmed to be the console port, so valid S3 endpoints exposed through LB/reverse proxy are preserved.
Also added regression coverage for both cases and verified with:

cargo fmt --all
cargo fmt --all --check
cargo test -p rustfs site_replication -- --nocapture

cxymds:
> Blocking: rustfs/src/admin/handlers/site_replication.rs:798 rejects every peer endpoint whose port is 9001 by comparing against DEFAULT_CONSOLE_PORT, but the S3 API port is configurable; a valid remote configured with --address :9001 will now fail target/config creation with InvalidRequest. Please compare against the actual console listener or only reject the known local console-derived endpoint while allowing peers that use global_rustfs_port().

Fixed in the latest update.

The peer-side validation no longer rejects endpoints just because their port matches the local console port. That check was too broad and could incorrectly reject a valid remote site whose S3 API is intentionally exposed on :9001 or any other port matching the local console listener.

The fix keeps the local console-port normalization only for deriving the local site replication endpoint, and no longer applies a generic port-based rejection to arbitrary peers. I also added regression coverage for the case where a valid remote peer uses the same numeric port as the local console.

Verified with:

cargo fmt --all
cargo fmt --all --check
cargo test -p rustfs site_replication -- --nocapture

houseme:
Addressed the earlier console-port concerns in the latest revision:

- the comparison now uses the runtime console configuration instead of hard-coding `9001`
- local endpoint rewriting now only happens when the incoming endpoint is actually using the console port
- a regression test was added to keep non-console ports unchanged, and another one covers a remote peer using the same numeric port as the local console

Would appreciate a quick re-check when you have a moment.

houseme (approved):
Re-review complete from my side. The two earlier blocking edge cases are addressed now:

- console port handling no longer hard-codes `9001`; it uses the runtime console configuration
- local endpoint port rewriting is now limited to the actual console-port case, so non-console external endpoints are preserved

The added regression coverage also matches those fixes well. Approved.

overtrue (approved):

PR #3257, opened 2026-06-07
fix(lifecycle): harden scanner ILM expiry accounting

marshawcoco:
## Related Issues
N/A

## Summary of Changes
- Add `missed` accounting to scanner source work snapshots.
- Record scanner-origin lifecycle expiry, newer-noncurrent expiry, and transition enqueue outcomes as lifecycle source work `queued` or `missed`.
- Return enqueue success from expiry and transition queue paths so scanner-driven ILM backlog accounting can distinguish queued work from missed work.
- Gate scanner ILM executed/timing accounting and local version/size accounting on successful enqueue, so missed work does not make scanner summaries optimistic.
- Replace the scanner source work positional argument list with `ScannerSourceWorkUpdate` and semantic helpers.
- Add regression coverage for lifecycle source missed accounting, no-worker/full-queue enqueue paths, enqueue-gated ILM accounting, and enqueue-gated local scanner accounting.

## Verification
- `cargo test -p rustfs-ecstore bucket::lifecycle::bucket_lifecycle_ops::tests --lib`
- `cargo test -p rustfs-scanner scanner_folder::tests --lib`
- `cargo test -p rustfs-scanner test_scanner_ilm_action_accounting_requires_enqueue_success --lib`
- `cargo test -p rustfs-scanner test_pending_scanner_accounting_requires_enqueue_success --lib`
- `cargo clippy -p rustfs-common --lib -- -D warnings`
- `cargo clippy -p rustfs-scanner --lib -- -D warnings`
- `cargo fmt --all`
- `cargo fmt --all --check`
- `git diff --check`

## Impact
- Scanner status source work can now expose missed scanner-driven lifecycle enqueue work.
- Scanner local object/version summaries stay conservative when lifecycle enqueue fails.
- No new config keys.
- No lifecycle deletion safety semantics are changed.

## Additional Notes
N/A

### DIFF
--- a/crates/common/src/metrics.rs
+++ b/crates/common/src/metrics.rs
@@ -319,14 +319,70 @@ struct ScannerSourceWorkCounters {
     missed: AtomicU64,
 }
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct ScannerSourceWorkUpdate {
+    pub checked: u64,
+    pub queued: u64,
+    pub executed: u64,
+    pub failed: u64,
+    pub skipped: u64,
+    pub missed: u64,
+}
+
+impl ScannerSourceWorkUpdate {
+    pub const fn checked(count: u64) -> Self {
+        Self {
+            checked: count,
+            queued: 0,
+            executed: 0,
+            failed: 0,
+            skipped: 0,
+            missed: 0,
+        }
+    }
+
+    pub const fn queued(count: u64) -> Self {
+        Self {
+            checked: 0,
+            queued: count,
+            executed: 0,
+            failed: 0,
+            skipped: 0,
+            missed: 0,
+        }
+    }
+
+    pub const fn executed(count: u64) -> Self {
+        Self {
+            checked: 0,
+            queued: 0,
+            executed: count,
+            failed: 0,
+            skipped: 0,
+            missed: 0,
+        }
+    }
+
+    pub const fn missed(count: u64) -> Self {
+        Self {
+            checked: 0,
+            queued: 0,
+            executed: 0,
+            failed: 0,
+            skipped: 0,
+            missed: count,
+        }
+    }
+}
+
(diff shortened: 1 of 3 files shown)


marshawcoco:
@codex review

houseme (changes requested):
Nice direction overall, but I think there is still one blocking inconsistency in the scanner path.

We now correctly gate `record_scanner_ilm_action()` on enqueue success, but `scanner_folder::apply_actions()` still updates local state as if the lifecycle work definitely happened before we know whether enqueue actually succeeded. For example, `remaining_versions -= 1`, `size = 0`, and the noncurrent delete batching all happen before the new `queued` result is checked.

That means a full/closed queue or missing expiry worker can still make the scanner undercount retained versions/size in `alert_excessive_versions()` and `SizeSummary`, while the new metrics correctly report the work as missed. In other words, the PR fixes the lifecycle metrics, but leaves the scanner's own object/version accounting on the old optimistic path.

I'd like to see the local size/version accounting stay unchanged unless the enqueue actually succeeds.

marshawcoco:
Fixed. Scanner local version/size accounting is now gated on enqueue success as well.

  I added pending scanner accounting for destructive lifecycle work, so `remaining_versions`, `SizeSummary`, and cumulative size are only updated as expired/deleted after expiry/transition/newer-noncurrent enqueue succeeds. If enqueue fails, the scanner keeps the object/version/size accounted as retained while the lifecycle source work reports `missed`. I also added a regression test covering the enqueue-success vs enqueue-failure local accounting behavior.

houseme (approved):
This addresses the scanner-side accounting mismatch I called out before.

The local retained size/version bookkeeping now stays on the retained path unless the lifecycle work is actually queued, so it no longer diverges from the new queued/missed lifecycle metrics. This looks good to me pending the remaining CI jobs.

PR #3263, opened 2026-06-07
fix(ecstore): avoid offline disks on admin timeout

### LINKED ISSUE #2525: The disk is showing as offline, but the data can still be uploaded normally.
I created a rustfs cluster with 4 nodes and a total of 16 disks using binary packages. The version is 1.0.0-alpha.93. 
However, when I opened the console to check, I found that the statistical information seemed to be incorrect.

Image

I created a bucket and uploaded an image to it. When viewed on the four servers, it was observed that all the disks successfully synchronized the image data.

Image

Image

Could you please help me figure out what this question is about?

cxymds:
## Related Issues
Fixes #2525

## Summary of Changes
- Raise the admin peer aggregation timeout from 2 seconds to 5 seconds to align with the drive info path.
- Cache the last successful peer `StorageInfo` and `ServerProperties` responses for admin aggregation.
- Avoid synthesizing offline disks or offline peer state on a single admin timeout by serving cached data when available.
- Only synthesize offline admin state after repeated consecutive admin-call failures.
- Add regression coverage for the admin-timeout path where data operations remain healthy.

## Verification
- `cargo test -p rustfs-ecstore notification_sys --lib`
- `cargo test -p e2e_test admin_timeout_regression_test -- --nocapture`
- `cargo fmt --all --check`
- `make pre-commit`

## Impact
Console/admin storage reporting no longer misclassifies an otherwise healthy peer as offline after a single slow admin call. No configuration changes are required.

## Additional Notes
The fix is intentionally scoped to the admin aggregation path. It does not change the underlying data path behavior for S3 operations.

### DIFF
--- a/crates/e2e_test/src/admin_timeout_regression_test.rs
+++ b/crates/e2e_test/src/admin_timeout_regression_test.rs
@@ -1,0 +1,208 @@
+// Copyright 2026 RustFS Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::common::{RustFSTestClusterEnvironment, init_logging, local_http_client};
+use aws_sdk_s3::primitives::ByteStream;
+use http::header::HOST;
+use reqwest::StatusCode;
+use rustfs_madmin::{ITEM_OFFLINE, InfoMessage, StorageInfo};
+use rustfs_signer::constants::UNSIGNED_PAYLOAD;
+use rustfs_signer::sign_v4;
+use s3s::Body;
+use serial_test::serial;
+use std::error::Error;
+use std::process::Command;
+use tokio::time::{Duration, sleep, timeout};
+use uuid::Uuid;
+
+const BUCKET: &str = "issue-2525-admin-timeout";
+
+async fn signed_admin_get(
+    url: &str,
+    access_key: &str,
+    secret_key: &str,
+) -> Result<reqwest::Response, Box<dyn Error + Send + Sync>> {
+    let uri = url.parse::<http::Uri>()?;
+    let authority = uri.authority().ok_or("request URL missing authority")?.to_string();
+    let request = http::Request::builder()
+        .method(http::Method::GET)
+        .uri(uri)
+        .header(HOST, authority)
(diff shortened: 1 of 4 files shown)


overtrue (changes requested):
Blocking: crates/ecstore/src/notification_sys.rs:888 and :918 add production Mutex::lock().unwrap() calls, but crates/AGENTS.md forbids unwrap/expect outside tests; a poisoned cache mutex would panic admin aggregation instead of returning cached/offline/initializing fallback. Please handle the lock error explicitly.

houseme (changes requested):
Thanks for chasing this regression. The cache-and-threshold direction makes sense, but I still see one blocking issue before this is ready:

In `crates/ecstore/src/notification_sys.rs`, both `handle_peer_failure()` and `handle_server_info_failure()` now call `cache.lock().unwrap()` on the production admin fallback path. That means a poisoned cache mutex will panic exactly where this change is trying to degrade gracefully. It also conflicts with `crates/AGENTS.md`, which explicitly disallows `unwrap`/`expect` outside tests.

Please handle the lock error explicitly and keep the fallback path non-panicking.

cxymds:
Updated the admin fallback path to handle poisoned cache mutexes without panicking.

Both `handle_peer_failure()` and `handle_server_info_failure()` now recover the inner state from `PoisonError`, log the condition, and continue returning the existing fallback states instead of panicking in the admin aggregation path.

Added a regression test covering poisoned admin cache mutexes in the `notification_sys` test suite.

Verification rerun:
- `cargo test -p rustfs-ecstore notification_sys --lib`
- `cargo fmt --all --check`
- `make pre-commit`

houseme (approved):
Re-reviewed the mutex-poisoning follow-up in `notification_sys`.

The previous blocker is resolved: the admin cache lock path now handles poisoned mutexes explicitly, keeps the fallback path non-panicking, and adds regression coverage for that case.

Code-wise, this looks good to me now.

overtrue (changes requested):
Blocking: crates/ecstore/src/notification_sys.rs:289 and :339 still ignore poisoned admin cache locks on successful peer calls via `if let Ok(mut c) = cache.lock()`, so after one poisoned cache mutex a later successful admin call will not refresh `last_*` or reset failure counters; the next transient failure can still hit `CONSECUTIVE_FAILURE_THRESHOLD` and synthesize offline state despite recovery. Please recover poisoned locks consistently on the success path too and add a poison -> success -> failure reset regression test.

cxymds:
Updated the successful admin peer-call paths to recover poisoned cache mutexes the same way as the fallback paths.

`storage_info()` and `server_info()` now refresh `last_*` and reset the corresponding failure counters even after a poisoned admin cache lock, so a recovered peer does not immediately trip `CONSECUTIVE_FAILURE_THRESHOLD` on the next transient failure.

Added a regression test covering the poison -> success -> failure-reset flow.

Verification rerun:
- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore notification_sys --lib`
- `make pre-commit`

overtrue (approved):
Looks good now.

Issue #3161, opened 2026-06-01
Single Node Multiple Disk Mode (SNMD) not initialized (1.0.0-beta.6)


flask-pro:
When rustfs is launched using 'docker compose', volumes are not initialized and the service does not start. The logs are empty. 

Version `1.0.0-alpha.86` works.
Version `1.0.0-beta.6` works only in Single Node Single Disk Mode (SNSD) .

My `compose.yaml`:
```yaml
services:
  # grant the necessary permissions to RUSTFS volumes path
  rustfs_perms:
    image: alpine
    user: root
    volumes:
      - /mnt/data:/fix_path
    command: chown -R 10001:10001 /fix_path

  rustfs:
    security_opt:
      - "no-new-privileges:true"
    image: rustfs/rustfs:1.0.0-beta.6
    container_name: rustfs
    ports:
      - "9000:9000" # S3 API port
      - "9001:9001" # Console port
    environment:
      - RUSTFS_VOLUMES=/data/rustfs{0...1} # Define 2 storage volumes
      - RUSTFS_ADDRESS=0.0.0.0:9000
      - RUSTFS_CONSOLE_ADDRESS=0.0.0.0:9001
      - RUSTFS_CONSOLE_ENABLE=true
      - RUSTFS_CONSOLE_CORS_ALLOWED_ORIGINS=*
      - RUSTFS_ACCESS_KEY=${RUSTFS_ACCESS_KEY:?Set RUSTFS_ACCESS_KEY to a non-default value}
      - RUSTFS_SECRET_KEY=${RUSTFS_SECRET_KEY:?Set RUSTFS_SECRET_KEY to a non-default value}
    depends_on:
      rustfs_perms:
        condition: service_completed_successfully
    volumes:
      - /mnt/data:/data
      - ./deploy/rustfs/logs:/app/logs
    networks:
      - legalact-net
    restart: unless-stopped
    healthcheck:
      test:
        [
          "CMD",
          "sh",
          "-c",
          "curl -f http://127.0.0.1:9000/health && curl -f http://127.0.0.1:9001/rustfs/console/health",
        ]
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 40s

networks:
  legalact-net:
    driver: bridge
    ipam:
      config:
        - subnet: 172.20.0.0/16
```

# 1) Basic environment

```
[...]

PR #3267, opened 2026-06-07
fix(ecstore): gate rustix fs diagnostics on windows

houseme:
## Related Issues
- #3161 
- #3265 

## Summary of Changes
- Gate the `rustix::fs::stat` diagnostic block in `crates/ecstore/src/endpoints.rs` behind `#[cfg(not(windows))]`.
- Preserve the existing local physical disk independence validation flow on Windows while avoiding compilation against the Unix-only `rustix::fs` module.
- Keep the change scoped to diagnostics only; no validation behavior was changed for supported platforms.

## Verification
- `cargo fmt --all`
- `cargo fmt --all --check`
- `make pre-commit`

## Impact
- Fixes Windows builds that compile `crates/ecstore` by avoiding references to the Unix-only `rustix::fs` module.
- No expected behavior change on non-Windows platforms beyond keeping the existing diagnostic output path unchanged.

## Additional Notes
- Local verification for `x86_64-pc-windows-msvc` could not be completed in this environment because the target is not installed, but the failing code path is now conditionally compiled only on non-Windows platforms.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -401,7 +401,7 @@ dependencies = [
  "arrow-ord",
  "arrow-schema",
  "arrow-select",
- "atoi",
+ "atoi 2.0.0",
  "base64 0.22.1",
  "chrono",
  "comfy-table",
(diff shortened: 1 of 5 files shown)


houseme:
## Review

The fix is correct and minimal. `rustix::fs` is gated behind `#[cfg(not(windows))]` in the rustix crate itself, so the attribute properly gates the `stat`, `major`, and `minor` calls.

On Windows after this fix:
- `diagnostic.device_numbers` stays `None`, which `summary()` already handles via `unwrap_or("(unavailable)")`
- The actual validation logic is unaffected — it uses `get_physical_device_ids()` (line 776) which has a dedicated Windows implementation

**One observation for follow-up:** the Windows `NotFound` fallback path (lines 731-756) calls `get_physical_device_ids` but doesn't push a diagnostic entry before `continue` (line 748). Not a blocker since PR #3265 had the same gap, but worth tracking if we want complete diagnostic coverage on Windows.

LGTM.

PR #3277, opened 2026-06-08
fix(scanner): publish partial usage for compacted scans

### LINKED ISSUE #3273: Bucket usage information is still pending.
If the number of objects is particularly large(I guess), it will keep looping and waiting until it times out.

Image

And finally like this.

Image

marshawcoco:
## Related Issues

  Fixes #3273

  ## Summary of Changes

  - Publish scanner usage progress while compacted parent directories are still scanning, so large compacted trees no longer wait until the whole parent
  finishes before producing backend usage updates.
  - Publish the first bucket-level partial usage snapshot immediately when the set-level usage cache is cold or missing that bucket, instead of waiting for the periodic scanner persistence tick.
  - Skip the initial scanner startup delay when persisted usage cache is cold and at least one bucket exists, while preserving the configured randomized delay for warm caches, empty deployments, and inspection failures.
  - Keep the change backend-only; no admin API or frontend behavior is changed.

  ## Verification

  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `git diff --check`
  - `cargo test -p rustfs-scanner test_scan_folder_compacted_parent_sends_partial_update -- --nocapture`
  - `cargo test -p rustfs-scanner apply_bucket_result_ -- --nocapture`
  - `cargo test -p rustfs-scanner scanner_io::tests -- --nocapture`
  - `cargo test -p rustfs-scanner scanner_folder::tests -- --nocapture`
  - `cargo test -p rustfs-scanner initial_scanner_delay_ -- --nocapture`
  - `cargo test -p rustfs-scanner --lib -- --test-threads=1`

  ## Impact

  Cold or missing data usage cache now starts producing backend usage data sooner after startup. Existing warm-cache deployments keep the configured staggered scanner startup behavior. No API, schema, configuration, or deployment changes are required.

  ## Additional Notes

  N/A

### DIFF
--- a/crates/scanner/src/scanner.rs
+++ b/crates/scanner/src/scanner.rs
@@ -118,14 +118,88 @@ fn randomized_cycle_delay_for(interval: Duration) -> Duration {
     delay.max(Duration::from_secs(1))
 }
 
-fn initial_scanner_delay() -> Duration {
-    initial_scanner_delay_for(scanner_start_delay().map(|duration| duration.as_secs()))
-}
-
 fn initial_scanner_delay_for(start_delay_secs: Option<u64>) -> Duration {
     start_delay_secs
         .map(|secs| randomized_cycle_delay_for(Duration::from_secs(secs)))
         .unwrap_or_else(randomized_cycle_delay)
+}
+
+fn initial_scanner_delay_for_startup(start_delay_secs: Option<u64>, usage_cache_is_cold: bool, has_buckets: bool) -> Duration {
+    if usage_cache_is_cold && has_buckets {
+        Duration::ZERO
+    } else {
+        initial_scanner_delay_for(start_delay_secs)
+    }
+}
+
+fn data_usage_info_is_cold(info: &DataUsageInfo) -> bool {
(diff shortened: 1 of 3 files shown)


houseme (changes requested):
This is close, but I don't think the immediate-publish gate matches the intended behavior yet. In `apply_bucket_result_to_cache`, `should_publish` is driven by `cache.info.last_update.is_none() || cache.find(&result.name).is_none()`. The set cache is rebuilt with `last_update: None` even when `old_cache` was warm, so the first result for an already-cached bucket still gets published immediately. That widens the change beyond the PR goal of only fast-tracking cold caches or buckets missing from the set cache.

I’d tighten this by carrying the warm/cold state from the loaded cache into the new set cache, or by making the predicate check whether the preloaded cache already had published data for that bucket before this cycle starts.

Also, `cargo clippy -p rustfs-scanner --lib --tests -- -D warnings` currently fails on the new `ConfigNotFound` match guards in `scanner.rs`, so this isn’t green to approve yet.

houseme (approved):
Thanks for tightening this up. The warm-cache immediate publish gate now behaves the way the PR describes, and the scanner-side cleanup is green on my side.

PR #3283, opened 2026-06-08
feat(table-catalog): refine table catalog permissions

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Refines the Iceberg REST table catalog authorization boundary.

  This PR changes warehouse-scoped table catalog routes to authorize with the target warehouse bucket instead of using unscoped admin authorization, and tightens the load-table route to require table metadata read permission.

  Key behavior:

  - catalog config remains catalog-level authorization.
  - namespace and table routes authorize against the `{warehouse}` bucket.
  - load table now uses `GetTableMetadataAction` because the response reads and returns table metadata JSON.
  - commit continues to require `CommitTableAction`.
  - create/register/delete/list namespace and table routes retain their existing table-aware actions.
  - no public maintenance route is added in this PR.

  The route/action mapping is now covered by handler-block tests so the test cannot pass merely because an action name appears elsewhere in the file.

  ## Verification

  - `cargo test -p rustfs table_catalog_handlers_require_table_admin_actions --lib`
  - `cargo test -p rustfs-policy test_table_admin_actions_are_valid --lib`
  - `cargo test -p rustfs-policy test_table_metadata_admin_actions_are_valid --lib`
  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `git diff --check`

  ## Impact

  Table catalog REST authorization is now scoped to the warehouse bucket for warehouse routes. Policies that previously relied on unscoped admin table
  catalog access may need to allow the same action for the target warehouse bucket.

  No data-plane object authorization, credential vending, table ARN matching, or maintenance API behavior is changed.

  ## Additional Notes

  This completes the MVP table catalog permission refinement.

### DIFF
--- a/crates/policy/src/policy/action.rs
+++ b/crates/policy/src/policy/action.rs
@@ -563,6 +563,31 @@ pub enum AdminAction {
 }
 
 impl AdminAction {
+    pub(crate) fn is_table_resource_scoped(&self) -> bool {
+        matches!(
+            self,
+            AdminAction::GetTableBucketAction
+                | AdminAction::SetTableBucketAction
+                | AdminAction::GetTableNamespaceAction
+                | AdminAction::SetTableNamespaceAction
+                | AdminAction::UpdateTableNamespacePropertiesAction
+                | AdminAction::DeleteTableNamespaceAction
+                | AdminAction::GetTableAction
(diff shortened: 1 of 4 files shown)


marshawcoco:
@codex review

Issue #3281, opened 2026-06-08
Can't retrieve user principalID via CLI or WebUI


Theoooooo:
**Is your feature request related to a problem? Please describe.**
I'm currently using RustFS with an external ClamAV instance using the notification destination events configured as a webhook.
The events are configured to PUT on my bucket, when i upload a new file it correctly send a notification to my ClamAV instance, ect ...

But in the data sent, i can't find a way to link the Put notification to a specific user. The only thing i can use for now is the data under the following key : Records[0].userIdentity['principalID'] which look something like that : L8JEX5VAR4L03GIF2VV4

I can't find a way to link this ID to any existing users. Listing the users via the RC command line doesn't return any ID i can potentially use, same thing via the WebUI, inspecting the network request/response when listing the users

Is it possible to add this feature ?

cxymds:
@overtrue  I am already working on this enhancement

houseme:
Thanks for reporting this.

The notification flow itself seems to be working correctly. The real issue is that `Records[0].userIdentity.principalID` appears to be exposed as an internal S3-compatible request identity, rather than a user-facing RustFS username or a directly queryable user ID.

Because RustFS currently does not provide a straightforward way to resolve that `principalID` back to the actual user in the CLI, WebUI, or event payload, downstream integrations such as ClamAV can receive the event but still cannot reliably attribute the upload to a specific user.

So the gap is less about event delivery, and more about missing identity mapping. For audit, security scanning, and external automation workflows, this makes the notification much less useful in practice.

A practical fix would be to either expose a resolvable user identity directly in the event payload, or provide an admin/API lookup path that maps `principalID` back to the corresponding user.

Theoooooo:
> A practical fix would be to either expose a resolvable user identity directly in the event payload, or provide an admin/API lookup path that maps `principalID` back to the corresponding user.

Thanks for the detailed shared, i think that's exactly what i'm aiming for
Since the identity is already being delivered through the event notification system, we just need to be able to look it

I was either thinking of :
- Displaying the ID inside the user cli list command (rc admin user info <ALIAS> <USER>
- Displaying the principalID and the related username/access_key of the ID inside the log

cxymds:
Implemented CLI support for inspecting any access key via the RustFS Admin API.

New command:

```bash
rc admin access-key info <ALIAS> <ACCESS_KEY>

```
This resolves whether the access key belongs to a regular user, service account, or STS credential, and supports both human-readable and JSON output.
[...]

PR #3285, opened 2026-06-08
fix: support regular users in access key info

cxymds:
## Summary
- extend `info-access-key` to return regular user details instead of rejecting non-derived access keys
- keep existing STS and service account behavior while tightening fallback visibility checks
- extract access-key identity projection into an internal helper module without changing existing service-account admin route contracts
- reuse the shared IAM error mapper for regular-user lookups so the refactor does not introduce new error-shape drift
- keep bulk access-key listing on the existing slim response shape while centralizing projection logic

## Testing
- cargo test -p rustfs --lib admin::access_key_identity::tests::regular_user_lookup_errors_match_shared_iam_mapper -- --exact
- cargo test -p rustfs --lib admin::handlers::service_account::tests
- cargo fmt --all --check
- make pre-commit


cxymds:
https://github.com/rustfs/backlog/issues/663

cxymds:
https://github.com/rustfs/rustfs/issues/3281

rustfs/src/admin/handlers/service_account.rs
```diff
@@ -848,7 +770,7 @@ impl Operation for InfoAccessKey {
                 return Err(s3_error!(AccessDenied, "access denied"));
             };
 
-            if request_user_name(&cred) != target_cred.parent_user {
+            if !can_fallback_view_access_key_info(&cred, target_cred) {
```
GatewayJ: One authorization edge case looks worth checking here. `is_allowed(... ListServiceAccountsAdminAction, deny_only: false)` returning false does not only mean the caller lacks an explicit allow. It can also mean an explicit Deny matched, the caller only has non-admin/S3 permissions such as `s3:GetObject`, an STS or service-account parent/session policy does not allow this admin action, policy conditions such as source IP or claims do not match, or the caller has no resolvable policy. The new ownership fallback treats all of those failures the same and still allows self/parent-owned `info-access-key` responses, so an explicit deny on `admin:ListServiceAccounts` can be bypassed in the fallback cases.

GatewayJ:
In `user.rs`, the self-service `deny_only` path explicitly excludes service accounts: regular users can act on themselves, temp/STS credentials can act on their parent user, but service accounts must have an explicit admin allow.

This PR’s `info-access-key` fallback seems to use `request_user_name(requester)` for both STS and service accounts. Since `request_user_name()` returns `parent_user` for derived credentials, a service account with `parent_user = alice` can query the regular user access key `alice` when there is no explicit `admin:ListServiceAccounts` Allow, as long as there is also no explicit Deny.

Is that intentional? My concern is that this lets a service account represent its parent user for `info-access-key`, while the existing `user.rs` boundary treats service accounts differently from temp/STS credentials. Should service accounts require an explicit `admin:ListServiceAccounts` Allow here, matching the existing self-service boundary?

cxymds:
Addressed the latest access-boundary concern in `InfoAccessKey`.

Service accounts no longer use the self/parent fallback path after the admin check fails. The fallback now remains limited to regular-user self lookups and STS/temp credentials representing their parent user, matching the existing `user.rs` deny-only boundary.

Added a regression test covering the service-account -> parent regular user case, and kept the existing no-explicit-deny fallback tests green.

Verification:
- `cargo fmt --all --check`
- `cargo test -p rustfs --lib fallback_access_key_info -- --nocapture`

GatewayJ (approved):

cxymds:
- Fixed service account identity resolution by checking `is_service_account()` before `is_temp()`. Service account credentials may contain a JWT `session_token`, so checking temp first can incorrectly classify them as STS.
- Updated the test fixture so signed service account credentials are stored through the user/service-account path, not the STS path, and the resolver now receives the signed credential shape that matches production.
- Unified the admin-authorized `InfoAccessKey` not-found response locally: missing regular users now return `InvalidRequest: access key not exist`, matching missing STS/service-account behavior.
- This does not change the shared `iam_error_to_s3_error()` mapper, so other S3/IAM handlers keep their existing error contracts.
- The non-admin fallback path still returns `AccessDenied` for missing target keys, preserving the existing anti-enumeration behavior.

GatewayJ (approved):

PR #3287, opened 2026-06-08
feat(table-catalog): support standard REST create commit

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  - Advertise standard Iceberg REST create and commit endpoints from catalog config now that the handler accepts standard request shapes.
  - Change create table handling to accept schema-driven Iceberg REST fields and generate the initial Iceberg table metadata JSON in the protected metadata directory.
  - Add standard commit handling for `requirements` plus `updates`, applying supported metadata updates before writing the next metadata JSON and using the existing catalog CAS path.
  - Return the updated metadata payload from commit responses while keeping token, generation, and commit id fields for RustFS diagnostics.
  - Keep the legacy MVP commit body path for existing internal/manual callers.
  - Add fail-closed tests for mismatched requirements, unknown updates, and `set-location` updates that try to move table metadata outside the current table bucket.

  ## Verification

  - `cargo fmt --all --check`

  ## Impact

  This removes the biggest REST Catalog protocol-shape gap after the MVP foundation: standard clients create tables from schema metadata and commit through requirements plus updates, rather than by sending a precomputed new metadata location.

  The implementation still routes final pointer changes through the existing catalog store CAS and commit log path, so the new protocol adapter does not duplicate concurrency control.

  ## Additional Notes

  This still does not claim full Iceberg REST parity. PyIceberg/Spark end-to-end verification, deeper manifest/data path validation, snapshot lineage conflict checks, standard idempotency header support, and client compatibility matrix work remain for follow-up.

### DIFF
--- a/rustfs/src/admin/handlers/table_catalog.rs
+++ b/rustfs/src/admin/handlers/table_catalog.rs
@@ -28,6 +28,7 @@ use rustfs_policy::policy::action::{Action, AdminAction};
 use s3s::{Body, S3Request, S3Response, S3Result, header::CONTENT_TYPE, s3_error};
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
 use std::collections::BTreeMap;
+use time::OffsetDateTime;
 use uuid::Uuid;
 
 const JSON_CONTENT_TYPE: &str = "application/json";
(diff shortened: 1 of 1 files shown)


marshawcoco:
@codex review

PR #3289, opened 2026-06-08
fix(cli): remove autogenerated `help` Clap subcommand

Tirka:
## Related Issues
N/A

## Summary of Changes
Currently Clap generates subcommand `help`.

`rustfs --help`:

```text
Main CLI parser

Usage: rustfs [COMMAND]

Commands:
  server  Start the object storage server (default when no subcommand is given)
  info    Display system information
  tls     Inspect TLS certificate directory layout and parsing status
  help    Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
```

This subcommand isn't handled in source code, rustfs tries to run `server` subcommand instead and fails.

`rustfs help`:

```text
{"timestamp":"2026-06-08T21:14:56.739294507+03:00","level":"ERROR","fields":{"message":"Server encountered an error and is shutting down: Volume not found"},"target":"rustfs","filename":"rustfs/src/main.rs","line_number":249,"threadName":"main","threadId":"ThreadId(1)"}
[FATAL] Server encountered an error and is shutting down: Volume not found
```

## Verification
Manual testing

`rustfs --help`:

```text
Main CLI parser

Usage: rustfs [COMMAND]

Commands:
  server  Start the object storage server (default when no subcommand is given)
  info    Display system information
  tls     Inspect TLS certificate directory layout and parsing status

Options:
  -h, --help     Print help
  -V, --version  Print version
```

## Impact
Better CLI user experience

## Additional Notes
Alternative solution could be actual implementation of `help` subcommand

I have read and agree to the CLA


Tirka:
> 
> Missing signatures:
> 
> * @Tirka
> 
> 
> 

houseme (changes requested):
This change hides the autogenerated `help` subcommand from `--help`, but it does not fix the original `rustfs help` behavior.

`preprocess_args_for_legacy()` still treats `help` as a legacy server argument because `KNOWN_SUBCOMMANDS` only includes `server`, `info`, and `tls`. That means `rustfs help` is still rewritten to `rustfs server help`, so we would keep falling into the server path instead of showing help.

Please either preserve `help` as a recognized command in the preprocessing layer or add explicit handling for it, and add a regression test for `rustfs help` so this path stays covered.

overtrue (changes requested):
Blocking: this only hides `help` from `--help`; `preprocess_args_for_legacy()` still rewrites `rustfs help` to `rustfs server help` because `KNOWN_SUBCOMMANDS` lacks `help`, so please fix that path and add a regression test.

PR #3295, opened 2026-06-09
feat(table-catalog): internalize catalog backing paths

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Move the object-backed table catalog control-plane state out of user table buckets and into the internal RustFS metadata bucket.

This PR updates the catalog store backing paths so that table bucket entries, namespace entries, table entries, commit logs, idempotency indexes, and table-entry write locks use `.rustfs.sys` with an internal `s3tables/catalog/table-buckets/<bucket-hash>/...` layout.

It also shares the internal catalog prefix/hash helper with bucket deletion and removes the matching `.rustfs.sys` catalog prefix after a bucket is successfully deleted. This prevents stale table catalog records from being reused if the same bucket name is created again.

Iceberg table metadata and maintenance/diagnostic object checks still operate on the user table bucket, because those files are part of the Iceberg data plane rather than the catalog entry source of truth.

The path layout hashes untrusted table bucket, table id, commit id, and idempotency values before using them in internal object paths.

## Verification

- `cargo test -p rustfs-ecstore store::bucket::tests:: --lib`
- `cargo test -p rustfs table_catalog::tests:: --lib`
- `cargo fmt --all --check`
- `git diff --check`

## Impact

No external REST API shape changes.

Catalog control-plane objects are no longer written into the user table bucket namespace. Deleting a bucket now also removes that bucket's internal table catalog prefix from `.rustfs.sys`.

Existing MVP object-backed catalog data under older paths is not migrated by this PR.

## Additional Notes

N/A

### DIFF
--- a/crates/ecstore/src/bucket/metadata.rs
+++ b/crates/ecstore/src/bucket/metadata.rs
@@ -30,6 +30,7 @@ use s3s::dto::{
     ServerSideEncryptionConfiguration, Tagging, VersioningConfiguration, WebsiteConfiguration,
 };
 use serde::Serializer;
+use sha2::{Digest, Sha256};
 use std::collections::HashMap;
 use std::io::{Read, Write};
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


marshawcoco:
@codex review

PR #3302, opened 2026-06-09
feat(table-catalog): add metadata maintenance control plane

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Adds the next S3 Tables maintenance slice for metadata-only maintenance.
 This PR extends the existing metadata maintenance planner with a job/report shape that includes:

  - maintenance job id
  - table bucket, namespace, and table identity
  - current metadata location and catalog generation
  - retain-recent metadata setting
  - cleanup safety window and cleanup watermark
  - separate `cleanup_candidate_locations` and `deletable_metadata_locations`

  It also adds a metadata maintenance control-plane route:

  `POST /{warehouse}/namespaces/{namespace}/tables/{table}/maintenance/metadata`

  The route is guarded by warehouse-scoped `RunTableMaintenanceAction` and accepts:

  - `retain-recent-metadata-files`
  - `delete`

  By default `delete` is false, so the route performs a dry-run. When `delete` is true, it runs the existing metadata-only controlled delete path, which still rechecks the current metadata pointer before deletion.

  This intentionally remains metadata-only. It does not delete data files, manifest files, or delete files, and it does not add a background worker, compaction, snapshot expiration, lease management, or persisted job status.

  ## Verification

  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `cargo test -p rustfs table_catalog::tests:: --lib`
  - `git diff --check`

  ## Impact

  Adds an internal table catalog maintenance control-plane endpoint for Iceberg metadata cleanup planning and metadata-only cleanup execution.

  The new maintenance route is not advertised in the standard Iceberg REST `/config` endpoint because it is a RustFS maintenance extension, not a standard Iceberg REST Catalog operation.

  ## Additional Notes

  The conservative delete boundary is intentional: LIST is only used to discover metadata candidates, and deletion remains limited to table metadata JSON objects after current pointer and retention checks.

### DIFF
--- a/rustfs/src/admin/handlers/table_catalog.rs
+++ b/rustfs/src/admin/handlers/table_catalog.rs
@@ -57,6 +57,7 @@ static REGISTER_TABLE_HANDLER: RestRegisterTableHandler = RestRegisterTableHandl
 static LOAD_TABLE_HANDLER: RestLoadTableHandler = RestLoadTableHandler {};
 static COMMIT_TABLE_HANDLER: RestCommitTableHandler = RestCommitTableHandler {};
 static DROP_TABLE_HANDLER: RestDropTableHandler = RestDropTableHandler {};
+static TABLE_METADATA_MAINTENANCE_HANDLER: RestTableMetadataMaintenanceHandler = RestTableMetadataMaintenanceHandler {};
 
 #[derive(Debug, Serialize)]
 struct CatalogConfigResponse {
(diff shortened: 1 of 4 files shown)


marshawcoco:
@codex review

houseme (approved):
Looks good to me. The new route, policy entry, report shape, and tests line up cleanly. No blocking issues from my side.

PR #3305, opened 2026-06-09
feat(table-catalog): add credential response boundary

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Adds the first permission/credential boundary step for the Iceberg REST catalog.

  - Introduces a `TableCatalogResource` context for warehouse, namespace, and table-scoped REST catalog routes.
  - Routes now build a catalog resource before authorization, while preserving the current warehouse bucket-scoped IAM behavior.
  - Adds the standard Iceberg REST `storage-credentials` response field to load table responses.
  - Marks credential vending as explicitly unsupported with `rustfs.credential-vending=unsupported`.
  - Keeps credential fields out of the response, so the catalog does not imply temporary S3 credential vending is available yet.

  This is intentionally a structural step: it prepares later table/namespace ARN handling and credential vending without changing current authorization semantics.

  ## Verification

  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `git diff --check`
  - Rust code quality scan on the changed production code
  - `TMPDIR=/data/tmp CARGO_TARGET_DIR=/data/rustfs-target cargo test -p rustfs table_catalog::tests:: --lib`

  ## Impact

  Iceberg REST load/create/register table responses now include an explicit credential boundary. Existing clients should continue to use their configured S3 credentials; RustFS still does not vend temporary credentials in this PR.

  ## Additional Notes

  N/A

### DIFF
--- a/crates/policy/src/policy/policy.rs
+++ b/crates/policy/src/policy/policy.rs
@@ -1588,6 +1588,79 @@ mod test {
     }
 
     #[tokio::test]
+    async fn test_table_admin_action_with_resource_is_limited_to_table_object_scope() -> Result<()> {
+        use crate::policy::action::{Action, AdminAction};
+
+        let data = r#"
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": ["admin:GetTableMetadata"],
+      "Resource": ["arn:aws:s3:::warehouse-a/namespaces/analytics/tables/events"]
+    }
+  ]
+}
+"#;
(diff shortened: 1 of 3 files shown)


marshawcoco:
@codex review

houseme (approved):
Looks good to me. The table-catalog auth path now carries namespace/table object scope into IAM matching, and the load-table response keeps the credential boundary explicit by returning no storage credentials and marking credential vending as unsupported. I did not find any blocking issues in this diff.

PR #3320, opened 2026-06-10
feat(table-catalog): add product API surface

marshawcoco:
## Related Issues
  N/A

  ## Summary of Changes
  Adds the RustFS table catalog product API surface on top of the existing Iceberg REST catalog MVP.

  - Adds table bucket enablement and discovery routes under both `/iceberg/v1` and `/_iceberg/v1`.
  - Adds metadata-location get/update APIs backed by the existing catalog CAS commit path.
  - Adds maintenance config and maintenance job status APIs.
  - Persists maintenance config and maintenance reports in the internal catalog backing store.
  - Adds catalog export, import, diagnostics, and rollback API routes.
  - Rejects `background-enabled: true` in maintenance config because background workers are not implemented in this PR.
  - Extends route registration and route policy coverage for all new canonical and compat catalog routes.

  ## Verification
  - `cargo fmt --all`
  - `cargo fmt --all --check`
  - `TMPDIR=/data/tmp CARGO_TARGET_DIR=/data/rustfs-target-p4-product cargo test -p rustfs table_catalog --lib -j2`
  - `TMPDIR=/data/tmp CARGO_TARGET_DIR=/data/rustfs-target-p4-product cargo test -p rustfs route_registration_test --lib -j2`
  - `TMPDIR=/data/tmp CARGO_TARGET_DIR=/data/rustfs-target-p4-product cargo clippy -p rustfs --lib -j2 -- -D warnings`
  - `git diff --check`

  ## Impact
  Adds new table catalog extension API routes. Automatic background maintenance remains unsupported; manual maintenance and job status are available through the new API surface.

  ## Additional Notes
  The metadata-location update and rollback APIs reuse the existing CAS commit implementation instead of introducing a second pointer update path.

### DIFF
--- a/rustfs/src/admin/handlers/table_catalog.rs
+++ b/rustfs/src/admin/handlers/table_catalog.rs
@@ -1720,6 +1720,7 @@ where
 
 async fn update_table_metadata_location_response<S>(
     store: &S,
+    metadata_backend: &impl crate::table_catalog::TableCatalogObjectBackend,
     bucket: &str,
     namespace: &crate::table_catalog::Namespace,
     table: &str,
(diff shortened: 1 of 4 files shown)


marshawcoco:
@codex review

houseme (changes requested):
Thanks for pushing this product API surface forward. I found one blocker before approval: `rollback_table_response` commits the requested metadata location before parsing the target metadata and checking its table `location`. The generic commit path only checks that the object exists under this table metadata directory, so an existing but invalid metadata file can still become the current pointer and then fail while building the response. Please mirror the pre-commit `read_table_metadata_json` + `validate_metadata_table_location_in_bucket` guard used by the metadata-location update/import paths.

Smaller follow-up: `enable_table_bucket_response` writes the bucket marker before creating the catalog entry, so a failed catalog write can leave `GET /buckets` reporting the bucket as enabled with no catalog entry. Swapping that order would make the enable call atomic.

houseme (approved):
Thanks for the follow-up. I rechecked the previous blockers and the latest commits address them: rollback now validates the target metadata before committing, and table-bucket enablement now writes the visible marker after the catalog entry is created. The focused table-catalog tests and clippy pass locally.

Issue #3057, opened 2026-05-22
Expand Pool Rustfs build on K8S


casperair77-tech:
Currently, I'm building Rustfs using K8S with 4 nodes and 4 PVCs/1 pod.

How would I expand the pool capacity on this cluster if I add 4 more nodes?

And currently, I'm using a storage class created and managed by DirectPV.

this is file value.yml of Helm Chart with firtst pool

casperair77-tech:
```
replicaCount: 4
config:
  rustfs:
    # Examples
    # volumes: "/data/rustfs0,/data/rustfs1,/data/rustfs2,/data/rustfs3"
    volumes: "http://rustfs-{0...3}.rustfs-headless:9000/data/rustfs{0...3} "
storageclass:
  name: local-ssd
  dataStorageSize: 50Gi
  logStorageSize: 15Gi
  pvcAnnotations: {}
```

majinghe:
Sorry to reply so late.

To be honest, expand pool just doesn't play well with K8s installation. If you want to expand the pool, the `RUSTFS_VOLUMES` will be `http://rustfs-pool-1-{0...3}.rustfs-pool-1-headless:9000/data/rustfs{0...3} http://rustfs-pool-2-{0...3}.rustfs-pool-2-headless:9000/data/rustfs{0...3}`, and the RustFS will start pool 1 and 2. However, with k8s installation, container is started by k8s instead of RustFS, which means, pool 2 can not be started with pool 1 at the same time.


I think the best practice for k8s expand is to adjust the size of pvc and resources limit for pods dynamically.

Rohmilchkaese:
This is now implemented in #3325 — chart-native pool expansion.

@majinghe regarding the startup-ordering concern ("pool 2 can not be started with pool 1 at the same time"): the PR handles this by keeping **all pools on one shared headless service**, so every pool expression resolves under the same DNS domain, and by relying on the server's own retry-through-restart behavior: all pool StatefulSets are created simultaneously (`podManagementPolicy: Parallel`), pods that come up before their peers are resolvable exit and get restarted by the kubelet, and the cluster converges on its own once every pod has a DNS entry.

Tested live on a k8s cluster (Longhorn storage, `1.0.0-beta.7`): a running single-pool deployment with data was expanded via plain `helm upgrade` to

```yaml
pools:
  enabled: true
  list:
    - {}                  # existing pool, keeps its StatefulSet/PVCs/data in place
    - replicaCount: 4     # new capacity
[...]

PR #3325, opened 2026-06-10
feat(helm): support multiple server pools for capacity expansion

Rohmilchkaese:
## What

Adds opt-in **server pool** support to the Helm chart, enabling horizontal capacity expansion of distributed deployments — the chart-side counterpart to the pool support the server already has (`RUSTFS_VOLUMES` is split on spaces with one pool expression each; `rc admin pool / expand / rebalance / decommission` already exist).

```yaml
pools:
  enabled: true
  list:
    - {}                  # pool 0: inherits top-level values, keeps existing names/PVCs/data
    - replicaCount: 4     # pool 1: new capacity (4 or 16)
      storageclass:
        dataStorageSize: 10Gi
```

## Design

- `pools.enabled=false` (default): rendering is **byte-identical** to the current chart.
- One StatefulSet per `pools.list` entry. Entries may set `replicaCount` (4 or 16, same layouts the chart supports today) and a `storageclass` block; omitted fields inherit the top-level values.
- **In-place expansion of existing deployments**: pool 0 keeps the legacy StatefulSet/pod/PVC names and its selector unchanged (StatefulSet selectors are immutable), so a live single-pool install can enable pools without renames or data loss. Additional pools render as `<fullname>-poolN` with a `rustfs.com/pool` label in their selector. Overlap between pool 0's legacy selector and later pools is safe for StatefulSets because pod ownership is name-prefix based.
- `RUSTFS_VOLUMES` / `RUSTFS_SERVER_DOMAINS` enumerate every pool (space-/comma-joined respectively).
- Pod anti-affinity is **scoped per pool** (`rustfs.com/pool` match) so total pods may exceed node count while each pool still spreads its own pods across distinct nodes — without this, expansion would be unschedulable on clusters sized for one pool.
- Validation fails fast: unsupported `replicaCount`, empty `pools.list`, pools with standalone mode.
- Pools are **append-only** (list index = StatefulSet identity); documented in `values.yaml` and the README, along with the `rc admin rebalance start` follow-up step.

## Testing

**Backward compatibility (render goldens)** — `helm template` output of this branch vs `main` is byte-identical for five configurations: chart defaults, `replicaCount=16`, `mtls.enabled=true`, `obs_log_directory=""`, and a real production values file. `helm lint` passes.

**Pools rendering** — two-pool values render: two StatefulSets (`rustfs`, `rustfs-pool1`), pool 0 with the legacy selector, pool 1 with the pool label; `RUSTFS_VOLUMES` = both expressions space-joined; anti-affinity carries the pool match; per-pool PVC size overrides honored. Validation errors fire as designed.

**Live expansion test** (k0s v1.35 cluster, Longhorn storage, image `1.0.0-beta.7`):

1. Installed single-pool (`pools.enabled=false`, replicaCount 4), wrote 20 objects.
2. `helm upgrade` to `pools.enabled=true` with `[{}, {replicaCount: 4}]`.
3. Pool 0's StatefulSet was updated **in place** — same UID before and after the upgrade (`98bc222a-…`), no recreation.
4. All 8 pods across both pools converged Ready (pods crash-restart until every peer resolves — the server exits on unresolvable peers; documented as expected during expansion).
5. Server recognizes the topology:
   ```
   Pools:
     Pool 0: http://pooltest-rustfs-{0...3}.pooltest-rustfs-headless.…:9000/data/rustfs{0...3}
       Usage: 5.02 MiB / 2.56 GiB
     Pool 1: http://pooltest-rustfs-pool1-{0...3}.pooltest-rustfs-headless.…:9000/data/rustfs{0...3}
       Usage: 3.00 MiB / 5.59 GiB
   ```
6. All 20 pre-expansion objects intact and readable; 10/10 new writes succeeded post-expansion.
7. `rc admin rebalance start` ran and moved 10 objects (608 KiB) across pools. (The rebalance later reported a server-side `Io error: timeout` for one bucket — server behavior on `1.0.0-beta.7`, orthogonal to this chart change; topology, data placement and reads/writes across pools all work.)

## Notes

- No chart version bump included — `Chart.yaml` is already staged at 0.8.0 and the publish workflow derives versions from release tags; happy to adjust if you prefer a bump here.

## Related

Addresses #3057 (open question about expanding a k8s helm deployment with additional nodes — the maintainer-described `RUSTFS_VOLUMES` layout is what this PR renders, with the startup-ordering concern handled via a shared headless service and converge-through-restart, verified live).


Rohmilchkaese:
Pushed two hardening follow-ups: pool StatefulSet names are now truncated DNS-safely (base shortened, `-poolN` suffix always preserved — verified unique at a 53-char release name with 3 pools), and the README now documents that the single PodDisruptionBudget intentionally spans all pools. All five backward-compat golden renders remain byte-identical to main; lint clean.

Rohmilchkaese:
Update: live-tested in-place expansion on a real 5-node cluster (k0s, Longhorn; 4-replica pool 0 → +4-replica pool 1, `rustfs 1.0.0-beta.8`) and found — then fixed in 52ea752 — an expansion deadlock with the original `requiredDuringScheduling` anti-affinity.

**The deadlock** (reproducible whenever the cluster has fewer nodes than the total pod count):

1. The not-yet-rolled pool-0 pods still carry the unscoped *required* anti-affinity, repelling every rustfs pod from their nodes → part of pool 1 stays **Pending** (no IP → no headless-DNS record).
2. Every pod that already has the new two-pool `RUSTFS_VOLUMES` exits fatally with `failed to lookup address information` for those Pending peers.
3. The StatefulSet rolling update waits for the crashing pod to go Ready before rolling the rest of pool 0 — which keeps repelling pool 1. Circular.

Worse, the state can't self-heal after a template fix: StatefulSets assign revisions per ordinal, so Pending ordinals are only recreated with the new template after higher ordinals go Ready — recovery requires deleting the new pool's StatefulSet (PVCs survive). Shipping soft affinity from the start avoids the state entirely.

**The fix:** in pool mode, anti-affinity is now *preferred* (weight 100, still pool-scoped) — pods spread when capacity allows but expansion converges on small clusters. `pools.enabled=false` still renders the existing required anti-affinity **byte-identically** (verified by diffing full `helm template` output against `main`).

With the fix, the full expansion was re-tested end-to-end: pool 0 StatefulSet kept its UID (true in-place, selector untouched), all pods converged, pre-expansion data stayed readable, and writes work post-expansion.

One upstream observation worth a separate issue: the server exiting fatally on unresolvable peers (instead of retrying resolution) is what turns a scheduling delay into a crash loop; the chart's `publishNotReadyAddresses: true` covers crashing pods, but Pending pods have no IP to publish. A retry-on-resolve at startup would make rollouts of any topology change much more forgiving.

majinghe:
@Rohmilchkaese Thanks for your contribution. I will test it locally to make sure the change is compatible.

Rohmilchkaese:
My pleasure - its mostly down to stuff i want to see happening @majinghe

Rohmilchkaese:
Resolved the merge conflict by merging the latest `main` into the branch.

The only collision was in `helm/rustfs/templates/statefulset.yaml`: #3513 (heal service-account binding) added a `serviceAccountName` line to the pod spec on the same line where this branch had switched the template context to `$` for the per-pool `range`. I kept #3513's `serviceAccountName` and adapted it to the pool range context (`{{ include "rustfs.serviceAccountName" $ }}`).

Verified after the merge:
- Default rendering (`pools.enabled=false`) is still **byte-identical** to `main`.
- With pools enabled, every pool's StatefulSet now carries the `serviceAccountName`.
- `helm lint` passes.

majinghe:
@Rohmilchkaese Sorry to review so late, i have tested locally, pool expansion can work fine. To be honest, for pool expansion, we perfer to use our [operator](https://github.com/rustfs/operator). We will merge this PR after you resolve the confilict.

Thanks again.

majinghe (changes requested):
Please fix the confilict first.

majinghe:
@Rohmilchkaese Thanks for your contribution.

/LGTM

/cc @houseme @loverustfs

majinghe (approved):

PR #3380, opened 2026-06-12
feat: stream object zip downloads

### LINKED ISSUE #3375: Console: folder download fails on large prefixes — add server-side streaming ZIP endpoint

Downloading a "folder" (object-key prefix) from the web console fails on large prefixes. The console builds the archive entirely in the browser: for each object it fetches the bytes into a `Blob` and adds it to JSZip, then calls `zip.generateAsync({ type: "blob" })`. All source objects plus the resulting ZIP are held in browser memory at once, and browsers cap a single `Blob` at roughly 2 GiB, so the download aborts with JSZip's error:

```
Uncaught (in promise) Error: Bug : can't construct the Blob.
```

Reproduced on a 3.3 GiB / 1231-object prefix (RustFS 1.0.0-beta.8). The relevant client code is in `rustfs/console`, `components/object/list.tsx` (per-object signed URL → `Blob` → `zip.file(...)`, then `generateAsync`). MinIO's console avoided this because the ZIP was assembled and streamed by the server, so the archive size was not bounded by browser memory.

A server endpoint that, given a bucket + prefix, lists the objects and writes a ZIP archive incrementally into the HTTP response body — no in-memory accumulation and no temp files — so server memory stays flat regardless of archive size. Sketch:

- Route on the admin API, e.g. `GET /rustfs/admin/v3/download-zip?bucket=<b>&prefix=<p>&compression=stored|deflate`.
- Authorize with the existing IAM path: `s3:ListBucket` on the bucket and `s3:GetObject` on the prefix.
- Stream a ZIP64 archive (archives may exceed 4 GiB and 65535 entries); default method `Stored` (no CPU cost), optional `Deflate`.
- A client disconnect cancels listing and reading (dropping the response body stops the producer).
- Follow-up: switch the console (`rustfs/console`) from JSZip to this endpoint.

I have a working prototype that implements exactly this — an admin handler plus a streaming ZIP64 writer in the `rustfs-zip` crate, with unit and integration tests; the full `make pre-commit` passes (clippy `--all-features -D warnings`, and the workspace test suite). I'm happy to open it as a PR once the design is agreed: https://github.com/ssubbotin/rustfs/tree/feat/prefix-zip-download

**Describe alternatives you've considered**

- Keep the download client-side but stream to disk via the File System Access API (or a service worker) instead of a `Blob`. This avoids a server change but re-introduces browser-side complexity and has limited browser support.
- A dedicated route on the console server (port 9001) instead of the admin API (port 9000).

**Additional context**

A few design questions for maintainers:

- **Route placement**: admin API (`/rustfs/admin/v3`) vs a console-specific route. The console already reaches the admin API with its STS credentials, which argues for the admin API.
- **Browser auth for the download**: admin routes use header-based SigV4 only (no presigned query auth), and a browser cannot set an `Authorization` header on a plain link navigation. So the console step needs either `fetch()` + the File System Access API, or adding presigned-query auth to this route. Guidance on the preferred approach would be welcome.
- **Limits**: cap the number of objects / total size (HTTP 413) or allow any size? I lean toward no hard cap, with logging of large exports.
- **ZIP writer**: the prototype uses the `async_zip` crate for ZIP64 streaming. Open to alternatives if a new dependency is a concern.

cxymds:
## Related Issues

Fixes #3375

## Summary of Changes

- Adds signed admin API support for creating short-lived object ZIP download tokens.
- Adds browser-navigable token download URLs that stream ZIP64 archives from the server with bounded buffers.
- Expands explicit objects and prefixes, preserves relative entry names, deduplicates keys, and re-checks `GetObjectAction` authorization before streaming.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs --lib object_zip_download --no-default-features`
- `cargo test -p rustfs --lib route_policy --no-default-features`
- `cargo test -p rustfs --lib route_registration_test --no-default-features`
- `cargo test -p rustfs --lib check_access_ --no-default-features`

Full non-lib `cargo test -p rustfs object_zip_download` was attempted earlier, but this local machine ran out of disk during binary/test linking (`ld: write() failed, errno=28`). The lib-focused checks above pass and cover the new route, token, authorization, and helper behavior.

## Impact

- Adds `POST /rustfs/admin/v3/object-zip-downloads` for signed token creation.
- Adds `GET /rustfs/admin/v3/object-zip-downloads/{id}.zip?token=...` for short-lived browser downloads.
- Adds `async_zip` and `futures-lite` dependencies for bounded async ZIP streaming.
- Reduces Console/browser memory pressure for large multi-object downloads when paired with the console PR.

## Additional Notes

The in-memory token records store only hashed token values and clear credential secret/session token fields from the authorization snapshot retained for GET-time object authorization.

---


majinghe (approved):

GatewayJ:
[P1] `collect_zip_items` gathers all keys into a `Vec` before authorization, preflight, and streaming. For large prefixes, this can grow unbounded and OOM, which undermines the bounded streaming goal. Please either authorize/write items page-by-page during listing, use streaming way.

[P2] Bucket/object/listing errors are wrapped as `InternalError`, so expected user-facing cases such as missing buckets or objects can become 500 responses. Please preserve the original S3 error semantics, for example by reusing `ApiError::from` or mapping to `NoSuchBucket`, `NoSuchKey`, and `AccessDenied` as appropriate.

cxymds:
Addressed the latest review in `7b312252`.

[P1] The download path no longer collects the full listing into memory before processing. It now validates explicit items and listed prefixes page-by-page, then streams the ZIP by re-listing page-by-page while only keeping the current page in memory.

[P2] Bucket/list/object preflight calls now preserve existing S3 error semantics via `ApiError::from(...)` instead of wrapping those storage errors as `InternalError`, so cases like `NoSuchBucket`, `NoSuchKey`, and `AccessDenied` are returned correctly.

Verification run:
- `cargo +1.96.0 fmt --all --check`
- `cargo +1.96.0 test -p rustfs object_zip_download::tests:: -- --nocapture`

cxymds:
Need expert review on a layering decision exposed by CI.

cc @overtrue @houseme @GatewayJ

Current CI failure:
`dep|rustfs/src/server/layer.rs|infra->interface|crate::admin::router::ADMIN_OBJECT_ZIP_DOWNLOADS_PATH`

Context:
- This PR added token redaction for object ZIP download URLs in `rustfs/src/server/layer.rs`
- The current implementation references `ADMIN_OBJECT_ZIP_DOWNLOADS_PATH` from `admin::router`
- `Check layered dependencies` flags that as an `infra -> interface` reverse dependency

I see two reasonable fixes:
1. Move the shared object-zip-download path constant to a neutral location that both `admin::router` and `server::layer` can depend on
2. Keep `server::layer` independent and use a local constant/helper there for the redaction path match

My current preference is `2` as the smallest fix unless there is already an approved neutral home for admin path constants.

Would appreciate guidance from folks familiar with the layering guard and the `server/` vs `admin/` boundary before I patch the CI failure.

houseme (changes requested):
## Review Summary

Great feature addition overall. I like the signed short-lived ZIP download flow and the page-by-page validation/streaming model. I did find one blocking issue from current tree:

- [Major] `rustfs/src/server/layer.rs` now depends on `crate::admin::router::ADMIN_OBJECT_ZIP_DOWNLOADS_PATH` (via the import used in `redact_sensitive_uri_query`). That introduces an `infra -> interface` reverse dependency and fails the repo's layer dependency gate (`./scripts/check_layer_dependencies.sh`). Please remove this cross-layer dependency (e.g., inline a neutral local constant in `server/layer.rs` or move the constant to a lower-layer module) so the build can pass CI.

I re-ran the review-coverage checks locally: `cargo clippy --all-targets --all-features -- -D warnings`, `cargo clippy -- -D clippy::perf`, `cargo check --all-targets`, and focused tests (`object_zip_download`, `route_policy`, `route_registration_test`, `check_access_`). Everything else passed cleanly.

houseme (changes requested):
I found one blocking issue in this PR that should be fixed before merge.

### Blocking

1) [rustfs/src/server/layer.rs:17] Avoid importing admin routing constants into server layer
- Issue: `crate::admin::router::ADMIN_OBJECT_ZIP_DOWNLOADS_PATH` introduces a reverse dependency `server -> admin` (`infra->interface`), which fails the repository layer dependency guard. This can break architectural boundaries and layer-order assumptions.
- Why this matters: The project’s layer policy appears intentionally enforced by script and CI. This change already causes `./scripts/check_layer_dependencies.sh` to fail.
- Fix: Move the path constant into a lower/shared module (or inline a local static in `server/layer.rs`) and keep a single source of truth only within compatible layers.

cxymds:
Updated the PR to address the follow-up review and AI expert findings.

What changed:
- Removed the `server::layer -> admin::router` dependency and kept ZIP-download token redaction local to `server::layer`
- Extended request-log token redaction to cover both `/rustfs/admin/...` and `/minio/admin/...` object ZIP download URLs
- Reworked ZIP download generation to fully prepare the ZIP archive before returning `200 OK`
- Stopped re-listing after pre-validation; listed pages are now validated and written during archive preparation so the response streams a prepared artifact instead of a second-pass listing
- Reduced duplicate tracking to explicit object keys plus normalized listing prefixes, avoiding the previous full-list `seen` growth across streamed listing passes

Why:
- Fixes the `infra -> interface` CI failure from `crate::admin::router::ADMIN_OBJECT_ZIP_DOWNLOADS_PATH`
- Closes the `/minio/admin` token redaction gap in request logs
- Avoids the authz/content-set TOCTOU where a second listing pass could stream objects that were not part of the validated set
- Avoids returning `200 OK` before archive generation has actually succeeded

Verification run:
- `cargo +1.96.0 check -p rustfs --quiet`
- `cargo +1.96.0 fmt --all`
- `cargo +1.96.0 fmt --all --check`
- `cargo +1.96.0 test -p rustfs object_zip_download::tests:: -- --nocapture`
- `cargo +1.96.0 test -p rustfs server::layer::tests::request_log_context_redacts_object_zip_download_tokens -- --nocapture`
- `./scripts/check_layer_dependencies.sh`

GatewayJ (approved):

houseme (approved):
I re-reviewed the latest commit (`d77559b9`) with the previous P1/P2 feedback in mind.

LGTM. The ZIP download handler now prepares the full archive before returning the response, validates request scope page-by-page, and avoids the `server -> admin` reverse dependency that had failed the layer check. I don’t see remaining blocking issues introduced by this patch.

Validation I ran:
- `cargo test -p rustfs object_zip_download`
- `cargo clippy -p rustfs --tests --message-format=short`
- `./scripts/check_layer_dependencies.sh`

PR #3391, opened 2026-06-12
docs(sftp): document server operations

simon-escapecode:
## Related Issues

Follow-up to the documentation suggestion in #3372.

## Summary of Changes

Adds docs/operations/sftp.md, an operator guide for the SFTP server: recommended configuration, the path model, host keys on Unix and Windows, the environment variable reference, session cleanup behaviour, IAM policy requirements per SFTP operation, client compatibility notes, multipart upload sizing and cleanup, and the log lines worth alerting on. Every value and behaviour statement in the guide was checked against the source.

Corrects documentation the platform change left stale. The sftp module overview and the UnsupportedPlatform error text still described Windows as unsupported. A source comment referenced a document that does not exist in the repository and now points at the new guide. The read-only mode doc comment adds SETSTAT and FSETSTAT to the rejected operation list. The changelog adds the two host-key reload variables missing from its environment list, describes the banner variable as the SSH identification string, adds fallback_watchdog.rs to the file inventory, and corrects the upload size cap and the compliance case count.

## Verification

- make pre-commit passes.
- The UnsupportedPlatform message change is wording only and no test asserts the message text, so no test changed.

## Impact

Documentation and comment wording. One user-visible string changes: the UnsupportedPlatform startup error on targets that are neither Unix nor Windows.

## Additional Notes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,11 +24,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - **32 unit and integration tests** covering middleware, auth handlers, task-local storage, and role detection
 - **SFTPv3 Protocol Support**: SSH-hosted SFTPv3 subsystem that translates each file operation into S3 calls against the local object store. Authentication uses IAM credentials (SSH username = access key, SSH password = secret key).
   - Full SFTPv3 packet coverage: open, read, write, stat, lstat, fstat, mkdir, rmdir, rename, remove, opendir, readdir, realpath, close, plus the rest of the 21-packet specification
-  - Streaming multipart write up to S3's 5 TiB per-file ceiling
+  - Streaming multipart write up to the part size times 10000 parts (156.25 GiB at the default part size)
   - Per-handle read-ahead cache with configurable window size and process-wide memory ceiling
   - Per-session liveness watchdog: Linux probes `/proc/net/tcp` and cancels wedged sessions on the order of 45 seconds; non-Linux falls back to an inactivity ceiling on the order of 30 minutes
   - 30-second SSH handshake deadline, per-call backend operation timeout, bounded multipart-abort fan-out, graceful-shutdown cascade
-  - 33 SFTPv3 compliance test cases under `crates/e2e_test/src/protocols/sftp_compliance.rs` spread across three entry points: `test_sftp_compliance_suite` (shared session), `test_sftp_compliance_readonly` (read-only mode), and `test_sftp_compliance_standalone` (one rustfs spawn per case)
+  - 34 SFTPv3 compliance test cases under `crates/e2e_test/src/protocols/sftp_compliance.rs` spread across three entry points: `test_sftp_compliance_suite` (shared session), `test_sftp_compliance_readonly` (read-only mode), and `test_sftp_compliance_standalone` (one rustfs spawn per case)
   - Four-layer regression-prevention tests guard against silent feature deletion: compile-time module assertion, module-presence unit test, cross-module `Protocol` enum assertion, end-to-end SSH banner test against the running binary
 
 ### Changed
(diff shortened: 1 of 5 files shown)


houseme (approved):

houseme:
**Review findings:**

Overall this is a solid PR — well-structured guide, and most claims verified against the source. Two issues need fixing before merge:

**1. `fallback_watchdog.rs` does not exist**

The CHANGELOG adds an entry for `crates/protocols/src/sftp/fallback_watchdog.rs`, but this file does not exist in the repo. The fallback behavior lives inside `wedge_watchdog.rs` (`WedgeReason::FallbackSilence` and `WEDGE_FALLBACK_KILL_SILENCE_SECS`). Either remove the line or point it at `wedge_watchdog.rs`.

**2. Watchdog log messages are wrong**

The docs claim non-Linux platforms log `"fallback watchdog cancelling session"`. In reality (`wedge_watchdog.rs:155`), the log is **always** `"wedge watchdog cancelling session"` on **all** platforms. There is no `fallback watchdog` message anywhere in the codebase. The `fallback_silence` reason is also not Linux-specific — it fires on every platform when the 30-minute silence threshold is hit.

In the "Log Lines Worth Alerting On" table: delete the `fallback watchdog cancelling session` row, and drop "Linux only" from the `wedge watchdog cancelling session` row.

**Everything else checks out** — SETSTAT/FSETSTAT read-only rejection, 10,000 parts limit, `SSH-2.0-` banner validation, idle timeout can't be 0, reload interval silently clamped to 5s, part size 5 MiB–5 GiB, read cache 256 KiB–64 MiB, 34 compliance test cases, etc. LGTM aside from the two items above.

simon-escapecode:
> **Review findings:**
> 
> Overall this is a solid PR — well-structured guide, and most claims verified against the source. Two issues need fixing before merge:
> 
> **1. `fallback_watchdog.rs` does not exist**
> 
> The CHANGELOG adds an entry for `crates/protocols/src/sftp/fallback_watchdog.rs`, but this file does not exist in the repo. The fallback behavior lives inside `wedge_watchdog.rs` (`WedgeReason::FallbackSilence` and `WEDGE_FALLBACK_KILL_SILENCE_SECS`). Either remove the line or point it at `wedge_watchdog.rs`.
> 
> **2. Watchdog log messages are wrong**
> 
> The docs claim non-Linux platforms log `"fallback watchdog cancelling session"`. In reality (`wedge_watchdog.rs:155`), the log is **always** `"wedge watchdog cancelling session"` on **all** platforms. There is no `fallback watchdog` message anywhere in the codebase. The `fallback_silence` reason is also not Linux-specific — it fires on every platform when the 30-minute silence threshold is hit.
> 
> In the "Log Lines Worth Alerting On" table: delete the `fallback watchdog cancelling session` row, and drop "Linux only" from the `wedge watchdog cancelling session` row.
> 
> **Everything else checks out** — SETSTAT/FSETSTAT read-only rejection, 10,000 parts limit, `SSH-2.0-` banner validation, idle timeout can't be 0, reload interval silently clamped to 5s, part size 5 MiB–5 GiB, read cache 256 KiB–64 MiB, 34 compliance test cases, etc. LGTM aside from the two items above.

Thanks for the review. Both findings match the tree from before #3372 merged. That PR introduced the platform split described in the guide, and the cited line number is the tell: wedge_watchdog.rs has that log at line 155 before the merge and at line 148 on main since 7a8514bd.

crates/protocols/src/sftp/fallback_watchdog.rs exists on main since 7a8514bd:
https://github.com/rustfs/rustfs/blob/7a8514bd/crates/protocols/src/sftp/fallback_watchdog.rs
It is the non-Linux session watchdog. server.rs spawns wedge_watchdog under cfg(target_os = "linux") and fallback_watchdog under cfg(not(target_os = "linux")):
https://github.com/rustfs/rustfs/blob/7a8514bd/crates/protocols/src/sftp/server.rs#L705-L710

The message "fallback watchdog cancelling session" is logged at fallback_watchdog.rs line 59, on non-Linux platforms only, which is what the table row says. On Linux the same 30-minute silence close runs inside wedge_watchdog with reason fallback_silence and logs "wedge watchdog cancelling session", matching the wedge row. Before #3372 the fallback path did live inside wedge_watchdog.rs, which is exactly what a pre-merge checkout shows.

So the CHANGELOG line and both table rows match main at the PR base. Happy to reword anything that reads ambiguously.

PR #3392, opened 2026-06-12
fix(heal): require disk RPC before remote recovery

marshawcoco:
## Related Issues
  N/A

  ## Summary of Changes
  - Require remote disk recovery probes to pass a lightweight `disk_info(noop=true)` RPC before marking a recovered disk online.
  - Evict stale cached gRPC channels when recovery probing fails due to connection, RPC, or timeout errors.
  - Add a regression test covering a node endpoint that accepts TCP connections but is not ready to serve disk RPCs.

  ## Verification
  - `cargo test -p rustfs-ecstore --lib test_remote_disk_recovery_requires_disk_rpc_readiness -- --nocapture`
  - `cargo test -p rustfs-ecstore --lib remote_disk -- --nocapture`
  - `cargo test -p rustfs-ecstore --lib get_online_disks_with_healing_and_info -- --nocapture`
  - `cargo fmt --all --check`
  - `git diff --check`

  ## Impact
  Remote disks are no longer marked recovered just because the node port accepts TCP connections. A restarted node must have its disk RPC path ready before other nodes treat the remote disk as online again.

  ## Additional Notes
  N/A

### DIFF
--- a/crates/ecstore/src/rpc/remote_disk.rs
+++ b/crates/ecstore/src/rpc/remote_disk.rs
@@ -352,19 +352,65 @@ impl RemoteDisk {
                     return;
                 }
                 _ = interval.tick() => {
-                    if Self::perform_connectivity_check(&addr).await.is_ok() {
-                        let became_online = health.mark_recovery_success(&endpoint, "connectivity_probe_success");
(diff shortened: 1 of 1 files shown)


marshawcoco:
@codex review

PR #3394, opened 2026-06-12
feat(table-catalog): retain protected ref metadata

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Adds conservative Iceberg `refs` retention to table metadata maintenance and diagnostics.

  Changes include:
  - Treat non-current snapshots referenced by Iceberg metadata `refs` as protected snapshots.
  - Keep metadata JSON files whose current snapshot or refs point at protected snapshots out of maintenance cleanup candidates.
  - Re-apply protected refs retention during metadata-only delete, so a stale maintenance report cannot delete metadata that is now protected by the current table refs.
  - Apply the same retention logic to catalog diagnostics orphan metadata scans.
  - Add regression coverage for maintenance dry-run and diagnostics behavior.

  ## Verification

  - `cargo test -p rustfs protected_snapshot_refs --lib -j2`
  - `cargo test -p rustfs metadata_maintenance --lib -j2`
  - `cargo test -p rustfs maintenance_dry_run --lib -j2`
  - `cargo test -p rustfs orphan_metadata_scan --lib -j2`
  - `cargo test -p rustfs table_catalog --lib -j2`
  - `cargo fmt --all --check`
  - `git diff --check`

  ## Impact

  This makes metadata cleanup and diagnostics more conservative for Iceberg tables that use branch/tag refs. No public API changes are introduced, and data, manifest, and delete files are not cleaned by this change.

  ## Additional Notes

  N/A


marshawcoco:
@codex review

PR #3401, opened 2026-06-13
fix(site-replication): bidirectional sync, pre-existing data back-fill, and related fixes

### LINKED ISSUE #3398: Site replication: existing buckets/objects not synced on `replicate add`
**Describe the bug**

Site replication does not perform an initial sync of pre-existing buckets and objects when a peer is added. After `replicate add`, only data created *afterward* is propagated (via the real-time make-bucket hook). Anything that already existed on either site before the `add` stays one-sided permanently, so the two sites diverge exactly when the operator expects them to converge.

Related secondary issues found in the same flow:
- The peer `sync_state` is hard-coded to `Unknown` and is never updated, so the console "Sync" column always shows `Unknown` even when replication is working.
- `admin replicate resync status` fails with `unexpected end of JSON input` (the status endpoint returns an empty/truncated body).
- The bucket in-sync counter underflows and prints `-1/N Buckets in sync`.
- When the internal service account is desynced, `peer/remove` returns 403 and there is no in-band way to force-clean a split-brained peer.

**To Reproduce**

1. Set up two independent sites (e.g. `https://siteA:30292` and `https://siteB:30292`) with identical root credentials, both restarted.
2. On site A, create a bucket with an object before linking: `mc mb A/preexisting && mc cp file.txt A/preexisting/file.txt`.
3. Link the sites: `mc admin replicate add A B`.
4. Wait, then list site B: `mc ls B` and `mc ls B/preexisting/`.
5. See that `preexisting` (and its object) never appears on B. `mc admin replicate status A` shows it as `✗` with the opposite column empty, and the header shows `-1/N Buckets in sync`.
6. The console "Sync" column shows `Unknown`; `mc admin replicate resync status A B` errors with `unexpected end of JSON input`.

**Expected behavior**

`replicate add` should trigger an initial sync so that all pre-existing buckets, their objects, and bucket metadata converge across all sites (matching MinIO's site-replication behavior). After linking, both sites should hold the same buckets/objects regardless of which data existed before the link, and the reported sync state should reflect actual replication health rather than `Unknown`.

**Screenshots**

Console shows peers linked with Sync = "Unknown" for every site, and `replicate status` reports `-1/N Buckets in sync` while pre-existing buckets remain on a single side only. (Attaching console screenshots of the Peer Sites table and the status output.)

 - OS: TrueNAS SCALE (host); container Ubuntu-based rustfs image
 - Browser: Chrome
 - Version: RustFS `1.0.0-beta.8` (server build `2026-06-10T07:58:22Z`); tested against `main` @ `b38e71fac3300f2ac9231c958c8c0dd2003093c1`

 - N/A

**Additional context**

Deployment: 2 independent sites on separate hosts, NodePort `30292`, valid wildcard TLS, clocks in sync, identical alphanumeric root credentials, both containers restarted before `add`.

What works (for reference): real-time propagation of buckets AND objects created *after* `add` works in both directions (an object copied post-link appeared on the peer within seconds, `Replicated: 1 / Received: 1`), and policy replication reports `5/5 in sync`. So the peer channel itself is healthy; the gap is specifically the initial/back-fill sync plus the status/state reporting.

Code pointers (from reading `main`):
- `SiteReplicationAddHandler::call` (`rustfs/src/admin/handlers/site_replication.rs`) ends with `persist_site_replication_state` and returns success with no pass over existing buckets — no `list_bucket` + `start_site_bucket_resync`. The manual-resync building blocks already exist (`start_site_bucket_resync`, same file) and could be invoked here and in the receiving `SRPeerJoinHandler`.
- `sync_state` is set to `SyncStatus::Unknown` at every construction site and never reassigned from real health (`crates/madmin/src/site_replication.rs` enum; handler sites in `site_replication.rs`). Could be derived from peer reachability + `site_replication_rule_complete`.
- Resync status route returns an empty body instead of a serialised `SRResyncOpStatus`.
- In-sync tally underflows in `merge_bucket_status_info` / `site_replication_config_mismatch`; clamp with `saturating_sub`.
- `SiteReplicationRemoveHandler` only collects `peer_errors` and still returns; `rm --force` should purge local state unconditionally and an admin command should allow re-pushing the `site-replicator-0` secret.

A note that cost us debugging time (not a bug): editing a site from a console port that is not a current member (a different rustfs instance) yields a raw `peer/edit ... 403 AccessDenied` dialog; a clearer message would help.

abdullahnah92:
## Related Issues

Fixes #3398

## Summary of Changes

Fixes seven bugs in site replication, diagnosed and reproduced on a live two-site deployment. The first six are the user-visible bugs; the seventh is a root-cause hang in the resync scanner that was silently masking several of them.

- **Bidirectional replication (Fix 6):** `ensure_site_replication_bucket_replication_config` bailed out on `Ok(_)` the moment any replication config existed. When the second site joined a bucket already configured on the first, the receiving side never added the rule back toward the first site, so objects written on the second site could never replicate back (`head_object ... service error`). The function now reconciles: it loads existing rules, builds the desired config via `build_site_replication_config`, appends only the missing `site-repl-*` rules, preserves non-site-repl rules, and re-numbers priorities.

- **Back-fill pre-existing data (Fix 1):** `SiteReplicationAddHandler` and `SRPeerJoinHandler` persisted state and returned without processing buckets that existed before linking. A new `backfill_existing_buckets_after_add` helper (run on both initiator and receiver) wires versioning/targets/replication config for each local bucket, broadcasts it to peers, and kicks a resync. Errors are logged but never abort the add.

- **Resync worker hang (root cause):** `resync_bucket` waited on `self.worker_rx.resubscribe().recv()`. `resubscribe()` positions the receiver at the broadcast channel's write-head, so the bootstrap signals pre-sent in `ReplicationResyncer::new` were invisible and every spawned task blocked forever — `resync start` reported "started" but objects never moved. The broken gate is removed; concurrency is already bounded by the inner per-object mpsc worker pool. Cancellation is preserved via `cancellation_token.is_cancelled()`.

- **sync_state always "Unknown" (Fix 2):** `sync_state` was hard-coded and never updated. `build_status_info` now derives it per peer: `Enable` (reachable + complete rules), `Disable` (reachable + incomplete), `Unknown` (unreachable).

- **In-sync counter underflow (Fix 3):** `replication_cfg_mismatch` was set for deployments that have no replication config, producing `-1/N`. It is now set only for deployments that own a config.

- **resync status returned empty body (Fix 4):** the `status` operation fell through to the wildcard arm and returned no serializable body, causing client `unexpected end of JSON input`. A `status` arm now returns the stored `SRResyncOpStatus`, or an explicit `{"status":"not-found"}` when no resync exists for the peer. Verified read-only (back-to-back calls return the same `resync_id`, no side-effect).

- **No recovery for a desynced service account (Fix 5):** a new `POST /v3/site-replication/rotate-svc-acct` handler (`SRRotateServiceAccountHandler`, registered as `Sensitive`, gated by `SiteReplicationOperationAction`) generates a fresh `site-replicator-0` secret, applies it locally, and pushes `peer/join` to each peer. It is idempotent and returns `status="Partial"` listing any failed peer endpoints so the operator can re-run until `Success`.

## Verification

Built and tested with the workspace toolchain (Rust edition 2024).

- `cargo build --release --bin rustfs` — clean (exit 0)
- All 53 `site_replication` inline unit tests pass; full existing suite passes.
- **Live two-site deployment** (two `rustfs` instances, separate data dirs/ports, identical credentials):
  - Back-fill: object created before `replicate add` appears on the peer after linking — PASS
  - Bidirectional: file written on site A and file written on site B both appear on both sites — PASS
  - Offline-recover: file written while one site was down replicates after it restarts — PASS
  - Delete sync: bucket deleted on one site is removed on the other — PASS
  - `replicate status` shows `2/2 Buckets in sync` (no `-1`) and `"sync":"enable"` — PASS
  - `resync/op?operation=status` (AWS4-signed PUT): HTTP 200 + valid JSON for both the existing-status and not-found paths; confirmed read-only — PASS

Unit tests cover the logic; live signed-request tests cover the routed end-to-end path.

## Impact

- **Behavior:** site replication now converges fully — pre-existing buckets/objects back-fill on `add`, replication works in all directions, deletes propagate, and a returning offline site catches up. Previously only data created after linking, from the initiator side, replicated.
- **API:** adds one admin endpoint, `POST /v3/site-replication/rotate-svc-acct` (Sensitive, `SiteReplicationOperationAction`). No existing endpoints change shape.
- **Compatibility:** no config or on-disk state format changes. Removed dead fields (`worker_tx`/`worker_rx`) are internal only.
- **Deployment/Docs:** the rotate-svc-acct endpoint is an operator recovery tool; it is idempotent and should be re-run until `status="Success"`.

## Additional Notes

- The back-fill uses a single shared `resync_id` across all buckets/peers for one `add`; resync status is tracked per peer deployment id, so this groups the operation without collisions. Flagging for reviewer awareness.
- Fix 4's not-found branch is covered by a unit test and was also exercised live; the existing-status path is verified live and read-only.
- Diagnosis originated from a real deployment where root credentials containing shell/URL-special characters desynced the `site-replicator-0` secret across sites (403 on every peer call). That is an operator-side configuration issue, but it motivated Fix 5 (in-band repair) and is worth documenting; happy to add a docs note if useful.


abdullahnah92:
## Live two-site test results

| # | Test | Verifies | Result |
|---|------|----------|--------|
| 1 | Back-fill | pre-link.txt (before linking) appears on siteB | PASS (live) |
| 2 | Bidirectional | f1 (siteA→both), f2 (siteB→both) | PASS (live) |
| 3 | Offline-recover | file written while siteB down replicates after restart | PASS (live) |
| 4 | Delete sync | bucket deleted on siteA removed from siteB | PASS (live) |
| 5 | No -1 counter | replicate status shows 2/2 in sync | PASS (live) |
| 6 | sync_state=Enable | "sync":"enable" for all peers | PASS (live) |
| 7a | resync status valid JSON | PUT operation=status → HTTP 200 + JSON | PASS (live) |
| 7b | status read-only | same resync_id on back-to-back calls | PASS (live) |
| 7c | not-found path | fresh state → {"op":"status","status":"not-found"} | PASS (live) |
| 8 | 53 unit tests | all site_replication inline tests | PASS |
| 9 | cargo build --release | clean build | PASS (4m 33s) |

All data-movement tests run against two live rustfs instances with AWS4-signed requests, not just unit tests.

abdullahnah92:

houseme (changes requested):
I found one blocking issue in the new backfill path. `backfill_existing_buckets_after_add()` always calls `site_replication_make_bucket_hook(name, false)`, so a pre-existing bucket with object lock enabled will be recreated on peers without `lockEnabled=true`. That loses the original bucket-creation semantics for locked buckets, and object lock cannot be added later after creation. Please preserve the real lock-enabled state when broadcasting `make-with-versioning` during backfill.

abdullahnah92:
Good catch, thanks @houseme. You're right — hard-coding lockEnabled=false would recreate object-lock buckets on peers without lock, which can't be fixed after creation. 

Fixed in <commit-hash>: backfill_existing_buckets_after_add now reads the bucket's real lock_enabled flag from its metadata and passes it through to site_replication_make_bucket_hook, so the object-lock setting is preserved on peers. Falls back to false with a warning only if metadata can't be read.

houseme (commented):
The latest update addresses my earlier object-lock concern in the backfill path. I do not see any new code issues in the delta. I am holding off on approval until the new CI run finishes cleanly.

houseme (approved):

PR #3418, opened 2026-06-13
fix(ecstore): preserve bucket visibility during rebalance

cxymds:
## Related Issues
N/A

## Summary of Changes
- Preserve bucket visibility during server-pool rebalance by reducing peer bucket operations against each pool's participating peers instead of the whole cluster.
- Filter local peer bucket operations to the pools owned by that peer so stale directories from unrelated pools cannot satisfy bucket visibility or create/delete quorum.
- Require local write quorum for local bucket listing and bucket info, and add best-effort healing when `MakeBucket` observes `BucketExists`.
- Avoid writing default bucket metadata during list operations, and complete started rebalance pools that have no bucket queue so empty participants do not remain stuck as running.

## Verification
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-ecstore rpc::peer_s3_client::tests::`
- `cargo test -p rustfs-ecstore rebalance::rebalance_unit_tests::test_complete_rebalance_pools`
- `cargo test -p rustfs admin::handlers::rebalance::rebalance_handler_tests::`
- `make pre-commit`

## Impact
This affects ECStore bucket metadata operations and rebalance coordination. It should make bucket visibility and bucket creation behavior consistent across arbitrary server-pool counts, including after adding a new pool and running rebalance. No API, configuration, migration, or deployment changes are expected.

## Additional Notes
The implementation follows the MinIO-style pool-local quorum model for bucket namespace operations: each pool is evaluated against its own participating peers, while the final bucket list is merged across pools. Local disk operations are scoped to the peer's pool membership to avoid unrelated pool state influencing quorum decisions.


houseme (approved):
Looks good to me. I walked through the pool-local quorum changes, the local-disk filtering, and the rebalance completion edge cases, and the behavior lines up with the stated intent. The targeted regressions plus the full green CI give good confidence here.

PR #3420, opened 2026-06-13
fix(bucket-repl): honor op_type in replicate_object so ExistingObject…

abdullahnah92:
## Related Issues
N/A — standalone bug fix.

## Summary of Changes
Fixes a bug in bucket replication where `replicate_object` re-derived its target list with a hard-coded `op_type: Object` / `existing_object: false`, ignoring `roi.op_type`. The resync worker sets `op_type = ExistingObject` (line ~889), but `filter_target_arns` never saw it, so the check that skips targets with `ExistingObjectReplicationStatus::DISABLED` was bypassed. Existing-object resync therefore pushed objects to targets that had explicitly opted out.

The fix passes `op_type: roi.op_type` and `existing_object: roi.op_type == ExistingObject` to `filter_target_arns`. Behavior is unchanged for Object/Heal/Metadata ops; only ExistingObject ops now correctly skip DISABLED targets. This matches MinIO, which honors the queued decision rather than re-deriving with default context.

Also adds `warn!` logging at the four MRF queue-overflow sites that previously dropped entries silently (observability only).

## Verification
- cargo build, cargo fmt --check, cargo clippy --all-targets --all-features -- -D warnings — all clean
- Two new unit tests: DISABLED target excluded for ExistingObject ops, included for Heal ops
- Live two-instance test: rule with ARN-A (existing-object ENABLED) + ARN-B (DISABLED); uploaded 5 objects; triggered resync. Result: ARN-A received 5/5, ARN-B stayed at 0/5. On the pre-fix binary the same test put objects in BOTH targets, confirming the bug.

## Impact
Behavior change only for ExistingObject resync ops: targets with existing-object replication DISABLED are now correctly skipped. No change to normal PUT, Heal, or Metadata replication. No config or on-disk format changes.

## Additional Notes
N/A


abdullahnah92:

PR #3429, opened 2026-06-14
test(table-catalog): verify vended credential data-plane scope

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Adds a PyIceberg smoke profile for RustFS catalog-vended table credentials and verifies the returned credentials against the S3 data plane.

This updates the table catalog smoke helper to:

- add a `rustfs-vended-credentials` profile
- call the standard table credentials endpoint after table creation
- require returned `storage-credentials` to include S3 access key, secret key, and session token
- verify the returned credential prefix exactly matches the created table warehouse location
- run a direct S3 scope probe with the returned temporary credentials
- verify `PutObject`, `HeadObject`, and `DeleteObject` inside the table warehouse prefix
- verify same-bucket `PutObject` outside the returned prefix is rejected
- reload the PyIceberg catalog with the returned table-scoped credentials before append, reload, and scan operations
- document server-side credential vending enablement, bootstrap limits, exact-prefix validation, and data-plane scope coverage

The profile still uses the configured principal for setup and REST request signing. The returned temporary credentials are checked against the created table warehouse location, checked directly against S3, and then used for PyIceberg data-plane operations after the table exists.

## Verification

- `python3 scripts/table-catalog/test_pyiceberg_smoke.py`
- `python3 -m py_compile scripts/table-catalog/pyiceberg_smoke.py scripts/table-catalog/test_pyiceberg_smoke.py`
- `rg -n "\b(PR|PG|P)[0-9]+\b" scripts/table-catalog`
- `git diff --check HEAD~1 HEAD`

## Impact

Adds client conformance coverage and documentation for catalog-vended table credentials, including exact-prefix validation and direct S3 data-plane scope validation. No server-side API behavior changes.

## Additional Notes

N/A


marshawcoco:
@codex review

marshawcoco:
@codex review

houseme (approved):

houseme (commented):
Thanks for the update — this is a clean and solid follow-through on catalog-vended credentials.

I reviewed the 3-file diff (README + smoke script + tests) against base 61a5500232c252e0281c3d9212cf912322a658de. The new flow reads the table credentials endpoint, validates scope exactly against the created table location, probes inside/outside data-plane access, and reroutes the PyIceberg catalog to the temporary credentials only when needed. The added tests cover the critical new parsing and validation paths, and I didn’t find any blocking issues.

Verdict: No changes requested.

houseme (commented):
## Conservative Review (No Blocking Issues)

I took another pass with a conservative lens and this still looks good to land, but I want to flag a couple of potential edge-case follow-ups before merging.

1) Prefix/location normalization
- The credential-prefix validation is strict and useful, but it compares the parsed `s3://...` URIs directly (with trailing `/` normalization).
- If upstream components emit equivalent but differently encoded variants (for example path normalization/URL-encoding differences), this could produce false mismatches.
- Suggestion: consider normalizing both sides more aggressively (for example, explicit percent-decoding before comparison) if we ever hit false negatives.

2) Scope probe key cleanup behavior
- The “outside-prefix” negative probe currently treats any non-denial error as unexpected and fails, which is correct for strict validation.
- In that path, we currently attempt best-effort cleanup with static credentials.
- This is fine for now; just noting the cleanup is best-effort and can only be warned.

3) Data-plane assertion coverage
- We now validate both inside-write/head and outside-write-rejection, which is strong.
- A slightly broader assertion set (e.g., one more read/list check inside/outside) would make the probe even harder to regress, but that would be a follow-up improvement rather than a required fix.

Bottom line: no blockers from this pass. I’m keeping this as a conservative "no-block" review with non-blocking follow-up notes.

houseme (commented):
### Conservative Review (No Blocking)

Looks good overall and aligns with the PR goal. I don’t see any blocking issues.

A few non-blocking follow-ups:
- URL/URI prefix matching may be strict across equivalent encodings; consider extra normalization only if false negatives appear.
- Outside-scope negative probe cleanup is best-effort, which is acceptable for smoke validation.
- Data-plane check set is solid; one extra inside/outside read/list assertion would further harden it, but this is optional.

Verdict: approve to land, with these as optional follow-ups.

marshawcoco:
Thanks for the careful follow-up! I agree these are good hardening points. I will keep this PR focused and address the suggested follow-ups in a later change:
  - stronger URI/prefix normalization if equivalent encodings become relevant
  - broader data-plane assertions, especially additional read/list checks inside and outside the scoped prefix
  - keeping the cleanup path best-effort but making the behavior clearer in the smoke coverage

PR #3430, opened 2026-06-14
refactor(logging): normalize admin telemetry and error messages

houseme:
## Related Issues
N/A

## Summary of Changes
This PR consolidates the recent RustFS logging-governance work across startup, protocol, notify/audit/targets/obs, admin control-plane handlers, and follow-up admin API error-response normalization.

It standardizes structured logging around stable `event + component + subsystem + state/result` fields, reduces noisy sentence-style success logs on hot and repetitive paths, expands the logging guardrails script to prevent regressions, and normalizes outward-facing admin `s3_error!` messages in the highest-value handlers.

## Verification
- `cargo fmt --all`
- `cargo check -p rustfs`

## Impact
Admin/control-plane diagnostics become more consistent and easier to scan, while customer-facing admin error messages become more uniform across handlers. Logging guardrail coverage is expanded, which may fail CI if retired legacy log patterns are reintroduced. No intentional API contract changes beyond wording normalization in admin error responses.

## Additional Notes
- `make pre-commit` was intentionally not run per request.
- `origin/main` has already been merged into this branch.
- `target_descriptor.rs` was handled as part of the error-response normalization track rather than the logging-governance track.


rustfs/src/admin/handlers/oidc.rs
```diff
@@ -500,21 +519,42 @@ impl Operation for OidcCallbackHandler {
...
             })?;
 
         info!(
-            "OIDC login successful: username='{}', email='{}', sub='{}' (provider: {})",
-            claims.username, claims.email, claims.sub, actual_provider_id
+            event = EVENT_ADMIN_OIDC_STATE,
+            component = LOG_COMPONENT_ADMIN,
+            subsystem = LOG_SUBSYSTEM_OIDC,
+            provider_id = %actual_provider_id,
+            username = %claims.username,
```
cxymds: [P1] This logs `username`, `email`, and `sub` as structured info fields on successful OIDC auth. Since this PR is supposed to normalize logging rather than expand telemetry exposure, I do not think we should emit user-identifying OIDC claims here. Please keep only operational fields such as `provider_id`, `state`, and aggregate counts, or use a redacted/hashed identifier if correlation is required.

crates/targets/src/target/webhook.rs
```diff
@@ -357,14 +393,17 @@ where
...
-        info!("Webhook sending queued payload to target: {}", self.id);
         debug!(
-            target = %self.id,
+            event = EVENT_WEBHOOK_DELIVERY_STATE,
+            component = LOG_COMPONENT_TARGETS,
+            subsystem = LOG_SUBSYSTEM_WEBHOOK,
+            target_id = %self.id,
             bucket = %meta.bucket_name,
             object = %meta.object_name,
             event = %meta.event_name,
```
cxymds: [P2] These logs assign `event` twice in the same tracing call: once for the stable log event key and again for `meta.event_name`. That makes the structured `event` field ambiguous and can break queries/alerts built on the new logging contract. Please keep `event` for the stable log identifier and move the payload value to something like `payload_event` or `notification_event`.

cxymds (approved):

houseme:
Applied. I removed the user-identifying OIDC claim fields from the success-path logs and kept only operational fields such as `provider_id`, `state`, and aggregate counts.

houseme:
Applied. I kept `event` for the stable log event identifier and renamed the payload field to `payload_event` to avoid ambiguity in structured queries.

cxymds (approved):

PR #3431, opened 2026-06-14
fix(policy): preserve IAM policy readback shape

GatewayJ:
## Related Issues
N/A

## Summary of Changes
- Preserve submitted Action/Resource ordering when IAM policies are read back.
- Deduplicate ActionSet/ResourceSet values through private push_unique helpers so legacy repeated values remain normalized without losing first-seen order.
- Normalize empty effective service-account policy versions to the default IAM policy version before serializing admin responses.
- Add regression coverage for ordering, optional fields, empty resource omission, and JSON round-trip compatibility.

## Verification
- `make pre-commit`
- `cargo fmt --all --check`

## Impact
IAM policy readback becomes deterministic with submitted order while preserving legacy JSON compatibility for single-string and array Action/Resource forms. No deployment or configuration changes.

## Additional Notes
N/A

### DIFF
--- a/crates/policy/src/policy/action.rs
+++ b/crates/policy/src/policy/action.rs
@@ -17,7 +17,7 @@ use serde::{
     Deserialize, Deserializer, Serialize,
     de::{self, Error as DeError, Visitor},
 };
-use std::{collections::HashSet, fmt, ops::Deref};
+use std::{fmt, ops::Deref};
 use strum::{EnumString, IntoStaticStr};
 
 use super::{Error as IamError, Validator, utils::wildcard};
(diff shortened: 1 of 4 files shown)


houseme (commented):
## Review Summary
The PR is in good shape: it fixes IAM policy serialization determinism, preserves input order, and updates service account policy defaulting to avoid empty versions in responses. I did not find correctness-breaking issues.

## Issues
### Minor (Performance)

1. `crates/policy/src/policy/action.rs:60` Linear ActionSet lookup can add O(n) work in policy hot path
- `ActionSet::is_match` now iterates a `Vec` every time it is called.
- This is on the request matching path (`Statement::request_reaches_condition_eval`), so large policies can add noticeable latency versus the prior hashed lookup.
- Consider a dual representation (ordered `Vec` for serialization + internal hash/set cache for membership checks) if policy cardinality can grow.

2. `crates/policy/src/policy/resource.rs:63` Linear ResourceSet lookup has similar hot-path impact
- `ResourceSet::is_match` now scans the ordered list on every request, which increases matching cost with more resource patterns.
- A cached fast-membership structure would keep deterministic serialization while restoring near O(1) checks.

## Verdict
Ready: Yes (minor performance items only)

PR #3436, opened 2026-06-14
feat(table-catalog): bridge table data-plane policy

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Bridges table-aware authorization into the S3 object data plane for registered table warehouse objects.

This change adds:

- table warehouse object prefix resolution back to the owning table resource
- an additional table resource authorization gate after normal S3 IAM or bucket policy authorization
- read-like S3 object operations mapped to table metadata read authorization
- write/delete/MPU/object-metadata mutation operations mapped to table metadata write authorization
- anonymous bucket-policy access rejection for registered table warehouse objects
- table resource actions in vended table credential session policies, so scoped temporary credentials can pass the new data-plane gate
- negative coverage for overlapping warehouse prefixes, ensuring the most specific registered table prefix wins
- negative coverage that scoped table credentials do not grant unrelated catalog mutation actions
- committed Iceberg metadata `location` validation that keeps the persisted table warehouse location in sync after table relocation
- tolerant data-plane resolution for legacy or invalid persisted warehouse locations, so one bad entry cannot deny unrelated objects or hide a later valid table match
- vended credential smoke hardening with canonical S3 URI prefix normalization, percent-decoding, inside read probes, seeded outside read-denial probes, and outside write-denial probes

The normal S3 IAM/bucket policy path still runs first. The table-aware check only applies when the target object resolves to an active table warehouse prefix in a table-enabled bucket.

## Verification

- `python3 scripts/table-catalog/test_pyiceberg_smoke.py`
- `python3 -m py_compile scripts/table-catalog/pyiceberg_smoke.py scripts/table-catalog/test_pyiceberg_smoke.py`
- `cargo fmt --all --check`
- `cargo test -p rustfs --lib table_data_plane -- --nocapture`
- `cargo test -p rustfs --lib table_credential_session_policy -- --nocapture`
- `cargo test -p rustfs --lib object_table_catalog_store_syncs_warehouse_location_from_committed_metadata -- --nocapture`
- `cargo test -p rustfs --lib table_data_plane_resource_skips_invalid_warehouse_locations -- --nocapture`
- `git diff --check`

Also checked the smoke scripts for internal roadmap-label references.

## Impact

Table warehouse objects now require both the ordinary S3 authorization decision and the corresponding table resource permission. This closes the gap where a broad bucket/object policy could otherwise access registered table data without table-aware authorization.

When a commit moves a table warehouse through Iceberg metadata, the catalog store validates that the new location stays inside the table bucket and persists the updated warehouse location before later data-plane lookups use it.

The data-plane resolver skips invalid legacy table warehouse entries during lookup. Those entries do not grant table access, and they no longer turn unrelated object requests into table-catalog access denials.

The PyIceberg smoke helper now performs stronger vended credential scope validation, including equivalent URI encoding normalization and read/write denial coverage outside the scoped prefix.

## Additional Notes

The table warehouse resolver currently scans active catalog namespaces and tables for the matching prefix. That keeps this change scoped to the existing catalog store abstraction; a reverse index or cache can optimize this later without changing the authorization contract.

### DIFF
--- a/rustfs/src/admin/handlers/table_catalog.rs
+++ b/rustfs/src/admin/handlers/table_catalog.rs
@@ -386,7 +386,7 @@ impl TableCredentialIssuer for IamTableCredentialIssuer {
             ));
         }
 
-        let policy = table_credential_session_policy(&request.entry.table_bucket, &request.object_prefix)?;
+        let policy = table_credential_session_policy(request.entry, &request.object_prefix)?;
         let policy_buf = serde_json::to_vec(&policy)
             .map_err(|err| s3_error!(InternalError, "failed to serialize table credential session policy: {}", err))?;
         let expiration = OffsetDateTime::now_utc().saturating_add(Duration::seconds(self.ttl_seconds));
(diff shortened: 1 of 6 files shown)


marshawcoco:
@codex review

marshawcoco:
@codex review

Issue #3421, opened 2026-06-13
Bucket replication: failed object replications are lost on restart (MRF persistence is a stub)


abdullahnah92:
**Describe the bug**
Bucket replication does not durably retry failed object replications. If a replication target is unreachable long enough for in-memory retries to exhaust, the affected objects are lost on process restart and are never replicated automatically. The root cause is that the MRF (failed-replication retry) persistence layer is unimplemented.

Specifically, in `crates/ecstore/src/bucket/replication/replication_pool.rs`:
- `start_mrf_persister` and `start_mrf_processor` (~line 756) are empty stubs. The `mrf_save_tx`/`mrf_save_rx` channel (100k capacity) is created and written to on every failed/overflowed replication, but `mrf_save_rx` is never drained or persisted to disk, so entries are lost on restart.
- `resize_failed_workers` (~line 459) only ever spawns 1 MRF retry worker because the mpsc receiver can only be taken once; subsequent increments inflate the counter without spawning workers.
- `load_resync` (~line 983) has no leader lock (explicit TODO), so every node in a cluster runs resync simultaneously on startup, multiplying target load by N.

**To Reproduce**
Steps to reproduce (two-instance setup, source bucket replicating to a target):
1. Configure a bucket replication rule from instance-1 (source) to instance-2 (target).
2. Stop instance-2 (the target).
3. While instance-2 is down, upload several new objects to the source bucket on instance-1.
4. Restart instance-2 and wait.
5. Observe: some/all of the objects uploaded while the target was down do not replicate automatically. A manual `mc replicate resync start` is required to recover them.

**Expected behavior**
Objects that fail to replicate while a target is temporarily unreachable should be persisted to disk and automatically retried once the target recovers, surviving a process restart — without requiring a manual resync.

**Additional context**
Observed behavior from a live two-instance test:

| Condition | Behavior |
|---|---|
| Target unreachable briefly (seconds) | Objects still replicate (in-memory queue survives) |
| Target down long enough for retries to exhaust | Objects NOT retried after target returns |
| Manual resync after target recovers | Reliably recovers missed objects |

So manual resync is currently the only reliable recovery path; durable automatic retry does not work.

Questions before contributing a fix:
1. Is MRF persistence intended to be implemented, or is there a different planned recovery design?
2. Would you accept a PR implementing disk-backed MRF persistence + reload-on-startup (and fixing the single-worker and leader-lock issues)?
3. Any preferences on the on-disk format / location?

[...]

PR #3455, opened 2026-06-14
fix(scanner): skip startup delay when replication is active so failed…

abdullahnah92:
## Related Issues
Relates to the bucket-replication recovery issue (objects not re-replicated after a source-node restart).
- https://github.com/rustfs/rustfs/issues/3421

## Summary of Changes
The scanner applies a randomized 27–33 min startup delay to spread load across nodes, but it fired unconditionally on every restart — so after a node restart the scanner would not re-scan FAILED-status objects for up to 33 minutes. Replication failures that occurred while a target was unreachable therefore went unhealed for a long time.

This change skips the startup delay when replication is active (at least one bucket has an active replication rule at startup), so the scanner runs its first cycle within ~7s of restart, reads xl.meta, sees replication_status=FAILED, and re-queues those objects for replication.

## Verification
- `cargo fmt --check`, `cargo clippy -p rustfs-scanner -- -D warnings`, `cargo test -p rustfs-scanner` — all clean; new unit tests for the delay-skip logic.
- Live two-instance test: target down, 5 objects fail replication, SIGKILL the source, restart it. Before this fix: 0/5 healed within the window (scanner asleep). After: 5/5 auto-healed within seconds, no manual resync.

## Impact
Only affects startup timing on nodes with active replication: the first scan runs immediately instead of after a 27–33 min delay. No other behavior change, no config or format changes.

## Additional Notes
A complementary MRF-persistence PR follows as a companion.

### DIFF
--- a/crates/scanner/src/scanner.rs
+++ b/crates/scanner/src/scanner.rs
@@ -140,8 +140,19 @@ fn initial_scanner_delay_for(start_delay_secs: Option<u64>) -> Duration {
         .unwrap_or_else(randomized_cycle_delay)
 }
 
-fn initial_scanner_delay_for_startup(start_delay_secs: Option<u64>, usage_cache_is_cold: bool, has_buckets: bool) -> Duration {
-    if usage_cache_is_cold && has_buckets {
+fn initial_scanner_delay_for_startup(
+    start_delay_secs: Option<u64>,
+    usage_cache_is_cold: bool,
+    has_buckets: bool,
+    has_active_replication: bool,
+) -> Duration {
+    // Skip the startup delay when the cache is cold (first ever scan) OR when active replication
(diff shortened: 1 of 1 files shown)


abdullahnah92:

houseme (commented):
## Review Summary
Looks good overall. I reviewed the diff plus related startup path and ran focused checks (`cargo check -p rustfs-scanner --all-targets`, `cargo clippy -p rustfs-scanner --all-targets --all-features -- -D warnings`); no blocking issues were found.

### Findings
None.

### Good Patterns
- Replication-aware startup-delay bypass is tightly scoped and reuses the already-cached maintenance-feature detection result.
- Added test coverage for both positive and negative paths introduced by `initial_scanner_delay_for_startup`.

### Verdict
Looks good to me.

PR #3456, opened 2026-06-14
fix(bucket-repl): persist MRF retry queue to disk and reload on startup

abdullahnah92:
## Related Issues
Relates to the bucket-replication recovery issue — durability of in-flight replication retries.
- https://github.com/rustfs/rustfs/issues/3421

## Summary of Changes
`start_mrf_persister` and `start_mrf_processor` were empty stubs: the MRF (failed-replication retry) channel was written to but never drained or persisted, so retries in-flight at crash time were lost on restart.

This implements disk persistence (`mrf.bin`, same header + msgpack format as `resync.bin`): the persister drains the channel and flushes every 10s or every 1000 entries; the processor reloads entries on startup and re-queues them. It also fixes `resize_failed_workers` (which only ever spawned 1 worker because the mpsc receiver was taken once), and adds a leader-lock plus a per-object ns-lock to prevent duplicate cross-node replication.

## Verification
- `cargo fmt --check`, `cargo clippy -p rustfs-ecstore -- -D warnings`, `cargo test -p rustfs-ecstore` — all clean; new encode/decode roundtrip tests.
- Note: this is a safety-net for the narrow window where a retry is in-flight in memory at crash time. The primary restart-recovery fix is the separate scanner startup-delay PR.

## Impact
No config or format changes. `mrf.bin` is written under the existing meta prefix. Behavior change only on crash/restart recovery.

## Additional Notes
Companion to the scanner startup-delay PR.

### DIFF
--- a/crates/ecstore/src/bucket/replication/replication_pool.rs
+++ b/crates/ecstore/src/bucket/replication/replication_pool.rs
@@ -1732,7 +1732,7 @@ mod tests {
             delete_marker: false,
         };
 
-        let encoded = encode_mrf_file(&[entry.clone()]).expect("encode");
+        let encoded = encode_mrf_file(std::slice::from_ref(&entry)).expect("encode");
         let decoded = decode_mrf_file(&encoded).expect("decode");
 
         assert_eq!(decoded.len(), 1);
@@ -1761,7 +1761,7 @@ mod tests {
             delete_marker: true,
         };
 
-        let encoded = encode_mrf_file(&[entry.clone()]).expect("encode");
+        let encoded = encode_mrf_file(std::slice::from_ref(&entry)).expect("encode");
         let decoded = decode_mrf_file(&encoded).expect("decode");
 
         assert_eq!(decoded.len(), 1);
--- a/crates/ecstore/src/bucket/replication/replication_pool.rs
+++ b/crates/ecstore/src/bucket/replication/replication_pool.rs
@@ -20,14 +20,15 @@ use crate::bucket::replication::ResyncStatusType;
 use crate::bucket::replication::replicate_delete;
 use crate::bucket::replication::replicate_object;
 use crate::bucket::replication::replication_resyncer::{
-    BucketReplicationResyncStatus, DeletedObjectReplicationInfo, REPLICATION_DIR, RESYNC_FILE_NAME, ReplicationConfig,
-    ReplicationResyncer, TargetReplicationResyncStatus, decode_resync_file, get_heal_replicate_object_info, save_resync_status,
+    BucketReplicationResyncStatus, DeletedObjectReplicationInfo, MRF_REPLICATION_FILE, REPLICATION_DIR, RESYNC_FILE_NAME,
+    ReplicationConfig, ReplicationResyncer, TargetReplicationResyncStatus, decode_mrf_file, decode_resync_file, encode_mrf_file,
+    get_heal_replicate_object_info, save_resync_status,
 };
 use crate::bucket::replication::replication_state::ReplicationStats;
-use crate::config::com::read_config;
+use crate::config::com::{read_config, save_config};
 use crate::disk::BUCKET_META_PREFIX;
 use crate::error::Error as EcstoreError;
-use crate::store_api::{NamespaceLocking, ObjectIO, ObjectInfo};
+use crate::store_api::{NamespaceLocking, ObjectIO, ObjectInfo, ObjectOptions};
 use lazy_static::lazy_static;
 use rustfs_filemeta::MrfReplicateEntry;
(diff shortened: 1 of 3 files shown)


abdullahnah92:

houseme (changes requested):
## Review summary
Thanks for pushing this — wiring MRF persistence + startup replay is very solid progress, and it clearly addresses the old stubbed behavior from #3421.

I found a few correctness blockers we should fix before merge:

### Major (blocking)
1) MRF replay loses delete operations
- File/areas: `crates/ecstore/src/bucket/replication/replication_pool.rs` (`start_mrf_processor`) and `crates/filemeta/src/replication.rs` (`MrfReplicateEntry`)
- Symptom: persisted entries are replayed as object heal tasks (`queue_replication_heal`) unconditionally.
- Why it blocks: both object and delete failures are persisted via `to_mrf_entry`, but `MrfReplicateEntry` only stores bucket/object/retry_count/version_id/size, with no operation discriminator. `to_mrf_entry` for `DeletedObjectReplicationInfo` also writes `version_id = None`.
- Impact: delete-version failures can be dropped on restart or become no-ops when object lookup fails.
- Suggested fix: persist op kind in `MrfReplicateEntry` and branch replay to `replicate_delete`/heal object accordingly.

2) Flush-on-disk can lose queued entries on flush error
- File: `crates/ecstore/src/bucket/replication/replication_pool.rs` (`start_mrf_persister` + `flush_mrf_to_disk`)
- Symptom: `pending.clear()` runs even when `flush_mrf_to_disk` fails to encode or write.
- Why it blocks: one transient storage error at flush boundary drops the current batch permanently, violating “durable retry” intent.
- Suggested fix: only clear the in-memory batch after successful write, or re-enqueue/retain failed payload for retry.

### Notes
- The PR still doesn’t add tests for encode/decode + recovery replay behavior. Given these are failure-path persistence changes, a focused test (or test-only fixture) is needed to lock in these semantics.

Also, this PR aligns with part of #3421 but does not cover all reported root causes (heal re-replication path and startup delay were still being investigated there), so we should treat that issue as not fully closed yet.

abdullahnah92:
Thanks for the thorough review — all three were valid. Addressed in the latest push:

1. **Delete ops in replay**: added an `MrfOpKind` discriminator (Object/Delete, defaulting to Object for backward compatibility with existing files) plus the delete-specific fields. `to_mrf_entry` for deletes now persists the real version_id and delete_marker_version_id (previously lost as None). `start_mrf_processor` branches: delete entries replay via `schedule_replication_delete` (skipping `get_object_info`, which would fail for already-removed markers), object entries via the heal path.

2. **Flush losing entries on error**: `flush_mrf_to_disk` now returns success, and the in-memory batch is cleared only on a successful write — a transient storage error retains the batch for the next tick.

3. **Tests**: added 6 focused tests covering object/delete/versioned-delete roundtrips, a mixed batch, op-routing, and legacy files without the op field decoding as Object.

On the issue scope: agreed, this PR is one piece — I've opened a separate PR for the scanner startup-delay fix (the main restart-recovery path), and the issue should stay open until both land and the remaining paths are confirmed.

houseme (approved):
Great progress on this PR. I re-ran against the latest commit and the core blockers are now fixed:  is persisted and replayed correctly, delete-queue MRF entries now round-trip their version/delete-marker metadata, and the persister only clears pending entries on successful flush to avoid losing retries across restart.

I also validated a focused test:  passed. 

One small follow-up: this is non-blocking, but  currently constructs the test payload via current , so it doesn’t actually prove decode-compatibility with truly legacy no- bytes. Overall, no blocking issues remain from the previous review.

houseme (approved):
Great progress on this PR. I re-ran against the latest commit and the core blockers are now fixed: MrfOpKind is persisted and replayed correctly, delete-queue MRF entries now carry version/delete-marker metadata, and the persister keeps pending entries unless flush succeeds.

I validated a focused test: mrf_entry_object_roundtrip passed.

One small follow-up: this is non-blocking, but mrf_legacy_file_without_op_field_decoded_as_object currently uses current encode_mrf_file, so it does not truly prove legacy no-op-field bytes.
Overall, no blocking issues remain from the previous review.

houseme:
@abdullahnah92 please fix ci.

abdullahnah92:
fixed

abdullahnah92:
Thanks for the careful review. Addressed the follow-up: mrf_legacy_file_without_op_field_decoded_as_object now constructs a hand-built msgpack payload that genuinely omits the op field (mimicking pre-MrfOpKind files), rather than round-tripping the current encoder. It asserts the entry decodes with op defaulting to Object, proving real backward compatibility.

PR #3462, opened 2026-06-15
feat(table-catalog): add maintenance reachability reports

marshawcoco:
## Related Issues

  N/A

  ## Summary of Changes

  Adds safer and more observable table metadata maintenance execution.

  - Adds effective maintenance config resolution with table bucket defaults and table-level overrides.
  - Persists maintenance job context, including config source, worker ID, lease ID, heartbeat, started time, finished time, status, and failure reason.
  - Adds stable `latest` and `current` maintenance job aliases for job lookup.
  - Records disabled delete requests as failed maintenance jobs without deleting objects.
  - Keeps metadata cleanup conservative with current pointer re-checks and existing retained/deletable object explanations.
  - Reports manifest-list references as manual-review referenced objects when manifest Avro reachability parsing is not supported, keeping them out of delete candidates.
  - Updates the admin maintenance helper to use the unified maintenance run path.

  ## Verification

  - `cargo fmt --all --check`
  - `git diff --check`
  - `cargo test -p rustfs --lib maintenance_ -- --nocapture`
  - `cargo test -p rustfs --lib test_admin_route_matrix_matches_registered_routes -- --nocapture`
  - `cargo test -p rustfs --lib route_policy_ -- --nocapture`

  ## Impact

  Adds additive maintenance job/report fields and stable latest/current job aliases. Metadata delete now requires both an explicit delete request and enabled maintenance delete configuration; disabled delete attempts return a persisted failed job report instead of deleting objects.

  ## Additional Notes

  Manifest-list reachability is reported fail-closed until full manifest Avro parsing and data/delete-file reachability cleanup are implemented.

### DIFF
--- a/rustfs/src/admin/handlers/table_catalog.rs
+++ b/rustfs/src/admin/handlers/table_catalog.rs
@@ -2464,27 +2464,17 @@ async fn table_metadata_maintenance_response<B>(
 where
     B: crate::table_catalog::TableCatalogObjectBackend,
 {
-    let report = if request.delete {
-        store
-            .delete_table_metadata_maintenance_candidates(
-                bucket,
-                &namespace.public_name(),
-                table,
-                request.retain_recent_metadata_files,
-            )
-            .await
-            .map_err(catalog_store_error)
-    } else {
-        store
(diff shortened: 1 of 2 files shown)


marshawcoco:
@codex review

PR #3488, opened 2026-06-15
fix(admin): reject access keys and names containing any whitespace

AdnanHussainTurki:
## Summary
`has_space_be` only stripped leading/trailing whitespace (`s.trim().len() != s.len()`), so an access key or name with an **internal** space (e.g. `my key`) passed the check, then failed deeper with a generic error instead of the intended message. Creating an access key named `my key` in the console therefore returned a generic failure rather than "access key has spaces".

## Fix
Detect **any** whitespace character (`s.chars().any(char::is_whitespace)`), so the existing specific messages fire correctly. The helper guards access keys plus user, group, and policy names, so all of them now reject internal whitespace consistently.

## Tests
Adds regression tests covering internal, leading, trailing, tab, and clean-identifier cases.

## Verification
- `cargo test -p rustfs --lib admin::utils` — green
- `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings` — clean
- Manually: creating an access key with a space now returns `InvalidRequest: access key has spaces` (HTTP 400).

## Agreement

image

### DIFF
--- a/rustfs/src/admin/utils.rs
+++ b/rustfs/src/admin/utils.rs
@@ -16,8 +16,13 @@ use crate::server::{MINIO_ADMIN_PREFIX, has_path_prefix};
 use rustfs_crypto::{decrypt_data, decrypt_stream_io, encrypt_stream_io};
 use s3s::{Body, S3Result, s3_error};
 
+/// Returns `true` if `s` contains any whitespace character.
+///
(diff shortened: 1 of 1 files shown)


AdnanHussainTurki:

PR #3515, opened 2026-06-17
fix(storage): harden rebalance decommission state

cxymds:
## Related Issues
N/A

## Summary of Changes
- Move remote-tiered rebalance versions by preserving metadata instead of treating them as skipped transfers.
- Harden rebalance start/stop rollback coordination, including expected-id mismatch handling and terminal metadata reload propagation.
- Make decommission start and terminal state updates retry-safe around pool metadata save failures.
- Tighten data-movement delete-marker resume equivalence across user metadata, tags, replication, and purge state.
- Keep the rebalance/decommission fixes compatible with the latest storage compat facade split from `main`.

## Verification
- `make pre-commit` (guardrails and clippy/check phase passed; test phase stopped locally with `No space left on device` while writing `target/debug` artifacts)
- `cargo fmt --all --check`
- `git diff --check`
- `git diff --cached --check`
- `cargo check -p rustfs`
- `cargo test -p rustfs-ecstore rebalance --lib`
- `cargo test -p rustfs-ecstore decommission --lib`
- `cargo test -p rustfs rebalance_handler_tests --lib`

## Impact
- Improves correctness for rebalance/decommission coordination and remote-tiered version movement.
- No API, configuration, or deployment changes expected.

## Additional Notes
- Synced with the latest `main`; the PR branch is mergeable and GitHub CI is running.
- Local build artifacts were cleaned after verification to recover disk space.


houseme (changes requested):
I dug through the rebalance/decommission changes pretty carefully and also sanity-checked the new delete-marker / pool-selection behavior against MinIO's current implementation. The delete-marker migration coverage looks good, and the decommission compatibility notes are helpful.

I still see one blocking issue before this is safe to merge:

- `crates/ecstore/src/rebalance/control.rs`: `init_and_start_rebalance()` now serializes starts with `start_gate`, but it does not re-check the current rebalance state after acquiring that gate. Two start requests that both pass the pre-checks can still race: the first request starts rebalance, then the second request enters the gate, calls `init_rebalance_meta()` again, replaces the in-memory `rebalance_meta`, and can start a second worker even though the cluster already has an active rebalance. That is especially risky here because `merge_rebalance_meta()` keeps the persisted remote metadata when the IDs differ, so the second request can leave local in-memory state out of sync with the saved cluster metadata.

I’d fix this by re-validating rebalance/decommission state inside `init_and_start_rebalance()` after taking `start_gate` (before calling `init_rebalance_meta()`), and by adding a regression test for two concurrent start requests hitting the same node.

cxymds:
Addressed the blocking start-gate race and synced the branch with the latest `main`.

- Re-checks decommission and active rebalance state inside `init_and_start_rebalance()` after acquiring `start_gate`, before initializing replacement metadata.
- Returns `RebalanceAlreadyRunning` when a second start sees an active or stopping rebalance.
- Added regression coverage for the second-start-after-gate path plus validator edge cases.
- Resolved the latest compat facade conflicts while preserving the rebalance stop-propagation re-exports.

Verification:
- `cargo fmt --all --check`
- `git diff --check`
- `git diff --cached --check`
- `cargo test -p rustfs-ecstore rebalance --lib`
- `cargo test -p rustfs rebalance_handler_tests --lib`

cxymds:
@houseme review response for https://github.com/rustfs/rustfs/pull/3515#pullrequestreview-4540780740

Fixed the blocking `init_and_start_rebalance()` race in `2c6d8ce2`.

- Re-validates decommission state and current rebalance metadata after acquiring `start_gate`.
- Performs that check before `init_rebalance_meta()`, so a second start request cannot replace in-memory `rebalance_meta` while an active or stopping rebalance is already present.
- Returns `RebalanceAlreadyRunning` for the second start path.
- Added regression coverage for `init_and_start_rebalance()` rejecting the second start before metadata initialization, plus validator edge cases for active, terminal, missing, and decommission-running states.

Verification:
- `cargo fmt --all --check`
- `git diff --check`
- `git diff --cached --check`
- `cargo test -p rustfs-ecstore rebalance --lib` (217 passed)
- `cargo test -p rustfs rebalance_handler_tests --lib` (29 passed)

There are no inline review threads on this PR to resolve, so this comment is the explicit response to the blocking review summary.

PR #3516, opened 2026-06-17
fix(scanner): restore prompt bucket usage scans

### LINKED ISSUE #3496: Bug: rustfs_bucket_api_usage_bytes and objects_total metrics always return 0 since 1.0.0-beta.7
**Describe the bug**
The Prometheus metrics `rustfs_bucket_api_usage_bytes` and `rustfs_bucket_api_objects_total` always return `0` in versions `1.0.0-beta.7` and `1.0.0-beta.8`.

These metrics were working as expected in version `1.0.0-beta.5` (version `1.0.0-beta.6` had different parameter names). This indicates a regression introduced after `beta.5` or `beta.6`.

**To Reproduce**

1. Start Fluent-Bit (v5.0) to collect metrics using the following command:
```bash
nerdctl run -d --name fluent-bit -p 4318:4318 -p 2021:2021 -v ./fluent-bit.yaml:/fluent-bit/etc/fluent-bit.yaml ghcr.io/fluent/fluent-bit:5.0 -c /fluent-bit/etc/fluent-bit.yaml

```

2. Start the `rustfs` container using version `1.0.0-beta.8` (or `beta.7`) with OpenTelemetry metrics enabled:
```bash
nerdctl run -d --name rustfs_container -p 9000:9000 -p 9001:9001 -e RUSTFS_ACCESS_KEY=rustfsadmin -e RUSTFS_SECRET_KEY=rustfsadmin -e RUSTFS_CONSOLE_ENABLE=true -e RUSTFS_OBS_METRIC_ENDPOINT="http://<fluent-bit-ip>:4318/v1/metrics" swr.cn-north-4.myhuaweicloud.com/ddn-k8s/docker.io/rustfs/rustfs:1.0.0-beta.8 --address :9000 --console-enable /data

```

3. Upload some objects and generate bucket usage.
4. Check the reported metrics.
5. See error: Both `rustfs_bucket_api_usage_bytes` and `rustfs_bucket_api_objects_total` remain `0`.

**Expected behavior**
The metrics should correctly reflect the actual bucket usage bytes and total object count, as they did in version `1.0.0-beta.5`.

**Additional context**
Here is the version history and behavior for these two metrics:

* **1.0.0-beta.5**: Works fine (returns normal and correct data).
* **1.0.0-beta.6**: Metrics names/parameters were changed.
* **1.0.0-beta.7**: Returns `0`.
* **1.0.0-beta.8**: Returns `0`.

*(Note: Please verify the image tag `1.0.0-alpha.99` used in the start command, as the issue happens on the `beta` track).*

marshawcoco:
## Related Issues
Fixes #3496.

## Summary of Changes
- Remove the single-disk scanner default that stretched normal scans to a 24h cycle.
- Treat a cold persisted usage cache as enough reason to skip scanner startup delay, even when no buckets existed at startup.
- Keep active replication/lifecycle-aware behavior intact while letting single-disk deployments use the regular scanner speed-based interval.
- Add regression coverage for the startup-without-buckets cold-cache path and the single-disk default scheduling policy.

## Verification
- `cargo fmt --all`
- `cargo fmt --all --check`
- `CARGO_TARGET_DIR=/tmp/rustfs-3496-target cargo test -p rustfs-scanner`
- `CARGO_TARGET_DIR=/tmp/rustfs-3496-target cargo build -p rustfs --bin rustfs`
- Local single-disk E2E: started RustFS with an empty data directory, created `twinslab-x`, uploaded one object after startup, confirmed scanner metrics reported `scan_object=1`, and `/rustfs/admin/v3/datausageinfo` reported the bucket/object usage.
## Impact
Single-disk deployments no longer wait up to the previous 24h default scanner cycle before producing bucket usage data after buckets are created post-startup. Bucket usage metrics can populate promptly from the scanner-generated `DataUsageInfo` cache. No user-facing configuration changes are required.

## Additional Notes
The local E2E reproduces the issue #3496 timing: RustFS starts with no buckets, then a bucket and object are created after startup. On this branch, scanner metrics and data usage both update promptly instead of staying empty.


marshawcoco:
@codex review

houseme (approved):

PR #3538, opened 2026-06-17
feat(scanner): prioritize dirty usage refresh work

marshawcoco:
## Related Issues

Follow-up to #3496.

## Summary of Changes

This PR adds dirty bucket usage refresh hints for bucket/object mutation paths so the scanner can prioritize recently changed buckets before normal namespace walk work.

Changes include:
- add a bounded in-memory dirty usage bucket generation tracker in the scanner
- snapshot dirty buckets at scanner cycle start and prioritize them in bucket scan dispatch order
- clear dirty markers only after a complete, non-budget-limited scanner cycle, preserving newer generations written during a scan
- mark buckets dirty after successful create bucket, put object, copy object, delete object, delete objects, extract upload, and complete multipart upload operations

This keeps full scanner cycles as the correctness fallback while making bucket usage freshness less dependent on waiting for the next ordinary full-cycle walk.

## Verification

- `cargo check -p rustfs`
- `cargo fmt --all --check`

## Impact

Bucket usage metrics can refresh sooner after writes/deletes because the scanner prioritizes dirty buckets in the next cycle. No data usage cache format changes, no lifecycle/replication/heal behavior changes, and no new synchronous scanner work is added to write paths.

## Additional Notes

This is the first long-term usage freshness step after #3496. A follow-up PR should expose usage freshness status, including dirty backlog and last usage save/scan reason, so operators can distinguish cold cache, stale cache, dirty pending work, and truly empty buckets.


marshawcoco:
@codex review

PR #3574, opened 2026-06-18
feat(table-catalog): expose backing migration contract

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Adds a table catalog backing migration contract so operators can inspect whether a table is ready to move away from the current object-backed catalog state.

- Adds backing migration manifest fields to catalog diagnostics and export responses.
- Reports migration blockers, required source and target capabilities, current backing type, target backing type, and operator guidance.
- Marks tables that still need recovery before migration as blocked.
- Extends the PyIceberg smoke probe output to include backing migration diagnostics.
- Requires `RegisterTableAction` when external catalog sync creates/registers a missing table.

## Verification

- `cargo test -p rustfs --lib table_catalog_handlers_require_table_admin_actions -- --nocapture`
- `cargo test -p rustfs --lib external_catalog_bridge -- --nocapture`
- `cargo fmt --all --check`
- `git diff --check`

## Impact

Adds diagnostic/export response fields for table catalog backing migration readiness. Existing catalog behavior remains compatible; this does not migrate storage backends by itself.

## Additional Notes

This keeps the current object-backed catalog store and exposes the readiness contract needed before introducing a stronger metadata backend.


marshawcoco:
@codex review

PR #3577, opened 2026-06-18
feat(scanner): expose usage freshness status

marshawcoco:
## Related Issues

Follow-up to #3496.

## Summary of Changes

Expose scanner usage freshness diagnostics through scanner/admin metrics so operators can distinguish cold usage cache, stale usage data, dirty pending buckets, and truly empty buckets.

This PR adds:
- dirty usage pending bucket count
- last dirty mark and clear timestamps
- last cycle dirty and cleared bucket counts
- last usage save timestamp
- last usage save result and result code

The scanner now records dirty usage backlog changes and data usage save outcomes, and realtime admin metrics preserve these fields through the madmin metrics DTO.

Follow-up fixes in this branch also reset the last-cycle cleared dirty bucket count whenever a new dirty usage cycle snapshot starts, preventing stale cleared counts from appearing on incomplete cycles. The path containment fuzz target now uses the same ASCII byte trimming semantics as the production object path validator, avoiding false crashes for accepted object keys that contain Unicode whitespace bytes before dot characters.

## Verification

- `cargo fmt --all --check`
- `CARGO_INCREMENTAL=0 cargo test -p rustfs-common dirty_usage_cycle_snapshot_resets_stale_cleared_count`
- `CARGO_INCREMENTAL=0 cargo test -p rustfs-common usage_freshness`
- `CARGO_INCREMENTAL=0 cargo clippy -p rustfs-common --all-targets -- -D warnings`
- `RUSTFLAGS="--cfg tokio_unstable -C target-feature=-crt-static" cargo +nightly run --manifest-path fuzz/Cargo.toml --bin path_containment -- /tmp/path-crash`
- `RUSTFLAGS="--cfg tokio_unstable -C target-feature=-crt-static" cargo +nightly run --manifest-path fuzz/Cargo.toml --bin path_containment -- -max_total_time=10 -artifact_prefix=/tmp/path_containment_artifacts/`

## Impact

Adds backward-compatible scanner/admin metrics fields. No storage format, scanner scheduling, lifecycle, replication, heal, or write-path behavior changes.

## Additional Notes

This is the observability follow-up after dirty usage refresh work. It helps diagnose bucket metrics appear as zero because usage freshness work has not completed yet.

Full `make pre-commit` was not run locally because this worktree had only 7.1 GiB free after fuzz target compilation; focused checks above covered the changed common metrics code and the failing fuzz target.

### DIFF
--- a/crates/common/src/metrics.rs
+++ b/crates/common/src/metrics.rs
@@ -753,6 +753,13 @@ pub struct Metrics {
     scanner_checkpoint_cleared: AtomicU64,
     scanner_checkpoint_ignored: AtomicU64,
     scanner_checkpoint_stale: AtomicU64,
+    scanner_dirty_usage_pending_buckets: AtomicU64,
+    scanner_dirty_usage_last_mark_unix_secs: AtomicU64,
+    scanner_dirty_usage_last_clear_unix_secs: AtomicU64,
+    scanner_dirty_usage_last_cycle_dirty_buckets: AtomicU64,
+    scanner_dirty_usage_last_cycle_cleared_buckets: AtomicU64,
+    scanner_usage_last_save_unix_secs: AtomicU64,
+    scanner_usage_last_save_result: AtomicU8,
     scanner_source_work: Vec<ScannerSourceWorkCounters>,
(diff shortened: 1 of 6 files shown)


marshawcoco:
@codex review

PR #3625, opened 2026-06-19
fix(heal): track root erasure-set rebuild status

marshawcoco:
## Related Issues

Related to the replaced-node heal investigation reported by administrators.

## Summary of Changes

This fixes the root heal path used for replaced-disk / replaced-node recovery tracking.

Previously, a root heal request could be handled as a direct format heal and return a fixed token, so the actual erasure-set rebuild work was not represented by the heal manager. Status queries for an unknown or expired token could also be reported as `finished`, which made it hard to distinguish a completed heal from a task that was never tracked.

Changes:
- Route root heal requests with explicit `pool` and `set` into tracked erasure-set heal tasks.
- Encode the erasure-set target as `pool_N_set_M` so the heal manager can schedule the rebuild path.
- Allow root status queries to look up the task directly by token when the path is empty.
- Return `notFound` for unknown or expired heal tokens instead of reporting them as `finished`.
- Add unit coverage for root erasure-set request construction, validation, and empty-path status queries.

## Verification

- `cargo test -p rustfs --lib root_heal -- --nocapture`
- `cargo test -p rustfs --lib root_erasure_set -- --nocapture`
- `cargo test -p rustfs-heal`
- `cargo fmt --all --check`
- `git diff --check`

## Impact

Admin root heal requests that target a specific erasure set can now be tracked with a real heal task token. Status output should no longer make an untracked or expired task look like a successful finished heal.

## Additional Notes

Root heal without an explicit `pool` and `set` still follows the existing direct root format heal path.

### DIFF
--- a/crates/heal/src/heal/channel.rs
+++ b/crates/heal/src/heal/channel.rs
@@ -262,7 +262,13 @@ impl HealChannelProcessor {
             "Heal query received"
         );
 
-        let (summary, detail, items) = match self.heal_manager.get_task_report_for_path(&heal_path, &client_token).await {
+        let report = if heal_path.trim_matches('/').is_empty() {
+            self.heal_manager.get_task_report(&client_token).await
+        } else {
+            self.heal_manager.get_task_report_for_path(&heal_path, &client_token).await
+        };
+
+        let (summary, detail, items) = match report {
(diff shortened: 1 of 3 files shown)


marshawcoco:
@codex review

PR #3630, opened 2026-06-19
feat(scanner): define replication boundary contract

marshawcoco:
## Related Issues
N/A

## Summary of Changes
- Adds scanner replication boundary metadata to replication repair snapshots:
  - `scanner_role`
  - `execution_owner`
- Defines bucket replication repair as scanner-discovered repair admission owned by the bucket replication queue.
- Defines site replication passive/active states as boundary signals owned by the site replication runtime.
- Propagates the boundary metadata through the madmin/admin metrics snapshot and all scanner replication repair mappings.
- Extends common, scanner, madmin, and ecstore tests to lock the source, kind, role, and owner contract.
- Updates scanner runtime controls and benchmark runbook guidance for replication backlog interpretation.

## Verification
- `git diff --check`
- `cargo fmt --all --check`
- `cargo test -p rustfs-common scanner_replication --lib`
- `cargo test -p rustfs-scanner scanner_replication --lib`
- `cargo test -p rustfs-madmin scanner_metrics_merge_aggregates_replication_repair_status --lib`
- `cargo test -p rustfs-ecstore scanner_metrics_mapping_preserves_distributed_status_fields --lib`
- `cargo test -p rustfs-ecstore --lib --features rio-v2 --no-run`

## Impact
- Admin scanner metrics now include two additional replication repair metadata fields: `scanner_role` and `execution_owner`.
- Existing deserialization remains compatible through serde defaults.
- No replication engine, scanner scheduling, deployment, or configuration behavior changes.

## Additional Notes
This PR strengthens scanner operational contract without moving active site replication resync ownership into the scanner.


marshawcoco:
@codex review

houseme (changes requested):
Thanks for pushing this through. The bucket-vs-site ownership split looks aligned with MinIO: scanner-discovered bucket repair is still queue admission work, while active site resync stays owned by the site replication runtime rather than the scanner.

I found one blocker before merge: the distributed `ScannerReplicationRepairSnapshot` merge still matches entries only by `(source, kind)`, and `merge()` only adds counters. That means a rolling-upgrade cluster can merge an older node with empty default `scanner_role` / `execution_owner` first and permanently keep those fields empty in the aggregated admin view, even when newer nodes provide the metadata. Please preserve non-empty metadata during merge and add a mixed-version regression test for this path.

marshawcoco:
> Thanks for pushing this through. The bucket-vs-site ownership split looks aligned with MinIO: scanner-discovered bucket repair is still queue admission work, while active site resync stays owned by the site replication runtime rather than the scanner.
> 
> I found one blocker before merge: the distributed `ScannerReplicationRepairSnapshot` merge still matches entries only by `(source, kind)`, and `merge()` only adds counters. That means a rolling-upgrade cluster can merge an older node with empty default `scanner_role` / `execution_owner` first and permanently keep those fields empty in the aggregated admin view, even when newer nodes provide the metadata. Please preserve non-empty metadata during merge and add a mixed-version regression test for this path.

Thanks, fixed in the latest push.

`ScannerReplicationRepairSnapshot::merge()` now preserves non-empty `scanner_role` and `execution_owner` metadata when the existing aggregated entry was created from an older node with empty defaults. I also added a mixed-version regression test covering `replication_repair`, `current_cycle_replication_repair`, and `last_cycle_replication_repair`

houseme (approved):
Thanks for the follow-up. The mixed-version merge fallback is now in place, and the new regression test covers the exact rolling-upgrade case I called out earlier.

Focused re-checks look good on my side.

PR #3341, opened 2026-06-11
fix(logging): clarify recovery and metacache warn messages

houseme:
## Related Issues
Fixes rustfs/backlog#664.

## Summary of Changes
- clarify cached gRPC connection eviction logs so they explain that RustFS is discarding a failed cached channel and will reconnect automatically
- downgrade the cache-removal follow-up message from warn to info to avoid double-warning on normal recovery actions
- replace misleading `decommission_pool` metacache resolve logs with neutral `metacache resolve` wording and demote routine reconciliation traces to debug

## Verification
- `cargo test -p rustfs-filemeta --lib metacache -- --nocapture`
- `cargo test -p rustfs-protos --lib -- --nocapture`
- `cargo test -p rustfs-common --lib -- --nocapture`
- `make pre-commit`

## Impact
- improves operator-facing log clarity during scanner/deep-bitrot related pressure and routine gRPC cache recovery
- does not change recovery behavior, quorum logic, or metadata resolution semantics

## Additional Notes
- This is a follow-up to the first backlog #664 mitigation PR and focuses only on log semantics and signal quality.

PR #3653, opened 2026-06-20
fix(logging): suppress scanner leader eviction noise

hkwi:
## Summary

- add explicit log-level control for cached gRPC connection eviction
- keep the existing `evict_failed_connection(addr)` behavior as WARN by default
- downgrade scanner leader-lock connection eviction to DEBUG so expected scanner contention does not emit WARN noise
- keep remote disk and peer connection failure eviction at WARN

## Context

This builds on #3341. That PR clarified the generic eviction message and downgraded the cache-removal follow-up log to INFO, but scanner leader-lock failures still route through the generic WARN-level `evict_failed_connection()` helper.

Scanner leader-lock contention is already treated as an expected DEBUG-level path in `remote_locker.rs`, so this change lets that caller pass DEBUG through to the lower-level eviction helper while preserving WARN for actionable disk/peer RPC failures.

## Verification

- `cargo fmt --all`
- `cargo test -p rustfs-protos --lib`


hkwi:

overtrue (approved):
Code review looks good. Merge is blocked until the branch is updated with main.

PR #3691, opened 2026-06-21
refactor(ecstore): collapse peer RPC success-check into macro

seuros:


houseme:
> 
> Missing signatures:
> 
> * @seuros
> 
> 
> 

@seuros

houseme (approved):
I reviewed the refactor in peer_rest_client.rs and I do not see a behavioral change from this patch. Every call site still checks the same success/error_info pair before touching the payload, so the consolidation looks safe to me. The only remaining blockers are on the PR side: please update the branch and clear the CLA check, then this should be good to merge.

seuros:
I have the read  the CLA and there is absolutely no reason to have it in this project.

The RustFS is codebase is Apache 2.0 licensed which already cover 100% of the need .

That bot is adding fiction and noise by sending notifications in every PR and failing the builds.

So instead we should remove it.

Issue #3520, opened 2026-06-17
use rust-doctor for branch `1.0.0-beta.8`


Pierre-Gronau-ndaal:
# rust-doctor Report — branch `1.0.0-beta.8`

- **Commit:** `64c0ede0` (branch and tag point to the same commit)
- **Tool:** rust-doctor 0.2.0
- **Scope:** full workspace, 41 members
- **Scan time:** ~6 min (clippy 124s, cargo-geiger 123s)

## Score: 98/100 — "Great"

rust-doctor scores by **unique rules violated per dimension**, not by
occurrence count. Only 4 distinct error rules and 7 warning rules fired,
so the penalty is small. The raw counts still matter for triage.

| Dimension       | Score   |
| --------------- | ------- |
| Security        | 98/100  |
| Reliability     | 94/100  |
| Maintainability | 99/100  |
| Performance     | 98/100  |
| Dependencies    | 100/100 |

**Findings:** 6,900 total — 139 errors, 3,220 warnings, 3,541 info.

**Skipped passes:** `coverage` (llvm-cov not installed); `cargo-geiger`
unsafe audit (ran 123s, yielded nothing).

## Error-severity breakdown (139)

| Rule                       | Count | Verdict                       |
| -------------------------- | ----- | ----------------------------- |
| `panic-in-library`         | 74    | Mixed — 25 in tests           |
| `hardcoded-secrets`        | 35    | All false positives           |
| `blocking-in-async`        | 25    | ~15 real, rest in test blocks |
| `tokio-spawn-without-move` | 5     | False positives               |

## CRITICAL — none real

All 35 `hardcoded-secrets` are false positives. They flag **env-var name
constants**, **HTTP header names**, and **message strings** — not secrets:

- `crates/config/src/notify/redis.rs:44` —
  `ENV_NOTIFY_REDIS_PASSWORD = "RUSTFS_NOTIFY_REDIS_PASSWORD"` (env-var name)
- `rustfs/src/auth.rs:75` — `AMZ_CREDENTIAL = "X-Amz-Credential"` (header)
- `rustfs/src/server/cors.rs:27` —
  `ACCESS_CONTROL_ALLOW_CREDENTIALS = "access-control-allow-credentials"`
- The rest follow the same shape (test fixtures like `user_secret`,
  `RPC_SECRET_REQUIRED_MESSAGE`, etc.)

Security is genuinely clean, and Dependencies scored 100/100 (cargo-deny
[...]

PR #3717, opened 2026-06-22
fix(object): bound snowball auto extract

houseme:
## Related Issues
Tracks [rustfs/backlog#691](https://github.com/rustfs/backlog/issues/691)
Original deleted tracker: `rustfs/rustfs#3552`
Parent tracking context: `rustfs/rustfs#3520`

## Summary of Changes
- route snowball auto-extract through explicit archive entry-count, per-entry-size, and total-unpacked-size limits
- re-check bucket quota against cumulative unpacked size during streaming extraction instead of only trusting the original archive object size
- keep `ignore-errors` limited to skipping individual malformed entries while still failing the full request on archive-limit and quota violations
- preserve the current `object_usecase` shape on latest `main` while adding the new extract limit enforcement

## Verification
- `cargo fmt --all`
- `cargo fmt --all --check`
- `cargo check -p rustfs`
- `cargo test -p rustfs execute_put_object_extract_rejects_invalid_storage_class -- --nocapture`
- `make pre-commit`

## Impact
- snowball extract uploads now fail closed when archive entry counts, unpacked bytes, or bucket quota would be exceeded
- streaming extraction no longer allows `ignore-errors` to bypass quota and unpacked-size enforcement
- no new API surface is introduced; the change is limited to extract-upload validation and enforcement behavior

## Additional Notes
- the issue-specific local planning note remains local-only and is not included in this PR
- the branch was rebased onto the latest `main` before opening the PR so the extract limit logic matches the current `object_usecase` shape


houseme:
Closing this PR because it duplicates the already-merged fix in #3715 for the same `#3552` / `rustfs/backlog#691` scope.

After rebasing, this branch no longer carries an effective diff against current `main`, so keeping it open would only duplicate review history.

PR #3735, opened 2026-06-22
perf(server): lighten internode data-plane stack

houseme:
## Related Issues
N/A

## Summary of Changes
- Split internode data-plane requests from the public HTTP path with an early `RPC_PREFIX` dispatch scaffold.
- Remove `RequestLoggingLayer` and the generic `TraceLayer` from the internode path while keeping the external HTTP stack unchanged.
- Replace the internode path's full `RequestContextLayer` with a lite request context that preserves only the minimal compatible fields.
- Add lightweight internode RPC duration metrics and move PUT diagnostics toward object-level stage summary logs instead of child request-level tracing.
- Keep the public S3/Admin/Console behavior intact while making the internode stack independently evolvable for later follow-up work.

## Verification
- `cargo fmt --all`
- `cargo check -p rustfs`
- `cargo check -p rustfs -p rustfs-io-metrics`
- `cargo check -p rustfs -p rustfs-ecstore`
- `cargo test -p rustfs path_dispatch_service_identifies_rpc_prefix -- --nocapture`

## Impact
- Low-risk internal performance and observability refactor for internode data-plane traffic.
- No intended external API or user-facing behavior change.
- Internode requests now avoid the generic request logging and tracing stack, and instead rely on lighter internode-specific metrics plus object-level PUT stage summaries.

## Additional Notes
- This branch intentionally did not run `make pre-commit`; verification was kept to focused `fmt`, `check`, and targeted test coverage for the changed surfaces.
- The branch contains the Batch 1 through Batch 6 incremental commits for the first-stage internode data-plane lightweighting sequence.


houseme:
@codex review

houseme:
Addressed review follow-ups in the latest branch state:

- kept the full `RequestContextLayer` on the public stack, with the lite context only on the internode path
- removed the generic `TraceLayer` from the internode stack and kept lightweight internode-specific metrics instead
- updated `internode_metrics` descriptor tests for the added duration metric
- tightened internode path matching to use a path-boundary-safe check and added a `/rustfs/rpcx` boundary assertion
- fixed `PathDispatchService::call()` to dispatch on the selected inner service directly instead of cloning a different instance

Latest follow-up commit: `a1f38d0a8`.

PR #3815, opened 2026-06-24
fix: harden lock timeouts, health, and heal paths

cxymds:
## Related Issues
Follow-up to #3814.

## Summary of Changes
- Decouple remote namespace lock RPC transport timeout from the distributed lock acquire attempt budget.
- Add `RUSTFS_OBJECT_LOCK_RPC_TIMEOUT_MS` with a 3000ms default while preserving the existing bounded lock acquire attempt behavior.
- Keep liveness health probes lightweight by avoiding dependency readiness collection for `/health`, `/health/live`, and `/minio/health/live`.
- Preserve readiness health probe behavior for dependency and quorum checks.
- Prevent scanner heal fallback from submitting unresolved trailing-slash object candidates such as `object/` as object heal requests.
- Normalize trailing slashes when matching heal object and prefix paths for status/cancel lookups.
- Fail recursive heal listing when storage reports a truncated page without a continuation token, instead of silently ending early.

## Verification
- `cargo +1.96.0 fmt --all --check`
- `cargo fmt --all --check`
- `git diff --check`
- `cargo +1.96.0 check -p rustfs-ecstore --lib`
- `cargo +1.96.0 test -p rustfs-lock --lib acquire_guard_uses_bounded_distributed_attempt_timeout -- --nocapture`
- `cargo +1.96.0-aarch64-apple-darwin test -p rustfs-scanner --lib test_resolve_object_heal_entry -- --nocapture`
- `cargo +1.96.0-aarch64-apple-darwin test -p rustfs-heal test_heal_type_matches_path_normalizes -- --nocapture`
- `cargo +1.96.0-aarch64-apple-darwin test -p rustfs-ecstore --lib test_remote_client_acquire_locks_batch_uses_rpc_timeout_and_evicts_connection -- --nocapture`
- `cargo +1.96.0-aarch64-apple-darwin check -p rustfs-ecstore --lib`
- `cargo +1.96.0-aarch64-apple-darwin test -p rustfs-heal --lib next_heal_listing_token -- --nocapture`
- `cargo +1.96.0-aarch64-apple-darwin test -p rustfs-heal --lib test_recursive_bucket_heal -- --nocapture`
- `cargo +1.96.0-aarch64-apple-darwin check -p rustfs-heal --lib`
- `cargo +1.96.0-aarch64-apple-darwin test -p rustfs-heal --lib -- --nocapture`

Not completed locally:
- `cargo +1.96.0 test -p rustfs-ecstore --lib remote_locker -- --nocapture` was interrupted after the local compile/link step hung for several minutes.
- `cargo +1.96.0 test -p rustfs test_liveness_dependency_readiness_report_is_lightweight_ready -- --nocapture` was interrupted after the main crate compile remained running for over 30 minutes.
- `make pre-commit` was not run locally.

## Impact
- Adds an optional remote lock RPC timeout environment variable: `RUSTFS_OBJECT_LOCK_RPC_TIMEOUT_MS`.
- Liveness probes now report process liveness without blocking on storage, IAM, or lock quorum readiness checks.
- Readiness probes continue to report dependency readiness.
- Scanner-initiated heal no longer retries unresolved trailing-slash candidates as missing object data.
- Heal status and cancellation path matching now tolerates trailing slash differences for object and prefix paths.
- Recursive heal fails visibly if the listing pagination contract is incomplete, preventing partial scans from being reported as successful.

## Additional Notes
- This PR remains draft because the full `make pre-commit` gate was not completed locally.
- Load balancers should use `/health/live` for fast liveness checks; `/health/ready` remains suitable for readiness and diagnostics.

---

### DIFF
--- a/crates/config/src/constants/object.rs
+++ b/crates/config/src/constants/object.rs
@@ -281,6 +281,19 @@ pub const ENV_OBJECT_LOCK_ACQUIRE_TIMEOUT: &str = "RUSTFS_OBJECT_LOCK_ACQUIRE_TI
 /// Default lock acquisition timeout: 5 seconds.
 pub const DEFAULT_OBJECT_LOCK_ACQUIRE_TIMEOUT: u64 = 5;
 
+/// Environment variable for remote namespace lock RPC transport timeout in milliseconds.
+///
+/// This timeout bounds the internode RPC call itself. It is intentionally
+/// separate from `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT` and the distributed lock
+/// per-attempt acquire budget so short lock-contention windows do not become
+/// aggressive network deadlines.
+///
+/// Default: 3000 milliseconds (can be overridden by `RUSTFS_OBJECT_LOCK_RPC_TIMEOUT_MS`).
+pub const ENV_OBJECT_LOCK_RPC_TIMEOUT_MS: &str = "RUSTFS_OBJECT_LOCK_RPC_TIMEOUT_MS";
+
+/// Default remote lock RPC transport timeout: 3000 milliseconds.
+pub const DEFAULT_OBJECT_LOCK_RPC_TIMEOUT_MS: u64 = 3000;
+
 /// Environment variable to enable object namespace lock diagnostics.
 ///
 /// When enabled, RustFS emits slow lock acquisition and long lock hold
--- a/crates/ecstore/src/rpc/remote_disk.rs
+++ b/crates/ecstore/src/rpc/remote_disk.rs
@@ -568,7 +568,7 @@ impl RemoteDisk {
     {
         // Check if disk is faulty
         if self.health.is_faulty() {
-            warn!(
+            debug!(
                 event = EVENT_REMOTE_DISK_HEALTH,
                 component = LOG_COMPONENT_ECSTORE,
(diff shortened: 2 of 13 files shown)


houseme (changes requested):
Thanks for the follow-up here. The scanner/heal fixes look good, but I found two blockers before this is ready:

- `RUSTFS_OBJECT_LOCK_RPC_TIMEOUT_MS` still does not actually cap the transport deadline. `RemoteClient::rpc_timeout()` takes `max(configured, acquire_timeout + 500ms)`, so the RPC remains coupled to the lock wait budget. With the default 5s acquire timeout, the new default 3000ms timeout becomes 5500ms, and even an operator override like `50ms` still resolves to at least `500ms`. That misses the decoupling goal and diverges from MinIO's lock REST client, which just honors the caller deadline.
- `/health`, `/health/live`, and the console/public liveness paths now synthesize dependency readiness as fully healthy in the JSON body. That keeps the probe lightweight, but it also makes `details.storage`, `details.iam`, and `details.lock` report false positives whenever those dependencies are actually degraded. MinIO's liveness handler avoids external checks without fabricating healthy dependency state.

I re-ran `cargo check --all-targets` successfully in an isolated tree while reviewing. Once the timeout semantics and liveness payload semantics are tightened, I'm happy to take another look.

PR #3836, opened 2026-06-25
feat(table-catalog): add durable strong backing

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Adds durable strong table catalog backing semantics.

- Adds a strong catalog store that persists table bucket, namespace, table/view, commit log, and idempotency index state into an internal snapshot object.
- Replays durable catalog state on store startup so table load, commit lookup, idempotency lookup, and recovery diagnostics survive restart.
- Uses snapshot object ETag preconditions (`IfAbsent` / `IfMatch`) to fence stale writers.
- Reloads durable state after successful writes or CAS conflicts so local cache state cannot overwrite newer committed catalog state.
- Adds stale-writer coverage proving an older store instance cannot advance a stale metadata pointer or pollute commit/idempotency state after another writer has committed.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs --lib strong_catalog_backing_ -- --nocapture`
- `cargo test -p rustfs --lib table_catalog_store_trait_ -- --nocapture`
- `cargo test -p rustfs --lib diagnostics_backing_manifest_requires_recovery_before_migration -- --nocapture`

## Impact

Adds durable strong backing implementation and tests for table catalog state. Existing object-backed catalog behavior remains the default; this does not switch production handlers or change public API behavior.

## Additional Notes

The new backing uses an internal snapshot object plus ETag CAS as the durable fence. Stale writers fail closed and reload durable state instead of overwriting newer table pointer, token, generation, commit log, or idempotency index state.


marshawcoco:
@codex review

PR #3844, opened 2026-06-25
fix(scanner): preserve dirty usage convergence

### LINKED ISSUE #3841: datausageinfo can report partial object counts before scanner convergence
## Summary

The ignored single-node data usage E2E can reproducibly observe partial object counts from `/rustfs/admin/v3/datausageinfo` before the scanner has converged on the just-written objects.

This is split out from #3492 because #3492 is now the broader distributed scanner validation plan. This issue tracks the concrete local reproduction and expected API/test behavior around usage freshness.

## Environment

- Local single-node, single-disk E2E test
- Local checkout commit: `ad6184b1266ef28c44d8201c25319c5b28cff2b2`
- Note: local `main` was behind `upstream/main` by 7 commits when this was reproduced
- Date: 2026-06-25 UTC

## Reproduction

Run the ignored E2E locally:

```bash
CARGO_INCREMENTAL=0 RUSTFLAGS='--cfg tokio_unstable -C debuginfo=0' \
  cargo test -p e2e_test data_usage_reports_all_objects -- --ignored --nocapture
```

I ran it twice after removing CI-covered unit/mapping/aggregation tests from the local validation list.

## Observed Result

Both runs started RustFS, uploaded 1000 objects, and received HTTP 200 from `/rustfs/admin/v3/datausageinfo`, but the returned usage was still partial:

- Run 1: expected at least 1000 objects, got 959
- Run 2: expected at least 1000 objects, got 981

The logs around the failed assertion still showed the scanner reading object metadata, for example `read_metadata` spans for `obj-xxxx/xl.meta`, and `store_data_usage_in_backend` / scanner cycle activity around the same window.

The failing assertion is:

```text
crates/e2e_test/src/data_usage_test.rs:60:5
total object count should be at least 1000, got <partial count>
```

## Expected Result

The behavior should be explicitly defined and testable. Either:

- `/rustfs/admin/v3/datausageinfo` should not make a fresh-looking complete-count assertion before scanner usage has converged, or
- the API/status/metrics should expose enough usage freshness state for callers and tests to know that the returned count is still partial, or
- this E2E should wait on an explicit scanner usage-save/freshness convergence signal before asserting that all 1000 objects are reflected in data usage.

## Why This Matters

This is the same class of operator confusion as the scanner/bucket metrics freshness work: an admin API can return success while usage is still catching up. Without an explicit freshness/convergence signal in the path used by tests and operators, partial usage can be mistaken for a real object-count mismatch.

marshawcoco:
## Related Issues
Fixes #3841.

## Summary of Changes
- Prevent scanner completion snapshots from being published when their dirty bucket generation snapshot no longer covers the current dirty usage state.
- Preserve request-path dirty bucket usage in the admin/datausageinfo memory overlay until a scanner snapshot catches up to the same core counts.
- Wake the scanner immediately when dirty usage work is already pending so skipped or stale cycles retry promptly.

## Verification
- `cargo fmt --all --check`
- `CARGO_INCREMENTAL=0 RUSTFLAGS='--cfg tokio_unstable -C debuginfo=0' cargo test -p rustfs-scanner dirty_usage_snapshot_detects_uncovered_generation -- --nocapture`
- `CARGO_INCREMENTAL=0 RUSTFLAGS='--cfg tokio_unstable -C debuginfo=0' cargo test -p rustfs-scanner completed_data_usage_info_requires_every_set_before_publish -- --nocapture`
- `CARGO_INCREMENTAL=0 RUSTFLAGS='--cfg tokio_unstable -C debuginfo=0' cargo test -p rustfs-scanner test_wait_for_next_scanner_cycle_sees_existing_dirty_usage -- --nocapture`
- `CARGO_INCREMENTAL=0 RUSTFLAGS='--cfg tokio_unstable -C debuginfo=0' cargo test -p rustfs-ecstore scanner_sync_preserves_dirty_memory_update_with_later_partial_snapshot -- --nocapture`
- `CARGO_INCREMENTAL=0 RUSTFLAGS='--cfg tokio_unstable -C debuginfo=0' cargo test -p e2e_test data_usage_reports_all_objects -- --ignored --nocapture`

## Impact
Admin data usage responses no longer regress to partial scanner counts while a scanner cycle is still converging after recent writes. This keeps bucket object and size totals monotonic with request-path mutations until persisted scanner usage catches up.

## Additional Notes
The ignored E2E reproduced the issue on upstream/main with a partial object count before this fix, then passed after the scanner publish gate and dirty overlay preservation changes.

### DIFF
--- a/crates/ecstore/src/data_usage.rs
+++ b/crates/ecstore/src/data_usage.rs
@@ -50,6 +50,10 @@ struct CachedBucketUsage {
     usage: BucketUsageInfo,
     refreshed_at: SystemTime,
     usage_updated_at: SystemTime,
+    // Set by request-path mutations until a scanner snapshot catches up to the same core counts.
+    dirty: bool,
+    // Set when a newer scanner snapshot was observed but did not include the dirty counts yet.
+    stale_snapshot_pending: bool,
 }
 
 type UsageMemoryCache = Arc<RwLock<HashMap<String, CachedBucketUsage>>>;
@@ -481,6 +485,8 @@ fn cached_bucket_usage_from_backend(usage: BucketUsageInfo, updated_at: SystemTi
         usage,
         refreshed_at: SystemTime::now(),
         usage_updated_at: updated_at,
+        dirty: false,
+        stale_snapshot_pending: false,
     }
 }
 
@@ -490,11 +496,20 @@ fn cached_bucket_usage_now(usage: BucketUsageInfo) -> CachedBucketUsage {
         usage,
         refreshed_at: now,
         usage_updated_at: now,
+        dirty: false,
+        stale_snapshot_pending: false,
     }
 }
 
 fn data_usage_info_updated_at(data_usage_info: &DataUsageInfo) -> SystemTime {
     data_usage_info.last_update.unwrap_or(SystemTime::UNIX_EPOCH)
+}
+
+fn bucket_usage_counts_match(left: &BucketUsageInfo, right: &BucketUsageInfo) -> bool {
+    left.size == right.size
+        && left.objects_count == right.objects_count
+        && left.versions_count == right.versions_count
+        && left.delete_markers_count == right.delete_markers_count
 }
 
 /// Fast in-memory update for immediate quota and admin usage consistency.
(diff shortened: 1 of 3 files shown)


marshawcoco:
@codex review

PR #4014, opened 2026-06-28
feat(get): Small-file GET performance optimization for 1KiB-1MiB objects.

houseme:
## Summary

Small-file GET performance optimization for 1KiB-1MiB objects.

Closes https://github.com/rustfs/backlog/issues/765

### Inline Fast Path (SF02 + SF05)
- Add `ObjectInfo::is_inline_fast_path_eligible()` — single source of truth for inline detection
- Version-aware thresholds: 128 KiB (non-versioned) / 16 KiB (versioned), matching PUT path
- Skip disk I/O semaphore for inline objects (`_disk_permit: None`)

### Bucket Validation Cache (SF01)
- Adaptive backend: `RwLock<HashMap>` (default) / `starshard::ShardedHashMap` (via `RUSTFS_BUCKET_CACHE_STARSHARD=1`)
- 5s TTL, invalidated on `make_bucket` / `delete_bucket`

### Metadata Cache (SF03)
- TTL increased from 250ms to 2s, max entries from 1024 to 4096

### Conditional Metrics (SF07)
- Gate hot-path metrics behind `get_stage_metrics_enabled()` (`AtomicBool::Relaxed`)

### Bug Fixes
- **Versioning write lock**: `BucketVersioningSys::get()` used write lock for read-only operation — changed to read lock
- **make_bucket cache**: Added cache invalidation on bucket creation
- **Erasure decode**: Validate `written` return value from inline decode
- **Fault tolerance**: Restored `tokio::spawn` + `JoinError` handling in metadata fanout
- **Lifecycle check**: Restored unconditional `resolve_put_object_expiration` (SF06 conditional was incorrect)

## Reverted
- SF04 (remove `tokio::spawn`): Restored for fault tolerance — single disk panic would unwind entire operation
- SF06 (conditional lifecycle): Restored — skipping lifecycle for objects without `x-amz-expiration` header is incorrect
- Metadata early-stop: Implementation has bugs, reverted to disabled default


houseme:
Closed in favor of #4016.

PR #4016, opened 2026-06-28
feat(get): Small-file GET performance optimization for 1KiB-1MiB objects

houseme:
## Related Issues

Closes rustfs/backlog#765

## Summary of Changes

Small-file GET performance optimization for 1KiB-1MiB objects.

### Inline Fast Path (SF02 + SF05)

- Add `ObjectInfo::is_inline_fast_path_eligible()` as the single source of truth for inline detection.
- Use version-aware thresholds: 128 KiB for non-versioned objects and 16 KiB for versioned objects, matching the PUT path.
- Skip disk I/O semaphore acquisition for inline objects.

### Bucket Validation Cache (SF01)

- Add a 5s bucket validation cache.
- Use the default `RwLock<HashMap>` backend with optional `starshard::ShardedHashMap` via `RUSTFS_BUCKET_CACHE_STARSHARD=1`.
- Invalidate the cache on bucket creation and deletion.

### Metadata Cache (SF03)

- Increase metadata cache TTL from 250ms to 2s.
- Increase max entries from 1024 to 4096.

### Conditional Metrics (SF07)

- Gate hot-path metrics behind `get_stage_metrics_enabled()`.

### Correctness Fixes

- Change `BucketVersioningSys::get()` from a write lock to a read lock for read-only versioning queries.
- Validate the `written` return value from inline erasure decode.
- Restore `tokio::spawn` plus `JoinError` handling in metadata fanout for fault tolerance.
- Restore unconditional `resolve_put_object_expiration` behavior.

## Verification

- Not run in this turn; this PR recreates the existing branch without code changes.

## Impact

- Improves small-object GET hot paths by reducing repeated bucket validation, metadata cache churn, inline-object disk semaphore overhead, and hot-path metrics overhead.
- Keeps lifecycle handling and metadata fanout fault tolerance aligned with existing behavior.

## Additional Notes

- Recreated after closing #4014 so the PR is explicitly linked to rustfs/backlog#765.


cxymds:
Potential blockers I noticed while reviewing the GET path changes:

1. The GET I/O semaphore appears to be acquired too late.

`prepare_get_object_read_execution()` now calls `prepare_get_object_read()` before acquiring the disk read permit. However, for the legacy non-inline path, `store.get_object_reader()` already creates the duplex reader and spawns the background task that runs `get_object_with_fileinfo()`, which performs the actual disk reads.

So for non-inline GETs, the disk read work can start before the semaphore is acquired, and the permit no longer protects the read path it was meant to limit. Under high concurrency this can bypass I/O backpressure and increase disk/memory pressure.

Can we keep acquiring the permit before `get_object_reader()` for all non-inline paths? If the goal is to skip the permit only for inline-direct reads, the storage layer probably needs to return an explicit reader path/result so the app layer does not infer it after the read task may already have started.

2. The inline direct path does not appear to preserve legacy read behavior.

The normal read path constructs erasure with:

`Erasure::new_with_options(..., fi.uses_legacy_checksum)`

and also maps legacy `HighwayHash256S` to `HighwayHash256SLegacy`. The new inline fast path uses `Erasure::new(...)` and `checksum_info.algorithm` directly.

That means inline objects written with the legacy shard/checksum format may be decoded with the wrong shard sizing or checksum algorithm, causing GET failures for older inline data.

Can we reuse the same erasure/checksum selection logic as the normal path here, and add a regression test for a legacy inline object?

PR #4042, opened 2026-06-29
fix(runtime): keep lock clients available during startup

### LINKED ISSUE #4040: Distributed startup can fail lock quorum before AppContext initializes
## Problem

A 4-node distributed RustFS cluster can fail to become ready during startup on current `main` when internode lock RPCs are handled before `AppContext` is fully initialized.

## Reproduction context

Observed while testing a 4-node / 4-disk distributed cluster on VM008-VM011:

- Nodes: `10.0.0.13`, `10.0.0.14`, `10.0.0.15`, `10.0.0.16`
- Mode: distributed 4x4 erasure setup
- Binary: current `main` build, observed around `9414d645`
- Test shape: clean distributed startup using per-node local disks

The same test environment can start and become ready with the beta.8 binary, but current `main` can fail during early storage initialization.

## Observed behavior

During startup, storage initialization reaches metadata/config loading, including `.rustfs.sys/pool.bin`. At that point, peer lock RPCs may fail with:

```text
Lock client not initialized. Please ensure storage is initialized first.
```

The local node can then fail namespace lock quorum while loading pool metadata/config, for example with only the local lock succeeding and peer locks rejected because the remote node's lock client resolver returned `None`.

The cluster remains not ready instead of completing distributed startup.

## Expected behavior

Once startup has initialized and published lock clients, internode lock RPC handlers should be able to serve lock requests even if the full `AppContext` has not finished initialization yet.

Distributed startup should not fail lock quorum solely because the AppContext-backed resolver is not available during the early HTTP/RPC startup window.

## Notes

The failure appears to be in the startup/runtime dependency boundary rather than scanner behavior. The relevant sequence is:

1. Startup storage foundation initializes endpoint pools and lock clients.
2. HTTP/RPC serving can begin before full AppContext initialization completes.
3. Early storage runtime initialization loads `.rustfs.sys/pool.bin` and needs distributed namespace locks.
4. Remote NodeService lock RPC handlers can resolve lock clients through the AppContext path and return `None` before AppContext is ready.
5. Lock quorum fails even though startup lock clients were already initialized earlier.

This was surfaced while investigating discussion https://github.com/orgs/rustfs/discussions/3850, but it is a distributed startup lock-client availability issue, not a data scanner issue.

marshawcoco:
## Related Issues

Fixes #4040
Related to https://github.com/orgs/rustfs/discussions/3850

## Summary of Changes

Fixes a distributed startup regression where the internode lock RPC path could return `Lock client not initialized` before `AppContext` was fully initialized.

During startup, storage foundation initializes and publishes lock clients before the HTTP/RPC server begins serving requests, but the AppContext-based resolver could still return `None` in that early window. This caused peer lock requests during `.rustfs.sys/pool.bin` loading to fail quorum and keep 4-node clusters from becoming ready.

This PR keeps AppContext as the preferred source, while allowing the lock client resolvers to fall back to the startup-initialized ecstore lock clients until AppContext is available.

## Verification

- `cargo test -p rustfs startup_fallback_before_app_context --lib`
- `cargo test -p rustfs resolver_helpers_are_context_first_and_empty_when_context_is_absent --lib`
- `cargo fmt --all --check`
- Validated on VM008-VM011 with a temporary 4-node cluster on `19100/19101`: all nodes reached `ready=true`, with `Lock client not initialized=0` and `namespace lock quorum=0`.

## Impact

Distributed clusters can complete startup while internode lock RPCs are served before AppContext initialization finishes. No API, config, or data format changes.

## Additional Notes

The fix is scoped to lock client resolution. Existing AppContext-first behavior is preserved once AppContext is initialized.


marshawcoco:
@codex review

PR #4163, opened 2026-07-02
fix(scanner): advance cycle counter when a cycle ends on budget

rafaelperoco:
## Related Issues

N/A

## Summary of Changes

`run_data_scanner_cycle` only advances and persists `cycle_info.next` when a cycle completes fully. Cycles stopped by the cycle budget (`RUSTFS_SCANNER_CYCLE_MAX_DURATION_SECS`, `RUSTFS_SCANNER_CYCLE_MAX_OBJECTS`, `RUSTFS_SCANNER_CYCLE_MAX_DIRECTORIES`) return early and leave the counter where it was.

That counter is stamped into each bucket cache as `next_cycle` before every disk scan (`scanner_io.rs`), and compacted folders are only rescanned when `hash % DATA_USAGE_UPDATE_DIR_CYCLES == next_cycle % DATA_USAGE_UPDATE_DIR_CYCLES` (`scanner_folder.rs`). While the counter is pinned, the same 1/16 window of compacted folders is rescanned on every cycle and the remaining 15/16 are skipped indefinitely, so lifecycle expiration stops running for objects in those folders and their usage entries never refresh.

Budget-limited cycles now advance and persist the counter before going idle, reusing the persistence block of the completed path (extracted into `persist_scanner_cycle_state`). Hard scan failures keep the existing behavior and do not advance the counter.

## Verification

- `cargo test -p rustfs-scanner --lib` (adds `test_finalize_partial_scan_cycle_advances_and_persists_counter`)
- `cargo clippy -p rustfs-scanner --all-targets`
- Manual: single node, 60 objects across 12 prefixes, `RUSTFS_SCANNER_CYCLE=3` and `RUSTFS_SCANNER_CYCLE_MAX_OBJECTS=5` so most cycles stop on budget. Cycle numbers from `Scanner cycle started` events over ~50s:
  - before: `0 1 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 3 3 4 5 ...` (pinned while cycles end on budget)
  - after: `0 1 2 3 4 5 6 7 8 9 10 11 ...` (advances on every cycle)

## Impact

Deployments that configure a scanner cycle budget recover lifecycle expiration and usage refresh for all compacted folders. No API or configuration changes; the persisted state format is unchanged.

## Additional Notes

N/A


cxymds (approved):
Reviewed the scanner cycle counter change. The diff is scoped to partial budget completion, preserves the existing failure/cancel paths, and adds regression coverage for persisting the advanced counter.

Local verification: `cargo test -p rustfs-scanner test_finalize_partial_scan_cycle_advances_and_persists_counter -- --nocapture` passed. Remote CI was still pending at review time.

cxymds:
@rafaelperoco I reviewed this change and the main fix looks sound: advancing and persisting `cycle_info.next` for budget-limited partial cycles prevents the scanner from reusing the same cycle window indefinitely.

One small consistency suggestion: `finalize_partial_scan_cycle()` currently persists the cycle state before calling `mark_scan_cycle_idle()`. That means the persisted `CurrentCycle` can still contain the previous non-zero `current` value, while the full-completion path clears `current` before persisting. The persisted `next` counter is correct, so this does not appear to break the main fix, but swapping the order would keep partial-cycle state consistent with completed-cycle state:

```rust
cycle_info.next += 1;
mark_scan_cycle_idle(cycle_info).await;
persist_scanner_cycle_state(storeapi, cycle_info).await;
```

It would also be useful to extend the regression test with `decoded.current == 0`, so the persisted idle state is covered as well.

PR #4224, opened 2026-07-03
feat(storage): harden internode data-path controls

houseme:
## Related Issues

Related to rustfs/backlog#797.

## Summary of Changes

- Harden the internode HTTP data-plane by propagating `HttpWriter` shutdown failures, adding classified internode HTTP metrics, and introducing conservative tuning profiles that keep the legacy behavior as the default.
- Tighten remote control-plane behavior by unifying remote lock RPC deadline/eviction handling and adding a gated `BatchReadVersion` RPC path for metadata fanout.
- Make read-path optimization opt-in and observable: harden `ReadMultiple` response decoding, add `SameNode` shard locality attribution, keep locality scheduling behind `RUSTFS_SHARD_LOCALITY_SCHEDULING`, and keep adaptive batch processor behavior behind `RUSTFS_BATCH_PROCESSOR_ADAPTIVE`.
- Add internode observability counters/histograms for HTTP/gRPC operations, shard locality, batch read-version gate decisions, and batch processor observations.
- Add a local 4-node / 16-disk A/B runner with bounded warn-level file logging and signed admin metrics capture.
- Keep mixed-version and rollout behavior conservative: performance gates default to `legacy`, `off`, or observe-only; correctness fixes remain always on.
- Keep issue-analysis documents local-only; they are not part of the PR diff.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-rio internode_http_tuning_profile_parses_known_values_and_falls_back_to_legacy`
- `cargo nextest run -p rustfs -p rustfs-protocols --features sftp`
  - Result: 2261 passed, 71 skipped.
- `cargo clippy -p rustfs-io-metrics -p rustfs-ecstore --all-targets -- -D warnings`
- `cargo clippy -p rustfs-io-metrics -p rustfs-ecstore -p e2e_test -p rustfs --all-targets -- -D warnings`
- `cargo test -p rustfs-ecstore batch_metadata_rpc_mode`
- `cargo test -p rustfs-ecstore batch_read_version_response_decode`
- `cargo test -p rustfs encode_batch_read_version_response_payloads_keeps_json_and_msgpack_in_sync`
- `cargo test -p e2e_test --no-run`
- `bash -n scripts/run_issue797_local_4node_16disk_ab.sh`
- `shellcheck scripts/run_issue797_local_4node_16disk_ab.sh`
- `scripts/run_issue797_local_4node_16disk_ab.sh --dry-run --out-dir /private/tmp/issue797-runner-dry-run-review-fix`
  - Result: dry-run confirmed default data root stays under `/tmp` for safer local cleanup.
- `scripts/run_issue797_local_4node_16disk_ab.sh --skip-build --base-port 19300 --duration 60s --sizes 4KiB,1MiB --concurrency 8 --profiles baseline,metrics_logging --out-dir target/bench/issue-797-local-4node-16disk-ab-warn-retest-20260703 --data-root /private/tmp/issue797-local-4node-16disk-ab-warn-retest-20260703`
  - Result: baseline and metrics_logging 4KiB/1MiB runs passed with `error_count=0`; signed admin metrics snapshots completed with `curl-sigv4`; metrics_logging logs remained bounded at about 676 KiB.
- `make pre-commit`
  - Result: passed.
- `make pre-pr`
  - Result: passed. Full gate included formatting, unsafe-code allowance checks, architecture migration guardrails, logging guardrails, doc path checks, clippy, 7330 nextest tests passed with 113 skipped, and workspace doctests passed.

## Impact

- Runtime compatibility: existing defaults preserve legacy internode HTTP behavior and keep new performance paths disabled unless explicitly enabled.
- Configuration: adds or documents these rollout switches:
  - `RUSTFS_INTERNODE_HTTP_TUNING_PROFILE=legacy|balanced|throughput`
  - `RUSTFS_INTERNODE_HTTP_PROXY=off|system`
  - `RUSTFS_SHARD_LOCALITY_SCHEDULING=off|observe|on`
  - `RUSTFS_METADATA_BATCH_READ=off|auto|on`
  - `RUSTFS_BATCH_PROCESSOR_ADAPTIVE=off|observe|on`
- Protocol/API: adds batch read-version RPC wiring and compatibility fallback behavior. `off` uses unary reads, `auto` may fall back on unimplemented peers, and `on` fails explicitly when unsupported.
- Operations: adds local A/B tooling for 4-node / 16-disk validation with bounded warn-level file logging. Issue notes and analysis docs remain local-only and are excluded from the PR diff.

## Additional Notes

- Review comments have been addressed and resolved.
- Backlog issue 797 has been updated with the `RUSTFS_METADATA_BATCH_READ=on` usage guide and rollout scenarios.
- Local verification is complete. GitHub-hosted required checks should be allowed to finish before merge.


houseme:
@codex review

crates/ecstore/src/cluster/rpc/remote_disk.rs
```diff
@@ -1462,6 +1604,90 @@ impl DiskAPI for RemoteDisk {
         .await
     }
 
+    #[tracing::instrument(skip(self, req))]
+    async fn batch_read_version(&self, req: BatchReadVersionReq) -> Result<Vec<BatchReadVersionResp>> {
```
overtrue, line 1615: As far as I can tell nothing calls `batch_read_version` — the only invokers are this client impl and the server-side RPC handler, so no node ever initiates the batch read. That means `RUSTFS_METADATA_BATCH_READ` gates unreachable code and the gate-decision metrics can never fire, while the PR body documents it as a rollout switch for metadata fanout.

Either wire an actual caller into the fanout path in this PR, or split the RPC scaffolding out until there is one.

crates/ecstore/src/cluster/rpc/remote_disk.rs
```diff
@@ -1462,6 +1604,90 @@ impl DiskAPI for RemoteDisk {
...
+                                event = EVENT_REMOTE_DISK_RPC,
+                                component = LOG_COMPONENT_ECSTORE,
+                                subsystem = LOG_SUBSYSTEM_REMOTE_DISK,
+                                endpoint = %self.endpoint,
+                                batch_metadata_rpc_mode = mode.as_str(),
+                                op = "batch_read_version",
+                                state = "fallback_unimplemented",
+                                "Remote disk BatchReadVersion unsupported; falling back to unary read_version"
+                            );
+                            return batch_read_version_one_by_one(self, req).await;
```
overtrue: This fallback runs inside `execute_with_timeout_for_op(..., get_max_timeout_duration())`, so a full 128-item batch against a pre-upgrade peer has to finish 128 *sequential* unary reads within the single 30s (default) window. If it doesn't, the whole batch times out and the outer wrapper marks the disk failed — even though every individual unary read was succeeding.

The `off` path above returns before entering the wrapper; the unimplemented fallback should do the same.

crates/ecstore/src/services/batch_processor.rs
```diff
@@ -19,17 +19,246 @@
...
+        concurrency: configured_concurrency,
+        reason: BATCH_SUGGESTION_REASON_STABLE,
+    }
+}
+
+fn decrease_batch_concurrency(configured_concurrency: usize) -> usize {
+    (configured_concurrency.saturating_mul(3) / 4).max(1)
+}
+
+fn increase_batch_concurrency(configured_concurrency: usize, batch_size: usize) -> usize {
```
overtrue: Since the next batch starts from the raised value, this ratchets: under sustained fast large batches, concurrency climbs 25% per cooldown until it reaches the largest batch size (16 → 128 for the read processor), and nothing brings it back toward the configured value short of errors/timeouts/slow batches. The configured limit is effectively gone in `on` mode.

Consider a hard ceiling relative to the configured concurrency (2–4x?).

crates/rio/src/http_reader.rs
```diff
@@ -292,20 +465,29 @@ fn should_bypass_proxy_for_url(url: &str) -> bool {
...
+        InternodeHttpProxyMode::NoProxy => true,
+        InternodeHttpProxyMode::Legacy => should_bypass_proxy_for_url(url),
+    }
+}
+
 async fn get_http_client(url: &str) -> Client {
-    // Reuse HTTP connection pools while keeping loopback traffic away from
-    // system proxies so local RPC/tests do not leak to proxy listeners.
-    let disable_proxy = should_bypass_proxy_for_url(url);
+    let tuning = InternodeHttpClientTuning::from_env();
```
overtrue: This now does 7 env lookups plus string parsing on every internode HTTP request, where the old code was one atomic load and a loopback check. Env vars can't change mid-process — parse once into a `LazyLock`, which also makes the `cached.tuning == tuning` cache check unnecessary.

Same per-call env parsing pattern in `batch_processor_adaptive_mode()` (every batch) and `batch_metadata_rpc_mode()`.

crates/ecstore/src/set_disk/read.rs
```diff
@@ -595,14 +595,55 @@ fn resolve_read_part_from_responses(
...
+    }
+
+    if !host_name.is_empty() && local_endpoint_hosts.iter().any(|host| host == host_name) {
+        return ShardReadCost::SameNode;
+    }
+
+    ShardReadCost::Remote
+}
+
+fn local_endpoint_hosts_for_shard_costs() -> Vec<String> {
```
overtrue: Two things here. This list is rebuilt on every GET while endpoint pools are static after startup — worth caching. And with `RUSTFS_SHARD_LOCALITY_SCHEDULING=observe` alone, the costs get collected but `record_get_object_shard_locality_observe_only` drops everything unless stage metrics are also enabled — so observe mode pays the per-read overhead and produces nothing. At minimum document the dependency on the metrics-export gate.

scripts/run_issue797_local_4node_16disk_ab.sh
```diff
@@ -0,0 +1,583 @@
...
+
+die() {
+  printf '[ERROR] %s\n' "$*" >&2
+  exit 1
+}
+
+arg_value() {
+  local flag="$1"
+  local value="${2:-}"
+  if [[ -z "$value" || "$value" == --* ]]; then
```
overtrue, line 90: This rejects any value starting with `--`, so `--warp-extra-args --noclear` — the flag's own documented default — dies with "missing value". Needs a special case for that flag or `--flag=value` support.

scripts/run_issue797_local_4node_16disk_ab.sh
```diff
@@ -0,0 +1,583 @@
...
+  local endpoint_loop_id
+  local endpoints=()
+  for ((endpoint_loop_id = 1; endpoint_loop_id <= NODE_COUNT; endpoint_loop_id++)); do
+    endpoints+=("$(endpoint_for_node "$endpoint_loop_id")")
+  done
+  local IFS=','
+  printf '%s' "${endpoints[*]}"
+}
+
+warp_hosts_csv() {
```
overtrue: Looks unused — nothing calls `warp_hosts_csv`.

overtrue (commented):
Went through the whole diff. The correctness fixes here look solid — HttpWriter shutdown propagation, the ReadMultiple decode hardening, and the lock RPC timeout/eviction unification are all good changes with decent test coverage. A few concerns inline; the main one is that the new BatchReadVersion path has no caller yet, and the auto-fallback would misbehave once it does.

Also a process note: the PR body says `make pre-pr` hasn't run yet — should be green before this leaves draft.

overtrue:
Pushed cb5d79c addressing the inline findings:

- the auto-mode unary fallback now runs outside the batch RPC deadline, so a slow pre-upgrade peer can no longer time out the whole batch and mark the disk failed
- adaptive concurrency is capped at 4x the configured baseline and can no longer ratchet up to batch size
- the HTTP tuning / adaptive / batch-read env gates are parsed once per process instead of on every request
- observe mode skips shard read-cost collection when stage metrics are disabled, and the local endpoint host list is cached instead of rebuilt per read
- the A/B runner accepts `--warp-extra-args --noclear` and the unused `warp_hosts_csv` is gone

Verified with `cargo fmt --all --check`, targeted tests in rustfs-ecstore (batch_processor, remote_disk, shard locality) and rustfs-rio, and `cargo clippy -p rustfs-ecstore -p rustfs-rio --all-targets -- -D warnings`.

Left alone: BatchReadVersion still has no production caller. That's a scope call for this PR — either wire the metadata fanout call-site or note it as follow-up work before this leaves draft.

PR #4237, opened 2026-07-03
perf(get,heal): fix GET hot-path overhead and heal checkpoint scaling

overtrue:
## Related Issues

- rustfs/backlog#800 — erasure-set heal checkpoint is O(N²)
- rustfs/backlog#801 — `DiskInfo.healing` is never populated
- rustfs/backlog#802 — object data cache probe runs after the data read
- rustfs/backlog#803 — GET hot-path redundant clones/lookups/metrics
- rustfs/backlog#804 — disk-read permit held for the whole body transfer

## Summary of Changes

Five verified fixes from a GET-path performance audit and a scanner/heal completeness audit (analysis, evidence and micro-benchmarks are in the linked backlog issues):

**Heal checkpoint scaling (backlog#800)** — `ResumeCheckpoint` object sets change from `Vec<String>` to `HashSet<String>` (`Vec::contains` was O(n) per healed object: measured 1.5 ms at n=1M vs 11 ns for `HashSet`), and per-object checkpoint/resume-state persistence is batched (1000 mutations or 5 s, whichever first) instead of rewriting the full checkpoint file per object. `complete_page()` prunes the sets at page boundaries so memory stays bounded; position updates and page boundaries still persist unconditionally. Object heal is idempotent, so a crash re-heals at most one throttle window. Legacy Vec-format checkpoint files still deserialize (covered by a unit test).

**Healing visibility (backlog#801)** — an erasure-set heal now writes a `healing.bin` marker into the meta bucket of the disks it is rebuilding and clears it on success. The triggering endpoints are plumbed via a new `HealRequest::heal_endpoints` field (set by the auto disk scanner; empty for admin/unclean-shutdown triggers, which intentionally mark nothing). `LocalDisk::disk_info` surfaces the marker through the existing 1 s cache, so scanner heal coordination (`should_heal`/`skip_healing`), lock-layer disk selection, and admin/metrics healing counts finally see rebuilds. Only the disks being rebuilt are marked — marking the whole set would make the lock layer avoid healthy disks.

**Cache probe ordering (backlog#802)** — new `GetObjectBodyCacheHook` trait in ecstore (registered once at startup via `OnceLock`, no new crate dependency) lets the app-layer object data cache serve full-object GET bodies inside `get_object_reader`, after metadata quorum resolution (etag/size known for identity validation) but before the erasure shard read/decode. Previously the lookup ran at output-build time, after all disk I/O had already happened, so a cache hit saved essentially nothing. The hook rejects range/part requests, remote/transitioned objects and SSE objects, and is a no-op `None` branch when the cache is disabled (the default).

**GET/HEAD hot-path redundancy (backlog#803)** — `ObjectInfo` is cloned for event notification only when an event will actually be built (`OperationHelper::wants_object_info()`); GET/HEAD currently suppress events so the clone was pure waste. `get_opts`/`put_opts`/`del_opts` resolve bucket versioning with one metadata-sys lookup instead of two. The `skip_verify_bitrot` and lock-acquire-timeout env reads are cached via `OnceLock` (test builds keep reading the env so `temp_env` overrides work). The io-priority assignment metric is no longer recorded twice when priority scheduling is enabled. `GetObjectInput` fields are cloned selectively instead of cloning the whole input.

**Disk permit starvation (backlog#804)** — permits from the disk-read semaphore (default 64) were held until the client finished downloading the body, so 64 slow clients could block every subsequent streaming GET while disks sat idle. The permit wait is now bounded (`RUSTFS_OBJECT_DISK_PERMIT_WAIT_TIMEOUT`, default 5 s; 0 restores the previous unbounded wait); on timeout the GET proceeds without a permit and the bypass is counted in a metric plus a structured warning. `DiskReadPermitReader` also releases the permit at body EOF instead of holding it until the stream drops.

## Verification

- `make pre-commit` — pass (fmt, unsafe-code allowances, architecture migration guardrails, logging guardrails, doc paths, workspace quick-check)
- `cargo clippy -p rustfs -p rustfs-ecstore -p rustfs-heal -p rustfs-config --all-targets -- -D warnings` — pass
- `cargo test -p rustfs --lib` — 2096 passed, 0 failed
- `cargo test -p rustfs-heal --lib` — 176 passed, 0 failed (includes new unit tests: checkpoint dedupe/page-pruning, legacy Vec-format checkpoint deserialization, persist-throttle batching)
- `cargo test -p rustfs --lib app::object_data_cache::hook` / `disk_read_permit_reader_releases_permit_at_eof` — new tests for the cache hook (hit + SSE skip) and permit EOF release — pass
- `cargo test -p rustfs-ecstore --lib set_disk` / `disk::` — 671 passed; 1 failure (`set_level_listing_trait_methods_use_existing_listing_implementation`) reproduced identically on unmodified main via `git stash` A/B (pre-existing global-state ordering flake, "No lock clients available")
- `cargo test -p rustfs-heal --tests` — `heal_integration_test` serial tests fail identically on unmodified main (pre-existing global `BucketMetadataSys` re-init flake), also verified via `git stash` A/B

## Impact

- New env var `RUSTFS_OBJECT_DISK_PERMIT_WAIT_TIMEOUT` (seconds, default 5, `0` = previous unbounded behavior). This is a behavior change for GETs that used to wait indefinitely for a disk permit: after the bound they proceed without one instead of timing out at the request level.
- New per-disk marker file `healing.bin` inside the meta bucket while an erasure-set heal rebuilds that disk; `DiskInfo.healing` (admin ServerInfo, realtime metrics) now reports `true` during rebuilds instead of always `false`.
- Heal checkpoint files now serialize the object sets from `HashSet` (order not stable); old files load fine, and a mid-upgrade crash re-heals at most one throttle window of objects (heal is idempotent).
- New metric `rustfs.get_object.disk_permit.bypass.total`; the `io_priority_assignment` metric loses its double-count (dashboards tracking it will show roughly half the previous rate when priority scheduling is enabled).
- Object data cache behavior only changes when the cache is enabled (`RUSTFS_OBJECT_DATA_CACHE_*`, default disabled): hits are now served before the erasure read, via the `body_cache` reader-path label in GET path metrics.

## Additional Notes

- The remaining items from backlog#803 (passing `HeaderMap`/`ObjectOptions` by reference through the `ObjectIO` trait, `Arc`-returning bucket config getters) need cross-crate trait signature changes and are tracked in the issue as follow-ups.
- backlog#802's issue text proposed an app-layer pre-`get_object_info` variant (option B); implementation went with option A (in-ecstore hook) because `read_data=false` metadata reads bypass the SetDisks metadata cache, which would have made every cache miss pay a second full xl.meta fan-out.
- Fix-landing comments with verification details were posted on all five backlog issues.


rustfs/src/storage/options.rs
```diff
@@ -45,7 +45,44 @@ use crate::auth::AuthType;
...
 use crate::storage::storage_api::options_consumer::StorageObjectOptions as ObjectOptions;
+use s3s::dto::VersioningConfiguration;
+
+/// Fetch the bucket's versioning configuration once so callers can derive
+/// enabled/suspended state without repeated metadata-sys lookups per request.
+async fn bucket_versioning_config(bucket: &str) -> VersioningConfiguration {
+    match BucketVersioningSys::get(bucket).await {
+        Ok(cfg) => cfg,
+        Err(err) => {
+            tracing::warn!("{:?}", err);
```
houseme, line 58: Relevant information is more comprehensive

houseme (approved):

PR #4250, opened 2026-07-03
fix(s3): allow CopyObject to restore a historical object version

### LINKED ISSUE #4238: CopyObject incorrectly rejects self-copy of a historical object version and loses metadata
**Describe the bug**
When migrating from MinIO to RustFS (beta 8), found an incompatibility in the CopyObject implementation when copying a non-current object version onto the current object. Our application uses CopyObject with sourceVersionId to restore a previous object version by copying it onto the same bucket/key. Example Java code using AWS SDK:

```java
client.copyObject(request -> {
    request.destinationBucket(BUCKET);
    request.destinationKey(key);

    request.sourceBucket(BUCKET);
    request.sourceKey(key);
    request.sourceVersionId("...");

    request.metadataDirective(MetadataDirective.REPLACE);
});
```

On RustFS beta 8 there are two different incorrect behaviors:
1. Without MetadataDirective.REPLACE, RustFS rejects the request with:
`Cannot copy an object to itself. Source and destination must be different. (Service: S3, Status Code: 400, Request ID: null) (SDK Attempt Count: 1)`
However, the source is not actually the same object instance—the request explicitly references a historical object version via sourceVersionId. The source and destination share the same bucket/key but refer to different object versions.
2. When MetadataDirective.REPLACE is specified, the request succeeds, but RustFS does not preserve metadata correctly:
- metadata from the source version is not copied;
- Content-Type is lost;
- Web console shows errors while object inspect because it no longer has the correct content type.

**To Reproduce**
1. Enable bucket versioning.
2. Upload an object.
3. Upload a second version of the same object.
4. Call CopyObject with:
identical bucket/key for source and destination;
sourceVersionId pointing to the first version.

Observe:
RustFS returns Cannot copy an object to itself, or
if MetadataDirective.REPLACE is used, the request succeeds but object metadata (including Content-Type) is lost.

**Expected behavior**
RustFS should behave like MinIO (and AWS S3 semantics):
1. CopyObject should allow copying an older object version onto the current object when sourceVersionId is specified.
2. The self-copy validation should distinguish between:
same bucket/key/current object;
same bucket/key but different version.
3. When the operation succeeds, metadata should be handled according to MetadataDirective.
4. Content-Type and other object metadata must not disappear after the copy.

**Additional context**
RustFS version: beta-8 glibc
Environmets:
```
RUSTFS_ACCESS_KEY: ${S3_ACCESS_KEY}
RUSTFS_SECRET_KEY: ${S3_SECRET_KEY}
RUSTFS_CONSOLE_ENABLE: true
RUSTFS_OBJECT_GET_SKIP_BITROT_VERIFY: true # https://github.com/rustfs/rustfs/pull/2110
```

I’m not entirely sure whether the `MetadataDirective.REPLACE` behavior I described is actually a RustFS bug or an ambiguity in S3 semantics. Our application has only been tested against MinIO, where restoring a previous object version by calling `CopyObject` with `sourceVersionId` works correctly and preserves metadata as expected.

The AWS S3 documentation for `x-amz-metadata-directive` only states that metadata is either copied from the source object (`COPY`, default) or replaced with metadata provided in the request (`REPLACE`), but it does not explicitly describe how this should behave when the source and destination have the same bucket/key and the source is a historical version of the object.

RamakrishnaChilaka:
## Related Issues

Fixes #4238

## Summary of Changes

`CopyObject` rejected a same-key copy that carried a source `versionId`, so
restoring a historical version onto the current key (default `COPY` metadata
directive) returned `400 InvalidRequest: Cannot copy an object to itself` and
dropped the object's `Content-Type` and user metadata.

- **Guard fix** (`rustfs/src/app/object_usecase.rs`): the self-copy rejection now
  also requires that no source `versionId` was supplied. A same-key copy that
  carries a source version is a version restore and is allowed, matching AWS S3
  and MinIO (which reject a self-copy only when `srcOpts.VersionID == ""`).
- **Copy-source versionId normalization** (same file): trim the id, map a
  case-insensitive `"null"` to the nil UUID, and reject malformed ids with
  `InvalidArgument`, mirroring the existing GET/HEAD `get_opts` handling and
  avoiding a `500` from a downstream UUID parse failure.
- **Consistent bytes for version restore** (`crates/ecstore/src/store/object.rs`):
  when a versioned same-key restore provides a reader, write the fetched bytes
  through `put_object` (with the source metadata) instead of the zero-copy
  metadata-only version copy. The zero-copy path shares the source version's
  `data_dir`; for SSE-encrypted or compressed objects the handler re-encrypts
  the reader, so sharing the old bytes would leave the new version
  undecryptable. This matches MinIO, which forces a full rewrite
  (`metadataOnly = false`) when the source or target is encrypted.

## Verification

- `make pre-pr` (fmt, arch/logging guards, clippy, workspace tests, doc tests) — all passed
- `cargo test -p rustfs --lib execute_copy_object` (handler unit tests)
- `cargo test -p rustfs-ecstore --lib copy_object` (store unit tests)
- `cargo test -p e2e_test --lib copy_object_version_restore` (plain restore e2e)
- `cargo test -p e2e_test --lib copy_object_version_restore_sse` (SSE-S3 restore e2e)

## Impact

- Fixes an S3 compatibility bug: restoring a historical object version via
  `CopyObject` now succeeds and preserves data, `Content-Type` and user
  metadata, creating a new current version while leaving existing versions
  intact.
- No API, configuration, or deployment changes. Behavior for non-version
  self-copies (`REPLACE` directive, storage-class change) is unchanged.

## Additional Notes

New regression tests, both failing on `main` before the fix and passing after:
- `crates/e2e_test/src/copy_object_version_restore_test.rs` (plain object)
- `crates/e2e_test/src/kms/copy_object_version_restore_sse_test.rs` (SSE-S3)


houseme:
@codex review

PR #4283, opened 2026-07-05
fix: stabilize s3-tests delete key-limit coverage

houseme:
## Related Issues

Refs rustfs/backlog#835.
Fixes rustfs/backlog#836.
Fixes rustfs/backlog#837.
Fixes rustfs/backlog#838.
Refs rustfs/backlog#839 as a follow-up performance track; it is intentionally not implemented in this PR.

## Summary of Changes

- Apply a repo-owned patch to the pinned ceph/s3-tests checkout so DeleteObjects key-limit coverage no longer preloads 1001 objects and then lists them before asserting the protocol limit.
- Replace upstream s3-tests `datetime.utcnow()` usage in the patched cases with a timezone-aware UTC helper.
- Add a RustFS usecase regression test proving 1001 DeleteObjects entries are rejected with `InvalidArgument` before store lookup.
- Make list-path worker sends cancellation-aware so limit-triggered cancellation cannot hang behind a full output channel.
- Consolidate ListObjects list-path diagnostics through a lightweight logging context, avoiding duplicated tracing fields and unnecessary full `ListPathOptions` clones.
- Add s3-tests JUnit failure summarization to make future CI failures easier to identify from logs.

## Verification

- `cargo fmt --all`
- `cargo fmt --all --check`
- `bash -n scripts/s3-tests/run.sh`
- `git -C s3-tests apply --check ../scripts/s3-tests/patches/0001-rustfs-key-limit-timeout-and-utcnow.patch`
- `git diff --check`
- `cargo test -p rustfs-ecstore send_or_cancel_returns_false_when_cancelled_while_output_is_full`
- `cargo test -p rustfs execute_delete_objects_rejects_more_than_one_thousand_objects_before_store_lookup`
- `make pre-pr`

## Impact

- Stabilizes implemented s3-tests CI by removing unrelated ListObjects pressure from DeleteObjects key-limit tests.
- Keeps the server-side DeleteObjects protocol limit unchanged and covered by a focused regression test.
- Prevents canceled list workers from waiting indefinitely on a full result channel after a consumer has already reached the requested limit.
- Adds debug/error diagnostic context for future ListObjects failures without raising normal success-path log verbosity.
- Does not implement the #839 high-performance ListObjects project; that remains a separate track.

## Additional Notes

- The s3-tests checkout is pinned; local patches are applied immediately after checkout so CI remains reproducible.
- A local-only tracking package is kept under `docs/issue-tracking/s3tests-deleteobjects-list-timeout/`; it is ignored by this repository and is not part of the PR diff.

### DIFF
--- a/crates/audit/src/pipeline.rs
+++ b/crates/audit/src/pipeline.rs
@@ -247,7 +247,7 @@ impl AuditRuntimeView {
         registry.list_targets()
     }
 
-    pub async fn get_target_values(&self) -> Vec<rustfs_targets::SharedTarget<AuditEntry>> {
+    pub async fn get_target_values(&self) -> Vec<SharedTarget<AuditEntry>> {
         let registry = self.registry.lock().await;
         registry.list_target_values()
     }
--- a/crates/audit/src/registry.rs
+++ b/crates/audit/src/registry.rs
@@ -113,7 +113,7 @@ impl AuditRegistry {
     ///
     /// # Returns
     /// * `Option<Box<dyn Target<AuditEntry> + Send + Sync>>` - The removed target if it existed.
-    pub async fn remove_target(&mut self, id: &str) -> Option<rustfs_targets::SharedTarget<AuditEntry>> {
+    pub async fn remove_target(&mut self, id: &str) -> Option<SharedTarget<AuditEntry>> {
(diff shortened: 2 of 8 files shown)


houseme:
@codex review

PR #4300, opened 2026-07-06
fix(ecstore): require write quorum for metadata early stop

GatewayJ:
## Related Issues
N/A

## Summary of Changes
- Require metadata early-stop candidates to satisfy the object's write quorum instead of a weaker read quorum.
- Preserve the 2+2 erasure-set rule by using `data + 1` as the default write quorum when data and parity are equal.
- Add a caller opt-in gate around metadata early-stop so read-before-write and metadata-only paths can force full fanout, while GET body paths keep the explicit call-site decision.

## Verification
- `RUSTUP_TOOLCHAIN=stable cargo fmt --all -- --check`
- `RUSTUP_TOOLCHAIN=stable CARGO_HTTP_TIMEOUT=600 CARGO_HTTP_LOW_SPEED_LIMIT=1 CARGO_HTTP_LOW_SPEED_TIME=600 cargo test -p rustfs-ecstore metadata_quorum_accumulator --lib`
- `RUSTUP_TOOLCHAIN=stable CARGO_HTTP_TIMEOUT=600 CARGO_HTTP_LOW_SPEED_LIMIT=1 CARGO_HTTP_LOW_SPEED_TIME=600 cargo check -p rustfs-ecstore --lib`
- gh-1 low-impact IOChaos run passed with image `docker.io/rustfs/rustfs:fault-44cc0466-2e176aab-glibc`, artifact root `/root/fault-runs/io-eio-low-44cc0466-2e176aab-20260708T060522Z`.

## Impact
- Reduces the chance that metadata recovery or validation reads proceed from an early-stop subset that is insufficient for write-quorum durability.
- No API or configuration changes.

## Additional Notes
- Higher-impact gh-1 fault runs were blocked by environment pressure: one run triggered local-path rootfs DiskPressure, and one run hit a transient node NotReady during active IOChaos. The low-impact replay completed with all committed objects verified and no data corruption.

### DIFF
--- a/crates/ecstore/src/disk/disk_store.rs
+++ b/crates/ecstore/src/disk/disk_store.rs
@@ -812,6 +812,7 @@ impl LocalDiskWrapper {
                     recursive: false,
                     immediate: false,
                     undo_write: false,
+                    undo_delete: false,
                     old_data_dir: None,
                 },
             )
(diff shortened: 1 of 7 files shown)


houseme:
@codex review

loverustfs (approved):

PR #4313, opened 2026-07-06
fix(runtime): avoid restricted Linux io-uring failures

occcat:
## Related Issues

N/A

## Summary of Changes

- Stop enabling Tokio's `io-uring` runtime backend by default through Linux target dependency sections. Restricted Linux environments can return `EACCES` from `io_uring_setup`, which Tokio surfaces as `PermissionDenied` on filesystem operations.
- Use the bound listener address for startup endpoint logging instead of probing local interfaces. This keeps startup logging from depending on interface enumeration in restricted/containerized Linux environments.

## Verification

- `cargo fmt --all`
- `cargo fmt --all --check`
- `cargo tree -p rustfs --target x86_64-unknown-linux-gnu -e features | rg 'io-uring|io_uring'` (no matches)
- `RUSTFLAGS='--cfg tokio_unstable' cargo check -p rustfs --all-targets`
- `RUSTFLAGS='--cfg tokio_unstable' make pre-commit`
- `RUSTFLAGS='--cfg tokio_unstable' CARGO_BUILD_JOBS=2 NUM_CORES=2 make pre-pr`

## Impact

- Linux builds no longer opt into Tokio's io-uring backend by default, so file I/O uses Tokio's regular backend unless another dependency explicitly enables it.
- Startup endpoint logging no longer tries to discover a separate local interface IP; wildcard binds are reported as wildcard and logged.

## Additional Notes

- Observed in an Android/proot environment: `io_uring_setup(256, ...) = -1 EACCES`, after which Tokio file reads returned `PermissionDenied` and RustFS surfaced `DiskAccessDenied`.


cxymds:
Hi, thank you for your contribution! We believe this PR provides a valuable fix, and we'd like to merge it.

To proceed, please sign the CLA as soon as possible. If we don't receive the signed CLA after one day, we'll have to close this PR and reimplement the fix ourselves.

Thank you for your understanding and contribution!

PR #4337, opened 2026-07-06
feat(internode): optimize gRPC transport

houseme:
## Summary

Internode gRPC transport optimization, staged **P0–P3** — transport tuning, control/bulk channel
isolation, serialization convergence, and connection lifecycle/observability. Closes gaps vs MinIO's
`internal/grid`. **Every behavior change is env-gated; defaults keep the base build byte-for-byte unchanged.**

> Full design, env-flag table, compatibility/rollout analysis, and benchmark status:
> **rustfs/backlog#877**

## Stages

- **P0** transport tuning (`tcp_nodelay` + HTTP/2 windows) + codec message limit (4 MiB → 100 MiB) + payload metrics
- **P1** control/bulk channel isolation (opt-in) — a large transfer can't head-of-line block a lock RPC
- **P2** serialization convergence: JSON-fallback counter + encode-buffer presizing + gated msgpack-only send + **additive** `DeleteVersion` `_bin` fields
- **P3** peer online/offline metric (MinIO parity) + prewarm + self-healing offline bypass + idempotent read-only retry
- Tooling: msgpack + benchmark runbooks, one-click A/B driver, four-node compose forwards `RUSTFS_INTERNODE_*`

## Compatibility (full analysis in rustfs/backlog#877)

- Proto change is **purely additive** (`DeleteVersion(s)` `_bin` fields, existing tags unchanged); delete
  **always dual-writes**, server decodes **bin-first with JSON fallback** → **two-way compatible** on rolling upgrade.
- All **P1/P2/P3 behavior defaults off**; only P0 transport tuning is on by default (no protocol/semantic break).
  The 4 MiB→100 MiB limit only relaxes decoding; oversized messages already failed pre-upgrade (not a new regression).
- The only way to break a mixed-version cluster is manually enabling `RUSTFS_INTERNODE_RPC_MSGPACK_ONLY` before its
  documented preconditions — and it defaults off (single-env rollback).
- No S3/admin API, on-disk (`xl.meta`) format, auth, or quorum/consistency changes.

## Verification

- `make pre-commit` green (cherry-picked cleanly onto `main`); per-stage cargo check/test + clippy during development.
- A/B orchestration validated end-to-end on a real four-node docker cluster (env toggles confirmed in-container).
  Real before/after perf numbers are deferred to a dedicated bench host using the included driver — see rustfs/backlog#877.


houseme:
Thanks @copilot-pull-request-reviewer — all 3 comments addressed in 361a0953e.

**1. `execute_read_with_retry` health action (remote_disk.rs)** — correct. Retries now go through `execute_with_timeout_for_op_and_health_action` with `FailureHealthAction::IgnoreFailure` for every attempt except the last; only the final attempt uses `MarkFailure`. So a transient error on an early attempt no longer flips the disk faulty (which would short-circuit the remaining retries), and failures are no longer over-counted. No default impact — `RUSTFS_INTERNODE_IDEMPOTENT_READ_RETRIES` defaults to `0`.

**2 & 3. Poisoned-mutex peer-health helpers (io-metrics)** — correct. `record_peer_reachable`, `record_peer_unreachable`, `cluster_peer_is_offline` and `cluster_peer_should_bypass` now recover a poisoned lock via `PoisonError::into_inner()` instead of early-returning, so the offline gauge and bypass state stay functional after a panic (already unlikely — those critical sections do no fallible work).

Verification: cargo check/test on io-metrics + ecstore (105 rpc tests green), clippy clean on the touched files, and `make pre-commit` green (rebased onto the latest `main` merge).

Issue #3191, opened 2026-06-03
ListObjectsV2 timeout corrupts metadata cache, cascades to NoSuchBucket across unrelated prefixes


jonerikr:
**Describe the bug**

ListObjectsV2 operations on directories containing thousands of objects exceed rustfs's internal ~5s timeout. When the timeout fires, the shared metadata cache enters a corrupted state and all subsequent list requests — including against completely unrelated prefixes — return NoSuchBucket (HTTP 404) for ~15 seconds, even though the bucket exists and is healthy. The system recovers automatically.

This appears related to #3175 (same list_path_raw canceled + BrokenPipe pattern) but triggered at runtime by large directory listings, not on startup.

**To Reproduce**

Create a bucket with multiple directories, each containing thousands of small objects under a common prefix (e.g. an append-only index directory with 10,000+ files that grows continuously)
Issue concurrent paginated ListObjectsV2 requests against those large prefixes using AWS SDK v2 / boto3:

```
import boto3, itertools, time
from botocore.config import Config
from botocore.exceptions import ClientError

TABLES = ['table-1', 'table-2', 'table-3', 'table-4', 'table-5',
          'table-6', 'table-7', 'table-8', 'table-9', 'table-10']
PREFIX_PATTERNS = [
    ('{t}/',         '/'),
    ('{t}/_index/',  '/'),
    ('{t}/_index/000000000000.dat', ''),
]

client = boto3.client(
    's3',
    endpoint_url='http://localhost:9000',
    aws_access_key_id='ACCESS_KEY',
    aws_secret_access_key='SECRET_KEY',
    region_name='us-east-1',
    config=Config(
        signature_version='s3v4',
        retries={'max_attempts': 0},
        s3={'addressing_style': 'path'},
    ),
)

while True:
    for table, (prefix_tpl, delim) in itertools.product(TABLES, PREFIX_PATTERNS):
        prefix = prefix_tpl.format(t=table)
        try:
            client.list_objects_v2(Bucket='my-bucket', Prefix=prefix, Delimiter=delim)
        except ClientError as e:
            print(f"{e.response['Error']['Code']} HTTP {e.response['ResponseMetadata']['HTTPStatusCode']} "
                  f"prefix={prefix!r} msg={e.response['Error'].get('Message','')!r}")
        time.sleep(0.1)
    time.sleep(30)

```

[...]

PR #3188, opened 2026-06-03
fix(iam): prevent transient IAM walk timeout from crashing startup

houseme:
## Summary

- IAM startup metadata walk timeout on `.rustfs.sys/config/iam/` no longer kills the process. The server enters degraded mode and retries with exponential backoff (5s → 10s → 20s → 30s cap) until IAM recovers.
- `/health/ready` correctly reports 503 during degraded mode; existing `IamSysNotInitialized` fail-closed behavior is preserved for all IAM-dependent request paths.
- Fix admin path boundary matching so `/minio/administrator` and `/minio/adminx` are not incorrectly treated as admin routes.
- Admin GET requests without a body now get `Content-Length: 0` automatically.

## Root Cause

The crash loop in #3175 was caused by:

1. Cluster reboot → disk pressure → metadata walk timeout on `.rustfs.sys/config/iam/`
2. `init_iam_sys()` returned error
3. `main.rs` treated IAM init failure as fatal → `process::exit(1)`
4. `systemd` restarted the service → same timeout → crash loop

The underlying timeout is a transient condition (disk warmup after reboot), but the startup path amplified it into a full service outage.

### Core: IAM startup resilience (`rustfs/src/startup_iam.rs`)

- `bootstrap_or_defer_iam_init()`: attempts IAM init; on failure, enters degraded mode and spawns a background recovery task
- `spawn_iam_recovery_task()`: retries with exponential backoff (`compute_backoff_interval`), escalates log level to ERROR after 12 attempts (~5 min)
- `IamBootstrapDisposition::ReadyInline | Deferred`: callers distinguish immediate success from deferred recovery
- Test injection via `RUSTFS_TEST_IAM_FAIL_INIT_ATTEMPTS` and `RUSTFS_TEST_IAM_RETRY_INTERVAL_MS` for deterministic testing

### Startup flow (`rustfs/src/main.rs`, `rustfs/src/embedded.rs`)

- Replace blocking `init_iam_sys().await?` with `bootstrap_or_defer_iam_init().await?`
- `publish_ready_when_runtime_ready` only called on `ReadyInline`; deferred mode publishes readiness from the recovery task
- `ServiceStateManager` promoted to `Arc` for sharing with recovery task

### Admin path matching (`rustfs/src/server/prefix.rs`)

- New `is_admin_path()` uses `has_path_prefix()` to require exact match or `/`-delimited child, fixing false positives on `/minio/administrator` etc.
- Moved from `admin/router.rs` to shared `prefix.rs` for reuse

[...]

PR #4356, opened 2026-07-07
feat(ecstore): reclaim orphaned data dirs during heal

houseme:
## Summary

Reclaims orphaned physical data directories left under an object by pre-#3510
unversioned overwrites, closing the gap that heal's dangling-object logic never
covered. This is the automatic cleanup for the historical leaks behind #3231
(and the ListObjects timeouts / NoSuchBucket cascade in #3191).

### Background

Before #3510, an unversioned overwrite of a >128KiB object leaked one UUID-named
data dir per PUT (`find_unshared_data_dir_for_version` filtered versions through
`uses_data_dir()`, which is only true for tier-restored objects, so a normal
object's previous data dir was invisible to the cleanup path). #3510 stops new
leaks, but existing strays stay on disk forever — heal only removes *whole*
objects whose data is missing, never surplus data dirs of a healthy object. That
accumulated junk is what made ListObjects scan tens of GB and hit the 5s walk
timeout, which (pre-#3188) then faulted the drive and cascaded to `NoSuchBucket`.

### Changes

- **`SetDisks::reclaim_orphan_data_dirs`** (`io_primitives.rs`) — a fail-closed,
  quorum-safe sweep mirroring `purge_orphan_dir_object`:
  - referenced set = **union** of `get_data_dirs()` across every online replica's
    `xl.meta`, so a dir named by *any* replica is kept;
  - if a disk holds the object dir but `xl.meta` is missing/unparseable, the
    object is treated as degraded and **nothing** is removed (heal must run first);
  - only UUID-named subdirectories are ever considered; deletes are best-effort.
- **Heal integration** (`heal.rs`) — invoked at `heal_object`'s healthy tail,
  under the object write lock, so the background heal scanner and admin heal
  recover leaked space automatically. A reclaim failure never fails the heal.

### Tests

Four unit tests in `set_disk/mod.rs`:
- removes an unreferenced data dir, keeps the live one + metadata;
- no-op when every physical dir is still referenced;
- fail-closed abort when metadata is missing;
- cross-replica union preservation.

### Verification

- `cargo test -p rustfs-ecstore --lib reclaim_orphan_data_dirs` → 4 passed
- `cargo check -p rustfs-ecstore`, `cargo fmt --all`, `cargo clippy -p rustfs-ecstore --lib` clean

Refs #3231, #3191.

### DIFF
--- a/crates/ecstore/src/set_disk/core/io_primitives.rs
+++ b/crates/ecstore/src/set_disk/core/io_primitives.rs
@@ -54,7 +54,7 @@ use crate::set_disk::shard_source::ShardReadCost;
 use futures::stream::{FuturesUnordered, StreamExt};
 use metrics::counter;
 use std::{
-    collections::{HashMap, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     future::Future,
     pin::Pin,
     sync::OnceLock,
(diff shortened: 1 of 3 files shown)


sliekens:
@houseme Thanks for putting this together. While tracing it against a running build I ran into a gap that stops the reclaim from firing for the objects it targets — details below.

**`reclaim_orphan_data_dirs` is unreachable for healthy objects — the `disks_to_heal_count == 0` early return shadows it**

In `heal_object` ([`crates/ecstore/src/set_disk/ops/heal.rs`](https://github.com/rustfs/rustfs/blob/b540c7e2d0e65d80eb224241b1036a225b77ff23/crates/ecstore/src/set_disk/ops/heal.rs)):

https://github.com/rustfs/rustfs/blob/f84ba243a195e84da35f73dd5bc4d2d58bedb8b0/crates/ecstore/src/set_disk/ops/heal.rs#L211-L213

https://github.com/rustfs/rustfs/blob/f84ba243a195e84da35f73dd5bc4d2d58bedb8b0/crates/ecstore/src/set_disk/ops/heal.rs#L660-L672

The reclaim call at L664 is only reached when at least one disk needs healing (`disks_to_heal_count > 0`). An object that's fully readable and consistent — the "object is healthy here" case the reclaim comment describes — hits the `disks_to_heal_count == 0` early return at L212 and returns before the reclaim runs.

On a healthy deployment (e.g. single node, no degraded disks), every object with a leaked/orphaned data directory but an otherwise-valid `xl.meta` takes that early return. The reclaim path is therefore dead code for the objects that carry the leaked data dirs: an admin-triggered heal walks and "heals" them, reclaims nothing, and emits no `reclaimed orphaned data directories` log lines.

Fix: run `reclaim_orphan_data_dirs` on the `disks_to_heal_count == 0` path as well (gated on `!opts.dry_run`), rather than only after the out-dated-disk handling. The reclaim is already best-effort and targets healthy objects, so it belongs before/at the healthy early return rather than after it. I can open a PR along those lines.

Two related points from the same trace:

- Write-path convergence heals that rely on `heal_object` to sweep residual data dirs are no-ops for healthy objects, for the same reason.
- Object directories with no `xl.meta` aren't enumerated by the heal walk (it lists objects), so `remove: true` doesn't reach their leaked data dirs either — those need a separate sweep.

Let me know if I've misread the control flow anywhere.

houseme:
@sliekens Confirmed — your trace is correct. The reclaim only lives in the post-heal tail, below the `disks_to_heal_count == 0` early return, so a healthy object with a leaked pre-#3510 data dir returns before reaching it — dead code for exactly the objects it targets.

Opened #4781: runs the reclaim on the healthy early-return path too (gated on `!opts.dry_run`), via a shared `reclaim_orphan_data_dirs_best_effort` helper so both exits behave identically; fail-closed safety unchanged. Adds an e2e regression that fails on the current control flow and passes with the fix.

Same root cause covers your write-path point. The no-`xl.meta` orphan dirs are a genuinely separate gap — left out of scope for a dedicated sweep. Thanks for catching this.

PR #4373, opened 2026-07-07
test(multipart): disable fsync in concurrent resend test to fix flake

overtrue:
## Related Issues

N/A (fixes a flaky test in the `Test and Lint (rio-v2)` CI job; internal ref: backlog#853).

## Summary of Changes

`bucket::lifecycle::bucket_lifecycle_ops::tests::concurrent_resend_same_part_commits_one_generation` (added in #4329) fails intermittently under nextest's high parallelism with an `unwrap_failed` panic at the `outcome.expect(...)` call. The failing `Result` is `put_object_part` returning `Err(Lock(Timeout { resource: ".rustfs.sys/multipart/…@latest" }))` — the per-`uploadId` `put_object_part_commit` write lock timing out.

Root cause is a **flaky test**, not a real race in #4329's serialization:

- The commit lock correctly wraps only `rename_part`.
- `rename_part` performs a per-disk `fsync` when `drive_sync_enabled()` (default `true`).
- Under parallel-CI IO starvation, `fsync` latency is unbounded, so the cumulative hold time across the serialized commits can exceed the wall-clock lock-acquire deadline and one waiter times out. The generation invariant the test asserts is unaffected; only the "every resend succeeds without lock timeout" guard trips.

Fix (test-only): run the test with durable (`fsync`) writes disabled via a `#[serial]`-scoped, panic-safe helper (`with_drive_sync_disabled`) that temporarily sets `RUSTFS_DRIVE_SYNC_ENABLE=false` around the whole test future and restores it afterwards. This moves the unbounded `fsync` latency out of the commit critical section so per-commit holds are microsecond-scale and the deadline has a large, load-independent margin. All correctness assertions (single visible generation, exact byte match) and the lost-wakeup `NOTIFY_WAIT_CAP` regression guard are kept. A deterministic unit test `drive_sync_disabled_helper_scopes_env` verifies the helper disables fsync in scope and restores it on exit.

No production code changed.

## Verification

- `make pre-commit`
- `cargo test -p rustfs-ecstore --features rio-v2 --lib concurrent_resend_same_part_commits_one_generation` — 5/5 pass
- `cargo test -p rustfs-ecstore --features rio-v2 --lib drive_sync_disabled_helper_scopes_env` — pass
- `cargo clippy -p rustfs-ecstore --features rio-v2 --lib --tests` — clean
- Reproduced the original failure on the unmodified test (deadline forced to 1s + heavy IO/CPU load): ~31/48 runs panicked with `Lock(Timeout)`. The fixed test under the same stress config: 178 runs, 0 failures.

## Impact

Test-only change; no user-facing, API, config, or runtime behavior impact.

## Additional Notes

N/A

### DIFF
--- a/crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs
+++ b/crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs
@@ -3513,6 +3513,71 @@ mod tests {
         }
     }
 
+    // Run `test_fn` with durable (fsync) writes disabled, restoring the original
+    // `RUSTFS_DRIVE_SYNC_ENABLE` afterwards even if the test panics. Disabling
+    // drive sync strips the unbounded per-commit `fsync` latency out of the
+    // serialized `put_object_part_commit` critical section, so a burst of
+    // concurrent same-part resends cannot pile up past the lock-acquire deadline
+    // under parallel-CI IO starvation (the flake fixed here). The generation
+    // invariant the caller asserts is independent of whether shards are fsynced.
+    //
+    // The env is mutated only in the synchronous prologue/epilogue around the whole
+    // test future — before any concurrent task is spawned and after
+    // `catch_unwind().await` has driven every task to completion — and the caller is
+    // `#[serial]`, so the shared multipart test store (which reads
+    // `RUSTFS_DRIVE_SYNC_ENABLE` only on active rename/write paths) never reads it
+    // while it is being written.
+    //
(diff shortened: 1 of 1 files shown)


overtrue:
Superseded by #4370, which was already merged to main and fixes the same flaky test (`concurrent_resend_same_part_commits_one_generation`) with the same root-cause analysis (six legitimately serialized cross-disk commits exceeding the small default lock-acquire timeout under CI disk load). This PR was authored against stale main and did not see #4370. Closing as redundant.

PR #4381, opened 2026-07-07
chore(experiments): add io_uring cancel-safety spike (backlog#894)

houseme:
## Related Issues

rustfs/backlog#894 (P2 io_uring read backend, deferred by the P1.5 NO-GO gate in rustfs/backlog#893; tracking: rustfs/backlog#897). This PR is the "Spike 0" the P2 design calls out as a prerequisite: the cancel-safety prototype, archived per plan option B — prototype only, no production integration, feature stays un-started until the restart conditions in backlog#894 are met.

## Summary of Changes

Adds `experiments/io-uring-cancel-spike`, a **standalone-workspace** crate (its own `[workspace]` table, so the io-uring dependency does not enter the main Cargo.lock, build graph, or CI) that proves the ownership model P2's io_uring read backend requires:

- The read buffer AND the file handle are owned by the driver's pending (orphan) table from SQE submission until the CQE. Dropping the caller-side future never reclaims the buffer — it only abandons the result and (by default) submits `IORING_OP_ASYNC_CANCEL` to accelerate the CQE. Reclamation happens at exactly one place: the CQE handler.
- Spike finding worth folding back into the backlog#894 design text: the orphan table must own the **file handle** too (`Arc<File>`), not just the buffer — otherwise dropping the future closes the fd mid-read and a recycled fd number would receive the kernel's write.
- Probe submits a **real `IORING_OP_READ`** and verifies the CQE + data before accepting work (`io_uring_setup` succeeding is not enough under gVisor/seccomp), and classifies failures against the EACCES/EPERM/ENOSYS/EINVAL/EOPNOTSUPP restriction set from the P2 degradation contract.
- Shutdown cancels all in-flight ops and drains to `in_flight == 0` before the ring is dropped/unmapped.

Five tests cover: byte-exact correctness vs the file content, drop-before-CQE on a blocked pipe read (op provably in flight; buffer held until the CQE finally arrives), ASYNC_CANCEL reclaim via ECANCELED, a 256-op stress run with half the futures dropped (`delivered + orphan_reclaimed == submitted`), and shutdown drain. See `experiments/io-uring-cancel-spike/SPIKE.md` for the model diagram, invariants, and measured results.

This complies with `scripts/check_no_tokio_io_uring.sh`, which bans only the tokio io-uring runtime feature and explicitly allows an application-level, runtime-probed io_uring integration — the model this spike validates.

## Verification

- `./run-docker.sh` (in `experiments/io-uring-cancel-spike`) on a Linux kernel via Docker, two legs: default seccomp — `io_uring_setup` fails EPERM (the #4313 incident environment), probe classifies it as an expected restriction and all 5 tests degrade to a graceful skip, suite green; `seccomp=unconfined` — all 5 cancel-safety tests pass.
- `cargo fmt --check` clean inside the spike crate (it is outside the workspace, so root `cargo fmt --all` does not cover it).
- `make pre-commit` passes on the branch (fmt + arch checks + quick-check), confirming the new crate trips no workspace guard.
- Full `make pre-pr` intentionally not run: the diff adds only a workspace-excluded experiment crate; no workspace code, product behavior, or CI surface changes. `make pre-commit` plus the crate-focused Docker test matrix covers the changed surface.

## Impact

None on the shipped binary or the build: the crate is outside the workspace, `io-uring` does not enter the main Cargo.lock, and nothing links it. P2/P3/P4 remain deferred; this only archives the highest-risk piece of P2 knowledge in executable, tested form so a future restart does not re-derive it.

## Additional Notes

Draft on purpose — merge is optional and low-priority; the alternative archive location is the branch itself. If merged, `experiments/` becomes the home for future workspace-excluded prototypes. The driver intentionally uses a coarse poll loop instead of eventfd + `AsyncFd` reaping; production remainders are listed at the end of SPIKE.md.


houseme:
Closing as planned: this spike is an archive, not a merge candidate. The P1.5 benchmark gate keeps P2 (io_uring read backend) deferred, so the prototype stays parked on branch `houseme/p2-spike0-uring-cancel-safety` with its validated cancel-safety model, tests, and SPIKE.md findings.

Reopen (or re-cut from the branch) when the restart conditions in rustfs/backlog#894 are met — local NVMe-class hardware plus a re-run of the benchmark gate showing >=15-20% attainable headroom. Conclusions and the fd-ownership design finding are recorded on rustfs/backlog#894.

PR #4386, opened 2026-07-07
fix(ecstore): make post-commit old data dir cleanup best-effort

overtrue:
## Related Issues

Refs rustfs/backlog#898 (core-storage audit finding writecommit-01; complements #864).

## Summary of Changes

After an overwrite reaches write quorum and `rename_data` atomically commits the new
version (the object is durable and immediately readable via GET/HEAD), `put_object` and
`CompleteMultipartUpload` deleted the superseded old data dir via
`commit_rename_data_dir(...).await?`. A below-quorum failure of that best-effort GC was
propagated with `?` and surfaced to the client as `ErasureWriteQuorum` → HTTP 503 — a
false-negative ACK for a write that already succeeded (and the stale-metadata cache was
left un-invalidated because the op was seen as failed).

This is a **deliberate divergence from MinIO**, which couples the same delete to the
write result (`erasure-object.go` `commitRenameDataDir`). The divergence is justified by
durability semantics (a committed, readable object must not be reported as failed), and
the disk-health signal MinIO derives from that 503 is preserved by other means (below).

Changes:
- `commit_rename_data_dir` returns a structured `OldDataDirCleanup` receipt and **never
  returns `Err`**. Pure, unit-testable helpers (`classify_old_data_dir_cleanup`,
  `map_cleanup_join_result`, `is_cleanup_not_found`) extracted. `JoinError` maps to a
  non-ignored `DiskError::other` (not `DiskNotFound`); `old == committed` and empty-dir
  are guarded; not-found normalizes to reclaimed.
- Call sites (`object.rs`, `multipart.rs`) consume the receipt instead of `?`, so a
  successful write stays `Ok` (restoring the metadata-cache invalidation self-heal).
- **Disk-health signal**: on residue, an object heal is enqueued via the existing heal
  channel (admission-deduped). The companion reclamation already exists on `main`
  (`heal_object` → `reclaim_orphan_data_dirs`), so the orphan old data dir is reaped —
  no permanent leak.
- **Leak metrics**: `rustfs_old_data_dir_cleanup_attempted/reclaimed_total`,
  `rustfs_old_data_dir_leaked_total`, `_below_quorum_total`.
- **Test-only fault-injection seam**: `#[cfg(test)]` object-keyed registry; the
  production hook is a `#[cfg(not(test))]` no-op — zero behavioral impact, no `Disk`
  enum change. Enables a real end-to-end test that forces the old-dir cleanup to fail on
  one disk and asserts the PUT returns 200 with the new object.

## Verification

- `make pre-pr` (green)
- `cargo test -p rustfs-ecstore --lib`: pure-function groups A/C, join-error mapping,
  end-to-end 200-on-cleanup-failure, `reclaim_orphan_data_dirs_*`, and the #864
  rollback-boundary guard — all pass.
- `cargo clippy -p rustfs-ecstore --all-targets`: 0 warnings.

## Impact

A below-quorum failure to GC the *superseded* old data dir no longer fails an
already-committed write; it is observable (metrics) and reconciled by heal. No change to
the commit/rollback path (#864 boundary untouched). No API or on-disk format change.

## Additional Notes

**P1 — pending maintainer review of the deliberate MinIO divergence before merge.** Found
by a multi-agent core-storage audit; design produced by an adversarial multi-role design
round (which corrected the initial "MinIO is best-effort" assumption). Open questions
(heal enqueue inline-await vs spawn; metric disk-index label; below_quorum trigger) are
defaulted per the finalized design and noted in code for confirmation. Design detail:
backlog#898.


houseme:
@copilot resolve the merge conflicts in this pull request

PR #4387, opened 2026-07-07
perf(ecstore): skip tmp parent dir fsync for write-then-rename paths

houseme:
## Summary

This is step 1 of the 4-step HP-1 fsync-convergence plan tracked in https://github.com/rustfs/backlog/issues/922 (parent overview: https://github.com/rustfs/backlog/issues/936). Steps 2-4 (merging/parallelizing the commit-point spawn_blocking calls, moving shard fdatasync into BitrotWriter shutdown, and closing the reliable_mkdir_all directory-entry gap) are intentionally out of scope here because they are gated on the crash-injection harness.

`write_all_internal` previously coupled file-content durability (fdatasync) and directory-entry durability (parent dir fsync) behind a single `sync: bool`. This PR replaces the bool with a `SyncMode` enum (`None` / `FileAndDir` / `FileOnly`) and switches exactly the two write-then-rename tmp write points to `FileOnly`, removing one parent-directory fsync per disk per non-inline PUT (4 per PUT on a 4-drive set).

## Call-site audit

All `write_all_internal` reachability was audited; only the two tmp write points whose file is immediately renamed away changed mode:

- `FileOnly` (changed): the tmp xl.meta write in the non-inline `rename_data` path (`write_all_private` call writing `{src_path}/xl.meta`, renamed onto the destination at the commit point), and the tmp write inside `write_all_meta` (uuid tmp file in `.rustfs.sys/tmp`, renamed onto the final xl.meta path right after; covers `delete_version`, `update_metadata`, `write_metadata`, and the delete-marker paths, all of which fsync the destination parent after the rename inside `write_all_meta` itself).
- `FileAndDir` (kept, unchanged behavior): `write_all_public` / `write_all` (format.json and other files that land directly with no rename) and the old-metadata rollback backup `xl.meta.bkp` in `rename_data` (stays where written and is the sole restore source for `undo_write`).
- Untouched (out of scope per the adversarial-review constraints): shard `sync_dir_files` before the data-dir rename, the commit renames themselves, the destination parent fsync after the commit rename, the inline `rename_data` branch, and `rename_part`.

## Durability argument

The persistence invariant of #4221 still holds: shard contents and xl.meta contents are durable before the commit rename (shards via `sync_dir_files` / `rename_part`, xl.meta via `sync_data` in `write_all_internal`, both unchanged), and the destination directory entry is durable before the PUT is acknowledged (the post-rename `fsync_dir` on the destination parent, unchanged). The tmp parent directory fsync we remove contributed nothing to that invariant: the safe-rename recipe is content fdatasync -> rename -> fsync destination parent; the rename itself removes the tmp directory entry, so persisting it is pure overhead, and if the machine crashes before the rename the PUT was never acknowledged, so there is no recovery obligation for the tmp file. With `RUSTFS_DRIVE_SYNC_ENABLE=false` the behavior is bit-for-bit unchanged (`SyncMode` collapses to `None` exactly where the old `sync && drive_sync_enabled()` collapsed to `false`).

## Tests

New unit tests use a test-only (`#[cfg(test)]`) fsync-dir recorder in `disk/os.rs`, since durability regressions are invisible to ordinary behavior assertions:

- `test_write_all_meta_skips_tmp_parent_dir_fsync_but_fsyncs_dst_parent`
- `test_write_all_public_still_fsyncs_parent_dir`
- `test_rename_data_non_inline_skips_tmp_parent_dir_fsync` (also asserts the destination parent and the rollback-backup parent are still fsynced)

## Verification

- [x] `cargo fmt --all`
- [x] `cargo check -p rustfs-ecstore`
- [x] `cargo clippy -p rustfs-ecstore --all-targets -- -D warnings`
- [x] `cargo test -p rustfs-ecstore --lib -- disk::local::test` (52 passed, includes all `rename_data` tests and the fail-before-backup injection test)
- [x] `cargo test -p rustfs-ecstore --lib -- disk::os:: set_disk::` (275 passed, covers the #4221 regression suites)
- [x] `cargo test -p rustfs-ecstore` (1987 passed; the 3 failures in `store::bucket::tests::bucket_delete_*` are a pre-existing parallel-run flake reproduced identically on clean main @ 62a31e4ec, `InsufficientWriteQuorum` from `make_bucket` under full-suite load, unrelated to this change)
- [x] `make pre-commit`

## Compatibility

- On-disk format: unchanged (no layout, naming, or xl.meta change; only the timing of one directory-entry fsync on tmp paths that are immediately renamed away).
- Public API / RPC / wire protocol: unchanged; `SyncMode` is crate-internal.
- Configuration: no new or changed options; `RUSTFS_DRIVE_SYNC_ENABLE=false` behavior is bit-for-bit identical to before.
- Default durability contract: unchanged — shard and xl.meta contents are durable before the commit rename, and the destination parent directory entry is durable before the PUT is acknowledged (argued in the durability section above and asserted by the new fsync-recorder tests).
- Failure/rollback paths: unchanged; the `xl.meta.bkp` rollback backup keeps its parent-directory fsync.

## Platform note

The profiling numbers motivating HP-1 were collected on macOS where `F_FULLFSYNC` and four volumes on one SSD amplify per-fsync cost, so absolute savings do not extrapolate; the structural saving (one fewer directory fsync per disk per non-inline PUT) is platform-independent, and Linux ext4/xfs benefit should be confirmed with a warp A/B run as noted in the backlog issue.


houseme:

houseme:
The failing `Test and Lint (swift)` check is a pre-existing breakage on main (clippy `type_complexity` in `crates/protocols/src/swift/expiration_worker.rs:661`, present on main@62a31e4e whose own CI run 28881271231 failed identically) and is unrelated to this PR, which does not touch the protocols crate. #4393 fixes it; once that merges, re-running this job should clear the check. The same applies to the sibling PRs #4389/#4392/#4394.

PR #4389, opened 2026-07-07
perf(ecstore): move speculative PUT-tail tmp cleanup off the hot path

houseme:
## Summary

Implements https://github.com/rustfs/backlog/issues/924 (HP-3, parent: https://github.com/rustfs/backlog/issues/936): the tail of `SetDisks::put_object` inlined an `await` on a purely speculative `delete_all(RUSTFS_META_TMP_BUCKET)` before returning. On the success path `rename_data` has already moved the data dir out of the tmp workspace, so this call is normally a no-op on a missing path (p50 ~49us), yet under fsync-heavy load the same-disk queueing behind it was measured to add ~9ms average (p99 77ms) directly to client-visible PUT latency.

## Changes

- **`crates/ecstore/src/set_disk/ops/object.rs`** (commit 1): on PUT success, the speculative tmp cleanup now runs on a `tokio::spawn`ed task off the response path. It is kept (not dropped) because it is the real backstop: `rename_data`'s `remove_std` only removes empty directories and silently ignores failures. The failure path (quorum loss / rollback) keeps the cleanup strictly inline, so a failed PUT never returns while its tmp shards are still on disk (preserves the state-residue hardening tracked by backlog#864 / backlog#898).
- **`crates/ecstore/src/disk/os.rs`** (commit 2, separate as required by the issue's adversarial-review notes): `reliable_rename_inner` no longer blindly retries `NotFound` — the retry does not recreate the missing source/parent, so the second rename always failed identically. The retry decision is extracted into `should_retry_rename` (any other error keeps the existing single retry). This helper is shared by the `rename_data` commit path via `rename_all`, so that path is covered by a dedicated regression test.

## Compatibility (no-breakage declaration)

- **Failure-path cleanup unchanged**: when `put_object` fails (writer-quorum loss, encode error, rename_data rollback), the `delete_all` cleanup is still awaited inline before returning — byte-for-byte the same call, same error handling, same log line. Only the success path moves to `tokio::spawn`.
- **Multipart untouched**: no change to `ops/multipart.rs`; its `delete_all` calls on `RUSTFS_META_MULTIPART_BUCKET` stay inline (they remove real leftovers, deferral would widen CompleteMultipartUpload/Abort races).
- **On-disk format unchanged**: no change to xl.meta, tmp/trash layout, volume names, or any wire/API surface. The same `delete_all` runs with the same arguments; only its scheduling changes on the success path.
- **Public behavior unchanged**: PUT results, error types, and object visibility are identical; the spawned cleanup targets a per-request UUID tmp dir that no other request can observe.
- **`reliable_rename_inner` NotFound no-retry is semantically equivalent for every caller.** The function has exactly two callers: `reliable_rename` (via `rename_all`) and `rename_all_ignore_missing_source`. The dst parent is created before the retry loop and nothing inside the loop recreates a missing source or parent, so a NotFound retry was guaranteed to fail identically — callers observed the same `FileNotFound` one syscall later. For `rename_all_ignore_missing_source` (move_to_trash/delete cleanup) the caller maps NotFound to `Ok(())` regardless, so outcomes are provably identical. For `rename_all` (all call sites enumerated: local.rs startup tmp/trash moves, write finalize at local.rs:2468, the rename_data commit path at local.rs:3711/3788, part/version moves at local.rs:4106/4162/4672) NotFound propagates as `FileNotFound` in both old and new code and every caller already handles it (rollback/quorum/log). Non-NotFound errors keep the existing single retry, so any transient-error rescue behavior is preserved. The only theoretical delta is a sub-attempt race where an external actor recreates the missing path between the two rename attempts — an accidental rescue no caller's contract relies on and impossible for the UUID-unique tmp paths involved.

## Scope notes (per adversarial verification on the issue)

- `ops/multipart.rs` `delete_all` calls on `RUSTFS_META_MULTIPART_BUCKET` are intentionally untouched: they delete real leftovers (not a no-op), and deferring them would widen the race window for duplicate CompleteMultipartUpload/Abort. Multipart is deferred to a separate evaluation.
- Crash safety: if the process dies before the spawned cleanup task runs, the stale tmp entry lives at most 24h — `cleanup_stale_tmp_objects` (24h expiry threshold, 5-minute background loop) reclaims it. Durability of committed data is unaffected.

## Regression tests

All new tests are hermetic: the `SetDisks` under test is constructed directly on formatted local disks (same pattern as the existing `ops/locking.rs` tests), so no global local-disk registry, lock clients, or ECStore instance is touched and the tests cannot interfere with the existing `store::bucket` shared test environment.

- `put_object_success_eventually_cleans_tmp_workspace`: a successful 1MiB PUT (non-inline, real tmp shards) leaves the `.rustfs.sys/tmp` workspace empty on all 4 disks, polling for the spawned task to land.
- `put_object_failure_cleans_tmp_workspace_inline`: a PUT into a missing bucket volume writes tmp shards, fails at `rename_data` with `VolumeNotFound`, and the tmp workspace is verified empty immediately after `put_object` returns — no polling, proving the failure path stayed inline.
- `rename_retry_never_retries_not_found` / `rename_retry_allows_single_retry_for_other_errors`: retry predicate regression guards.
- `rename_all_moves_existing_directory_tree`: guards the `rename_data` commit path that shares `reliable_rename_inner`.

## Verification

- [x] `cargo fmt --all`
- [x] `cargo check -p rustfs-ecstore`
- [x] `cargo clippy -p rustfs-ecstore --all-targets` (clean)
- [x] `cargo test -p rustfs-ecstore` — all new tests pass; the new `set_disk::ops::object::put_object_tmp_cleanup_tests` and `disk::os::tests` modules are green, as is everything this PR touches
- [x] `make pre-commit` (fmt + arch checks + quick-check)
- Pre-existing full-suite flake, not introduced here: 3 `store::bucket::tests::bucket_delete_*` tests fail with `InsufficientWriteQuorum` when the whole `--lib` suite runs in one process, on this branch (1989 passed / 3 failed) and identically on clean main @ 62a31e4ec (1984 passed / 3 failed, same 3 tests); they pass when run in isolation on both refs. The failures come from cross-test global local-disk registry interference among the suite's shared ECStore environments, which is exactly why this PR's new tests were written hermetically (they construct `SetDisks` directly and touch no global state).

## Platform note on expected gains

The profiling numbers were collected on macOS, where `F_FULLFSYNC` amplifies sync cost (~9ms average queueing measured on the PUT tail; nosync shows the same call at ~43us). The change removes that measured wait from the response path on macOS; Linux ext4/xfs gains are structurally expected (same-disk IO queueing is platform-independent, especially on HDD/slow disks) but must be confirmed by warp A/B re-measurement — no Linux numbers are claimed here.

Closes rustfs/backlog#924. Ref rustfs/backlog#936.

### DIFF
--- a/crates/ecstore/src/disk/os.rs
+++ b/crates/ecstore/src/disk/os.rs
@@ -217,7 +217,7 @@ async fn reliable_rename_inner(
     let mut i = 0;
     loop {
         if let Err(e) = super::fs::rename_std(src_file_path.as_ref(), dst_file_path.as_ref()) {
-            if i == 0 {
+            if should_retry_rename(&e, i) {
                 i += 1;
                 continue;
             }
@@ -237,6 +237,19 @@ async fn reliable_rename_inner(
     }
 
     Ok(())
+}
+
+/// Whether a failed `rename` in [`reliable_rename_inner`] should be retried.
+///
+/// Only the first failure is retried, and `NotFound` is never retried: the
+/// retry does not recreate the missing source or parent directory, so a second
+/// attempt is guaranteed to fail identically. Skipping it spares speculative
+/// cleanup renames (e.g. `move_to_trash` on an already-removed tmp path) a
+/// pointless second syscall. This predicate is shared by the `rename_data`
+/// commit path via `rename_all`, so any relaxation here must keep genuine
+/// transient errors retryable.
+fn should_retry_rename(err: &io::Error, attempt: usize) -> bool {
+    attempt == 0 && err.kind() != io::ErrorKind::NotFound
 }
 
 pub async fn reliable_mkdir_all(path: impl AsRef<Path>, base_dir: impl AsRef<Path>) -> io::Result<()> {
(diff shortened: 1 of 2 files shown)


houseme:

PR #4392, opened 2026-07-07
perf(ecstore): data-shards-only lockstep GET reads with stripe-aligned deferred parity engagement (opt-in)

houseme:
## Summary

Implements https://github.com/rustfs/backlog/issues/923 (HP-2, part of https://github.com/rustfs/backlog/issues/936): PR #4289's lockstep fix made every reconstruction-verifying GET read all data+parity shards per stripe, and the parity blocks are read, bitrot-hashed and then discarded (`erasure.rs` drops them when no data shard is missing). On a healthy 2+2 layout that is a deterministic 2x amplification of per-GET read bytes, IOPS and HighwayHash CPU (profiling: 16 bitrot reads/GET where data-only needs 8; 32 vs 16 on multipart), plus per-stripe waits on max(4 disks) instead of max(2).

This PR restores data-shards-only reads on the healthy lockstep path while keeping the backlog#832 alignment invariant, behind a new opt-in gate that defaults to current main behavior.

## What changed

- New `DeferredReaderStripeHandle` (`io_support/bitrot.rs`): a deferred bitrot reader now shares its pending-open state, and the handle can advance the unopened source by whole bitrot blocks using the exact `bitrot_encoded_range` geometry the reader was created with (identity mapping when `hash_size == 0`). An already-opened reader rejects the advance.
- `ParallelReader::read_lockstep` (`erasure/coding/decode.rs`) tracks per-slot engagement: with the gate on, only data slots are engaged while the object is healthy; parity slots stay unopened deferred readers. When a data shard is missing at stripe start, enough parity is engaged up front (in parallel with the surviving data reads); when a data shard dies during a stripe, additional parity is engaged after the fact — in both cases realigned to the current stripe through the stripe handle before its first read.
- Degraded stripes engage one parity beyond the decode quorum so reconstruction verification keeps an extra source to check against: `erasure.rs` only verifies when `available_shards > data_shards`, so engaging exactly the quorum would silently disable the inconsistent-source check (adversarial-review point 2 in the backlog issue).
- `fill_deferred_bitrot_readers` (`set_disk/core/io_primitives.rs`) records stripe handles for deferred slots and, gate-on only, swaps eagerly opened parity readers for unopened deferred ones so they remain engageable mid-object; the ready/error bookkeeping used by quorum decisions is untouched.
- Both GET paths carry the handles from reader setup: legacy duplex via `Erasure::decode_with_stripe_handles`, codec streaming via `ParallelReader::with_deferred_parity_handles` (`set_disk/read.rs`).

## Gate

`RUSTFS_GET_LOCKSTEP_DATA_SHARDS_ONLY_ENABLE` (bool, default `false`). Default behavior is identical to current main: every live shard reader is read on every lockstep stripe, and eagerly opened parity readers are kept as-is. A dedicated test (`test_lockstep_default_reads_all_shards_per_stripe`) locks the default read-all-shards behavior by byte counting.

## Compatibility

- On-disk format: no change. No xl.meta, shard layout, or bitrot block format changes; the stripe handle only recomputes the existing `bitrot_encoded_range` offsets.
- RPC / wire protocol: no change.
- Public API: no change. `Erasure::decode` and `create_deferred_bitrot_reader` keep their signatures; all new entry points are `pub(crate)`.
- Config semantics: one new env var (`RUSTFS_GET_LOCKSTEP_DATA_SHARDS_ONLY_ENABLE`), default off; no existing config item changes meaning.
- Default runtime behavior: unchanged, locked by tests that run the same scenarios in both gate modes.

## Invariants preserved (both gate modes)

- Lockstep alignment (backlog#832): a reader never contributes a block from a different stripe than the rest of the set; a parity reader that cannot be realigned (no pending deferred handle) is retired instead of read out of position.
- Short read -> `UnexpectedEof` -> whole-object reader retirement (backlog#799 B2): untouched in `BitrotReader::read`; an engaged parity reader that errors is retired for the rest of the object exactly like a data reader.
- Inconsistent-source rejection: reconstruction verification stays active on degraded reads (quorum + 1 engagement); a bitrot-valid but erasure-inconsistent parity shard engaged mid-object fails the read with `inconsistent read source shards` (new regression test).
- Per-stripe geometry tracking (backlog#799 B2): the offset/short-final-stripe accounting in `read_lockstep` is unchanged.

## Expected benefit and acceptance criteria

With the gate on, per-GET read bytes, IOPS and bitrot hash CPU are deterministically halved on healthy 2+2 objects (call-count proof: `test_lockstep_healthy_get_reads_only_data_shards` asserts parity streams contribute zero read bytes), and per-stripe waits shrink from max(data+parity) to max(data) disks. Wall-clock throughput gains depend on the bottleneck: near 2x is plausible for cold/disk-bandwidth-bound reads, less when the client or network is the limit. Acceptance should be judged by io metrics (per-GET read bytes halved) on disk-bound or cold-cache benchmarks, not by wall clock on the original profiling rig (macOS, loopback-limited, ~10% worker CPU).

## Coordination with backlog#868 item 2

The codec-bridge parity-verification false-reject fix tracked in https://github.com/rustfs/backlog/issues/868 (item 2, `erasure/codec/bridge.rs:198-280`) modifies the adjacent verification logic. This PR deliberately does not touch `bridge.rs`; it only changes which shards reach the verifier and keeps the `available > data_shards` contract. The two changes are independent, but whichever lands second should re-run the degraded-read regression tests here (`test_erasure_decode_rejects_inconsistent_parity_engaged_midstream`, `test_erasure_decode_recovers_when_data_shard_dies_midway`) plus the codec streaming bridge tests to confirm the combined behavior.

## Verification

- `cargo fmt --all` clean; `cargo check -p rustfs-ecstore` clean; `cargo clippy -p rustfs-ecstore --all-targets` clean; `make pre-commit` passed.
- `cargo test -p rustfs-ecstore`: 1991 passed, 3 failed. The 3 failures (`store::bucket::tests::bucket_delete_*`) reproduce identically on clean main (62a31e4ec) under full-suite parallelism and pass when run in isolation on this branch — pre-existing flake unrelated to this change.
- All existing `erasure::coding::decode` tests green; new tests cover: stripe-handle offset conversion for the streaming checksum format and the `hash_size == 0` identity format (`io_support::bitrot` tests), bitrot verification after a mid-object advance, advance rejection after open, healthy-path data-shards-only call counting, default-mode read-all-shards locking, mid-object data-shard death recovery (both formats, both gate modes), mid-stream inconsistent-parity rejection (both gate modes), and gate-controlled parity reader conversion in reader setup.
- e2e in-process fault injection (`cargo test -p e2e_test reliability_disk_fault_test`, single-node 4-disk 2+2: degraded reads with offline disk, bitrot read-through, fresh-disk heal): run repeatedly in both modes. All data verifications (baseline, degraded read with a disk offline, degraded-write read-back, bitrot read-through, post-heal) pass in both modes. The module has one intermittent failure — `Os { code: 66, DirectoryNotEmpty }` from the harness's `bring_disk_online` (`remove_dir_all`+`rename` racing the server's background read-repair heal recreating the disk directory) — which reproduces at a similar rate with the gate off (2 of 6 default-mode runs; the default mode is behaviorally identical to main, locked by the gate-off tests), always after all object verifications succeeded: pre-existing harness race, not a data-path failure.

## Residual risks and requested merge precondition

- The gate-on path replaces eagerly opened parity readers with deferred ones; a degraded stripe pays one extra open+seek when parity is engaged (accepted cost per the backlog issue).
- A parity reader without a stripe handle cannot be engaged past stripe 0 with the gate on and is retired conservatively; in production both GET paths always carry handles, and the only handle-less caller (inline-direct fast path) decodes single-stripe objects where stripe-0 engagement needs no handle.
- This is the same region as the backlog#832 and backlog#799-B2 incidents. Recommended merge precondition: re-run PR #4289's 4-node 64-concurrency fault-injection soak with `RUSTFS_GET_LOCKSTEP_DATA_SHARDS_ONLY_ENABLE=true` and confirm unexpected-EOF stays at the 0/1 baseline before considering any default flip.


houseme:

houseme:
Diagnosis of the three failing checks, none caused by this PR: (1) `Test and Lint (swift)` is the pre-existing main breakage fixed by #4393 (clippy type_complexity in the swift expiration worker tests; this PR does not touch protocols). (2) `Test and Lint` and `Test and Lint (rio-v2)` both fail on `rustfs-ecstore bucket::lifecycle::bucket_lifecycle_ops::tests::concurrent_resend_same_part_commits_one_generation` after ~30-32s — that is the lock-timeout guard from #4370 firing under full-suite parallel load on CI runners. Evidence it is unrelated to this change: the identical test fails on #4394 whose default build is a no-op relative to main, it passed on main@62a31e4e's own run, and it passes locally in 0.5s across 3 isolated runs on the same merge base. Re-running after #4393 merges should clear swift; the resend flake is tracked separately in the backlog.

PR #4397, opened 2026-07-07
perf(ecstore): add strict/relaxed/none durability modes over the drive-sync switch

houseme:
## Related Issues

- https://github.com/rustfs/backlog/issues/926 (HP-5, this PR is phase 1)
- Parent overview: https://github.com/rustfs/backlog/issues/936

## Summary of Changes

This is **phase 1 of HP-5**: a global, per-process durability tier configured by environment variable. Per-bucket tiers (bucket metadata + admin API, the second phase in rustfs/backlog#926) are deliberately out of scope and will come as a separate PR.

Introduces `RUSTFS_DURABILITY_MODE=strict|relaxed|none` as a tiered replacement for the binary `RUSTFS_DRIVE_SYNC_ENABLE` escape hatch. The tier machinery is built directly on the `SyncMode` enum introduced by #4387 (no parallel mechanism): each of the six drive-sync gate points in `crates/ecstore/src/disk/local.rs` (`write_all_meta`, `write_all_public`/`write_all_internal`, `rename_part`, and the shard-sync, commit-dir-fsync and inline gates of `rename_data`) now resolves its `SyncMode` from an `effective_durability(volume)` helper instead of re-reading a boolean from the environment.

- `strict` (default): the current mainline write path, preserved bit for bit. The three fsync-recorder tests from #4387 run unchanged under a pinned strict override and pass, which is the regression proof that strict is byte-identical to main.
- `relaxed`: keeps the payload fdatasync for erasure shard files and multipart parts, but leaves metadata commits (xl.meta contents, rollback backups, directory entries of commit renames) to the page cache. This aligns with MinIO's default durability posture and makes competitor comparisons semantically explainable.
- `none`: no fsync on the object data path; the tier equivalent of the old escape hatch, minus the system-critical exposure.
- The mode is resolved **once per process** and cached in a `OnceLock` (the old switch called `getenv` on every check, a dozen times per PUT, and could even flip mid-operation), with a serialized `#[cfg(test)]` override hook as required by the issue's adversarial correction 6. The resolved mode is logged at startup (`disk_local_durability_mode` event).
- `rename_data` resolves the tier once per commit, so a single commit can never be half-synced.

Per the issue's adversarial correction 4, all six gate points were audited and converted, including `rename_part` (UploadPart durability), which keeps the part payload fdatasync under `relaxed`.

## Compatibility

**Default behavior is zero-change**: with no environment variables set, the effective mode is `strict`, and every fsync decision is identical to current main. All new tiers are opt-in.

Legacy switch mapping (the old variable keeps working, unrenamed, with unchanged behavior):

| `RUSTFS_DURABILITY_MODE` | `RUSTFS_DRIVE_SYNC_ENABLE` | Effective mode |
| --- | --- | --- |
| unset | unset | `strict` (current default, bit-identical) |
| unset | `true` | `strict` (bit-identical to today's `true`) |
| unset | `false` | `legacy-off` (bit-identical to today's `false`: no fsync anywhere, system metadata included) |
| valid value | anything | the named mode wins |
| invalid value | any | warning logged, falls back to the legacy switch, then the default |

**System-critical pinning**: writes committing into `.rustfs.sys` (format.json, IAM/cluster config, bucket metadata) and `.minio.sys` are pinned to `strict` regardless of the configured tier. The scratch namespaces `.rustfs.sys/tmp` and `.rustfs.sys/multipart` stage in-flight user object data and follow the tier; pinning is keyed on the destination volume at commit time, so IAM objects staged in tmp still commit fully synced. Only `legacy-off` bypasses pinning, because its contract is the exact historical `false` semantics. The mode is server-side configuration only and cannot be influenced by any request header.

Power-loss semantics per tier (full matrix and deployment guidance in `docs/operations/durability-modes.md`):

| Tier | Shard/part payload | xl.meta / backups / dir entries | Power-loss guarantee |
| --- | --- | --- | --- |
| `strict` | fdatasync | fsync | Acknowledged writes survive whole-node power loss (current behavior) |
| `relaxed` | fdatasync | page cache | A drive can lose recently acknowledged versions entirely (not just a dirent rollback); inline objects get no per-object fsync at all; recovery relies on EC redundancy across independent power domains plus the #4221 unclean-shutdown heal. Multi-node with independent power feeds only; single-node must stay strict |
| `none` | page cache | page cache | Acknowledged objects can vanish wholesale; system-critical writes still pinned |
| `legacy-off` | page cache | page cache | Historical `RUSTFS_DRIVE_SYNC_ENABLE=false`: nothing synced, including system metadata; deprecated |

Per the issue's adversarial correction 1, this PR claims **no performance numbers for `relaxed`**: the published 26x figure was measured with everything off (`none`-equivalent) on macOS and must not be used to size `relaxed`; the docs say so explicitly. Relaxed benchmarking on Linux ext4/xfs is follow-up work before any recommendation.

## Verification

- `cargo fmt --all --check`
- `cargo clippy -p rustfs-ecstore --all-targets` (no warnings)
- `cargo test -p rustfs-ecstore` — 2013 passed, 4 failed; the 4 failures (`store::bucket::tests::bucket_delete_*` x3 and `bucket::migration::tests::migrates_real_minio_bucket_metadata_end_to_end`) are the known load-related flakes tracked in https://github.com/rustfs/backlog/issues/937: all four fail identically on the clean merge-base without this change and all pass in isolation on this branch
- `make pre-commit` — passed
- New unit tests: legacy-switch/env mapping matrix (`test_resolve_durability_mode_mapping`), per-tier gate truth table (`test_durability_mode_sync_gates`), system-critical volume classification and pinning (`test_system_critical_volume_classification`, `test_effective_durability_pins_system_volumes`)
- New fsync-recorder behavior tests (reusing the #4387 test infrastructure): relaxed skips metadata-commit dir fsyncs but keeps the shard-data sync (`test_rename_data_relaxed_keeps_shard_sync_skips_commit_fsyncs`, `test_write_all_meta_relaxed_skips_dst_parent_dir_fsync`), relaxed pins system volumes (`test_write_all_public_relaxed_pins_system_volume`), legacy-off syncs nothing including system metadata (`test_rename_data_legacy_off_skips_all_fsyncs`)
- Strict-equivalence regression proof: the three pre-existing #4387 fsync-recorder tests run under a pinned `strict` override with their assertions unchanged and pass

## Impact

- New environment variable `RUSTFS_DURABILITY_MODE` (opt-in; default `strict` keeps current behavior exactly).
- `RUSTFS_DRIVE_SYNC_ENABLE` remains supported with its historical semantics; existing deployments need no configuration change and see no behavior change.
- New operator documentation: `docs/operations/durability-modes.md` (per-tier power-loss guarantee matrix, deployment rules, pinning rules).
- The mode is now cached per process: changing it requires a restart (previously the boolean was re-read per call, which was never a supported way to retune a live instance and could leave a single PUT half-synced).

## Additional Notes

- Follow-ups tracked in rustfs/backlog#926: phase 2 per-bucket tiers, Linux A/B benchmarks for `relaxed`, and crash-injection harness coverage for the documented per-tier semantics.
- The durability enum is positioned so the io_uring write path (rustfs/backlog#896) can consume it directly.

### DIFF
--- a/crates/ecstore/src/disk/local.rs
+++ b/crates/ecstore/src/disk/local.rs
@@ -233,12 +233,18 @@ const ENV_RUSTFS_OBJECT_MMAP_READ_METHOD: &str = "RUSTFS_OBJECT_MMAP_READ_METHOD
 const RUSTFS_OBJECT_MMAP_READ_METHOD_MMAP_COPY: &str = "mmap_copy";
 const RUSTFS_OBJECT_MMAP_READ_METHOD_DIRECT_READ_COPY: &str = "direct_read_copy";
 
-/// Fsync writes and commit-point renames so acknowledged data survives power loss.
-/// Disabling trades durability for latency: data acknowledged with 200 OK may only
-/// live in the page cache and be lost on a whole-node power failure.
+/// Legacy binary switch for commit-point durability (fsync writes and renames).
+/// Kept for compatibility: `true` maps to the `strict` durability mode (the
+/// default), `false` keeps its historical semantics of disabling every fsync
+/// on this disk, system-critical metadata included. Superseded by
+/// `RUSTFS_DURABILITY_MODE`, which takes precedence when both are set.
 /// Default: true.
 const ENV_RUSTFS_DRIVE_SYNC_ENABLE: &str = "RUSTFS_DRIVE_SYNC_ENABLE";
 const DEFAULT_RUSTFS_DRIVE_SYNC_ENABLE: bool = true;
+
+/// Durability tier for object data-path writes: `strict` (default) | `relaxed` | `none`.
+/// See docs/operations/durability-modes.md for the power-loss guarantee matrix.
+const ENV_RUSTFS_DURABILITY_MODE: &str = "RUSTFS_DURABILITY_MODE";
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 enum LocalReadCopyMethod {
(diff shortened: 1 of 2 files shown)


houseme:
Linux 4-node × 4-drive (XFS) tier A/B on this branch's build (same binary, `RUSTFS_DURABILITY_MODE` only, fresh cluster per run, `warp mixed --concurrent 64 --duration 5m --noprefix`):

**10MiB objects — tiers are equivalent (within noise), as the mechanism predicts** (shard fdatasync dominates and relaxed keeps it): strict Total 5010.0 MiB/s, PUT avg/p99 160.0/306.5 ms; relaxed Total 4936.0 MiB/s, PUT avg/p99 155.9/323.6 ms.

**256KiB objects — relaxed delivers the small-object win**: PUT avg 41.6 → 29.9 ms (−28.1%), PUT p50 38.7 → 28.4 ms (−26.6%), PUT p99 96.1 → 60.7 ms (−36.8%), PUT throughput 377.4 → 416.1 obj/s (+10.3%), total mixed ops 2515.9 → 2774.0 obj/s (+10.3%); GET and DELETE are neutral (GET avg 18.9 → 18.2 ms, DELETE avg 59.0 → 57.5 ms).

Interpretation: per-object metadata/directory fsyncs amortize poorly on small objects, which is exactly the segment backlog#645 flagged against MinIO; relaxed removes them there while large-object durability cost (shard fdatasync) is retained by design. Numbers are single 5-minute runs per cell on shared Azure managed disks — direction is consistent across all four op types but the usual noise-band caveat applies until a nightly A/B (HP-14 item 4) exists.

PR #4407, opened 2026-07-08
perf(ecstore): support per-bucket durability tier overrides

houseme:
Phase 2 of the durability-tier work (phase 1 was #4397). Lets a bucket override the process-wide `RUSTFS_DURABILITY_MODE` with its own `strict`/`relaxed`/`none` tier, resolved at every commit point through `effective_durability(volume)`, so operators can relax small-object-heavy buckets while keeping stricter defaults elsewhere. Backs https://github.com/rustfs/backlog/issues/938, tracked under https://github.com/rustfs/backlog/issues/936.

## What

A per-bucket override is stored as a `durability.json` RustFS extension entry in the bucket metadata file (read/written through the existing metadata subsystem, so it inherits the `x-rustfs-internal-*` / `x-minio-internal-*` dual-key MinIO interop convention). The override is published to the disk layer through the same bucket-metadata cache-invalidation path that already syncs targets and other per-bucket config — no new channel. `effective_durability` resolves in this order: the legacy full-off switch short-circuits first and keeps its historical semantics (never pinned, per-bucket overrides do not apply under it); system-critical volumes (`.rustfs.sys`, `.minio.sys`) are pinned to `strict` and can never carry an override; otherwise a valid per-bucket override wins over the process-wide mode; otherwise the global mode applies. An admin GET/PUT handler exposes the per-bucket configuration. `legacy-off` is deliberately not representable per bucket — it only exists for the process-wide `RUSTFS_DRIVE_SYNC_ENABLE=false` switch. Semantics are documented in `docs/operations/durability-modes.md`.

## Compatibility

- On-disk format: additive only — a new optional `durability.json` extension entry; absent/empty/invalid all mean "inherit the global mode". No change to existing metadata layout.
- Default behavior: unchanged. With no bucket override, a bucket follows the global mode, which defaults to `strict` and stays byte-for-byte identical to before (verified by the phase-1 fsync-recorder tests, which pin the global mode to strict and still pass unchanged).
- System buckets: cannot be relaxed; pinned to `strict` regardless of any override attempt (asserted by a dedicated test).
- Legacy switch: `RUSTFS_DRIVE_SYNC_ENABLE=false` keeps its exact prior behavior; per-bucket overrides are inert under it.
- Public API / RPC / wire protocol: unchanged except the additive admin durability endpoint.

## Verification

- `cargo fmt --all`; `cargo clippy -p rustfs-ecstore -p rustfs --all-targets` clean
- `cargo test -p rustfs-ecstore --lib durability` 11 passed; `bucket::durability` 3 passed; new `metadata_sync_publishes_and_clears_durability_override` covers publish/clear on install/delete
- `cargo test -p rustfs --lib durability` 3 passed (admin handler: mode normalization, inherit-serializes-as-null, request parsing)
- `make pre-commit` passed

Per-bucket relaxed tiers should be validated on the Linux 4-node bench (per #926's tier A/B, which showed 256KiB PUT p99 −36.8% for global relaxed); this PR is the mechanism, the bench follow-up applies it selectively.


houseme:
The failing `Test and Lint (rio-v2)` check is the current main-wide test-build breakage — `missing field \`ctx\` in initializer of store::ECStore` at crates/ecstore/src/store/multipart.rs:410, introduced when #4437's test constructor raced with the PR that added the required `ctx` field. It is unrelated to this PR (HP-5b does not touch that constructor) and affects every open PR's test build. The fix is pending in #4442; once it merges to main, re-running this job will clear the check. No change needed here.

PR #4411, opened 2026-07-08
perf(ecstore): Linux O_DIRECT shard-write path off the commit critical section

houseme:
## Summary

`create_file` on the local disk writes erasure shard and multipart part data through the page cache, so the commit-point `sync_dir_files` fdatasync inside the `rename_data` critical section has to flush the whole shard's dirty pages. Hotpath profiling of the PUT path (create_file avg 67.52ms vs 1.51ms with no sync; BitrotWriter::write p95 40.86ms) shows this as a concentrated cold-flush of ~2MiB of dirty pages per drive, and under concurrent PUTs on one device the writebacks interfere and lengthen every fdatasync in the rename critical section. MinIO writes the same data files with O_DIRECT so its commit-point fdatasync only leaves inode metadata plus a device FLUSH.

This PR adds an opt-in Linux O_DIRECT streaming writer for `create_file`, reusing the O_DIRECT read infrastructure landed in #4365 (statx STATX_DIOALIGN alignment probe, aligned bounce buffer, a per-disk `supported` latch, and a one-time fallback log) rather than building a second alignment/fallback stack. Incoming bytes are memcpy'd into an aligned bounce buffer on the reactor; each full aligned batch and the shutdown tail are flushed on the blocking pool, the same offloading posture as the buffered `tokio::fs::File` writer it replaces. The trailing sub-alignment remainder is written buffered after clearing O_DIRECT with fcntl (MinIO's recipe). Because the shard bytes reach the device during the streaming write phase, the unchanged commit-point `sync_dir_files` fdatasync degrades from flushing ~2MiB of dirty pages to a cheap metadata plus device FLUSH, shortening the `rename_data` critical section.

## Compatibility

Default off: the path is gated by the new `RUSTFS_OBJECT_DIRECT_IO_WRITE_ENABLE` environment variable (default false), whose naming mirrors the existing `RUSTFS_OBJECT_DIRECT_IO_READ_ENABLE`. When disabled the write path is byte-for-byte the previous buffered writer.

Linux only: the O_DIRECT writer is `#[cfg(target_os = "linux")]`. On non-Linux hosts the gate has no effect and `create_file` always uses the buffered writer.

Durability barrier preserved: no fsync is added or removed on the object data path. The commit-point `sync_dir_files` still fdatasyncs every shard file before the rename, so the change is durability-neutral; it only removes the dirty-page transfer cost from that fdatasync. Skipping the shard fdatasync entirely is a separate proposal (HP-1 step 3) and is deliberately not done here.

Durability tier interaction: the strict / relaxed / none tiers over `sync_dir_files` (from #4397) are unchanged. O_DIRECT direct-writes the shard bytes regardless of tier, and the tier still decides whether the commit-point fdatasync runs, so each tier keeps its exact power-loss guarantee.

Safe fallback, never a fake missing shard: filesystems that reject O_DIRECT (tmpfs, overlayfs, 9p) return EINVAL/EOPNOTSUPP, which latches the path off for the whole disk, logs once, and falls back to the buffered writer. An O_DIRECT open/write EINVAL is never surfaced as `InvalidInput`, because `to_file_error` maps `InvalidInput` to `FileNotFound`, which would masquerade as a missing shard and trigger a spurious erasure-coding rebuild (the trap confirmed in backlog#897); any such error is remapped to a generic write error so the write-quorum machinery sees the real failure.

## Testing

Unit tests cover the environment gate, the alignment / staging-capacity and tail-split math, an end-to-end `create_file` round trip across sizes crossing the alignment and multi-batch staging boundaries (buffered fallback on macOS, O_DIRECT on a block device), and the writer state machine driven over a plain file on Linux. Verified on macOS: `cargo fmt --all --check`, `cargo check -p rustfs-ecstore`, host `cargo clippy -p rustfs-ecstore --all-targets`, the new tests, and `make pre-commit`. Verified for Linux by cross-compiling with `cargo zigbuild --target x86_64-unknown-linux-gnu` (lib and `--tests`) and cross-`cargo clippy --target x86_64-unknown-linux-gnu --all-targets`, all clean.

Real O_DIRECT throughput on Linux NVMe (ext4/xfs, warp A/B) is deferred to the azure 4-node bench per the issue's GO/NO-GO gate; the development host is macOS and cannot exercise O_DIRECT, so no Linux performance numbers are fabricated here.

Refs: rustfs/backlog#927 (HP-6), rustfs/backlog#936.

### DIFF
--- a/crates/ecstore/src/disk/local.rs
+++ b/crates/ecstore/src/disk/local.rs
@@ -227,6 +227,20 @@ const DEFAULT_RUSTFS_OBJECT_DIRECT_IO_READ_ENABLE: bool = false;
 /// Default: 4MB.
 const ENV_RUSTFS_OBJECT_DIRECT_IO_READ_THRESHOLD: &str = "RUSTFS_OBJECT_DIRECT_IO_READ_THRESHOLD";
 const DEFAULT_RUSTFS_OBJECT_DIRECT_IO_READ_THRESHOLD: usize = 4 * 1024 * 1024;
+
+/// Enable O_DIRECT for erasure shard / multipart part data writes (Linux only).
+/// When enabled, `create_file` streams shard bytes straight to the device with
+/// O_DIRECT, so the commit-point `sync_dir_files` fdatasync no longer flushes
+/// ~2 MiB of dirty pages inside the `rename_data` critical section (it degrades
+/// to a cheap metadata/device FLUSH). Aligned whole blocks are written direct;
+/// the trailing sub-alignment remainder falls back to a buffered write after
+/// clearing O_DIRECT (MinIO's recipe). Durability is unchanged: the file is
+/// still fdatasynced at the commit point by the unchanged `sync_dir_files`.
+/// EINVAL/EOPNOTSUPP (tmpfs, overlayfs, 9p, ...) latch the path off and fall
+/// back to buffered writes for the whole disk. Non-Linux always falls back.
+/// Default: false (buffered writes via the page cache, as before).
+const ENV_RUSTFS_OBJECT_DIRECT_IO_WRITE_ENABLE: &str = "RUSTFS_OBJECT_DIRECT_IO_WRITE_ENABLE";
+const DEFAULT_RUSTFS_OBJECT_DIRECT_IO_WRITE_ENABLE: bool = false;
 const ENV_RUSTFS_OBJECT_MMAP_POPULATE_ENABLE: &str = "RUSTFS_OBJECT_MMAP_POPULATE_ENABLE";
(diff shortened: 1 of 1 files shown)


houseme:
The three failing `Test and Lint*` checks are the current main-wide test-build breakage (`missing field \`ctx\` in initializer of store::ECStore` at crates/ecstore/src/store/multipart.rs:410, from #4437 racing with the ctx-field PR). Unrelated to this PR — HP-6 only touches disk/local.rs and its non-test build (`cargo check -p rustfs-ecstore`) is clean. Every open PR's test build is affected; fix pending in #4442. Re-run after it merges. No change needed here.

PR #653, opened 2025-10-15
feat:  adapt to s3s typed etag support

overtrue:
## Summary
- bump `s3s` to commit f76013a52de2c8d44f6aa43182f1f1a938d2ad03 (upstream PR s3s-project/s3s#349) so the SDK natively formats RFC 7232 entity tags; this supersedes the manual quoting we added in rustfs/rustfs#592 and rustfs/rustfs#627
- update `crates/ecstore/src/client/object_api_utils.rs`:
  - `format_etag` now returns a typed `s3s::dto::ETag`, handling both strong (`"abc"`) and weak (`W/"abc"`) tags without double quoting
  - `extract_etag` keeps the existing precedence (`etag` metadata → `md5Sum`) but leaves the stored string untouched; the caller receives the typed ETag
- update `crates/ecstore/src/store_api.rs` to preserve backward compatibility:
  - when transforming `s3s::dto::CompletedPart`, convert `ETag::Strong/Weak` back to the quoted string format expected by the current metadata/persistence code paths
- make the networking helpers testable without real DNS:
  - add an injectable resolver (`set_mock_dns_resolver` / `reset_dns_resolver`) in `crates/utils/src/net.rs`
  - unit tests now rely on mock responses instead of hitting the live network, eliminating the flakiness we observed on bare-metal

## Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [x] Refactor
- [ ] Other:

## Related Issues

s3s-project/s3s#349, #592 and #627

## Checklist
- [x] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [x] Passed `make pre-commit`
- [x] Added/updated necessary tests
- [x] Documentation updated (if needed)
- [x] CI/CD passed (if applicable)

## Impact
- [ ] Breaking change (compatibility)
- [ ] Requires doc/config/deployment update
- [x] Other impact:

## Additional Notes

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

PR #4426, opened 2026-07-08
refactor(ecstore): add instance-scoped runtime context carrier

overtrue:
## Summary

Phase 5 of the runtime-identity isolation epic (backlog#939), **Slice 1 of 8**. Phases 1–4 (#653) unified access to ECStore's runtime globals behind accessor APIs but left the underlying `OnceLock`/`RwLock` statics process-global — so deleting the single-instance guard (rustfs#3243) would not actually isolate anything. This slice introduces the isolation *carrier* those phases deferred and migrates one representative group onto it, with **zero single-instance behavior change**.

## What changed

- **New `runtime::instance::InstanceContext`**, carried as an `Arc<InstanceContext>` field (`ctx`) on each `ECStore`. The isolation vehicle is the **object graph**, not `tokio::task_local` (which does not cross `tokio::spawn`, and ecstore has 155+ internal spawns): spawn closures already `move` the object `Arc`s, so method bodies read `self.ctx` without threading a `&ctx` argument through call sites.
- **Collapsed three erasure booleans into one source of truth.** `GLOBAL_IS_ERASURE` / `GLOBAL_IS_DIST_ERASURE` / `GLOBAL_IS_ERASURE_SD` were three independent `RwLock<bool>` truth sources kept consistent only by an implicit derivation inside `update_erasure_type`. They are now a single `RwLock<SetupType>` with *derived* predicates, removing the torn intermediate states that design allowed.
- **Backward-compatible facade.** `is_erasure()` / `is_dist_erasure()` / `is_erasure_sd()` / `update_erasure_type()` keep their signatures and forward through `current_ctx()`, which resolves to the live store's `ctx` once published and to a process `bootstrap_ctx()` before then.
- **Bootstrap adoption.** `ECStore::new` (and both test construction sites) adopt the same `bootstrap_ctx()` `Arc`, so the startup-time write and the post-construction read hit one cell — byte-for-byte single-instance equivalence.
- **Instance-scoped `setup_is_*(&self)` predicates** reading `self.ctx` — the seam the data path moves onto in Slice 2.
- **Arch guard** updated to allow the migrated identifiers in the new `runtime/instance.rs` owner file.

## Tests (6 new, all green)

- `erasure_predicates_match_legacy_derivation` — replays the old `update_erasure_type` derivation and checks every `SetupType` byte-for-byte (incl. `DistErasure ⇒ is_erasure`).
- `fresh_context_is_all_false`, `bootstrap_ctx_is_stable_singleton`, `distinct_contexts_do_not_share_state`.
- `store_adopts_bootstrap_context` — `Arc::ptr_eq` proof the constructor lands on the bootstrap cell.
- `instance_context_carrier_isolates_two_stores` — **non-strawman** end-to-end proof: two real `ECStore`s with distinct `ctx` read their setup type through the real `&self` predicates; DistErasure vs ErasureSD do not pollute each other.

## Verification

- `make pre-pr` green: fmt, unsafe-code, architecture-migration, logging-guardrails, tokio-io-uring, doc-paths, clippy, and the full `cargo nextest` suite (**7852 passed, 0 failed, 116 skipped**) + doc tests.

## Scope / follow-ups

This is deliberately one slice. Per the issue design each slice is independently mergeable/rollback-able, with hard ordering constraints (Slice 5's heal `'static` contract change is a separate prerequisite PR; Slice 7's per-instance cancellation must land before Slice 8's multi-instance acceptance). Slices 2–8 follow in separate PRs.

Refs #939

### DIFF
--- a/crates/ecstore/src/config/com.rs
+++ b/crates/ecstore/src/config/com.rs
@@ -1740,6 +1740,7 @@ mod tests {
                 endpoints,
                 crate::disk::format::FormatV3::new(1, 2),
                 lockers,
+                crate::runtime::instance::bootstrap_ctx(),
             )
             .await;
 
--- a/crates/ecstore/src/core/sets.rs
+++ b/crates/ecstore/src/core/sets.rs
@@ -33,6 +33,7 @@ use crate::{
     error::StorageError,
     layout::endpoints::{Endpoints, PoolEndpoints},
     object_api::{GetObjectReader, ObjectInfo, ObjectOptions, PutObjReader},
+    runtime::instance::InstanceContext,
     runtime::sources as runtime_sources,
     set_disk::SetDisks,
     store::init_format::{check_format_erasure_values, get_format_erasure_in_quorum, load_format_erasure_all, save_format_file},
@@ -77,6 +78,10 @@ pub struct Sets {
     pub default_parity_count: usize,
     pub distribution_algo: DistributionAlgoVersion,
     exit_signal: Option<Sender<()>>,
+    /// Per-instance runtime context, inherited from the owning `ECStore`
+    /// (issue #939, Slice2). Each `SetDisks` in `disk_set` shares this same
+    /// `Arc`, so the whole pool reads one instance's runtime identity.
+    pub(crate) ctx: Arc<InstanceContext>,
 }
 
 impl Drop for Sets {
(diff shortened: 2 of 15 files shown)


overtrue:
Superseded by #4413, which merged the same backlog#939 Slice 1 (InstanceContext carrier + erasure SetupType consolidation) with an evolved implementation. Resolving this branch's conflicts against main would leave an empty/regressive diff (it predates #4406's lock-loss fence and #4435's no_lock handling), so closing instead. Slices 2–8 will continue on top of main's version.

Issue #3775, opened 2026-06-23
Why do the console object count and size still as 0? The rustfs version is 1.0.0-beta.8 and system is windows 10


zhenglipeng:
<img width="1668" height="280" alt="Image" src="https://github.com/user-attachments/assets/48ea9564-f154-45e2-b189-e0046be89655" />

[rustfs.log](https://github.com/user-attachments/files/29231815/rustfs.log)

majinghe:
Did you try to add env vars `RUSTFS_SCANNER_ENABLED=true` and try again?

marshawcoco:
Thanks for the report. This may be related to the data usage / scanner path, but the attached log only shows `ListObjects` returning `AccessDenied`, so we first need to rule out a permission issue.

Could you please confirm:

1. Are you logging in to the console with root credentials or another user/service account?
2. After uploading objects, does the console stay at `0` for several minutes, or does it recover after a short delay?
3. Is `RUSTFS_SCANNER_ENABLED` explicitly set? If not, please try setting `RUSTFS_SCANNER_ENABLED=true` and restart RustFS once.

I tested the backend data usage path locally. Immediately after upload, the count may be briefly stale, but it caught up within a few seconds:

```text
t=00s total=872
t=01s total=960
t=02s total=1000
```
So a short delay is expected, but persistent `0` plus the `AccessDenied` log points more toward permissions or scanner not running.

zhenglipeng:
> 你尝试添加环境变量`RUSTFS_SCANNER_ENABLED=true`并重试了吗？

How to configure it on Windows?

zhenglipeng:
<img width="668" height="189" alt="Image" src="https://github.com/user-attachments/assets/ed72b8f2-8475-496b-914b-68f6b100ff7a" />

<img width="1880" height="945" alt="Image" src="https://github.com/user-attachments/assets/420fb0ed-947a-43f7-ae79-a24756d21cc8" />

<img width="1415" height="610" alt="Image" src="https://github.com/user-attachments/assets/a2509a8b-a583-422d-bea7-600c3fee6d24" />

[rustfs.log](https://github.com/user-attachments/files/29326790/rustfs.log)
How to resolve permission issues?

zhenglipeng:
@marshawcoco

marshawcoco:
Thanks for the updated screenshots and log.

[...]

PR #4490, opened 2026-07-08
fix(admin): refresh datausage live bucket usage

marshawcoco:
## Related Issues
Follow-up to #3775.

## Summary of Changes
Refresh live bucket usage from the object layer when serving admin data usage info, not only versioned buckets. This keeps `/rustfs/admin/v3/datausageinfo` aligned with the authoritative bucket usage path that already backs `/rustfs/admin/v3/accountinfo`.

This addresses the remaining Windows reproduction after #4472: `/accountinfo` stayed correct after restart, but `/datausageinfo` and the Console bucket list could still regress to a partial scanner snapshot such as `agent` showing fewer objects than the object layer and `cccc` temporarily showing `0 / 0`.

## Verification
- `cargo fmt --all --check`
- `cargo test -p rustfs app::admin_usecase::tests::execute_query_data_usage_info_returns_internal_error_when_store_uninitialized`

## Impact
Admin data usage responses and Console bucket stats should no longer expose stale or partial scanner cache values when live bucket usage can be refreshed from the object layer. The query may perform additional per-bucket usage refresh work when `/datausageinfo` is requested.

## Additional Notes
This does not change metacache/listing error semantics. The observed `DiskOngoingReq` list failures remain a separate follow-up from the data usage stats freshness path.

### DIFF
--- a/crates/ecstore/src/set_disk/core/io_primitives.rs
+++ b/crates/ecstore/src/set_disk/core/io_primitives.rs
@@ -613,6 +613,10 @@ pub(in crate::set_disk) fn resolve_read_part_from_responses(
 }
 
 pub(in crate::set_disk) fn shard_read_costs_for_disks(disks: &[Option<DiskStore>]) -> Vec<ShardReadCost> {
+    if disks.is_empty() {
+        return Vec::new();
+    }
+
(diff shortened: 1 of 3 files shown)


marshawcoco:
@codex review

PR #4499, opened 2026-07-08
fix(object-capacity): mark partial-failure refreshes degraded and merge over disk cache

overtrue:
## Problem

A full capacity refresh with partial disk failures (S06 in the audit at rustfs/backlog#1010) commits the surviving subset's sum as a fresh **exact** cluster total:

- `into_capacity_update` drops `per_disk` when `had_partial_errors`, so `update_capacity` skips the disk-cache reconcile block and writes the partial sum straight into the cache with `is_estimated=false`, `source=Scheduled`. No field carries the "some disks failed" fact.
- Meanwhile the complete disk cache keeps the failed disk's last-known bytes, so the next dirty-subset refresh recomputes the merged (higher) total — the reported capacity **oscillates** between the partial sum and the cache-merged total while a disk is flapping, with no degradation marker on either side. On a 2×100GB cluster this looks like 100GB↔200GB flapping in admin StorageInfo.

## Fix

- Add a `degraded` flag to `CapacityUpdate` and `CachedCapacity`, set when the scan behind the update had partial errors. `is_estimated` keeps its sampling-only semantics.
- On a degraded full refresh, surface only the disks whose own scan fully succeeded (a disk with intra-disk errors under-counts, so its stale cache entry is closer to the truth) and merge them over a complete disk cache: failed disks keep their last-known values, so the published total no longer dips to the surviving-subset sum and bounces back. The disk cache is never replaced or marked complete from a degraded refresh (preserves the rustfs/backlog#805 non-pollution guarantee).
- Without a complete disk cache the partial sum is still the best available value; it is committed unchanged but now visibly marked degraded.
- Expose the flag in refresh/admin capacity logs and a new `rustfs_capacity_degraded_readings_total{source}` counter.

## Tests

- `test_update_capacity_degraded_full_refresh_merges_cache_and_does_not_oscillate` — full→partial→full rounds report 200→200→200, degraded flag toggles true only for the middle round.
- `test_update_capacity_degraded_with_empty_per_disk_serves_merged_cache` — no trustworthy per-disk entries: the complete cache still backs the published total.
- `test_update_capacity_degraded_without_complete_cache_keeps_partial_sum` — no cache to merge: partial sum committed, marked degraded, cache not marked complete.
- `test_partial_scan_report_produces_degraded_update` — real scan with one dead disk yields a degraded update that never replaces the disk cache and only clears dirty flags of surviving disks.
- `test_into_capacity_update_partial_failure_filters_intra_disk_errors` — a disk with intra-disk errors is excluded from cache merging.

## Verification

- `cargo test -p rustfs-object-capacity` — 51 passed
- `cargo clippy -p rustfs-object-capacity -p rustfs-io-metrics --all-targets` — clean
- `cargo check -p rustfs` (service.rs consumer) — pass
- `cargo fmt --all` + all five architecture guard scripts — pass

Ref: rustfs/backlog#1014 (S06+relates to #805 O2), parent audit rustfs/backlog#1010

### DIFF
--- a/crates/io-metrics/src/capacity_metrics.rs
+++ b/crates/io-metrics/src/capacity_metrics.rs
@@ -55,6 +55,13 @@ pub fn record_capacity_update_completed(source: &'static str, duration: Duration
     .increment(1);
 }
 
+/// Record a committed capacity reading that is degraded: the refresh behind it
+/// had partial disk failures, so some disks kept last-known values.
+#[inline(always)]
+pub fn record_capacity_degraded_reading(source: &'static str) {
+    counter!("rustfs_capacity_degraded_readings_total", "source" => source).increment(1);
+}
+
 /// Record failed capacity update.
 #[inline(always)]
 pub fn record_capacity_update_failed(source: &'static str) {
--- a/crates/object-capacity/src/capacity_manager.rs
+++ b/crates/object-capacity/src/capacity_manager.rs
@@ -30,9 +30,9 @@ use rustfs_config::{
     ENV_CAPACITY_WRITE_TRIGGER_DELAY,
 };
 use rustfs_io_metrics::capacity_metrics::{
-    record_capacity_current_bytes, record_capacity_dirty_disk_count, record_capacity_refresh_inflight,
-    record_capacity_refresh_joiner, record_capacity_refresh_result, record_capacity_update_completed,
-    record_capacity_update_failed, record_capacity_write_operation,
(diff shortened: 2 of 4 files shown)


overtrue:
CI failure `shard_read_costs_for_empty_disk_set_are_empty` is pre-existing main breakage fixed by #4496 (merged); rebased onto latest main to pick it up and re-run CI. Unrelated to this PR's changes.

PR #4516, opened 2026-07-08
fix(obs): count scanned versions independent of ILM

houseme:
## What

`rustfs_scanner_versions_scanned_total` and `rustfs_ilm_versions_scanned_total` were both fed from the same helper `scanner_lifecycle_checked_versions`, which reads the `checked` counter of the scanner's `Lifecycle` work source. Nothing in the production scan path ever records `checked` for the `Lifecycle` source (only `executed`/`queued`/`missed`/`failed` are recorded there, and ordinary object scans record `checked` on the `Usage` source instead). As a result both metrics were effectively pinned at zero: a cluster could scan billions of versions and still report `versions_scanned = 0` while `objects_scanned` climbed, making scan-coverage dashboards and the versions/objects ratio meaningless. This is backlog#995 (OBS-09) from the observability audit.

The two metrics also have different intended meanings that were conflated: `rustfs_scanner_versions_scanned_total` is documented as "object versions scanned" (all versions the scanner walks, independent of ILM), whereas `rustfs_ilm_versions_scanned_total` is "object versions checked for ILM actions" (only the subset evaluated on buckets that have lifecycle rules).

## How

- `crates/common/src/metrics.rs`: add a lifetime counter `lifetime_versions_scanned` with `record_scanner_versions_scanned(count)`, surface it as `ScannerMetricsReport.versions_scanned`, and populate it in `report()`. Pure addition — no existing counter semantics change.
- `crates/scanner/src/scanner_io.rs` (`get_size`): record every version of each scanned object (`object_infos.len()`) before `apply_actions`, so the count accrues for all buckets regardless of lifecycle configuration.
- `crates/scanner/src/scanner_folder.rs`: after the ILM `Evaluator::eval` succeeds (reached only for lifecycle-configured buckets), record `object_opts.len()` on the `Lifecycle` source's `checked` counter — the honest "versions checked for ILM actions" number, including versions that evaluate to `NoneAction`.
- `crates/obs/src/metrics/stats_collector.rs`: the scanner collector now reads the real total from `metrics.versions_scanned`; the ILM collector keeps reading `scanner_lifecycle_checked_versions` (the ILM-checked subset). Each metric now measures what its name and help text claim.

## Verification

- `cargo test -p rustfs-common` — new `report_counts_scanned_versions_independent_of_lifecycle` (scanned versions accrue with zero lifecycle activity) and `report_tracks_lifecycle_checked_versions_separately` (scanner total vs ILM-checked subset diverge correctly), plus the existing `scanner_lifecycle_checked_versions_*` tests still pass.
- `cargo check -p rustfs-common -p rustfs-scanner -p rustfs-obs`.
- No change to the hot path beyond two relaxed atomic adds per scanned object / per ILM evaluation batch.

Closes backlog#995.

### DIFF
--- a/crates/common/src/metrics.rs
+++ b/crates/common/src/metrics.rs
@@ -750,6 +750,11 @@ pub struct Metrics {
     last_scan_cycle_duration_millis: AtomicU64,
     last_scan_cycle_objects_scanned: AtomicU64,
     last_scan_cycle_directories_scanned: AtomicU64,
+    /// Lifetime count of object versions walked by the data scanner, counted
+    /// for every version regardless of whether any lifecycle rule applies.
+    /// This is the honest source for `rustfs_scanner_versions_scanned_total`;
+    /// ILM-checked versions are tracked separately on the `Lifecycle` source.
+    lifetime_versions_scanned: AtomicU64,
     last_scan_cycle_bucket_drive_scans: AtomicU64,
     last_scan_cycle_bucket_drive_failures: AtomicU64,
     last_scan_cycle_yield_events: AtomicU64,
@@ -1106,6 +1111,9 @@ pub struct ScannerMetricsReport {
     pub oldest_active_path_age_seconds: u64,
     pub life_time_ops: HashMap<String, u64>,
     pub life_time_ilm: HashMap<String, u64>,
+    /// Lifetime object versions scanned, independent of ILM configuration.
+    #[serde(default)]
+    pub versions_scanned: u64,
     pub last_minute: ScannerLastMinute,
     pub active_paths: Vec<String>,
     pub current_scan_mode: String,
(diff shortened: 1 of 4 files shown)


houseme:
### Relationship to #4443 — this completes an incomplete fix, it is not a duplicate

#4443 ("correct versions scanned metrics") switched the data source from `life_time_ilm.values().sum()` to `scanner_lifecycle_checked_versions()`, which reads the `checked` counter of the scanner's `Lifecycle` work source. However, **nothing in the production scan path ever records `checked` for the `Lifecycle` source** — only `executed`/`queued`/`missed`/`failed` are recorded there, and ordinary object scans record `checked` on the `Usage` source. So after #4443 the metric still reads a permanently-zero counter: `rustfs_scanner_versions_scanned_total` and `rustfs_ilm_versions_scanned_total` remain 0 even while `objects_scanned` climbs.

This PR supplies the missing recording points and separates the two conflated meanings:

- `scanner_folder.rs`: record `Lifecycle.checked += object_opts.len()` after the ILM evaluator runs (lifecycle-configured buckets only) → `rustfs_ilm_versions_scanned_total` becomes the real "versions checked for ILM actions".
- `scanner_io.rs` + `common/metrics.rs`: add a `lifetime_versions_scanned` counter incremented for every version the scanner walks (any bucket, independent of ILM) → `rustfs_scanner_versions_scanned_total` becomes the real total scan coverage.
- `stats_collector.rs`: the scanner collector now reads the new total; the ILM collector keeps `scanner_lifecycle_checked_versions` (the ILM-checked subset).

Verified `git cherry origin/main` shows this branch is not upstream, and `record_scanner_versions_scanned` does not exist on `main`. Without this change #4443's source stays at zero in production.

PR #4521, opened 2026-07-08
fix(obs): shut down global telemetry guard

houseme:
Fixes the telemetry-guard shutdown gap from the observability audit (OBS-02, rustfs/backlog#988).

The `OtelGuard` — the only ordered shutdown point for the tracer/meter/logger providers, the Pyroscope agent, and the `tracing_appender` non-blocking `WorkerGuard` — was parked in a never-dropped `static`, so its `Drop` never ran and telemetry was never flushed on exit. Subscriber install also panicked on double-init instead of returning an error.

- Store the global guard as `Arc<Mutex<Option<OtelGuard>>>` and add `shutdown_global_guard()` which `take()`s and drops it, flushing providers before exit.
- Wire `shutdown_observability_guard()` into the server and embedded shutdown paths (including before the error-path exits).
- Switch subscriber installation from `.init()` to `.try_init()`, mapping failure to `TelemetryError::SubscriberInit` instead of panicking on a second init.

Addresses rustfs/backlog#988.

### DIFF
--- a/crates/obs/src/error.rs
+++ b/crates/obs/src/error.rs
@@ -27,9 +27,11 @@ pub enum GlobalError {
     #[error("Failed to set a global recorder: {0}")]
     SetRecorder(#[from] metrics::SetRecorderError<crate::Recorder>),
     #[error("Failed to set global guard: {0}")]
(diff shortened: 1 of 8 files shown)


houseme:

PR #4524, opened 2026-07-08
fix(obs): export resettable metrics as gauges

houseme:
Fixes the counter-semantics distortion from the observability audit (OBS-05, rustfs/backlog#991).

Non-monotonic / resettable values (IAM sync durations and per-minute plugin auth stats, audit and notification-target message totals) were declared as counters, so `WrappedCounter::absolute` (fetch_max semantics) masked every decrease/reset and inflated the exported series.

- Export the resettable/non-monotonic IAM, audit and notification-target values as gauges (`gauge.set(value)`) instead of counters, keeping genuinely monotonic totals as counters.
- Align the corresponding schema descriptors so metric type matches the emitted semantics.

Addresses rustfs/backlog#991.


houseme:

PR #4525, opened 2026-07-08
fix(obs): stop exporting fake cpu categories

houseme:
Fixes the fabricated CPU-time categories from the observability audit (OBS-07, rustfs/backlog#993).

`build_system_cpu_stats` hardcoded `user`/`nice`/`steal`/`avg_iowait` to 0 and put the total CPU utilization into the `system` field, while the schema exported these as independent CPU-time categories — so `cpu_user` was always 0 and `cpu_system` over-reported.

- Stop exporting the unmeasured CPU categories (and the mislabeled total-as-system), reporting only what is actually measured, with schema help text updated to match.

Addresses rustfs/backlog#993.


houseme:

PR #4526, opened 2026-07-08
fix(obs): drop placeholder drive series

houseme:
Fixes the placeholder drive metrics from the observability audit (OBS-08, rustfs/backlog#994).

`DriveDetailedStats` hardcoded inode counts, IO/timeout/availability error totals, waiting IO, API latency and all iostat rates to 0 while the schema advertised them as measured; and `perc_util` was filled with space-usage percent under an IO-busy-time name.

- Drop the placeholder drive series that were hardcoded to 0 (so the series are absent rather than falsely reporting 0) and correct the `perc_util` semantics.

Addresses rustfs/backlog#994.


houseme:

PR #4591, opened 2026-07-09
fix(ecstore): only evict remote lock channel on transport failures

### LINKED ISSUE #4567: Leader-election lock contention tears down healthy gRPC channels (rejection conflated with connection failure)
**Version:** 1.0.0-beta.8 (official `rustfs/rustfs:1.0.0-beta.8` image)
**Deployment:** Docker Swarm, MNMD, 4 servers x 1 disk, EC 2+2, plain HTTP between nodes.
Internode transport already softened via `RUSTFS_INTERNODE_HTTP2_KEEPALIVE_TIMEOUT_SECS=20`
and `RUSTFS_INTERNODE_RPC_TIMEOUT_SECS=30` (with the beta.8 defaults of 3s/10s the pattern
below escalates into a much larger eviction storm with lock RPC timeouts).

## Summary

During routine `leader.lock` election rounds, an expected/normal negative outcome of
`acquire_exclusive` (lock already held) results in `Evicting failed gRPC connection` for the
per-peer channel, at `crates/protos/src/lib.rs:211`, followed by
`Evicted stale connection from cache` (`crates/common/src/globals.rs:112`). The channel was
healthy; the lock was simply busy. The channel is then re-dialed and the cycle repeats on the
next election round, producing a constant background of WARN-level evictions (roughly one
burst per member per minute, one eviction per non-holder peer) on an otherwise idle, healthy
cluster.

## Repro signature (every member shows this; healthy 4-node cluster, no user traffic)

```text
INFO ... remote_locker.rs:267: remote acquire_exclusive for .rustfs.sys/leader.lock@latest   (x3 peers)
INFO ... remote_locker.rs:348: remote release for .rustfs.sys/leader.lock@latest:<owner-uuid>
WARN ... protos/src/lib.rs:211: Evicting failed gRPC connection: http://<peer>:9000/data
WARN ... globals.rs:112: Evicted stale connection from cache: http://<peer>:9000/data
```

Notably the evictions consistently target the NON-holder peers, while the channel to the
current lock holder stays clean - i.e. the failure that triggers eviction is the negative
acquire outcome, not a transport error. `curl` to the same peers' `/health` and 33KB HTTP
transfers succeed 100% of the time while the evictions occur; DNS resolution is verified
fresh and correct.

## Why this hurts

1. Constant channel churn: every election round pays connection re-establishment to most
   peers, and any in-flight RPC multiplexed on the evicted channel fails collaterally.
2. Operator noise: hundreds of WARN evictions/hour on a healthy cluster make real transport
   failures impossible to spot.
3. Interaction with the admin info aggregation (separate issue): a snapshot taken while a
   member's channel is in its evicted instant silently drops that member from `servers[]`,
   so the console membership list rotates randomly.
4. With the beta.8 default internode timeouts (3s keepalive-timeout / 10s RPC), this churn
   escalates under load into lock RPC timeout storms (sub-second lock budgets + shared
   channel head-of-line blocking) that ejected a member on our cluster and prevented rejoin.

## Suggested behavior

- Distinguish error classes in the gRPC client wrapper: only transport-level errors
  (connect/reset/deadline on the channel itself) should evict the cached channel. An
  application-level negative result (lock busy/denied) must not.
- The channel-isolation work visible on `main` (separate control/bulk connections so bulk
  transfers cannot head-of-line block lock RPCs) looks related and very welcome - this issue
  is the complementary client-side error-classification half.

## Environment notes

No clock skew (NTP verified on all nodes), no resource pressure (~100MiB RSS of 2GiB limit,
~0% CPU), names resolve correctly and instantly at every check.

## Related (checked, not duplicates)

- #3097 - lock RPC timeouts under high-concurrency PutObject via tokio thread starvation
  (closed). Different trigger: that report needs heavy load and CPU starvation; this one
  reproduces on a fully idle cluster and is about error CLASSIFICATION (a routine negative
  acquire evicting a healthy channel), not about latency under load.
- #3116 - warp test write-lock failures (closed); same lock subsystem, different symptom.

overtrue:
## Summary

Fixes #4567.

During routine `leader.lock` election rounds an idle, healthy cluster produced a constant background of WARN-level gRPC channel evictions (one per non-holder peer per round). The primary cause — a busy/denied lock being conflated with a connection failure — is already fixed on `main`: a denied lock returns a normal `Ok(GenerallyLockResponse { success: false, .. })` response and never reaches the eviction path.

This change closes the remaining error-classification gap. The remote lock RPC wrapper (`RemoteClient::execute_rpc`) evicted the cached channel on **any** `tonic::Status`. Genuine transport failures (connect refused/reset, keepalive/deadline on the channel) surface as a `tonic::Status`, but so do server-produced application statuses returned on a perfectly healthy channel:

- `Unauthenticated` / `PermissionDenied` from the signature interceptor
- `Internal` / `FailedPrecondition` when the peer's lock service is not ready yet
- `InvalidArgument`, `ResourceExhausted`, `Unimplemented`, `Aborted`, ...

Evicting the channel for those cannot help — the channel is fine — and every spurious eviction re-dials the connection **and** advances the peer toward the offline threshold via `record_peer_unreachable`, i.e. needless churn plus a nudge toward marking a healthy peer offline.

## Change

- Add `RemoteClient::is_transport_failure(&tonic::Status)`: true when the status carries an underlying transport error as its `source`, or its code is one of `Unavailable` / `DeadlineExceeded` / `Unknown` / `Cancelled`. Server-produced application statuses are reconstructed from grpc-status trailers on the client and have no `source`, so they classify as non-transport.
- Gate the `Ok(Err(status))` eviction in `execute_rpc` on `is_transport_failure`; the RPC still fails (returns a `LockError`) but the channel is only evicted for real transport failures. The client-side timeout arm still evicts, unchanged.
- This mirrors the transport-vs-application (`should_evict`) distinction already used on the remote-disk RPC path.

## Testing

- `cargo test -p rustfs-ecstore --lib cluster::rpc::remote_locker` → 7 passed.
  - New unit test `test_is_transport_failure_classifies_only_channel_level_errors` asserts transport codes / sourced statuses evict, and application statuses (`Unauthenticated`, `Internal`, `InvalidArgument`, ...) do not.
  - Existing eviction tests (connection-refused → `Unavailable`, client timeout) still pass.
- `cargo clippy -p rustfs-ecstore --lib --all-features` → clean.

### DIFF
--- a/crates/ecstore/src/cluster/rpc/remote_locker.rs
+++ b/crates/ecstore/src/cluster/rpc/remote_locker.rs
@@ -93,6 +93,29 @@ impl RemoteClient {
         resource_summary == ".rustfs.sys/leader.lock@latest"
     }
 
+    /// Classify a `tonic::Status` as a transport-level failure of the cached channel.
+    ///
+    /// Only genuine connection problems (connect refused/reset, keepalive/deadline
+    /// on the channel itself, cancelled in-flight streams) justify evicting and
+    /// re-dialing the cached channel. A broken transport surfaces either as one of
+    /// the codes below or carries the underlying hyper/h2 error as its `source`.
+    ///
+    /// Server-produced application statuses (`Unauthenticated`/`PermissionDenied`
+    /// from the signature interceptor, `Internal`/`FailedPrecondition` when the
+    /// peer's lock service is not ready yet, `InvalidArgument`, `ResourceExhausted`,
+    /// `Unimplemented`, ...) are reconstructed from grpc-status trailers on the
+    /// client and have no `source`. Evicting the channel for those cannot help —
+    /// the channel is healthy — and only churns the connection while advancing the
+    /// peer toward the offline threshold. See issue #4567.
+    fn is_transport_failure(status: &tonic::Status) -> bool {
+        use tonic::Code;
+        std::error::Error::source(status).is_some()
+            || matches!(
+                status.code(),
+                Code::Unavailable | Code::DeadlineExceeded | Code::Unknown | Code::Cancelled
+            )
+    }
+
     async fn evict_connection(&self, op: &'static str, reason: &str, resource_summary: &str) {
         let log_level = if Self::is_scanner_leader_lock(resource_summary) {
             debug!(
@@ -148,6 +171,12 @@ impl RemoteClient {
             Ok(Ok(response)) => Ok(response),
             Ok(Err(err)) => {
                 let reason = err.to_string();
+                // Only evict (and re-dial) the cached channel when the failure is a genuine
+                // transport problem. A server-produced application status (auth denied, peer
+                // lock service not ready, invalid args, ...) arrives on a perfectly healthy
+                // channel; evicting it just churns the connection and pushes the peer toward
+                // the offline threshold for no benefit. See issue #4567.
+                let transport_failure = Self::is_transport_failure(&err);
                 if Self::is_scanner_leader_lock(resource_summary) {
                     debug!(
                         addr = %self.addr,
@@ -156,6 +185,7 @@ impl RemoteClient {
                         resource_summary,
                         tonic_code = ?err.code(),
                         tonic_message = err.message(),
+                        transport_failure,
                         "Remote lock RPC returned tonic error for scanner leader lock"
                     );
                 } else {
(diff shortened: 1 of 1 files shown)


lump:
Verified: Zero WARN/ERROR lines in the logs, where beta.8's baseline was 4 evictions/minute/peer. Any eviction warning we see from now on is a real signal.
Thank you.

PR #3179, opened 2026-06-02
fix(object-lock): allow locked objects to receive new versions

marshawcoco:
## Related Issues

Fixes #3174

## Summary of Changes

- Allow put-like writes to bypass current-version Object Lock only when the destination write actually creates a new version: versioning enabled, not suspended, and no explicit version ID.
- Validate CopyObject Object Lock overwrite protection against destination write options, and build destination options from the destination version ID.
- Preserve AccessDenied behavior for unversioned writes, suspended-versioning overwrites, and explicit target-version overwrites of protected objects.
- Add regression coverage for suspended versioning and CopyObject with a retained destination version.

## Verification

- `cargo fmt --all`
- `cargo fmt --all --check`
- `git diff --check`
- `CARGO_TARGET_DIR=/tmp/rustfs-3179-review-target CARGO_BUILD_JOBS=2 cargo test -p rustfs --lib validate_existing_object_lock -- --nocapture`
- `CARGO_TARGET_DIR=/tmp/rustfs-3179-review-target CARGO_BUILD_JOBS=2 cargo test -p rustfs --lib copy_object_allows_new_version_for_locked_destination_but_blocks_explicit_overwrite -- --ignored --nocapture`

## Impact

Versioned buckets can create new versions over retained or legal-held objects, matching S3/MinIO Object Lock semantics. Suspended or unversioned overwrites and explicit target-version overwrites of protected objects remain blocked.

## Additional Notes

`make pre-commit` was not run for the latest review update per requester direction. A started pre-commit run was terminated before completion and did not leave additional code changes.

PR #4598, opened 2026-07-09
perf(ecstore): backfill rename_data old size and gate the PUT prelookup

overtrue:
## Related Issues

Closes rustfs/backlog#1009 (rename_data old-size backfill infrastructure, HP-7 re-scope). Analysis lineage: rustfs/backlog#928 (HP-7, closed as re-scoped), parent tracking rustfs/backlog#936. Prerequisites rustfs/backlog#922 (HP-1) and rustfs/backlog#935 (crash harness) are closed.

## Summary of Changes

Threads the destination key's previous current-version size out of the dst `xl.meta` that `rename_data` already reads, so the PUT path can skip its unconditional pre-PUT `get_object_info` full-disk fanout on buckets without object locking. Per the HP-7 re-scoped analysis, that prelookup costs one namespace read-lock plus one 4-disk `read_version` fanout per PUT (~4-5% of PUT latency for mixed-256k objects on fresh keys, where the negative lookup can never be served from cache).

**Disk layer (`RenameDataResp`)**: new `OldCurrentSize` enum (`Absent` / `Present(i64)`) and a `#[serde(default)] old_current_size: Option<OldCurrentSize>` field. `None` means unknown (pre-upgrade peer or unparseable metadata) and is never conflated with `Absent`. Wire compatibility is msgpack-named + serde-default in both directions.

**`LocalDisk::rename_data`** (non-inline and inline commit branches): before `add_version`, observe the destination's current version from the already-loaded dst `xl.meta` via `observe_old_current_size`. The observation reproduces the per-disk `read_version`/`get_file_info` pipeline: missing xl.meta → `Absent`; all-free-versions → `Absent`; zero-version meta → `Present(0)` (mirroring `get_file_info`'s synthetic deleted FileInfo); resolvable latest — live object, delete marker, or purge-pending — → `Present(fi.size)`; undecodable latest (including the `all_parts=true` part-array length guard, same flag as the lookup) or unparseable meta → unknown (no vote). The single deliberate divergence — `read_data=false`, so a corrupt inline-data map votes the version's size instead of abstaining — is documented at the helper; the voted size matches what healthy disks report, and it avoids cloning inline bytes per disk per PUT. The commit syscall sequence is unchanged; the observation is a pure in-memory read of the already-loaded buffer.

**`SetDisks::rename_data`**: per-disk observations are quorum-reduced like `reduce_common_data_dir` — a value must be reported by ≥ write_quorum disks to win, otherwise the set-level result is unknown. Unknown per-disk entries never vote, so a rolling-upgrade cluster still reaches quorum from its upgraded disks.

**Surface without touching `ObjectInfo`** (hard red line: event notification payloads, replication, and ILM consume `ObjectInfo` verbatim): new parallel inherent methods `SetDisks::put_object_with_old_current_size` → `Sets::…` → `ECStore::…` returning `(ObjectInfo, Option<OldCurrentSize>)`; the `ObjectIO::put_object` trait methods delegate and drop the extra value. No trait signature changes, so all other implementors and callers are untouched.

**App layer (PUT usecase)**: the prelookup now runs only when it has a consumer — the bucket has object locking (WORM validation; the gate fails closed: a bucket-metadata read error keeps the lookup) or the PUT carries an explicit `version_id` (internal replication overwrite, where the lookup targets a specific version the backfill does not observe). Otherwise accounting consumes the backfill through `previous_current_size_from_backfill`, which maps `Present(size)` → `Some(size.max(0))` and `Absent` → `None`, exactly what the prelookup produced. CompleteMultipartUpload and CopyObject keep their prelookups (out of scope; the infrastructure is in place for a follow-up).

**Unknown-backfill fallback**: when neither source can determine the previous state (peers predating the field during a rolling upgrade, or sub-quorum divergence), the new `record_bucket_object_write_unknown_previous_memory` records only the components that are correct regardless of the previous state (bytes always added; a versioned write always adds a version) and leaves `objects_count` to the next authoritative scanner refresh, which replaces the in-memory cache wholesale. A debug log marks each degraded record.

### Semantic finding vs. the issue text

The issue asked to replicate "latest / delete-marker / missing → None" semantics. Reading and testing the actual lookup showed RustFS's `SetDisks::get_object_info` does **not** convert a delete-marker latest to not-found (unlike MinIO): it returns `Ok` with `size 0`, and `record_bucket_delete_marker_memory` never decrements `objects_count`. Mapping markers to `Absent` would therefore double-count objects on every versioned PUT-over-marker. The backfill instead reproduces the lookup's real behavior — `Present(0)` for marker-latest — which the issue's governing requirement (accounting numbers bit-identical for every PUT shape) demands. The end-to-end equivalence test caught this divergence before it shipped.

### Behavior notes (deliberate, documented)

- A PUT to a key whose existing metadata is corrupt/quorum-divergent used to fail at the prelookup with a read-quorum error on all buckets; on locking-disabled buckets it now proceeds (the commit re-checks write quorum; per-disk overwrite of unparseable dst meta is pre-existing rename_data behavior) and takes the degraded-accounting path. This is MinIO-aligned (MinIO PUTs never prelookup) and turns a hard failure into a self-healing write. CopyObject and CompleteMultipartUpload keep their fail-closed prelookups, so corrupt-dst behavior now differs across write paths until they adopt the backfill.
- The parity target is the set-level (`SetDisks`) lookup. Two app-visible lookup variants deviated from it and from each other, and both over-incremented `objects_count` relative to delete-marker accounting: the multi-pool `ECStore` path converts a delete-marker latest to not-found, and directory-key lookups pin the nil version. The backfill standardizes on the self-consistent set-level answer (`Present(0)`), so multi-pool marker-resurrection PUTs and directory re-PUTs stop over-counting; documented on `OldCurrentSize`.
- Multi-pool: the pool router sends overwrites to the pool that already holds the key (a marker-latest pool counts as existing), so the backfill observes the right pool except in degenerate duplicate-key states left by interrupted decommissions.

### Adversarial validation

Six independent reviewer passes (correctness, security, concurrency/durability, compatibility, performance, test-coverage) ran against the final diff. Highlights: the commit syscall sequence and lock-loss fence are unchanged; the backfill is read under the commit write lock, so it is strictly more atomic than the old prelookup (which raced concurrent PUT/DELETE outside the lock); wire compatibility was proven empirically in both directions against the locked rmp-serde version, including the JSON compat path; the WORM gate is strictly more conservative than the shipped HP-8 delete gate (fail-closed vs fail-open on metadata errors), and lock-enabled buckets are always versioned so the skipped validation is a no-op for every write shape that can reach the skip path. Review findings that changed code: `all_parts=true` in the observation (restores per-disk abstention on corrupt part arrays), the quorum-mutant-killing and non-inline-unparseable tests, the WORM-gate truth-table tests, and the sub-quorum blackbox test.

## Verification

- `make fmt-check unsafe-code-check architecture-migration-check logging-guardrails-check tokio-io-uring-check doc-paths-check` — pass.
- `cargo clippy --all-targets --all-features -- -D warnings` (workspace) — pass, re-run after the final test additions.
- `cargo nextest run -p rustfs-ecstore` (full crate suite) — 2374 passed, 1 skipped.
- `cargo nextest run -p rustfs` (full crate suite) — pass (includes the new `put_prelookup_gating_test` against a real 4-disk ECStore and the backfill-mapping unit test).
- Untouched workspace crates are covered by the workspace-wide clippy compile (`--all-targets`); no source outside `rustfs-ecstore` and `rustfs` changed except re-export facades compiled by both.
- Key regression tests added:
  - `ecstore_validation_blackbox::old_current_size_backfill::backfill_matches_prelookup_for_every_put_shape`: for every PUT shape (fresh key, unversioned inline overwrite, non-inline overwrite both directions, versioned first/second write, PUT over delete marker) asserts the backfill equals what `get_object_info` reported immediately before the PUT — the invariant that keeps accounting unchanged.
  - `backfill_is_unknown_below_quorum_but_put_succeeds`: garbage xl.meta on 2 of 4 disks → the real set-level fanout yields unknown while the PUT still commits.
  - `LocalDisk` unit + e2e tests: observation semantics per shape, inline and non-inline branches, unparseable dst meta → unknown (both branches), corrupt part arrays → abstain.
  - Quorum-reduction tests including the "unknown counts as an Absent vote" mutant killers (`all_unknown_is_unknown`, `unknown_majority_does_not_become_absent_quorum`).
  - `put_prelookup_gating_test`: lock bucket keeps the prelookup, plain bucket skips, missing bucket metadata fails closed.
  - Wire-compat: msgpack-named round trip for all three field states; legacy payload without the field decodes to unknown.
  - `data_usage`: the unknown-previous record function adds size (and version count when versioned) and never moves `objects_count`.

## Impact

- Performance: on buckets without object locking, every PUT saves one namespace read-lock + one full-set `read_version` fanout (~4-5% PUT latency on mixed-256k per the HP-7 profiling); absolute numbers to be confirmed by the warp A/B gate on Linux.
- Compatibility: internode `RenameDataResp` gains one `#[serde(default)]` field; mixed-version clusters are wire-compatible in both directions (old peers' responses read as unknown and simply don't vote). No on-disk format change, no `ObjectInfo`/event-payload/API change.
- During a rolling upgrade, PUTs whose set cannot reach a quorum of upgraded disks record a degraded usage delta (size always counted; `objects_count` deferred to the next scanner refresh). Self-correcting; marked by a debug log.
- Object locking buckets and explicit-version (replication) PUTs are byte-for-byte on the pre-existing path.

## Additional Notes

Follow-up candidates (not in scope): adopt the backfill in CompleteMultipartUpload and CopyObject (both keep their prelookups; multipart currently discards the fifth `rename_data` tuple element); `put_object_lock_configuration` enables locking on an existing versioned bucket without a peer metadata-reload fan-out (defense-in-depth observation from the security pass, harmless for WORM here).

### DIFF
--- a/crates/ecstore/src/api/mod.rs
+++ b/crates/ecstore/src/api/mod.rs
@@ -261,9 +261,9 @@ pub mod data_usage {
         DATA_USAGE_CACHE_NAME, apply_bucket_usage_memory_overlay, init_compression_total_memory_from_backend,
         load_compression_total_from_memory, load_data_usage_from_backend, record_bucket_delete_marker_memory,
         record_bucket_object_delete_memory, record_bucket_object_version_write_memory, record_bucket_object_write_memory,
-        record_compression_total_memory, refresh_bucket_usage_from_object_layer,
-        refresh_versioned_bucket_usage_from_object_layer, remove_bucket_usage_from_backend,
-        replace_bucket_usage_memory_from_info, store_compression_total_in_backend,
+        record_bucket_object_write_unknown_previous_memory, record_compression_total_memory,
+        refresh_bucket_usage_from_object_layer, refresh_versioned_bucket_usage_from_object_layer,
+        remove_bucket_usage_from_backend, replace_bucket_usage_memory_from_info, store_compression_total_in_backend,
     };
 }
 
@@ -273,9 +273,9 @@ pub mod disk {
     pub use crate::disk::{
         BATCH_READ_VERSION_MAX_ITEMS, BUCKET_META_PREFIX, BatchReadVersionItem, BatchReadVersionReq, BatchReadVersionResp,
         CheckPartsResp, DeleteOptions, Disk, DiskAPI, DiskInfo, DiskInfoOptions, DiskLocation, DiskOption, DiskStore,
-        FileInfoVersions, FileReader, FileWriter, HEALING_MARKER_PATH, RUSTFS_META_BUCKET, ReadMultipleReq, ReadMultipleResp,
-        ReadOptions, RenameDataResp, STORAGE_FORMAT_FILE, UpdateMetadataOpts, VolumeInfo, WalkDirOptions, new_disk,
-        validate_batch_read_version_item_count,
+        FileInfoVersions, FileReader, FileWriter, HEALING_MARKER_PATH, OldCurrentSize, RUSTFS_META_BUCKET, ReadMultipleReq,
+        ReadMultipleResp, ReadOptions, RenameDataResp, STORAGE_FORMAT_FILE, UpdateMetadataOpts, VolumeInfo, WalkDirOptions,
+        new_disk, validate_batch_read_version_item_count,
     };
     pub use bytes::Bytes;
     pub use endpoint::Endpoint;
--- a/crates/ecstore/src/cluster/rpc/remote_disk.rs
+++ b/crates/ecstore/src/cluster/rpc/remote_disk.rs
@@ -3060,6 +3060,7 @@ mod tests {
         let response = RenameDataResp {
             old_data_dir: Some(Uuid::new_v4()),
             sign: Some(vec![1_u8; 32]),
+            old_current_size: Some(crate::disk::OldCurrentSize::Present(4096)),
         };
         let json = serde_json::to_vec(&response).expect("rename data response json should encode");
         let named_msgpack = encode_msgpack_named(&response).expect("rename data response named msgpack should encode");
--- a/crates/ecstore/src/core/sets.rs
+++ b/crates/ecstore/src/core/sets.rs
@@ -389,6 +389,22 @@ impl crate::storage_api_contracts::object::ObjectIO for Sets {
     }
 }
 
+impl Sets {
+    /// `put_object` plus the rename_data old-size backfill
+    /// (rustfs/backlog#1009); see `SetDisks::put_object_with_old_current_size`.
+    pub async fn put_object_with_old_current_size(
+        &self,
+        bucket: &str,
+        object: &str,
+        data: &mut PutObjReader,
+        opts: &ObjectOptions,
+    ) -> Result<(ObjectInfo, Option<crate::disk::OldCurrentSize>)> {
+        self.get_disks_by_key(object)
+            .put_object_with_old_current_size(bucket, object, data, opts)
+            .await
+    }
+}
+
 #[async_trait::async_trait]
 impl BucketOperations for Sets {
     type Error = Error;
--- a/crates/ecstore/src/data_usage/mod.rs
+++ b/crates/ecstore/src/data_usage/mod.rs
@@ -689,6 +689,34 @@ async fn record_bucket_object_write_memory_inner(
     entry.stale_snapshot_pending = false;
 }
 
+/// Degraded in-memory update for an object write whose previous current size
+/// could not be determined (rustfs/backlog#1009: the pre-PUT lookup was
+/// skipped and the rename_data backfill came back unknown — mixed-version
+/// peers or sub-quorum metadata divergence). Applies only the components that
+/// are correct regardless of the previous state: the new bytes always count,
+/// and a versioned write always adds a version. objects_count (and the
+/// non-versioned overwrite's old-size subtraction) are left to the next
+/// scanner refresh, which replaces this cache with authoritative numbers.
+pub async fn record_bucket_object_write_unknown_previous_memory(bucket: &str, new_size: u64, creates_new_version: bool) {
+    ensure_bucket_usage_cached(bucket).await;
+
+    let mut cache = memory_cache().write().await;
+    let entry = cache
+        .entry(bucket.to_string())
+        .or_insert_with(|| cached_bucket_usage_now(BucketUsageInfo::default()));
+
+    entry.usage.size = entry.usage.size.saturating_add(new_size);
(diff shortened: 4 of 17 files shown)


overtrue:
The remaining Typos failure comes from current main, not this branch: `unparseable` in `rustfs/src/app/bucket_usecase.rs:329` (introduced by #4592 before the typos dictionary gained the rule; the workflow pins `crate-ci/typos@master`). This PR's own occurrences were already renamed in ca532eac7. Fix for main: #4602 — once it merges, a re-run here goes green.

overtrue:
Cross-review from the parallel implementation (#4599, now closed in favor of this PR — both were opened for rustfs/backlog#1009 within minutes of each other).

I independently verified the load-bearing semantic finding here and can confirm it against the code: `record_bucket_delete_marker_memory` (crates/ecstore/src/data_usage/mod.rs) only increments `delete_markers_count` and never decrements `objects_count`, and the existing e2e (`data_usage_reports_versioned_objects_and_delete_markers`) pins `objects_count == 2` while alpha is marker-latest — so the ledger and the scanner both keep marker-latest keys counted. Mapping marker-latest to `Present(0)` is the only choice consistent with both; #4599 mapped it to `None` (multi-pool prelookup parity) which would have double-counted `objects_count` on every marker-resurrect PUT relative to the ledger. The blackbox equivalence oracle in this PR (backfill vs. real prelookup per PUT shape) is exactly the test design that catches this class of bug — #4599's tests asserted the vote value directly and missed it.

Two small things from #4599 worth considering as follow-ups (not blockers):

1. An ignore-gated e2e in `crates/e2e_test/src/data_usage_test.rs` asserting a non-versioned overwrite reports exactly the new size from the first usage poll (never old+new). This PR's `put_prelookup_gating_test` covers the gate against a real store; the e2e adds the full-server accounting path. Happy to port it onto this branch if wanted.
2. During review of #4599 we found a pre-existing inconsistency unrelated to this change: `test_put_object_overwrite_blocked_by_legal_hold` (and the copy variant) in crates/e2e_test still assert overwrite rejection, contradicting the versioned-new-write early-return from #3179; the e2e crate is excluded from regular CI so it never surfaced. Filed for separate follow-up.

CI note: the Typos failure is inherited from main (#4592's `unparseable`, fix in flight at #4602); on #4599 the "S3 Implemented Tests" job was cancelled at 1h05m (never concluded) — if the same happens here it's worth a re-run rather than reading it as a regression.

PR #4599, opened 2026-07-09
perf(ecstore): backfill previous current size from rename_data and skip the pre-PUT lookup

overtrue:
Implements rustfs/backlog#1009 (HP-7 rescoped): every PUT used to issue an unconditional `get_object_info` before reading the body, whose only surviving consumers are usage accounting (`previous_current_size`) and WORM validation. `rename_data` already parses the destination xl.meta on every disk during the commit, so the old size was read twice per PUT — the pre-lookup costs one namespace read-lock plus a full-set `read_version` fanout (~4-5% of PUT latency on mixed-256k objects, per the HP-7 profiling in backlog#928).

## What changed

- **Wire (`RenameDataResp`)**: new `#[serde(default)] old_current_size: Option<RenameOldCurrentSize>` where `RenameOldCurrentSize { live_size: Option<i64> }`. Outer `None` means "disk did not report" (pre-upgrade peer); inner `None` means "no live current version" (fresh key or delete-marker latest). Keeping the two `None`s separate is what makes mixed-version clusters safe: an absent field can never be misread as "the object did not exist". Both msgpack-named and JSON legs stay compatible in either direction, and the third state (reported-but-no-live-version) survives both formats (all test-pinned).
- **Disk (`LocalDisk::rename_data`)**: both commit branches (non-inline and inline) observe the destination's latest version from the already-parsed dst `FileMeta` before `add_version`, via `FileMeta::into_fileinfo` with an empty version id — the same latest-version selection the read path uses, so delete-marker/missing/free-version semantics match `get_object_info` exactly. No new syscall, no new failure mode in the commit sequence (all observation errors collapse to `None`).
- **Set (`SetDisks::rename_data`)**: the per-disk observations are voted like `reduce_common_data_dir`: a value must reach write quorum to be trusted (`Quorum`); disagreement below quorum is `NoQuorum`; any successful disk that did not report poisons the vote to `Unsupported`. Two distinct values can never both reach write quorum (parity is capped at N/2 on every config path, so 2·wq > N).
- **Surface**: `ObjectOptions` gains an in-process out-parameter `put_old_current_vote: Option<Arc<OnceLock<RenameOldCurrentVote>>>` that `SetDisks::put_object` fills after the commit. `ObjectInfo` is untouched — nothing new flows into events, replication, or ILM payloads (the hard red line from the issue). `ObjectOptions` has no serde derive, so the field can never cross a wire.
- **App (`put_object` usecase)**: when the bucket has no object-lock configuration and the PUT carries no explicit `?versionId=`, the pre-PUT `get_object_info` fanout is skipped and accounting uses the backfilled size. Fail-closed gates:
  - unreadable bucket metadata counts as locked (keeps the pre-lookup + WORM validation path); the WORM half of the gate is a pure function with a fail-closed unit-test matrix;
  - `NoQuorum` falls back to fresh-key accounting for that write only (over-counts, never under-counts, until the authoritative scanner reconciles — conservative for quota); warns once per process, then debug, so a heal-lagged set cannot flood the hot-path log;
  - `Unsupported` (pre-upgrade peer) additionally flips a process-wide sticky flag that restores the original pre-lookup path until restart, so a rolling upgrade degrades to at most one imprecise record per upgraded node.

`copy_object`'s pre-lookup and CompleteMultipartUpload are out of scope per the issue (multipart discards the vote explicitly).

## Accounting equivalence

For every PUT shape the backfilled `previous_current_size` is bit-identical to what the pre-lookup produced: fresh key → `None`; non-versioned overwrite → `Some(old stored size)` (exact size delta); versioned write → previous latest live size (drives `objects_count` ±1); zero-byte, compressed (both paths use the stored `fi.size`), inline and non-inline branches all covered. One deliberate divergence: on **single-pool** deployments a delete-marker-latest used to yield `Some(0)` because `SetDisks::get_object_info` returns `Ok` with `delete_marker: true` (multi-pool translates it to `ObjectNotFound` → `None`). The backfill always reports `None`, matching the multi-pool and semantically correct side: a PUT that resurrects a deleted key now counts the object again on single-pool too.

The observation is taken under the per-object write lock at commit time, so concurrent overwrites of the same key now account against the immediately-prior committed version — strictly more accurate than the old unlocked pre-lookup, which could double-read the same stale size.

## Compatibility

- Old peer → new node: absent field decodes as `None` ("not reported") → vote is `Unsupported` → sticky fallback to the original path.
- New node → old peer: the unknown msgpack/JSON field is ignored (test-pinned against a legacy-shaped decoder). The proto schema is unchanged — evolution happens inside the opaque serde blob.
- WORM: buckets with object-lock enabled keep the exact original pre-lookup + `validate_existing_object_lock_for_write` path. Non-lock buckets cannot hold retention/legal-hold metadata (invariant already relied on by `can_skip_delete_objects_pre_stat`), so skipping the validation there is a no-op. Enabling object-lock on an existing bucket is monotonic and takes effect for the next PUT — the same window the old path had. Conditional writes (If-Match/If-None-Match) were never served by this pre-lookup — they run inside `SetDisks::put_object::check_write_precondition`.

## Tests

- `disk::local::test::rename_data_old_current_size_semantics::*` — latest/newest-wins/delete-marker/missing semantics against hand-built `FileMeta`s.
- `set_disk::tests::test_reduce_rename_old_current_vote` — quorum/no-quorum/unsupported/failed-disk vote matrix.
- `set_disk::ops::object::put_object_old_current_backfill_tests::*` — end-to-end on a 4-disk hermetic set: the backfilled vote equals the pre-write `get_object_info` size for inline, non-inline, and versioned writes; a PUT over a real committed delete marker votes "no live current"; slot-less PUTs unaffected.
- `cluster::rpc::remote_disk::tests::rename_data_resp_old_current_size_wire_compat_with_pre_upgrade_peers` — both wire directions, both formats, plus the reported-fresh-key third state.
- `app::object_usecase::tests::put_prelookup_backfill_lock_gate_is_fail_closed` — the WORM gate matrix (locked / unlocked / unreadable metadata).
- `app::object_usecase::tests::resolve_backfilled_previous_current_size_matrix_and_sticky_fallback` — vote→accounting translation, negative-size clamp, and sticky-flag behavior.
- `e2e_test::data_usage_test::data_usage_reports_overwrite_size_delta_exactly` (ignore-gated like its siblings) — a real-server overwrite must report exactly the new size from the first usage poll, never old+new.

## Adversarial validation

Six independent reviewer passes over the final diff (correctness, security, concurrency/durability, compatibility, performance, test-coverage). Correctness, security, concurrency/durability, and compatibility returned null reports (attacks included quorum boundaries at wq−1, two-values-at-quorum reachability, delete-marker topology divergence, commit-sequence purity/panic-safety on hostile xl.meta, lock-loss fencing, rolling-upgrade decode in both directions, WORM TOCTOU via late lock enablement, i64::MIN size payloads). Performance flagged the per-PUT NoQuorum warn (fixed: once-per-process latch, then debug) and noted the observation runs on all rename_data callers (accepted: single-version parse on already-read bytes; per-disk gating would need another wire field). Test-coverage findings all addressed (gate matrix unit test, hermetic delete-marker case, wire third-state round-trip, e2e overwrite accounting).

One pre-existing inconsistency surfaced during review, untouched by this PR and filed separately: `test_put_object_overwrite_blocked_by_legal_hold` in `crates/e2e_test` still asserts overwrite rejection, contradicting the versioned-new-write early-return introduced by #3179; the e2e crate is excluded from regular CI so it never surfaced.

## Verification

- `make fmt-check unsafe-code-check architecture-migration-check logging-guardrails-check tokio-io-uring-check doc-paths-check`: PASS
- `cargo clippy --all-targets --all-features -- -D warnings`: PASS
- `cargo nextest run --all --exclude e2e_test` + doc tests: PASS

Closes rustfs/backlog#1009.

### DIFF
--- a/crates/e2e_test/src/data_usage_test.rs
+++ b/crates/e2e_test/src/data_usage_test.rs
@@ -99,6 +99,57 @@ async fn data_usage_reports_all_objects() -> Result<(), Box<dyn std::error::Erro
     Ok(())
 }
 
+/// Regression test for rustfs/backlog#1009: on a non-locked, non-versioned
+/// bucket the PUT usecase skips its pre-PUT lookup and accounts overwrites
+/// with the size backfilled by rename_data. From the moment the bucket shows
+/// up in data usage, an overwrite must report exactly the new size (never
+/// old+new double-counted) and a single object.
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+#[ignore = "Starts a rustfs server and requires awscurl; enable when running full E2E"]
+async fn data_usage_reports_overwrite_size_delta_exactly() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    init_logging();
+
+    let mut env = RustFSTestEnvironment::new().await?;
+    env.start_rustfs_server(vec![]).await?;
+
+    let client = env.create_s3_client();
+    let bucket = "data-usage-overwrite";
+    let first_size = 1024_u64;
+    let overwrite_size = 4096_u64;
+
+    client.create_bucket().bucket(bucket).send().await?;
+    client
+        .put_object()
+        .bucket(bucket)
+        .key("resized")
+        .body(ByteStream::from(vec![1u8; first_size as usize]))
+        .send()
+        .await?;
+    client
+        .put_object()
+        .bucket(bucket)
+        .key("resized")
+        .body(ByteStream::from(vec![2u8; overwrite_size as usize]))
+        .send()
+        .await?;
+
+    // The in-memory record is synchronous with the PUT responses, so the
+    // first poll that knows the bucket must already be exact; a transient
+    // first_size + overwrite_size reading would mean the overwrite was
+    // double-counted and only later repaired by the scanner.
+    let usage = wait_for_bucket_usage(&env, bucket, |usage| usage.buckets_usage.contains_key(bucket)).await?;
+    let bucket_usage = usage.buckets_usage.get(bucket).expect("bucket usage should exist");
+    assert_eq!(
+        bucket_usage.size, overwrite_size,
+        "overwrite must account exactly the new size, not double-count the previous one"
+    );
+    assert_eq!(bucket_usage.objects_count, 1, "overwrite must not create a second object");
+
+    env.stop_server();
+    Ok(())
+}
+
 /// Regression test for issue #3898.
 /// Versioned buckets should expose versions and delete markers through admin data usage.
 #[tokio::test(flavor = "multi_thread")]
--- a/crates/ecstore/src/api/mod.rs
+++ b/crates/ecstore/src/api/mod.rs
@@ -274,8 +274,8 @@ pub mod disk {
         BATCH_READ_VERSION_MAX_ITEMS, BUCKET_META_PREFIX, BatchReadVersionItem, BatchReadVersionReq, BatchReadVersionResp,
         CheckPartsResp, DeleteOptions, Disk, DiskAPI, DiskInfo, DiskInfoOptions, DiskLocation, DiskOption, DiskStore,
         FileInfoVersions, FileReader, FileWriter, HEALING_MARKER_PATH, RUSTFS_META_BUCKET, ReadMultipleReq, ReadMultipleResp,
-        ReadOptions, RenameDataResp, STORAGE_FORMAT_FILE, UpdateMetadataOpts, VolumeInfo, WalkDirOptions, new_disk,
-        validate_batch_read_version_item_count,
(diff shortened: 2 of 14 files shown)


overtrue:
Closing in favor of #4598 — the same backlog#1009 scope was implemented in parallel (both branches are mine; race condition between sessions). Cross-review verdict posted on #4598: its delete-marker-latest semantics (`Present(0)`, matching the in-memory ledger where `record_bucket_delete_marker_memory` never decrements `objects_count`) are more correct than this PR's `None` mapping, which would double-count `objects_count` on marker-resurrect PUTs; its mixed-version handling (unknowns abstain instead of poisoning the vote) also degrades more gracefully. The two useful extras from this PR (e2e overwrite-usage assertion, the stale legal-hold e2e discovery) were offered on #4598 / filed separately.

PR #4604, opened 2026-07-09
feat(utils): make DNS resolver cache TTL configurable via RUSTFS_DNS_CACHE_TTL_SECS

### LINKED ISSUE #4568: DNS cache TTL is hardcoded to 300s; make it configurable (or honor 0 = disabled) for orchestrated networks
**Version:** 1.0.0-beta.8 (also present on current `main`)
**Deployment:** Docker Swarm, MNMD, stable service DNS names, dynamic container IPs.

## Summary

`crates/utils/src/net.rs` maintains a process-global resolver cache with a hardcoded TTL:

```rust
static DNS_CACHE: LazyLock<Mutex<HashMap<String, DnsCacheEntry>>> = ...;
const DNS_CACHE_TTL: Duration = Duration::from_secs(300); // 5 minutes
```

There is no environment variable or config option to tune or disable it (the only bypass,
`has_custom_dns_resolver()`, is not reachable from configuration in beta.8).

## Why this matters on orchestrated networks

On Docker Swarm and Kubernetes, the NAME is the only durable identity of a peer; container/pod
IPs legitimately change whenever a task is rescheduled, and the platform DNS
(Docker's embedded DNS at 127.0.0.11, or kube-dns/CoreDNS) always serves the current answer
with no meaningful caching of its own. A fixed 300s application-side cache means that after a
peer restarts, a member can keep dialing a dead IP for up to five minutes even though a fresh
lookup would have returned the live address immediately. For a distributed store whose
internode RPC deadlines are measured in seconds, five minutes of dialing a corpse is a long
outage window, and it composes badly with fail-fast behaviors elsewhere (e.g. startup exits
fatally when a peer does not resolve or respond).

Rust deliberately has no runtime-level DNS cache or global TTL knob (unlike the JVM's
`networkaddress.cache.ttl`), so this application cache is the only cache in the stack on
musl-based images - and currently the only one that cannot be turned off.

## Suggested change

Read the TTL from an environment variable, keeping the current default:

- `RUSTFS_DNS_CACHE_TTL_SECS` (u64, default 300)
- `0` disables caching entirely (every `get_host_ip` consults the system resolver)

This is a small, backward-compatible change confined to `crates/utils/src/net.rs`, in the
same spirit as the existing `RUSTFS_INTERNODE_*` transport variables added for cluster
resilience.

## Context

Encountered while debugging cluster membership on Docker Swarm: we ultimately proved the cache
was NOT the culprit in our incident (resolutions were fresh), but only by reading the source -
during triage, an unconfigurable invisible cache between rustfs and the platform DNS is one
more thing that cannot be ruled out from the outside. Making it tunable (and logging the TTL at
startup) removes a whole diagnostic branch for anyone running rustfs on dynamic networks.

houseme:
## Summary

Closes #4568.

The resolver cache in `crates/utils/src/net.rs` hardcoded a 300s TTL with no way to tune or disable it (the only bypass, `has_custom_dns_resolver()`, is not reachable from configuration). On orchestrated networks (Docker Swarm, Kubernetes) a peer's DNS name is its only durable identity while its container/pod IP legitimately changes on reschedule, and the platform DNS (Docker's embedded DNS, kube-dns/CoreDNS) always serves the current answer. A fixed 300s application-side cache can keep a member dialing a dead IP for up to five minutes after a peer restarts — a long outage window for a store whose internode RPC deadlines are measured in seconds, and it composes badly with fail-fast behaviors elsewhere. Rust has no runtime-level DNS cache, so on musl images this was the only cache in the stack and the only one that could not be turned off.

## Changes

- Read the TTL from `RUSTFS_DNS_CACHE_TTL_SECS` (u64, default 300), in the same spirit as the existing `RUSTFS_INTERNODE_*` transport knobs. The value is resolved once via the existing `get_env_u64` helper.
- `0` disables caching entirely: every `get_host_ip` consults the system resolver and nothing is cached.
- Log the resolved TTL (or "disabled") once at first use, so operators debugging cluster membership on dynamic networks can rule the cache in or out from the logs alone — removing a whole diagnostic branch.
- Drop the two per-lookup `info!` logs on the resolve path: they would flood the log when caching is disabled, and were hot-path noise otherwise.

Backward compatible (default behavior unchanged at 300s) and confined to `net.rs`.

## Test

- `cargo test -p rustfs-utils --features net --lib -- net::test` — 13 passed, including a new deterministic `test_dns_cache_entry_expiry` covering the TTL comparison at the cache-entry level (no reliance on the process-global cache/TTL singletons, to avoid flakiness).
- `cargo clippy -p rustfs-utils --features net --lib` — clean.
- `cargo fmt --check` and `scripts/check_logging_guardrails.sh` — pass.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -45,6 +45,8 @@ nonexisted = "nonexisted"
 consts = "consts"
 # Swift API - company/product names
 Hashi = "Hashi"  # HashiCorp
+# Accept alternate spelling used in parser/XML comments.
+unparseable = "unparseable"
 
 [files]
 extend-exclude = []
--- a/crates/utils/src/net.rs
+++ b/crates/utils/src/net.rs
@@ -48,7 +48,39 @@ impl DnsCacheEntry {
 }
 
 static DNS_CACHE: LazyLock<Mutex<HashMap<String, DnsCacheEntry>>> = LazyLock::new(|| Mutex::new(HashMap::new()));
-const DNS_CACHE_TTL: Duration = Duration::from_secs(300); // 5 minutes
+
+/// Environment variable to tune the DNS resolver cache TTL, in seconds.
+///
+/// Rust has no runtime-level DNS cache or global TTL knob (unlike the JVM's
+/// `networkaddress.cache.ttl`), so this application cache is the only cache in the stack on
+/// musl-based images. On orchestrated networks (Docker Swarm, Kubernetes) a peer's DNS name is
+/// its only durable identity while its IP changes on reschedule; a fixed cache can keep a member
+/// dialing a dead IP after a peer restarts. Making the TTL tunable removes an invisible,
+/// unconfigurable cache between rustfs and the platform DNS.
+///
+/// `0` disables caching entirely (every `get_host_ip` consults the system resolver).
+const ENV_DNS_CACHE_TTL_SECS: &str = "RUSTFS_DNS_CACHE_TTL_SECS";
+const DEFAULT_DNS_CACHE_TTL_SECS: u64 = 300; // 5 minutes
+
+/// Resolved DNS cache TTL, read once from `RUSTFS_DNS_CACHE_TTL_SECS` at first use.
+///
+/// A value of `Duration::ZERO` means caching is disabled. The configured value is logged once so
+/// operators debugging cluster membership on dynamic networks can rule the cache in or out from
+/// the logs alone.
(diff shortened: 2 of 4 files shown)


lump:
#4604 verified: node-1 logs DNS resolver cache TTL set to 5s (RUSTFS_DNS_CACHE_TTL_SECS) at startup: knob visible in the logs, exactly as requested.

Thank you.

Issue #1066, opened 2025-12-08
Veeam VBR - S3 returned empty versionId


acolada-gt:
Starting with version alpha.72, following bug #985,  a S3 object repository can be created successfully in Veeam Backup and Replication and no longer gives an error when selecting a bucket folder.

The issue is that no data can be written to the target S3 repository, because it gives an error when verifying the repository:

Error: S3 returned empty versionId for newly uploaded object [/Veeam/Backup/<bucket folder>/Clients/{<id....>}/<id...>/CloudStg/Meta/Blocks/History/CheckpointHistory.{<id...>}], check that versioning is enabled on the bucket Agent failed to process method {Cloud.EnableBackupFeatures}.  

<img width="1025" height="641" alt="Image" src="https://github.com/user-attachments/assets/2885e3ae-3804-47be-8c67-0dcc152b01e7" />

The error is the same for any method of upload (moving backup files through Veeam or starting a new backup).
https://helpcenter.veeam.com/docs/vbr/userguide/os_backup_job.html?ver=13

loverustfs:
Hi @acolada-gt ,

Thank you so much for the incredibly thorough and professional testing! Your contributions are huge in helping RustFS mature faster and benefit the wider community.

loverustfs:
Hey @acolada-gt ,

Could you please provide detailed steps? We do not use Veeam.

Thanks.

acolada-gt:
> Hey [@acolada-gt](https://github.com/acolada-gt) ,
> 
> Could you please provide detailed steps? We do not use Veeam.
> 
> Thanks.

Do you mean the steps for creating the backup job and starting it?
If so, I have sent a link from Veeam about it, in the previous message.

loverustfs:
> > Hey [@acolada-gt](https://github.com/acolada-gt) ,
> > Could you please provide detailed steps? We do not use Veeam.
> > Thanks.
> 
> Do you mean the steps for creating the backup job and starting it? If so, I have sent a link from Veeam about it, in the previous message.

Ok, thanks.  

@shiroleeee  this link : https://helpcenter.veeam.com/docs/vbr/userguide/os_backup_job.html?ver=13

acolada-gt:
It really depends on what you are going to test Veeam with. If you want to backup a bucket from another repository or a windows/linux file, or even a virtual machine, let me know and I will send the link from veeam site.
[...]

Issue #1064, opened 2025-12-08
Add an API for Admin verification


loverustfs:
Title: Add an API for Admin verification 
Description/Fix: Add an API endpoint that allows the frontend to identify if a user has Admin privileges. 
Source: https://github.com/rustfs/rustfs/issues/1020 
Expected outcome: The frontend can determine if a specific username is an administrator.
OS: Linux & Windows 
Email: hello@rustfs.com

PR #1063, opened 2025-12-08
feat: add is-admin user api

yxrxy:
https://github.com/rustfs/rustfs/issues/1020
closes: https://github.com/rustfs/rustfs/issues/1064

## Type of Change
- [x] New Feature

## Checklist
- [x] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [x] Passed `make pre-commit`

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

PR #4625, opened 2026-07-09
fix(experiments): import io_uring cancel-safety spike and apply backlog#1051 audit remediation

houseme:
## What

Imports the Spike 0 io_uring cancel-safety prototype (from the closed PR #4381 branch) as a baseline, then applies all 15 remediation items from the multi-expert adversarial audit tracked in [rustfs/backlog#1051](https://github.com/rustfs/backlog/issues/1051), one commit per sub-issue.

The crate is a standalone workspace and stays out of the main `Cargo.lock` / build graph (NOT production code). The audit found no impact on the current production binary; its value is that rustfs/backlog#1048 declares the ownership model "directly reusable for P2", so these defects would otherwise ride into P2 production verbatim. This PR hardens the model before that reuse.

## Validation

- `experiments/io-uring-cancel-spike/run-docker.sh` — both legs pass: leg 1 deterministically degrades (explicit seccomp profile blocks io_uring → all 11 tests SKIP), leg 2 runs real io_uring under `seccomp=unconfined` (all 11 tests pass, zero skips). The harness now fails on either a non-degrading leg 1 or a vacuous-pass leg 2.
- `cargo check --tests` green on Linux (`rust:1-bookworm`).
- macOS host cannot build this Linux-only crate; all verification ran in Docker.

## Fixes (one commit per rustfs/backlog sub-issue)

**Memory safety / soundness**

- #1053 — drain the probe SQE to its CQE before releasing the buffer (probe-path UAF); EINTR retries the wait, drain failure leaks over UAF.
- #1054 — abort on driver-thread panic before field destructors run, instead of freeing in-flight buffers (unwind-UAF); `catch_unwind` is insufficient.

**Hangs**

- #1055 — bound the shutdown drain (leak-over-UAF on timeout) so a hung disk cannot make graceful shutdown block forever; record the ASYNC_CANCEL three-state result.
- #1056 — assert NODROP at probe, monitor the CQ-overflow counter, handle submit() EBUSY as backpressure instead of swallowing it.

**Correctness**

- #1057 — reject reads above MAX_RW_COUNT so the `len as u32` cast can't silently truncate a 4 GiB read into a false EOF.
- #1058 — resubmit short reads for positioned (whole-range) reads to the final CQE; stream reads keep read(2) semantics.

**errno / resource / security**

- #1059 — scope the restriction-errno class to probe time, guard offset > i64::MAX at submit.
- #1060 — backpressure with permits released at the CQE (not at future drop), bounding in-flight below CQ capacity.
- #1061 — open the probe file via O_TMPFILE (fallback O_EXCL|O_NOFOLLOW+unlink) instead of a predictable, symlink-followable temp path.

**Design docs**

- #1063 — correct invariant 2's fd-reuse mechanism (real window is SQE construction → io_uring_enter) and add invariants 6/7/8 + the errno three-class and bounded-drain contracts.
- #1062 — pin the reused-buffer content-hygiene invariant (caller-visible bytes ⊆ [0, cqe.res)) for P3.

**Tests / harness**

- #1064 — directly pin fd ownership (F_GETFD after caller drop) and orphan-integrity, replacing counter-only proxies.
- #1065 — cover CQ-overflow safety under load and read boundaries (len==0, EOF, cross-EOF short read, huge-len/offset rejection, pipe half-close).
- #1066 — cover the Drop-without-shutdown path and de-flake the exact `cancel_stress` split.
- #1067 — make `run-docker.sh` assert each leg's real path (deterministic seccomp block + no-skip check).

## Not implemented

Per the audit, this only hardens the spike. P2's own leftovers (eventfd/AsyncFd reaping, per-disk latch, O_DIRECT, `LocalIoBackend` integration, singleton ring lifecycle) remain P2 work; the design constraints they must honor are now written into `SPIKE.md`.


houseme:
## Superseded by the standalone repo → https://github.com/rustfs/uring

Per the decision to keep io_uring out of `rustfs/rustfs` and validate it in isolation, this implementation has been migrated to a dedicated repository **[rustfs/uring](https://github.com/rustfs/uring)** (crate `rustfs-uring`), preserving the full per-issue commit history from this PR.

Why the standalone repo is better here:
- **Independent CI that actually runs io_uring** — the main workspace excludes this crate (its `Test and Lint` passed in ~4s without building it). rustfs/uring CI is green on three jobs: `fmt + clippy`, a native `test` leg that runs the suite with **real io_uring on the runner and fails on any skip**, and a `docker` leg exercising both the graceful-degradation (seccomp-blocked) and real-io_uring paths.
- **No pollution of `rustfs/rustfs`** — main stays clean; the crate evolves on its own until it is production-ready.

Next step is integration as a git dependency behind runtime probing (`LocalIoBackend`), tracked in rustfs/backlog#1051. Closing this PR in favor of that path.

PR #4631, opened 2026-07-09
feat(ecstore): converge startup topology under a wait-mode policy

houseme:
## Related Issues

Tracked by rustfs/backlog#1105

## Summary of Changes

Multi-node startups (Kubernetes StatefulSets, bare-metal, VMs) can hit transient DNS/local-host resolution failures while peers and headless-service records are still coming up. Previously endpoint resolution retried DNS for a fixed 90s window and then returned an error, which propagated out of `run()` and exited the process; in Kubernetes this drove repeated kubelet restarts even though the cluster eventually converged, which reads to operators as an unstable service.

This adds a startup topology convergence policy so recoverable DNS/topology errors surface as "process Running, readiness=false, waiting" instead of a process exit, while genuine configuration errors still fail fast. The policy has three modes:

- `orchestrated`: wait effectively indefinitely for recoverable DNS/topology errors so the process stays Running (readiness stays false) instead of exiting; also defer the DNS-IP cross-port validation because headless-service records can still be flapping.
- `bounded`: wait a finite window (default 3m) then fail with an action-oriented error listing host, stage, attempts, elapsed, last error and remediation hints.
- `fail-fast`: fail on the first non-transient resolution error (CI/local debugging).

Mode is auto-detected (Kubernetes via `KUBERNETES_SERVICE_HOST` -> orchestrated; distributed URL endpoints on non-Kubernetes -> bounded; local path endpoints -> fail-fast) and can be overridden via `RUSTFS_STARTUP_TOPOLOGY_WAIT_MODE`, `RUSTFS_STARTUP_TOPOLOGY_WAIT_TIMEOUT` and `RUSTFS_STARTUP_TOPOLOGY_RETRY_MAX_DELAY`.

Additional behavior: divergent local/remote verdicts for endpoints that share a `host:port` are normalized (local wins), retry warnings are throttled (first attempt plus at most once per 30s), and DNS-independent checks (local same-path, physical-disk independence) still run in every mode. The change is confined to endpoint resolution in `crates/ecstore/src/layout/endpoints.rs` plus new env constants in `crates/config`; public API signatures (`from_volumes`, `create_server_endpoints`, `create_pool_endpoints`) are unchanged, and the fix stays inside the Rust startup chain (no `entrypoint.sh` dependency and no forced TCP/RPC preflight).

Files:

- `crates/config/src/constants/app.rs`: new env constants (`ENV_STARTUP_TOPOLOGY_WAIT_MODE` / `_WAIT_TIMEOUT` / `_RETRY_MAX_DELAY`, `ENV_KUBERNETES_SERVICE_HOST`) and defaults (180s wait, 8s retry cap).
- `crates/ecstore/src/layout/endpoints.rs`: `StartupTopologyWaitMode` / `StartupTopologyPolicy` (env-resolved via `rustfs_utils::get_env_opt_str`, test-injectable), duration parser, per-policy DNS retry deadline/backoff cap, same-host local normalization, orchestrated deferral of DNS-IP validation, throttled/action-oriented logging, and a one-line convergence summary.
- `crates/utils/src/envs.rs`: minor cleanup — simplify the `log_once` poisoned-lock branch to `unwrap_or_else(PoisonError::into_inner)`, matching the existing convention in `crates/obs` and `crates/iam`.

## Verification

- `cargo test -p rustfs-ecstore --lib layout::endpoints` (17 tests pass, including 5 new ones; uses injectable policy/sleep/deadline, no real long sleeps)
- `cargo test -p rustfs-utils --lib envs` (13 tests pass; covers the simplified `log_once`)
- `cargo clippy -p rustfs-ecstore -p rustfs-config -p rustfs-utils` (no warnings)
- `cargo check -p rustfs` (downstream binary builds)
- `scripts/check_logging_guardrails.sh`, `scripts/check_layer_dependencies.sh` (pass)

## Impact

- Deployment: Kubernetes/orchestrated multi-node startups no longer exit on transient DNS/topology errors, so pods stay Running with readiness false and stop accumulating kubelet restarts. Operators should keep a generous `startupProbe` `failureThreshold`, consider `publishNotReadyAddresses: true` on the headless service, and `podManagementPolicy: Parallel` on the StatefulSet; these are recommendations only, not runtime dependencies.
- Configuration: three new optional env vars (`RUSTFS_STARTUP_TOPOLOGY_WAIT_MODE|WAIT_TIMEOUT|RETRY_MAX_DELAY`), all defaulting to `auto` behavior.
- Compatibility: no on-disk or on-wire format change; setup-type classification is unchanged; mixed-version rolling upgrades are unaffected (each node resolves its own topology). One intentional behavior change: the non-Kubernetes distributed default wait grows from the previous fixed 90s to a bounded 180s; CI or environments that want immediate failure can set `RUSTFS_STARTUP_TOPOLOGY_WAIT_MODE=fail-fast`.
- API: no public API signature change.

## Additional Notes

Docker is intentionally not treated as orchestrated by auto-detection (container presence alone does not change behavior); Docker Swarm/compose distributed deployments fall into `bounded` or can opt into `orchestrated` explicitly. Non-transient and configuration errors (malformed `RUSTFS_VOLUMES`, mixed style/scheme, duplicate/root/empty paths, bare `host:port` without scheme, invalid port, local path reuse, shared physical disks, and later auth/RPC-secret/disk-format/IAM-quorum errors) are unchanged and still fail fast.

Adversarial Validation (high risk: init/distributed path) — one-line verdict per role:

- Correctness: attacked orchestrated skip vs setup-type derivation, empty/single-drive early return, and normalize-before-demotion ordering — no break found (covered by `orchestrated_policy_defers_dns_ip_same_path_check`).
- Security: only operator-controlled env parsing, no untrusted input/secret/injection surface — no finding.
- Concurrency/durability: single-threaded startup path, no locks, no shared mutable state, no persistence change — no finding.
- Compatibility: no format/wire change, rolling upgrade safe; documented the intentional 90s->180s bounded-default change. Endpoint-resolution tests inject a deterministic bounded policy so they no longer depend on the CI runner's ambient `KUBERNETES_SERVICE_HOST` (verified by running the suite with that variable set) — no unexpected break.
- Performance: startup-only, O(n) over a small endpoint set, no per-request/per-object hot-path impact — no finding.
- Test coverage: every behavior has a revert-fails test; two end-to-end gaps (indefinite orchestrated wait, fail-fast transient) covered indirectly via `resolve_from` (`Duration::MAX`/`ZERO`) plus the timeout/backoff unit tests to avoid hangs and real DNS.

### DIFF
--- a/crates/config/src/constants/app.rs
+++ b/crates/config/src/constants/app.rs
@@ -143,6 +143,41 @@ pub const ENV_MINIO_CI: &str = "MINIO_CI";
 /// Default flag value for bypassing local physical disk independence checks.
 pub const DEFAULT_UNSAFE_BYPASS_DISK_CHECK: bool = false;
 
+/// Environment variable selecting how startup waits for DNS/topology
+/// convergence in multi-node setups. One of `auto` (default), `orchestrated`,
+/// `bounded`, or `fail-fast`.
+///
+/// - `orchestrated`: wait effectively indefinitely for recoverable DNS/topology
+///   errors so the process stays Running (readiness stays false) instead of
+///   exiting and triggering a Kubernetes pod restart loop.
+/// - `bounded`: wait for a finite window (see
+///   `ENV_STARTUP_TOPOLOGY_WAIT_TIMEOUT`) then fail with an actionable error.
+/// - `fail-fast`: fail on the first non-transient resolution error (CI/local).
+/// - `auto`: Kubernetes -> orchestrated, distributed URL endpoints -> bounded,
+///   local path endpoints -> fail-fast.
+pub const ENV_STARTUP_TOPOLOGY_WAIT_MODE: &str = "RUSTFS_STARTUP_TOPOLOGY_WAIT_MODE";
+
+/// Environment variable bounding how long startup waits for topology/DNS
+/// convergence before failing in `bounded` mode. Accepts durations such as
+/// `3m`, `90s`, `500ms`, or a bare number of seconds.
+pub const ENV_STARTUP_TOPOLOGY_WAIT_TIMEOUT: &str = "RUSTFS_STARTUP_TOPOLOGY_WAIT_TIMEOUT";
+
+/// Environment variable capping the per-attempt backoff delay while retrying
+/// startup topology/DNS convergence. Accepts durations such as `8s`.
+pub const ENV_STARTUP_TOPOLOGY_RETRY_MAX_DELAY: &str = "RUSTFS_STARTUP_TOPOLOGY_RETRY_MAX_DELAY";
+
+/// Environment variable injected into every Kubernetes pod; its presence marks
+/// an orchestrated deployment for startup topology auto-detection.
+pub const ENV_KUBERNETES_SERVICE_HOST: &str = "KUBERNETES_SERVICE_HOST";
+
+/// Default bounded-mode wait window (seconds) for non-Kubernetes distributed
+/// deployments before startup topology convergence is declared failed.
+pub const DEFAULT_STARTUP_TOPOLOGY_WAIT_TIMEOUT_SECS: u64 = 180;
+
+/// Default per-attempt backoff cap (seconds) while retrying startup
+/// topology/DNS convergence.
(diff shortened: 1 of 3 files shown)


houseme:

PR #1048, opened 2025-12-08
feat(admin): Add file history API endpoints

tennisleng:
Splitting #1045: Admin API handlers.

### Key Differentiators

| Feature | Existing `ListObjectVersions` | This PR's Endpoints |
|---|---|---|
| **Point-in-time query** | ❌ Lists all versions, client must filter | ✅ `GetObjectAtTime` returns exact version at timestamp T |
| **Directory snapshot** | ❌ Not supported | ✅ `ListObjectsAtTime` shows entire directory as it appeared at time T |
| **User-friendly history** | ❌ Raw version data | ✅ `GetFileHistory` returns formatted history with `is_latest`, `is_deleted` flags |

PR #4632, opened 2026-07-09
feat(ecstore): runtime-probed io_uring read backend, gray-off by default (backlog#1104)

houseme:
## What

Wires the standalone [rustfs-uring](https://github.com/rustfs/uring) crate into `LocalIoBackend` as an **opt-in, gray-off** io_uring read backend — the first integration step of rustfs/backlog#1104 (P2 io_uring read path, following the #1048/#1051 audit).

## Behavior

- **Default build is unchanged.** The io_uring backend is used only when `RUSTFS_IO_URING_READ_ENABLE=true` **and** the per-disk io_uring probe succeeds.
- On a restricted host (seccomp/gVisor/old kernel) the probe fails and the disk uses `StdBackend`; on **any** per-read driver error a read falls back to `StdBackend` — byte-for-byte, so enabling it can never regress correctness.
- Only positioned reads (`pread_bytes`) go through the cancel-safe `UringDriver`; the other three `LocalIoBackend` methods delegate to the inner `StdBackend`.

## Design

- `Cargo.toml`: `rustfs-uring` as a **Linux-only git dependency** (pinned rev `39018c0`). The guard `scripts/check_no_tokio_io_uring.sh` explicitly allows an application-level io-uring integration; only the tokio `io-uring` runtime feature is banned.
- `UringBackend::pread_bytes` mirrors `StdBackend::pread_bytes`'s resolution / access-check / bounds preamble exactly and only swaps the raw byte read for `driver.read_at`.
- Backend chosen at `LocalDisk` construction via `build_local_io_backend`.

## Verification

- `cargo check -p rustfs-ecstore` on Linux (git dep resolved, cfg(linux) `UringBackend` compiled).
- New differential test `uring_backend_reads_match_std` passes under **real io_uring** (`seccomp=unconfined`): every positioned read returns byte-identical data (driver-served when io_uring is available, `StdBackend` fallback otherwise).

## Not in this PR (tracked separately)

- Runtime errno three-class degradation latch + per-disk probe cache → rustfs/backlog#1101.
- eventfd/AsyncFd reaping, O_DIRECT, async backpressure, singleton ring → rustfs/backlog#1102.
- 5.15-kernel EBUSY + fault-injection verification → rustfs/backlog#1103.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -112,4 +112,9 @@ jobs:
         with:
           fail-on-severity: moderate
           allow-ghsas: GHSA-2f9f-gq7v-9h6m
+          # rustfs-uring is a first-party Apache-2.0 crate pulled as a git
+          # dependency (backlog#1104); its license cannot be read from the git
+          # source, so allow it explicitly to avoid a spurious unknown-license
+          # warning on every Cargo.lock change.
+          allow-dependencies-licenses: pkg:cargo/rustfs-uring
           comment-summary-in-pr: always
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5327,6 +5327,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "io-uring"
+version = "0.7.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9080b15e63775b9a2ac7dca720f7050a8b955e092ea0f6020a4a80f69998cdc0"
+dependencies = [
+ "bitflags 2.13.0",
+ "cfg-if",
+ "libc",
+]
+
+[[package]]
 name = "ipconfig"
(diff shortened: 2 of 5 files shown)


houseme:
Re: Dependency Review — `rustfs-uring` "unknown license".

Addressed: `rustfs-uring` is a first-party Apache-2.0 crate (declares `license = "Apache-2.0"` and ships a LICENSE file at https://github.com/rustfs/uring); Dependency Review just can't read the SPDX from a git source. Added `allow-dependencies-licenses: pkg:cargo/rustfs-uring` to the dependency-review-action config in `.github/workflows/audit.yml` so the spurious unknown-license warning is suppressed on this and future Cargo.lock changes. No vulnerable/incompatible-license packages were flagged.

PR #4634, opened 2026-07-09
feat: add an opt-in NATS JetStream publish path for the notify and audit targets

simon-escapecode:
## Summary

This adds an opt-in JetStream publish path to the NATS notify and audit targets. With it enabled,
each event is published to a pre-provisioned JetStream stream and the local store-and-forward queue
entry clears only after the server returns a durable publish acknowledgement. The path is off by
default and the disabled path is unchanged.

## Problem

The NATS notify and audit targets publish through NATS Core. A NATS Core publish is fire and forget:
the client returns as soon as the message is written to the connection, before the server confirms
receipt, and the server retains nothing once the message is delivered. If the broker restarts, or the
connection drops between the publish and the flush, the event is lost, even though the on-disk send
queue has already treated it as delivered and removed it. No application-level acknowledgement exists
to gate that removal.

NATS Core is therefore not a reliable at-least-once transport for bucket notifications: events can be
dropped across routine broker maintenance or a transient network fault, with no error and no retry.

NATS provides at-least-once delivery through JetStream, where a publish targets a persisted stream and
returns a PublishAck once the server has durably stored the message. The change adds an opt-in
JetStream publish path to the NATS notify and audit targets. A queued event clears only after its
PublishAck, so delivery survives a broker restart or a reconnect. The default NATS Core behaviour is
unchanged.

## Design

When the feature is enabled on a target, the publish goes through JetStream and awaits the real
PublishAck before the queue entry is cleared:

- Publish through a cached JetStream context with the deduplication header set, and await the real
  PublishAck. The acknowledgement means the stream leader has accepted and sequenced the message.
- Clear the local store-and-forward queue entry only on the durable acknowledgement. A timeout is a
  retry with the same identifier, never a clear and never a silent drop. A crash before the
  acknowledgement leaves the entry in the queue, and it replays after restart.
- Set a stable Nats-Msg-Id on every publish, minted once when the event is queued and reused across
  every retry and replay of that entry, so the server collapses duplicates within the stream
  duplicate window. A retry after a slow acknowledgement reuses the same identifier and is not
  delivered twice.
- Validate the stream up front. At target init and in the admin validation path, a single stream
  lookup asserts the stream exists, captures the configured subject, acknowledges writes (no_ack
  false), is not sealed, and has a duplicate window of at least the worst-case retry span. RustFS
  never creates the stream. A missing or misconfigured stream fails validation rather than starting
  in a state where every publish would silently fail.
- Record events that can never be delivered. A terminal error (a bad subject, an authorization
  failure, a message that exceeds the maximum size) fails fast, and a transient error that does not
  clear within the bounded retries is exhausted. In both cases the event is moved to an on-disk
  failed-events store inside the queue directory, with the event body, the deduplication identifier,
  an error class tag, the failure time, and the retry count, and an error-level log line is written.
  The failed store is bounded by count and age and is kept separate from the live queue limit. A
  failed record is diagnostic only and is never republished. Only a permanent rejection is recorded,
  a retryable condition keeps its entry on the live queue until delivered.

The acknowledgement timeout defaults to 30 seconds. A server replicating to multiple replicas
acknowledges only after the replication and the deferred fsync, which legitimately takes well over
100 milliseconds at the tail, so a short timeout would spuriously fail mid-replication.

## Configuration

Three keys are added to both the notify NATS target and the audit NATS target:

- jetstream_enable. Boolean, default off. Turns the JetStream path on for this target.
- jetstream_stream_name. String, required when enable is on. The pre-provisioned stream to publish
  to.
- jetstream_ack_timeout_secs. Integer seconds, range 10 to 120, default 30. How long to wait for a
  publish acknowledgement before retrying.

Each key has the matching RUSTFS_NOTIFY_NATS_ and RUSTFS_AUDIT_NATS_ environment variable form. When
enable is on, both a stream name and a queue directory must be set, validated at startup and in the
admin validation path. The disabled path is byte-identical to the prior behaviour, so a deployment
that does not set jetstream_enable is unaffected.

## Operator requirement

The operator pre-provisions the stream with its retention, storage, and replication policy. The
stream must capture the publish subject, acknowledge writes, not be sealed, and set a duplicate
window of at least the worst-case retry span.

The duplicate window must cover the worst case so a late retry of an already-persisted event is
recognised as a duplicate rather than delivered a second time:

    duplicate_window >= (retry_count * ack_timeout_secs) + backoff_sum

There are 5 retry attempts and the backoff sum is 124 seconds. At the default 30 second
acknowledgement timeout the required window is (5 * 30) + 124 = 274 seconds. The span scales with
the acknowledgement timeout, so raising jetstream_ack_timeout_secs requires raising the stream
duplicate window in step. Validation rejects a stream whose duplicate window is below this span.

For durability across a node failure, provision the stream with a replica count of at least 3, since
a single-replica stream is durable only against a clean restart.

Full operator guidance is in docs/operations/nats-jetstream.md.

## Durability trade-offs

Two points in the failed-events store path look like durability gaps on a first read and are not. They
are called out here so a reviewer reads them as the intended behaviour.

The failed store writes each entry to a temp file in the target directory, fsyncs the file, then
renames it onto the final path. Rename within a single directory is atomic on the supported
filesystems, so a crash at any point leaves either the old file, no file, or the complete new file,
never a partial. The containing directory is fsynced after the rename on the live and failed
paths alike, and the failed directory's parent is fsynced after its lazy creation, so a completed
write survives a power loss on the supported journaling filesystems.

Moving a live queue entry to the failed store writes the failed entry first and deletes the live entry
only after the write succeeds. The two stores are separate, so the move is not a single atomic step. A
crash between the two leaves the entry in both: the failed copy is recorded and the live copy survives.
The surviving live copy replays after restart, and the stable Nats-Msg-Id minted when the event was
queued collapses it within the stream duplicate window, so the replay is recognised as a duplicate
rather than delivered twice. The terminal-class filter then routes it back to the failed store without
a second delivery attempt. The reverse order, deleting the live entry before the failed write, could
lose the event if the crash fell between the two. The chosen order favours a recoverable duplicate over
a loss, which is the correct bias for an at-least-once path.

## Testing

Unit tests cover:

- Configuration parsing and validation, including the acknowledgement timeout range, the
  stream-name-required-when-enabled rule, and the queue-directory-required-when-enabled rule.
- The worst-case retry span calculation and the duplicate-window enforcement at the default and
  boundary acknowledgement timeouts.
- Stream validation: subject binding by literal and wildcard, no_ack rejection, sealed rejection,
  and the too-small and zero duplicate-window rejections.
- The acknowledgement timeout resolution and its application on context rebuild.

A system reliability proof exercises the end-to-end behaviour against a real server:

- Kill-and-restart zero-loss. The server is killed mid-publish and restarted, and every event is
  accounted for with none lost.
- Deduplication. A replay of an already-persisted event is collapsed within the duplicate window
  rather than delivered twice.
- Terminal-to-store. A terminal failure moves the event to the failed-events store with the correct
  error class.
- Flag-off baseline. With the feature off, behaviour matches the NATS Core path.

## Platforms

The change is portable Rust with no platform-specific code on the publish path. The durable store
avoids filenames reserved on other filesystems and does not fsync directories, so its behaviour holds
across Linux, macOS, and Windows filesystems. It builds and its tests pass on all three.

- Linux: the crate unit tests, the live-server reliability rig, and the TLS broker tests all pass.
- macOS on Apple Silicon: the crate unit tests pass and the full server binary builds.
- Windows 11 on x86-64: the crate unit tests pass and the full server binary builds.

The live-server rig and broker tests run on Linux. Their coverage is the platform-independent publish
protocol, already proven against a real server there, while the platform-sensitive on-disk store is
covered by the unit tests that pass on all three.


overtrue (changes requested):
Thanks for putting this together. The JetStream direction makes sense, but I don't think this is safe to merge yet. A few durability issues need fixing first:

- Stream validation fails in NATSTarget::init(), but the generic store-backed activation path ignores that error and still starts replay. A missing stream or an undersized duplicate window therefore doesn't actually fail fast.
- Legacy queue entries fall back to a SHA-256 of the body for Nats-Msg-Id. Two separate queued events with identical bodies get the same ID, so the second one can be deduplicated and then removed locally. The stable queue key should be part of the fallback ID.
- The failed-store move fsyncs the temp file, renames it, and then deletes the live queue entry without fsyncing the failed directory. A power loss in that window can leave neither copy durable.
- Several temporary JetStream states, including stream offline / cluster not leader, fall through as terminal. They are moved out of the live queue, but there is no failed-store read or requeue path, and those entries are pruned after 72 hours.
- The last failed attempt still sleeps for another 64 seconds even though there is no next retry. Since replay is serial, a backlog of failed entries can take hours or days longer than necessary.

The checks are green, but the broker-backed NATS tests are ignored and the kill/restart, reconnect, and replay cases described in the PR aren't exercised in CI. Once the issues above are fixed, please add a real broker-backed regression covering those failure paths.

houseme:
@codex review

crates/targets/src/error.rs
```diff
@@ -52,6 +52,9 @@ pub enum TargetError {
     #[error("Request error: {0}")]
     Request(String),
 
+    #[error("JetStream publish error ({}): {detail}", if *.retryable { "retryable" } else { "terminal" })]
+    JetStreamPublish { retryable: bool, detail: String },
```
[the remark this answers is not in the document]
simon-escapecode: The format expression is valid thiserror syntax. A leading dot expands to the field binding, so if *.retryable dereferences the retryable field into the boolean the conditional needs. The crate compiles and the full CI build passes, confirming the Display impl generates correctly.

crates/targets/src/target/mod.rs
```diff
@@ -593,6 +651,133 @@ pub(crate) fn delete_stored_payload(
...
+) -> Result<Vec<u8>, TargetError> {
+    let failed_at_unix_ms = SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or_default().as_millis() as u64;
+    queued.meta.failure = Some(FailedEntryMeta {
+        error_class,
+        error_detail: build_failed_error_detail(error),
+        nats_msg_id: queued.meta.dedup_id.clone(),
+        failed_at_unix_ms,
+        retry_count,
+    });
+    queued.encode()
```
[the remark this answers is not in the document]
simon-escapecode: Fixed. The failed-store move now resolves the dedup id the same way the publish path does (resolve_dedup_id over the entry meta and key) and records that resolved id as nats_msg_id in the failed entry and the moved-to-failed-store log, so an entry queued before enable carries the id the publish attempted rather than an empty string. The fallback derives from the entry on-disk key rather than a hash of the body, so two events with identical bodies keep distinct identifiers.

crates/targets/src/store.rs
```diff
@@ -838,6 +999,76 @@ where
...
+                Err(err) if err.kind() == std::io::ErrorKind::NotFound => {}
+                Err(err) => return Err(StoreError::Io(err)),
+            }
+        }
+        Ok(pruned)
+    }
+
+    fn failed_len(&self) -> usize {
+        self.failed_entries_oldest_first().map(|entries| entries.len()).unwrap_or(0)
+    }
```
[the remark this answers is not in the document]
simon-escapecode: Resolved. failed_len now returns a read_dir count that skips temp files, without the per-file stat or the mtime sort. The ordered scan is kept only where ordering matters, removing expired entries and trimming the store to its capacity bound. Removal of a residual temp file also stays with the ordered scan, so the count only reads the directory and never deletes anything.

crates/targets/src/target/nats.rs
```diff
@@ -350,6 +460,285 @@ where
...
+/// entries queued before JetStream was enabled. The hash is over the decoded body the publish sends,
+/// so it is identical across every replay of the same entry.
+fn resolve_dedup_id(meta_dedup_id: &str, body: &[u8]) -> String {
+    if !meta_dedup_id.is_empty() {
+        return meta_dedup_id.to_string();
+    }
+    let digest = Sha256::digest(body);
+    let mut id = String::with_capacity(digest.len() * 2);
+    for byte in digest {
+        id.push_str(&format!("{byte:02x}"));
```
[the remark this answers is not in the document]
simon-escapecode: Resolved. The fallback no longer hashes the body. For an entry queued before JetStream was enabled, resolve_dedup_id now derives the id from the entry on-disk key (to_key_string, a per-entry UUIDv7 name), unique per distinct entry and stable across replays of the same entry. Two events with identical bodies therefore get distinct Nats-Msg-Id values, and a genuine retry of one entry still deduplicates. Tests cover the distinct-entries and same-entry-stability cases.

crates/targets/src/target/mod.rs
```diff
@@ -593,6 +651,133 @@ pub(crate) fn delete_stored_payload(
...
+        object = %meta.object_name,
+        event_name = %meta.event_name.as_str(),
+        nats_msg_id = %meta.dedup_id,
+        retry_count,
+        "event moved to the failed-events store"
+    );
+
+    // Delete the live entry only after the failed entry is durably written and the alarm is emitted.
+    delete_stored_payload(store, key)?;
+    target.record_final_failure();
```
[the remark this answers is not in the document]
simon-escapecode: Resolved. The record_final_failure call inside the failed-store move is removed. The replay worker already reports PermanentFailure or RetryExhausted for the same failure and the notify and audit hooks record it, so a failed JetStream delivery counts once. Every failed-store move is paired with a hook event, so no path under-counts.

simon-escapecode:
> Thanks for putting this together. The JetStream direction makes sense, but I don't think this is safe to merge yet. A few durability issues need fixing first:
> 
> * Stream validation fails in NATSTarget::init(), but the generic store-backed activation path ignores that error and still starts replay. A missing stream or an undersized duplicate window therefore doesn't actually fail fast.
> * Legacy queue entries fall back to a SHA-256 of the body for Nats-Msg-Id. Two separate queued events with identical bodies get the same ID, so the second one can be deduplicated and then removed locally. The stable queue key should be part of the fallback ID.
> * The failed-store move fsyncs the temp file, renames it, and then deletes the live queue entry without fsyncing the failed directory. A power loss in that window can leave neither copy durable.
> * Several temporary JetStream states, including stream offline / cluster not leader, fall through as terminal. They are moved out of the live queue, but there is no failed-store read or requeue path, and those entries are pruned after 72 hours.
> * The last failed attempt still sleeps for another 64 seconds even though there is no next retry. Since replay is serial, a backlog of failed entries can take hours or days longer than necessary.
> 
> The checks are green, but the broker-backed NATS tests are ignored and the kill/restart, reconnect, and replay cases described in the PR aren't exercised in CI. Once the issues above are fixed, please add a real broker-backed regression covering those failure paths.

Thanks for the detailed review. All really good points. I have addressed all of this now (taken a while!):

Stream validation now gates the publish path itself through a per-context verdict. A target whose stream fails validation cannot publish and queued entries stay queued. Validation runs again on a later scan, so a repaired stream resumes publishing without a restart, and is_active reports the validation state instead of connectivity alone. The generic store-backed activation path is unchanged for the other targets.

resolve_dedup_id derives the legacy fallback from the entry's stable on-disk key (a per-entry UUIDv7 name) instead of a body hash, so two queued events with identical bodies keep distinct Nats-Msg-Id values while a genuine retry of one entry stays deduplicated. Unit tests cover the distinct-entries and same-entry-stability cases.

The failed-store write now syncs the failed directory after the rename and the parent directory after the lazy create, matching the durability discipline of the live queue write path. The failed filename also derives from the entry key, so a repeated move of one entry overwrites its own file instead of accumulating copies.

The transient cluster and stream conditions are reclassified retryable: leaderless cluster, peer membership change, offline stream, JetStream not ready, and the resource and quota exhaustion codes, plus a back-stop that treats any server rejection reporting a 5xx status as retryable even when it carries no specific error code. The complete async-nats error surface was audited for the classification, the classifier tests pin the set in both directions, and the operations guide now documents the outcome families and the policy. One note: the no-responders case was already retryable, the misclassified set was the cluster and stream state codes.

The backoff sleep after the final attempt is removed. Retries keep their exponential spacing, but the fifth failure moves the entry onward immediately, so a backlog of failing entries clears without the trailing dead time per entry.

The kill and restart, reconnect, and replay paths were verified against a live broker: queued events persist on disk with the broker down, replay drains them exactly once on restart, a re-published identifier is deduplicated by the server, and a cold-connect outage exhausts the bounded budget into the failed events store instead of looping. The repository's broker-backed tests cover these paths and run against a live server (they are ignored by default so the unit suite stays hermetic). Wiring a NATS service container into CI so they run there too is a workflow change, guidance on the preferred pattern is welcome.

crates/targets/src/target/nats.rs
```diff
@@ -350,6 +481,384 @@ where
...
+/// surfaces as a raw Network or Configuration error, which the replay worker treats as permanent
+/// and leaves in the live queue, reported as a failure on every scan without bound. Routing it as a
+/// retryable JetStream error bounds the same-id retry span to the retry lifetime and moves the
+/// entry to the failed store on exhaustion. The Configuration case covers material that degrades
+/// after start, unreadable TLS files being the concrete producer, and an operator repairing the
+/// file heals the target without a restart. Other errors pass through unchanged.
+fn jetstream_connect_error(err: TargetError) -> TargetError {
+    match err {
+        TargetError::Network(detail) | TargetError::Configuration(detail) => {
+            TargetError::JetStreamPublish { retryable: true, detail }
```
overtrue: This mapping is what routes a plain broker outage into the failed store: a connect failure becomes a retryable `JetStreamPublish` error, the replay loop exhausts its five attempts within minutes, and the entry leaves the live queue for a store that prunes at 72h. The flag-off path keeps the same entry queued indefinitely and delivers it on reconnect. My preferred fix is to keep `Network` errors as `NotConnected` here (entry stays queued, same as the core path) and reserve the bounded-retry → failed-store funnel for publish-level errors and validation failures.

Separate smaller thing on the same lines: `detail` carries the raw connect/configuration error text. `build_failed_error_detail` treats the `JetStreamPublish` detail as allowlisted-verbatim ("constructed to exclude the server body"), but this path can put an address or the credentials-file path into it (e.g. "Failed to load NATS credentials file: …"). A fixed category string ("connect: network" / "connect: configuration") keeps the invariant — the concrete cause is already logged at the connect site.

crates/targets/src/target/mod.rs
```diff
@@ -593,6 +651,143 @@ pub(crate) fn delete_stored_payload(
...
+        TargetError::Timeout(value) => format!("timeout: {}", redacted_secret(value)),
+        TargetError::Storage(value) => format!("storage: {}", redacted_secret(value)),
+        TargetError::Authentication(_) => "authentication".to_string(),
+        TargetError::Configuration(_) => "configuration".to_string(),
+        other => format!("error: {}", redacted_secret(&other_error_category(other))),
+    };
+
+    let mut detail = summary;
+    if detail.len() > FAILED_ERROR_DETAIL_MAX_LEN {
+        detail.truncate(FAILED_ERROR_DETAIL_MAX_LEN);
```
overtrue: `String::truncate` panics when the cut lands inside a multi-byte character. The only arm that feeds free text into this function is `JetStreamPublish`, and via the connect-error mapping that text can be a localized OS error string — non-ASCII is realistic on the Windows builds this PR explicitly supports. A panic here takes down the replay worker task for the target and replay silently stops until restart.

Needs a char-boundary-safe cut, e.g.:

```rust
if detail.len() > FAILED_ERROR_DETAIL_MAX_LEN {
    let mut cut = FAILED_ERROR_DETAIL_MAX_LEN;
    while !detail.is_char_boundary(cut) {
        cut -= 1;
    }
    detail.truncate(cut);
}
```

crates/targets/src/store.rs
```diff
@@ -838,6 +1019,117 @@ where
...
+        let _fs_guard = self
+            .fs_guard
+            .read()
+            .map_err(|_| StoreError::Internal("Failed to acquire read lock on store filesystem".to_string()))?;
+
+        let now = SystemTime::now();
+        let mut pruned = 0usize;
+        for (path, written_at) in self.failed_entries_oldest_first()? {
+            let age = now.duration_since(written_at).unwrap_or_default();
+            if age < FAILED_STORE_TTL {
```
overtrue, line 1160: For `Terminal` entries a TTL is fine — replaying them without a config fix just repeats the failure. For `TransientExhausted` entries this prune is the point where an unattended outage becomes permanent loss: these are exactly the events a recovered server would accept, and there is no read or requeue path anywhere in the tree, so "recorded, not silently dropped" holds for 72 hours only.

If connect failures stay in the live queue (direction 1 in my review) this matters much less. Otherwise `TransientExhausted` should either be exempt from the TTL or be requeued into the live store here instead of removed — the stable dedup id makes the requeue safe within the duplicate window.

docs/operations/nats-jetstream.md
```diff
@@ -0,0 +1,242 @@
...
+- Validation fails at startup with a missing-stream error. The stream named in
+  jetstream_stream_name does not exist on the server, or is not reachable.
+  Provision it, or correct the name.
+- Validation fails with a subject, no_ack, sealed, or duplicate-window error. The
+  pre-provisioned stream does not capture the publish subject, has no_ack set, is
+  sealed, or has a duplicate window below the worst-case retry span. Correct the
+  stream configuration to satisfy the pre-provisioning requirements above.
+- Repeated stream-not-found errors in the log. The stream disappeared or was
+  never created. The publish is retried within a bound and then recorded in the
+  failed store. Provision the stream so queued events recover on the next replay.
```
overtrue: This recovery claim only holds for entries still in the live queue. Entries already moved to the failed store are not replayed by anything — they sit until the 72h prune deletes them. Either the behaviour or this sentence needs to change; if the behaviour stays, this section needs an explicit recovery procedure for failed-store entries (today there is none, and the PR description's "inspect-and-replay path" doesn't exist in the code).

overtrue (changes requested):
Thanks for the thorough follow-up. I re-verified all five points from the last round against ecd0a108d, and four of them are fixed exactly as described: the per-context validation verdict now gates every publish (plus the admin check and is_active), the dedup fallback derives from the stable on-disk key, the failed-store writes sync the directories, the classifier rework is pinned in both directions, and the trailing backoff is gone. Unit suite and clippy are clean locally. The paused-clock replay tests and the byte-identical flag-off pin are nice touches.

One point from the last round is still open, and it's why I'm keeping this at request changes:

- The failed store still has no way back — no read, no requeue, no admin surface, no documented manual procedure (`put_failed_raw`/`prune_failed_store`/`failed_len` is the entire API) — and the connect-error mapping now funnels ordinary broker outages into it. Once the bounded retry budget is spent (minutes), entries leave the live queue for a store that evicts oldest at 10k and permanently deletes at 72h. With the flag off, the same outage keeps every entry in the live queue indefinitely and delivers on reconnect with zero operator action. So the durability feature is currently weaker than the core path for exactly the scenario it exists to survive: an outage nobody notices within 72 hours.
- Related, three descriptions don't match the implementation and need correcting whichever way the behaviour goes: the PR description's "inspect-and-replay path" for recovery (it doesn't exist in the code), "queued entries stay queued" for validation failures in your reply (they move to the failed store after five attempts — `jetstream_connect_failure_retries_then_writes_transient_exhausted` pins exactly that), and the ops guide's "queued events recover on the next replay" (only true for entries still in the live queue).

Suggested directions, in order of my preference — any one of them resolves the block:

1. Map context-acquisition/connect failures to `NotConnected` so a plain outage keeps the flag-off semantics: the entry stays queued and is retried across scans indefinitely. Keep the bounded-retry → failed-store funnel for publish-level errors and validation failures, where fail-fast-and-record is the right call. Smallest diff, cleanest semantics.
2. Auto-requeue `TransientExhausted` entries into the live queue on the prune tick. The dedup ids are stable, so the duplicate window absorbs the replays; `Terminal` keeps the current lifecycle.
3. At minimum: exempt `TransientExhausted` from the TTL prune (keep the count bound), document a real recovery procedure, and fix the three descriptions above.

Two smaller items are inline: a char-boundary panic in the failed-detail truncation that can kill a replay worker, and raw connect-error text reaching the failed-entry detail against the stated allowlist invariant.

On the CI question you asked: `services:` containers can't take command arguments, so `-js` won't fit there. The simple pattern is a plain step on the Linux test job:

```
docker run -d -p 4222:4222 nats:2 -js
cargo test -p rustfs-targets --test nats_jetstream_regression_guards --test nats_jetstream_validation_integration -- --ignored
```

With direction 1 (or 2/3) plus the doc corrections I'm happy to approve — the rest of this is in good shape.

houseme (approved):

simon-escapecode:
> Thanks for the thorough follow-up. I re-verified all five points from the last round against [ecd0a10](https://github.com/rustfs/rustfs/commit/ecd0a108da281610b96bbf5934405a42deb199b1), and four of them are fixed exactly as described: the per-context validation verdict now gates every publish (plus the admin check and is_active), the dedup fallback derives from the stable on-disk key, the failed-store writes sync the directories, the classifier rework is pinned in both directions, and the trailing backoff is gone. Unit suite and clippy are clean locally. The paused-clock replay tests and the byte-identical flag-off pin are nice touches.
> 
> One point from the last round is still open, and it's why I'm keeping this at request changes:
> 
> * The failed store still has no way back — no read, no requeue, no admin surface, no documented manual procedure (`put_failed_raw`/`prune_failed_store`/`failed_len` is the entire API) — and the connect-error mapping now funnels ordinary broker outages into it. Once the bounded retry budget is spent (minutes), entries leave the live queue for a store that evicts oldest at 10k and permanently deletes at 72h. With the flag off, the same outage keeps every entry in the live queue indefinitely and delivers on reconnect with zero operator action. So the durability feature is currently weaker than the core path for exactly the scenario it exists to survive: an outage nobody notices within 72 hours.
> * Related, three descriptions don't match the implementation and need correcting whichever way the behaviour goes: the PR description's "inspect-and-replay path" for recovery (it doesn't exist in the code), "queued entries stay queued" for validation failures in your reply (they move to the failed store after five attempts — `jetstream_connect_failure_retries_then_writes_transient_exhausted` pins exactly that), and the ops guide's "queued events recover on the next replay" (only true for entries still in the live queue).
> 
> Suggested directions, in order of my preference — any one of them resolves the block:
> 
> 1. Map context-acquisition/connect failures to `NotConnected` so a plain outage keeps the flag-off semantics: the entry stays queued and is retried across scans indefinitely. Keep the bounded-retry → failed-store funnel for publish-level errors and validation failures, where fail-fast-and-record is the right call. Smallest diff, cleanest semantics.
> 2. Auto-requeue `TransientExhausted` entries into the live queue on the prune tick. The dedup ids are stable, so the duplicate window absorbs the replays; `Terminal` keeps the current lifecycle.
> 3. At minimum: exempt `TransientExhausted` from the TTL prune (keep the count bound), document a real recovery procedure, and fix the three descriptions above.
> 
> Two smaller items are inline: a char-boundary panic in the failed-detail truncation that can kill a replay worker, and raw connect-error text reaching the failed-entry detail against the stated allowlist invariant.
> 
> On the CI question you asked: `services:` containers can't take command arguments, so `-js` won't fit there. The simple pattern is a plain step on the Linux test job:
> 
> ```
> docker run -d -p 4222:4222 nats:2 -js
> cargo test -p rustfs-targets --test nats_jetstream_regression_guards --test nats_jetstream_validation_integration -- --ignored
> ```
> 
> With direction 1 (or 2/3) plus the doc corrections I'm happy to approve — the rest of this is in good shape.

Thanks for the re-review. Note it ran against ecd0a10, and the branch was rebuilt after that point: the current content is the single commit c86e07617, under the branch update merge.

The blocking point is resolved there, direction 1 and further: connect failures map to NotConnected and stay queued indefinitely, and the failed store now records only non-retryable rejections, so no recoverable event can enter it. One divergence: validation failures also stay queued rather than funneling after five attempts, since an unprovisioned stream is recoverable. A test asserts that only a non-retryable error ever produces a failed-store entry.

Both inline items are fixed (character-boundary truncation with a multi-byte test, fixed-label details only). All three descriptions are corrected, including the PR description.

Happy to add the suggested CI step in this PR or a follow-up, whichever is preferred.

crates/targets/src/target/nats.rs
simon-escapecode: Resolved in commit c86e07617 with the suggested mapping: Network and Configuration connect failures map to NotConnected, so the entry stays queued indefinitely and delivers on reconnect, the same behaviour as the core NATS path. In addition, the failed store now records only permanent rejections, so no recoverable event can reach it.

On the detail point: a stored record now carries only a short fixed label plus the numeric server code where one exists. Anything outside the fixed label set is replaced with a placeholder before it is written, so an address or a credentials-file path never reaches a stored record. The full error text still appears in the log at the connection site.

crates/targets/src/target/mod.rs
simon-escapecode: Resolved in commit c86e07617 with the suggested fix: the cut moves back to the nearest character boundary before truncating, covered by a unit test using multi-byte characters straddling the cap. The free-text source is also closed, stored details are fixed labels and numeric codes rather than raw error strings, so the truncation now guards a bounded input.

crates/targets/src/store.rs
simon-escapecode: Resolved in commit c86e07617. Connect failures stay in the live queue, and entries that run out of retries are no longer moved at all, they remain queued and keep retrying on later scans. The class named in the comment no longer exists, the failed store records only permanent rejections, so the TTL applies only to entries that would fail the same way if replayed. A recovered server never has anything to accept from the failed store, because a recoverable event never enters it.

docs/operations/nats-jetstream.md
simon-escapecode: Resolved in commit c86e07617, both the behaviour and the sentence changed. Stream-not-found is retryable now, the entry stays in the live queue until the stream exists, and nothing recoverable reaches the failed store. The guide states that a failed record is diagnostic only and never republished, and the inspect-and-replay claim is removed from the PR description.

overtrue (approved):

simon-escapecode:
@houseme : The Cargo Deny check is not from this PR. It fails on the spin crate, which the advisories database now reports as withdrawn from the registry. The dependency is unchanged from main and nothing here pulls it in. Cargo Deny runs on pull requests, so this looks to affect open PRs in general and needs a lockfile update on main to clear.

PR #1130, opened 2025-12-12
render imagePullSecrets in Deployment/StatefulSet

czaloumis:
## Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
#1123 

## Checklist
- [ ] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [ ] Passed `make pre-commit`
- [ ] Added/updated necessary tests
- [ ] Documentation updated (if needed)
- [ ] CI/CD passed (if applicable)

## Impact
- [ ] Breaking change (compatibility)
- [ ] Requires doc/config/deployment update
- [ ] Other impact:

## Additional Notes

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

PR #1127, opened 2025-12-12
add image pull secret support

majinghe:
## Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues

https://github.com/rustfs/rustfs/issues/1123

## Checklist
- [ ] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [ ] Passed `make pre-commit`
- [ ] Added/updated necessary tests
- [ ] Documentation updated (if needed)
- [ ] CI/CD passed (if applicable)

## Impact
- [ ] Breaking change (compatibility)
- [ ] Requires doc/config/deployment update
- [ ] Other impact:

## Additional Notes

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

PR #1115, opened 2025-12-11
Update build.yml

ghost:
## Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [x] Refactor
- [ ] Other:

## Checklist
- [x] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [ ] Passed `make pre-commit`
- [x] Added/updated necessary tests
- [x] Documentation updated (if needed)
- [x] CI/CD passed (if applicable)

## Impact
- [x] Breaking change (compatibility)
- [x] Requires doc/config/deployment update
- [x] Other impact:

## Additional Notes

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

Issue #1114, opened 2025-12-11
Failed to get cluster info using mc


whls:
**Describe the bug**
Failed to get cluster info using mc. 
Error message:  mc: <ERROR> Unable to get service info. The specified bucket does not exist.

**To Reproduce**
1. set alias rustfs  with mc. 
2. list bucket successful with command "mc ls rustfs".
3. use command "mc admin info rustfs" to get cluster info.
4. Failed with error message:  mc: <ERROR> Unable to get service info. The specified bucket does not exist.
5. with parameter debug
[root@rustfs-node01 ~]# mc admin info rustfs --debug
mc: <DEBUG> GET /minio/admin/v3/info?metrics=false HTTP/1.1
Host: 172.20.57.161:9000
User-Agent: MinIO (linux; amd64) madmin-go/3.0.70 mc/RELEASE.2025-05-21T01-59-54Z
Accept-Encoding: zstd,gzip
Authorization: AWS4-HMAC-SHA256 Credential=rustfsadmin/20251211//s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20251211T065456Z

mc: <DEBUG> HTTP/1.1 404 Not Found
Content-Length: 132
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: *
Content-Type: application/xml
Date: Thu, 11 Dec 2025 06:54:56 GMT
Vary: origin, access-control-request-method, access-control-request-headers
X-Request-Id: 2085c5a1-66d1-4951-b255-cfdc64d7ba6f

<?xml version="1.0" encoding="UTF-8"?><Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist</Message></Error>mc: <DEBUG> Response Time:  2.961669ms

**Expected behavior**
Based on the official documentation, the mc admin info command can obtain cluster information

**Screenshots**

![Image](https://github.com/user-attachments/assets/c328a8bc-9460-4ac4-84b3-6badff058853)

**Desktop (please complete the following information):**
 - OS: CentOS Linux release 8.4.2105
 - mc version:  RELEASE.2025-05-21T01-59-54Z
 - rustfs Version:  rustfs 1.0.0-alpha.72

loverustfs:
Hey @whls ,

[...]

PR #4655, opened 2026-07-10
fix(object-data-cache): make memory container-aware and bound cache config

houseme:
## Summary

The object data cache sized itself — and decided when to stop filling — from **host** memory, and its `validate()` left several numeric knobs unbounded. On Kubernetes, where pod limits sit far below node RAM, both the sizing input and the safety valve were wrong by the node/pod ratio. Separately, one env var could panic the process at startup.

## The defects

**Capacity and the memory gate ignored cgroup limits** (backlog [#1110](https://github.com/rustfs/backlog/issues/1110)). `resolved_max_bytes()` derived capacity from `System::total_memory()`, and the fill gate compared `available_memory()` against `min_free_memory_percent`. In a 2 GiB pod on a 256 GiB node, capacity resolved to ~12.8 GiB and the gate saw ~200 GiB free, so it never paused fills: ordinary GET traffic grew the cache past the pod limit and the kernel OOM-killed the process.

Both paths now resolve through one helper in `memory.rs`. `select_effective_memory()` is pure and unit-tested; a cgroup only overrides when it actually constrains below the host, so an unlimited cgroup transparently falls back to host values. The resolved capacity and its basis (`host` / `cgroup`) are logged once at startup.

**An over-long TTL panicked the builder** (backlog [#1113](https://github.com/rustfs/backlog/issues/1113)). `validate()` rejected only zero durations. `Duration::from_secs(u64)` from `RUSTFS_OBJECT_DATA_CACHE_TTL_SECS` flowed straight into moka, which **asserts** above ~1000 years. `from_config_or_disabled` catches `Err`, not `panic`, so a single mistyped env var aborted `AppContext::new` — a CrashLoopBackoff from a setting every other invalid value merely warns about. Now bounded at 30 days, so the adapter degrades to no-op as documented.

**`max_entry_bytes` inflated total capacity** (backlog [#1114](https://github.com/rustfs/backlog/issues/1114)). `clamp_derived_max_bytes` applied it as a **floor** on total capacity, and `safe_cap` was itself floored by it — so setting only the per-entry cap silently raised the whole cache above the percent-of-RAM/64 GiB clamp that function exists to enforce (512 MiB RAM + `MAX_ENTRY_BYTES=1 GiB` → a 1 GiB cache). The floor is gone; an entry that cannot fit is now a config error, never a reason to grow the cache.

Three further `validate()` gaps: an entry larger than an explicit `max_bytes` was admitted, inserted, immediately evicted, and reported as `Inserted` forever (backlog [#1115](https://github.com/rustfs/backlog/issues/1115)); `max_entry_bytes` above `u32::MAX` silently bypassed moka's u32 weigher (backlog [#1127](https://github.com/rustfs/backlog/issues/1127)); `time_to_idle > ttl` is inert and now warns (backlog [#1140](https://github.com/rustfs/backlog/issues/1140)).

**Malformed numeric env values enabled the cache with defaults** (backlog [#1130](https://github.com/rustfs/backlog/issues/1130)). `from_env_or_disabled` documents "falling back to no-op on invalid input", which held for bad mode strings and `validate()` failures — but `get_env_opt_u64` and friends returned `None`, so the cache ran **enabled** with settings the operator never chose. `get_env_parse_outcome::<T>()` (additive, in `rustfs-utils`) now distinguishes absent from malformed; malformed keys invalidate the whole config and yield the disabled adapter with one aggregated warning.

`identity_keys_max >= 2` is also enforced: the identity index admits a new key by evicting the oldest, so a budget of 1 evicts the previous key on every fill and can never hold two live keys of one object. That policy change lands in the sibling index-races PR; this validation is its config-side half (backlog [#1128](https://github.com/rustfs/backlog/issues/1128)).

## Notes for review

- `crates/object-data-cache` gains a `tracing` dependency; it had no logging, and #1110 requires reporting the resolved capacity and its basis where capacity is resolved.
- `Cargo.lock` also syncs `rustfs-uring` to `f577ba0`. This is a **pre-existing** inconsistency on `main`: `crates/ecstore/Cargo.toml` already pins that rev while the committed lock still held `39018c0`, so every build silently rewrote the lockfile. Unrelated to this change, but required for a lockfile that stays clean.

## Verification

```
cargo test -p rustfs-object-data-cache   49 passed; 0 failed
cargo test -p rustfs object_data_cache   28 passed; 0 failed
cargo test -p rustfs-utils envs          14 passed; 0 failed
cargo clippy -p rustfs-object-data-cache --all-targets   0 warnings
make pre-commit                          passed
```

The TTL guard was mutation-tested: short-circuiting it makes `overlong_ttl_builds_disabled_adapter_without_panic` fail with the real panic it exists to prevent —

```
panicked at moka-0.12.15/src/common/builder_utils.rs:11:9:
time_to_live is longer than 1000 years
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9575,6 +9575,7 @@ dependencies = [
  "sysinfo",
  "thiserror 2.0.18",
  "tokio",
+ "tracing",
 ]
 
 [[package]]
@@ -10030,7 +10031,7 @@ dependencies = [
 [[package]]
 name = "rustfs-uring"
 version = "0.1.0"
-source = "git+https://github.com/rustfs/uring?rev=39018c09cde98b380658f1a1511349a3574daddf#39018c09cde98b380658f1a1511349a3574daddf"
+source = "git+https://github.com/rustfs/uring?rev=f577ba0bfd6c3d74f9bdc54b573b8fcb45d6bc22#f577ba0bfd6c3d74f9bdc54b573b8fcb45d6bc22"
 dependencies = [
  "io-uring",
  "libc",
--- a/crates/object-data-cache/Cargo.toml
+++ b/crates/object-data-cache/Cargo.toml
@@ -35,6 +35,7 @@ starshard.workspace = true
 sysinfo = { workspace = true, features = ["multithread"] }
 thiserror.workspace = true
 tokio = { workspace = true, features = ["sync"] }
+tracing.workspace = true
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["macros", "rt", "time"] }
--- a/crates/object-data-cache/src/cache.rs
+++ b/crates/object-data-cache/src/cache.rs
@@ -348,6 +348,9 @@ mod tests {
         let config = ObjectDataCacheConfig {
             mode: ObjectDataCacheMode::FillBufferedOnly,
             max_bytes: 8_388_608,
+            // Fill must not depend on the live memory reading, which differs
+            // between a developer host and a CI container.
+            min_free_memory_percent: 0,
             ..ObjectDataCacheConfig::default()
         };
         ObjectDataCache::new(config).expect("fill-enabled cache config should initialize")
--- a/crates/object-data-cache/src/config.rs
+++ b/crates/object-data-cache/src/config.rs
@@ -85,7 +85,9 @@ pub struct ObjectDataCacheConfig {
     /// See [`ttl`](Self::ttl): expiration uses `min(ttl, time_to_idle)`, so
     /// setting `time_to_idle` above `ttl` is inert.
     pub time_to_idle: Duration,
-    /// Minimum free memory percent before fill is paused.
+    /// Minimum free memory percent before fill is paused. Zero disables the
+    /// memory gate entirely, which makes fill admission independent of the
+    /// host's or container's live memory reading.
     pub min_free_memory_percent: u8,
     /// Fill concurrency multiplier applied to CPU count.
     pub fill_concurrency_per_cpu: u16,
@@ -210,7 +212,8 @@ impl ObjectDataCacheConfig {
             );
         }
 
-        if self.min_free_memory_percent == 0 || self.min_free_memory_percent > 100 {
+        // Zero is a deliberate opt-out of the memory gate, not an invalid value.
+        if self.min_free_memory_percent > 100 {
             return Err(ObjectDataCacheConfigError::InvalidMinFreeMemoryPercent);
         }
 
@@ -344,6 +347,30 @@ mod tests {
             .expect_err("max fill concurrency must not be smaller than per-cpu factor");
 
         assert_eq!(err, ObjectDataCacheConfigError::FillConcurrencyMaxTooSmall);
+    }
+
+    #[test]
+    fn validate_accepts_zero_min_free_memory_percent_as_gate_opt_out() {
+        let config = ObjectDataCacheConfig {
+            min_free_memory_percent: 0,
+            ..ObjectDataCacheConfig::default()
+        };
+
+        assert!(config.validate().is_ok());
+    }
+
+    #[test]
+    fn validate_rejects_min_free_memory_percent_above_100() {
+        let config = ObjectDataCacheConfig {
+            min_free_memory_percent: 101,
+            ..ObjectDataCacheConfig::default()
+        };
+
+        let err = config
+            .validate()
+            .expect_err("a free-memory floor above 100% is unsatisfiable");
+
+        assert_eq!(err, ObjectDataCacheConfigError::InvalidMinFreeMemoryPercent);
     }
 
     #[test]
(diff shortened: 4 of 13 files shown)


houseme:
CI caught a real regression, and it is worth being precise about what it proved.

`Test and Lint` failed on `body_lookup_returns_hit_when_cache_contains_matching_body`: the fill returned `SkippedMemoryPressure` instead of `Inserted`. This is **the container-aware memory gate working as intended**. CI runs on self-hosted Kubernetes, so the gate now reads the pod's cgroup free memory, finds it below the 20% floor, and refuses the fill. On a developer host there is no cgroup, sysinfo falls back to host memory, and the tests pass — exactly the host/container divergence this PR exists to fix, surfacing first in the test suite.

The gate was right; the tests were the ones depending on a live memory reading.

Fixed in `51a97a81c` by making `min_free_memory_percent = 0` a deliberate opt-out rather than an invalid value. `allows_fill` now returns early before touching any snapshot, so fill admission is independent of where the suite runs, and operators gain the same escape hatch. Every test that requires a fill to succeed sets the floor to `0`.

Critically, the tests that *exercise* the gate keep it enabled through a new `memory_gated_config()`, so the ODC-05 coverage is preserved rather than short-circuited. Both directions were mutation-tested:

| mutation | result |
|---|---|
| remove the `min_free == 0` early return | `zero_min_free_percent_disables_the_gate` fails |
| point `moka_backend_skips_fill_under_memory_pressure` back at the ungated config | that test fails — proving it was not silently defanged |

```
cargo test -p rustfs-object-data-cache   52 passed; 0 failed
cargo test -p rustfs object_data_cache   28 passed; 0 failed
make pre-commit                          passed
```

houseme:
Merged latest `main` (`50fb9204e`) — not a routine rebase. There was a real cross-PR hazard worth spelling out.

#4657 landed on `main` with new tests (`moka_backend_refill_after_expiry_...`, `moka_backend_concurrent_fills_same_identity_both_insert`, ...) that require a fill to **succeed**, and they run under `enabled_config()`, which on that branch carried `min_free_memory_percent: 20`. Those tests passed CI there because #4657 does **not** contain the container-aware memory gate — that is this PR. Each PR was green in isolation, yet the merged result could have failed on CI: the gate would read the pod's cgroup free memory and refuse those fills.

`git merge` reports no textual conflict, which proves nothing about semantics. So I verified the joint state locally rather than letting CI find out:

- `enabled_config()` resolves to `min_free_memory_percent: 0` after the merge (this PR's change wins; #4657 never touched that line), so #4657's fill-dependent tests are gate-independent.
- Its two tests that build a custom config do `let mut config = enabled_config()`, inheriting the opt-out.
- `moka_backend_skips_fill_under_memory_pressure` still uses `memory_gated_config()`, so the ODC-05 coverage is not short-circuited.
- #4657's `moka_backend_identity_budget_evicts_oldest_and_caches_newest` sets `identity_keys_max = 2`, which satisfies the `>= 2` validation this PR adds. Had it used `1`, the merge would have broken it.

Joint state on top of `main` (now including #4654 and #4657):

```
cargo test -p rustfs-object-data-cache                          61 passed; 0 failed
cargo test -p rustfs object_data_cache                          30 passed; 0 failed
cargo test -p rustfs-ecstore --lib body_cache_hook_gate_tests    8 passed; 0 failed
cargo fmt --all --check                                          clean
```

The PR is now verified against the code it will actually merge into, not against a stale base.

houseme:
Second `Test and Lint` failure, same root cause, and this one is on me: my previous sweep was not exhaustive.

Fixing the first batch moved nextest's fail-fast point forward, and CI surfaced the next set — `build_get_object_body_with_cache_materializes_once_and_hits_later` and `..._uses_cached_body_without_reader_preread`. Both assert a fill lands, and both build their adapter inside `object_usecase.rs`, which my earlier grep did not cover (I had searched the `object_data_cache/` module and the engine crate only). Six construction sites there still carried the default 20% free-memory floor.

Fixed in `6d183a311`: floor set to `0` on all six, matching the sibling test modules.

Two things I did differently this time, since "the tests pass on my machine" proves nothing here — a developer host has no cgroup, so it passes either way.

**Proof the exemption actually decouples from memory.** I pinned the gate's snapshot to `total=1000, available=0` — 0% available, harsher than any CI pod — and re-ran:

```
build_get_object_body_with_cache*                6 passed
rustfs object_data_cache                        30 passed
rustfs-object-data-cache                        61 passed
```

They pass because `allows_fill` returns before touching any snapshot. That is the property that matters, and it now has a demonstration rather than an assertion.

**An exhaustive sweep, not a targeted grep.** A script walked every `ObjectDataCacheConfig` in the tree that selects a `Fill*` mode and checked each for an explicit `min_free_memory_percent`. It flagged four remaining sites; all four are false positives, verified individually:

- `config.rs:133` — the `matches!` arm inside `fill_enabled()`, not a config literal.
- `adapter.rs:409` / `:427` — two tests that assert the adapter comes out **disabled**, so they never fill.

I did not blanket-apply the exemption to those. The gate-exercising test still runs under `memory_gated_config()` with the floor at 20%.

```
make pre-commit   passed
```

PR #4659, opened 2026-07-10
fix(ecstore): roll back delete on set-level quorum failure

overtrue:
## Related Issues

Refs rustfs/backlog#864 — the **delete half**. The overwrite half was already fixed by #4107; this adds the symmetric rollback for the delete path.

## Summary of Changes

On a set-level delete quorum failure, the disks that already applied the delete were not
rolled back, so a DELETE that returns 5xx could still make the object disappear (a failed
delete becoming a real delete). This mirrors the #4107 overwrite `undo_write` model onto
the delete path.

Three-phase `prepare → quorum → undo | commit`:
- **prepare** (`stage_rollback`): before the destructive local change, back up the full
  `xl.meta` and **same-disk rename** (metadata-level, not copy) the deleted data dir into a
  per-token rollback dir (`.rustfs.sys/tmp/.rollback/{token}/`). The create-from-absent leg
  (force-del-marker on a missing object) records a `.created_from_absent` sentinel instead.
- **quorum**: unchanged (`reduce_write_quorum_errs`).
- **undo | commit**: undo/commit fan out to **all** disks (not `err.is_none()`-filtered);
  each disk **self-decides from its token manifest** — has `xl.meta` → restore it and the
  staged data dirs; has the absent sentinel → delete the just-created marker; no token dir →
  idempotent no-op. This eliminates ghost markers, batch-delete `None` ambiguity, and
  half-applied errored disks.

Scope: single-object (`delete_object_version`) and batch (`delete_objects`, per-(op,object)
unique token aligned to the versions array), disk layer (`delete_version`,
`delete_versions_internal`) and set layer. Gated by `RUSTFS_DELETE_ROLLBACK_ENABLE`
(default on; `off` restores the previous MinIO-style fail-forward). `DeleteOptions` gains 4
`Default`-compatible, RPC-serialized fields; the `undo_write` overwrite path is untouched.

Deliberate design choices (please review): RustFS already chose "roll back successful disks
on quorum failure" for overwrite (#4107); this extends the same house philosophy to delete
(MinIO is fail-forward here). Staging via same-disk rename keeps undo fsync-free and
metadata-level; undo is best-effort (warn on failure, residue left to heal), introducing no
worse failure class than today.

## Verification

- `make pre-pr` (green — full workspace incl. RPC consumers of `DeleteOptions`)
- Disk-layer + set-layer TDD tests for: undo restores object on quorum failure, commit
  cleans rollback on success, self-describing idempotent no-op on unstaged disks,
  create-from-absent marker removal, batch per-object token alignment.

## Impact

A delete that fails set quorum no longer leaves the object deleted on the successful disks;
it is restored. Successful deletes are unchanged (byte-identical when `stage_rollback` is
off). Behind `RUSTFS_DELETE_ROLLBACK_ENABLE` (default on). No on-disk format change; the
rollback dir lives under `.rustfs.sys/tmp/.rollback`, distinct from the overwrite `.bkp`.

## Additional Notes

**P1 — pending maintainer review before merge** (delete critical path). Design produced by a
multi-agent research + adversarial design round (3 NEEDS_CHANGES rounds); full design and
open questions (fault-injection granularity, orphan-cleaner age bound, detached-commit
tradeoff — defaults adopted, noted in code) are in backlog#864.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10030,7 +10030,7 @@ dependencies = [
 [[package]]
 name = "rustfs-uring"
 version = "0.1.0"
-source = "git+https://github.com/rustfs/uring?rev=39018c09cde98b380658f1a1511349a3574daddf#39018c09cde98b380658f1a1511349a3574daddf"
+source = "git+https://github.com/rustfs/uring?rev=f577ba0bfd6c3d74f9bdc54b573b8fcb45d6bc22#f577ba0bfd6c3d74f9bdc54b573b8fcb45d6bc22"
 dependencies = [
  "io-uring",
  "libc",
--- a/crates/ecstore/src/disk/disk_store.rs
+++ b/crates/ecstore/src/disk/disk_store.rs
@@ -813,6 +813,7 @@ impl LocalDiskWrapper {
                     immediate: false,
                     undo_write: false,
                     old_data_dir: None,
+                    ..Default::default()
                 },
             )
             .await?;
--- a/crates/ecstore/src/disk/local.rs
+++ b/crates/ecstore/src/disk/local.rs
@@ -19,8 +19,9 @@ use crate::disk::{
     BUCKET_META_PREFIX, CHECK_PART_FILE_CORRUPT, CHECK_PART_FILE_NOT_FOUND, CHECK_PART_SUCCESS, CHECK_PART_UNKNOWN,
     CHECK_PART_VOLUME_NOT_FOUND, CheckPartsResp, DeleteOptions, DiskAPI, DiskInfo, DiskInfoOptions, DiskLocation, DiskMetrics,
     FileInfoVersions, FileReader, FileWriter, MmapCopyStageMetrics, OldCurrentSize, RUSTFS_META_BUCKET, RUSTFS_META_TMP_BUCKET,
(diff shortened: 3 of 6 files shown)


overtrue:
Closing as **superseded**. During rebase onto latest `main` it emerged that **#4300 ("require write quorum for metadata early stop") already landed a set-level delete-quorum rollback** for this exact problem — `DeleteOptions.undo_delete: bool` + `old_data_dir`, disk helpers `write_metadata_rollback_backup` / `restore_delete_rollback` / `restore_delete_rollback_after_error` (local.rs), and set-level undo fan-out in `delete_object_version` / `delete_objects` (object.rs). So backlog#864's **delete half is fixed on main**.

This PR implements a competing full rollback (independent `.rustfs.sys/tmp/.rollback/{token}` staging, `undo_delete: Option<Uuid>`, self-describing all-disk fan-out) that collides head-on with #4300 (duplicate `undo_delete` field, two rollback mechanisms). Merging it would duplicate #4300's just-merged work, so it is closed rather than force-merged.

One genuine residual hardening from this design is worth keeping: #4300's undo loop skips errored disks (`if err.is_some() { continue }`), so a disk that stages + applies the delete and *then* errors is not rolled back. Tracked as a targeted follow-up (see backlog).

Issue #1146, opened 2025-12-14
The logs suddenly reached 57 GB in size and the Object Browser issue


codedByCan:
My first problem: object browser

<img width="871" height="340" alt="Image" src="https://github.com/user-attachments/assets/621c8bcf-fa8d-428c-96c4-23b216be7d31" />

I have a public bucket, but I cannot connect and view the files inside. I can view them when I connect using another browser or code.

My second problem: logs suddenly becoming bloated

I couldn't open and read the logs because my server's storage suddenly filled up, but I had copied the latest logs from the console.

[flexi_logger][ERRCODE::Write] writing log line failed, caused by Custom { kind: Other, error: "Send" } See https://docs.rs/flexi_logger/latest/flexi_logger/error_info/index.html#write [flexi_logger][ERRCODE::Write] writing log line failed, caused by Custom { kind: Other, error: "Send" } See https://docs.rs/flexi_logger/latest/flexi_logger/error_info/index.html#write [flexi_logger][ERRCODE::Write] writing log line failed, caused by Custom { kind: Other, error: "Send" } See https://docs.rs/flexi_logger/latest/flexi_logger/error_info/index.html#write [flexi_logger][ERRCODE::Write] writing log line failed, caused by Custom { kind: Other, error: "Send" } See https://docs.rs/flexi_logger/latest/flexi_logger/error_info/index.html#write [flexi_logger][ERRCODE::Write] writing log line failed, caused by Custom { kind: Other, error: "Send" } See https://docs.rs/flexi_logger/latest/flexi_logger/error_info/index.html#write [flexi_logger][ERRCODE::Write] writing log 

Version: v1.0.0-alpha.69

loverustfs:
What deployment method did you use?

Is the log level info or error?

codedByCan:
I'm deploying via Docker. I'm not using Compose or anything like that.

Log level: Error

Additional logs;

`ERROR timestamp=2025-12-14T13:25:32.332946186Z fields={"message":"get_object_with_fileinfo  tema/bdf486/pages/user/support.hbs err ObjectNotFound(\"tema\", \"bdf486/pages/user/support.hbs\")"} target=rustfs_ecstore::set_disk filename=crates/ecstore/src/set_disk.rs line_number=3784 threadName=rustfs-worker threadId=ThreadId(7)

[...]

Issue #1126, opened 2025-12-12
Bucket Configuration Center Optimization Plan


cxymds:
# Bucket Configuration Center Optimization Plan

## Problem Description / Background

Currently, bucket configurations (**bucket configuration, bucket events, bucket lifecycle, bucket replication**) are scattered across different modules of the system, with deep access paths and complex operational logic, leading to:

1. **Inconsistent user experience**: Configuration entry points are hidden, making it difficult for users to find quickly, or user operations are inconvenient.
2. **Function overlap**: Bucket configuration-related operations are tightly coupled with the object browser, resulting in blurred functional boundaries.
3. **Maintenance difficulties**: Related logic is distributed across multiple modules, with high code duplication and significant modification risks.

## Optimization Goals

Migrate the **bucket events, bucket lifecycle, and bucket replication** functions from the existing modules and integrate them with the basic bucket configuration to form a unified **"Bucket Configuration Center"**, achieving:

- **Unified entry point**: All bucket configuration functions are centralized in one page/module.
- **Clear logic**: Each configuration function has distinct responsibilities with no cross-dependencies.
- **Convenient operations**: Users can complete all bucket configuration operations in a unified interface.
- **Easy scalability**: Future new configuration functions can be directly integrated into the existing architecture.

Issue #1109, opened 2025-12-11
After restarting the operating system, it is not possible to log in with the default account key


GaoLe996:
OS: Rocky linux 9
RustFS: rustfs 1.0.0-alpha.72

/etc/default/rustfs:
`
RUSTFS_ACCESS_KEY=rustfsadmin
RUSTFS_SECRET_KEY=rustfsadmin
RUSTFS_VOLUMES="/data/rustfs0"
RUSTFS_ADDRESS=":9000"
RUSTFS_CONSOLE_ENABLE=true
RUST_LOG=error
RUSTFS_OBS_LOG_DIRECTORY="/var/logs/rustfs/"
`

/etc/systemd/system/rustfs.service:
`
[Unit]
Description=RustFS Object Storage Server
Documentation=https://rustfs.com/docs/
After=network-online.target
Wants=network-online.target

[Service]
Type=notify
NotifyAccess=main
User=rustfs-user
Group=rustfs-user

WorkingDirectory=/usr/local
EnvironmentFile=-/etc/default/rustfs
ExecStart=/usr/local/bin/rustfs $RUSTFS_VOLUMES

LimitNOFILE=1048576
LimitNPROC=32768
TasksMax=infinity

Restart=always
RestartSec=10s

OOMScoreAdjust=-1000
SendSIGKILL=no

TimeoutStartSec=30s
TimeoutStopSec=30s

NoNewPrivileges=true
 
ProtectHome=true
PrivateTmp=true
PrivateDevices=true
ProtectClock=true
ProtectKernelTunables=true
ProtectKernelModules=true
ProtectControlGroups=true
RestrictSUIDSGID=true
RestrictRealtime=true

# service log configuration
StandardOutput=append:/var/logs/rustfs/rustfs.log
StandardError=append:/var/logs/rustfs/rustfs-err.log

[Install]
WantedBy=multi-user.target
`

Question: 
It was possible to log in normally with the default account just after deployment, but after restarting the operating system, it now prompts that login is not possible for some reason.

<img width="1610" height="1184" alt="Image" src="https://github.com/user-attachments/assets/c33904aa-8ad1-43f2-97f0-25e930fb9f01" />

loverustfs:
Do you have more logs?

loverustfs:
### RockyLinux works fine on my end.


## OS Version

[...]

Issue #1108, opened 2025-12-11
Bug fix listObjectsV2


loverustfs:
Title: Add an API for Admin verification
Description/Fix: ListObjectsV2 does not return nested prefix when folder name is repeated
Source: https://github.com/rustfs/rustfs/issues/1100
Expected outcome: Fix this bug
OS: Linux & Windows
Email: [hello@rustfs.com](mailto:hello@rustfs.com)

PR #4675, opened 2026-07-10
test(cache): end-to-end regressions for the body-cache hook P0s

houseme:
## Summary

Two P0 data-corruption fixes shipped in [#4654](https://github.com/rustfs/rustfs/pull/4654) with only **predicate-level unit tests**. Those tests cannot catch a caller that opens a *new* shortcut returning `buffered_body` directly — which is precisely the original form of both bugs. The blocker was the hook itself: a process-wide `OnceLock` where the first registration wins, making a deterministic register/unregister harness impossible.

This PR removes the blocker and cashes in the coverage.

## The hook slot (backlog [#1126](https://github.com/rustfs/backlog/issues/1126))

`AppContext::new` builds a fresh `ObjectDataCacheAdapter` and registers it on every construction, but the `OnceLock` kept the first one forever. If AppContext is ever rebuilt (config reload, test reinit), the ecstore GET probe keeps reading adapter #1's cache while every usecase-layer fill and invalidation targets adapter #2 — a silent degradation to a 0% hook hit rate, with residual entries in cache #1 living out their TTL unreachable by invalidation.

Replaced with `RwLock<Option<Arc<dyn GetObjectBodyCacheHook>>>`. Registration swaps atomically and WARNs (via `Arc::ptr_eq`) when it replaces a *different* instance.

**Not `ArcSwap`, despite it being a workspace dep**: arc-swap's `RefCnt` is `unsafe impl<T> RefCnt for Arc<T>` — `Sized` only, since `as_ptr` returns a thin pointer — so it cannot hold `Arc<dyn Trait>` without a Sized newtype and a second `Arc`. The probe reads the slot once per full-object GET and clones one `Arc` under a read guard, negligible beside the metadata quorum fan-out that already completed before it; registration is a rare startup event.

## The actual prize: end-to-end regressions

`body_cache_hook_e2e_tests` drives the **real `get_object_reader`** — gate, probe, reader construction, served bytes — not the `full_object_plaintext_len` predicate.

- Harness: `make_local_set_disks(4, 2)`, genuine erasure-coded local disks.
- Fixture: genuinely compressed through the same `compression_reader` codec the PUT path uses (660000 → 2972 bytes), with `assert_ne!(compressed, plaintext)` so the fixture cannot silently degenerate.
- The stand-in `GetObjectBodyCacheHook` *is* the production injection point.

| test | asserts |
|---|---|
| `raw_data_movement_read_serves_stored_bytes_not_cached_plaintext` | backlog [#1108](https://github.com/rustfs/backlog/issues/1108) — decommission's raw read yields STORED compressed bytes, never the cached plaintext |
| `compressed_cache_hit_publishes_decompressed_size` | backlog [#1109](https://github.com/rustfs/backlog/issues/1109) — a compressed hit publishes the decompressed length as `object_info.size` and the stream length matches (plus a no-hook control proving real decompression) |
| `restore_read_serves_stored_bytes_not_cached_plaintext` | backlog [#1146](https://github.com/rustfs/backlog/issues/1146) — a restore read yields stored bytes |

## Mutation testing

Each test was verified to bite. The first row is the one that matters:

| mutation | failing test |
|---|---|
| **delete the hit-site `object_info.size = plaintext_len` — predicate untouched** | `compressed_cache_hit_publishes_decompressed_size` (`left: 2972, right: 660000`) |
| drop `raw_data_movement_read` from the gate | `raw_data_movement_read_serves_stored_bytes_not_cached_plaintext` |
| drop `restore_request_active` from the gate | `restore_read_serves_stored_bytes_not_cached_plaintext` |

The first mutation leaves the allow-list predicate completely correct and breaks only the hit site. A predicate-level test is blind to it; the e2e test catches it immediately. That is the gap this PR exists to close — and it is exactly the shape of the two original P0s.

## Scope note

The e2e tests use a stand-in hook rather than the real `ObjectDataCacheAdapter`, which is unavoidable: the adapter lives in the `rustfs` crate and ecstore cannot depend on it. This is not "faking it via the predicate" — the tests drive `get_object_reader` end to end through the production trait. Adapter-level fill/invalidation coverage remains the job of tests in `rustfs/src/app/object_data_cache/`.

## Verification

```
cargo test -p rustfs-ecstore body_cache   11 passed; 0 failed   (8 predicate + 3 e2e)
make pre-commit                           passed
```

Verified on top of the current `main` (merged, no conflicts), not a stale base.


houseme:
CI failure fixed, but the cause is not in this PR.

All four `Test and Lint` variants failed to compile:

```
error: call to `std::mem::drop` with a value that does not implement `Drop`.
  --> rustfs/src/startup_entrypoint.rs:41:5
   |
41 |     drop(dial9_guard);
   = note: `-D clippy::drop-non-drop` implied by `-D warnings`
```

`startup_entrypoint.rs` is not a file this branch touches. `git log` puts that line in `00536da80` (#4663, dial9 telemetry), already on `main`; this branch only inherited it by merging `main`.

**The bug is real, and it affects every PR based on current `main`.** There are two `Dial9SessionGuard` types: the one in `enabled.rs` implements `Drop` and flushes/seals the trace segment; the one in `disabled.rs` is a unit struct with no `Drop`. Default-feature builds — i.e. every CI job — compile the disabled variant, so `drop()` is a no-op and clippy rejects it under `-D warnings`.

Deleting the `drop()` call would be the wrong fix: it is load-bearing under the `dial9` feature, because `process::exit` runs no destructors. Instead the disabled guard now carries the same `Drop` contract, so the call site means one thing under both cfgs — flush when there is something to flush, no-op otherwise.

Verified by reverse mutation: removing the new `impl Drop` makes `clippy::drop_non_drop` fire again (2 hits); restoring it clears it.

```
cargo clippy -p rustfs --lib --tests -- -D warnings   0 errors, 0 warnings
cargo clippy -p rustfs-obs --all-targets -- -D warnings   clean
cargo test -p rustfs-obs dial9                        15 passed
```

Issue #1137, opened 2025-12-13
Nix Development shell and shebangs compatibility


mhkarimi1383:
**Is your feature request related to a problem? Please describe.**

I want to contribute and I'm a nix user, also having nix development shell will help contributors to work on project easier since it with prepare a nice shell will everything needed

**Describe the solution you'd like**
shebang fixes and nix develop files

**Describe alternatives you've considered**

**Additional context**
I have made these in my own fork and I will make a PR if that's OK

loverustfs:
Hey @Mic92 ,

Is this demand strong in NixOS?

Thank you very much.

Mic92:
Haven't reviewed the flake.nix, fixing shebangs makes sense not just for NixOS. It's also needed for potential FreeBSD support and on MacOS /bin/bash is super old - using env shebangs allows for using homebrew. In any case, https://github.com/rustfs/rustfs/pull/1096 should also enable `nix develop` devshells.

loverustfs:
> Haven't reviewed the flake.nix, fixing shebangs makes sense not just for NixOS. It's also needed for potential FreeBSD support and on MacOS /bin/bash is super old - using env shebangs allows for using homebrew. In any case, https://github.com/rustfs/rustfs/pull/1096 should also enable `nix develop` devshells. 

Hey @Mic92 ,

Thank you so much! #1096 has already been merged. Thank you for your contribution!

loverustfs:
Hey @mhkarimi1383 ,

NixOS is an awesome OS, and we really love its design philosophy. That said, I haven't done a deep dive into it myself yet, so I'm not fully up to speed on the best and safest way to integrate it with RustFS for our users.

If this is in your wheelhouse, we would absolutely welcome a PR. @Mic92  actually submitted our first NixOS-related PR recently. Since this isn't my area of expertise, maybe the two of you could collaborate on this? Thanks a ton!

mhkarimi1383:
I will checkout the PR and I will try to add missing things if needed

Thanks

loverustfs:
> I will checkout the PR and I will try to add missing things if needed
> 
> Thanks

Ok, Thanks.

loverustfs:
Hey @mhkarimi1383 ,

Thanks for the PR! Using #!/usr/bin/env bash is indeed a more portable approach and follows best practices for cross-platform compatibility, especially for NixOS users. I'm happy to merge this.

[...]

Issue #1135, opened 2025-12-13
SECURITY.MD contains placeholder content


WauHundeland:
The [SECURITY.md](https://github.com/rustfs/rustfs/blob/main/SECURITY.md) file currently contains placeholder content, such as 
> Use this section to tell people about which versions of your project are currently being supported with security updates.

or 

> Use this section to tell people how to report a vulnerability.
>
> Tell them where to go, how often they can expect to get an update on a reported vulnerability, what to expect if the vulnerability is accepted or declined, etc.

This should be replaced with actual content about where security vulnerabilities should be reported.

loverustfs:
Hey @WauHundeland ,

Could you submit a PR to us? If it's reasonable, we can accept the merger.

lunalna:
I have created a pull request: #1140

loverustfs:
> I have created a pull request: [#1140](https://github.com/rustfs/rustfs/pull/1140)

Hey @WauHundeland ,

Merged.
Thank you for you contribution.

loverustfs:
> I have created a pull request: [#1140](https://github.com/rustfs/rustfs/pull/1140)


@WauHundeland ,

If you have any other thoughts or suggestions, please don't hesitate to share them. We really appreciate it!

Issue #1123, opened 2025-12-11
imagePullSecrets configured in values.yaml but not rendered in Deployment/StatefulSet templates


czaloumis:
**Describe the bug**
The `imagePullSecrets` value configured in `values.yaml` is not being rendered into the Deployment or StatefulSet templates, causing pods to fail pulling images from private registries despite correct configuration.

**To Reproduce**
1. Add `imagePullSecrets` to `values.yaml`:
```
imagePullSecret: 
  - my-secret-name
```
2. Run `helm template .` or `helm install`
3. Inspect generated manifests: Deployment/StatefulSet `spec.template.spec.imagePullSecrets` missing
4. Pods that needs authentication for pulling image fail with: `ErrImagePull` / `ImagePullBackOff`

**Expected behavior**  
When `imagePullSecrets` is set in values.yaml, the generated Deployment and StatefulSet manifests should include:
```
spec:
  template:
    spec:
      imagePullSecrets:
        - name: my-registry-secret # <- This should be rendered
```  

**Environment**  
- Helm version: 
version.BuildInfo{Version:"v4.0.1", GitCommit:"12500dd401faa7629f30ba5d5bff36287f3e94d3", GitTreeState:"clean", GoVersion:"go1.25.4", KubeClientVersion:"v1.34"}
- Chart version: 1.0.3

**Additional context**  
This prevents deployments from private registries. Most Helm charts inject `{{ .Values.imagePullSecrets }}` into Deployment/StatefulSet templates. Chart templates appear to lack this rendering logic.

majinghe:
@czaloumis Thanks for you feedback, will check this issue and update later.

majinghe:
@czaloumis Already add support for pulling image from private registry. #1127 . After the pr is merged, you can pull the latest helm chart code to deploy. Thx.

czaloumis:
@majinghe, I reviewed your commit—really nice that we can now configure imagePullSecrets via the Helm chart! However, I think it would be better to also support referencing an existing imagePullSecret rather than only creating a new one.

Would you be open to me submitting an MR for this enhancement?

majinghe:
@czaloumis Good suggestion, existing imagePullSecret should be considered, please free feel to open a pr to enhance helm chart. Thx.

PR #1122, opened 2025-12-11
helm: refactor helm chart

jurim76:
## Type of Change
- Refactor

## Summary of Changes

- Update NOTES.txt
- Render `RUSTFS_VOLUMES` with a template helper, allow to override RUSTFS_VOLUMES in values file
- Add `commonLabels` support
- Add `config.rustfs.obs_environment` config parameter
- Replace selector labels with `rustfs.selectorLabels` template, replace `app` label with `app.kubernetes.io/name` label
- Fix `podSecurityContext` indent
- Refactor service.yaml
- Refactor init container command
- Add affinnity `podAntiAffinity` and `nodeAffinity` support 
- Add `nodeSelector` support
- Add tolerations support
- Allow to redefine probes
- Allow to redefine logs dir
- Fix helm tests
- Update README.md

## Checklist
- [ ] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [ ] Documentation updated (if needed)

## Impact
- Breaking change (compatibility)
- Requires doc/config/deployment update

## Additional Notes

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

PR #4678, opened 2026-07-10
perf(object-data-cache): take fill off the GET path and fix the metrics

houseme:
## Summary

The object data cache filled synchronously inside the GET response future, enforced none of its concurrency knobs, did a blocking `/proc/meminfo` read on the async fill path, and reported metrics that were wrong by construction. This PR fixes the hot path and makes the metrics mean what their names say.

## Hot path

**Fill blocked the response** (backlog [#1120](https://github.com/rustfs/backlog/issues/1120)). `build_get_object_body_with_cache` awaited the fill before building the response blob, even though the complete body was already in hand and the fill result is discarded. Worse, a non-leader awaited `waiter.wait()` — blocking *this* request until *another* request's leader finished its memory-gate check, prune scan, index insert and moka insert. A waiter gains nothing: its bytes are already materialized.

Fills now run as a detached task and the response returns immediately. `singleflight` is redesigned from a `watch`-channel waiter model to lock-free `Leader | Busy` election: a non-leader skips rather than waits, and the `Waiter`/`wait()` machinery is gone.

The inner `tokio::spawn` inside `fill_body` is **kept**. Detaching at the call site removes the latency coupling, but once the index key is registered the recheck/undo must still complete even if the fill future is aborted. Removing it would break `moka_backend_cancelled_fill_still_undoes_lost_race` and weaken the invalidation-race undo from [#4657](https://github.com/rustfs/rustfs/pull/4657).

**Concurrency knobs were decorative** (backlog [#1116](https://github.com/rustfs/backlog/issues/1116)). `fill_concurrency_per_cpu` / `fill_concurrency_max` were validated with three dedicated error variants and plumbed from env, yet had no consumer anywhere: singleflight only dedups per key, so distinct-key fills were unbounded. Added a `Semaphore` sized `min(per_cpu × available_parallelism, max)`, acquired with `try_acquire_owned` after winning leadership and before the memory gate. Saturation returns `SkippedFillConcurrency` — it does **not** queue, since queueing would reintroduce the latency coupling #1120 just removed.

**The memory gate blocked a tokio worker** (backlog [#1119](https://github.com/rustfs/backlog/issues/1119)). `allows_fill` → `refresh_if_stale` took two `std::Mutex`es and ran `System::refresh_memory()` inline, with no staleness re-check after acquiring the system lock — so at each 5s boundary every concurrent fill serialized on that mutex and each performed its own redundant blocking read. Replaced with a periodic refresher (`interval` + `spawn_blocking`) updating an atomic snapshot; `allows_fill` now only reads atomics. The `min_free_memory_percent == 0` short-circuit still runs before any snapshot read (see `51a97a81c`).

**Metrics emission under the singleflight mutex** (backlog [#1137](https://github.com/rustfs/backlog/issues/1137)). `acquire`/`remove_entry` hashed a label set and entered the recorder while holding the global `fills` lock. Now they capture `len()` under the guard, drop it, then emit.

**Materialize read was unbounded and warn-and-served** (backlog [#1112](https://github.com/rustfs/backlog/issues/1112)). Re-assessed severity first: there is **no** HTTP framing corruption, because the body is wrapped with `remaining = response_content_length`. Two real issues remained: `read_to_end` had no upper bound, and a length mismatch only `warn!`ed while the direct-memory path treats the identical inconsistency as a hard error. Now bounded by `take(cap + 1)` with a hard error on mismatch. No timeout added — GET I/O deadlines belong to a higher layer.

## Metrics that told the truth

**`requests_total` counted 2-4× per GET** (backlog [#1122](https://github.com/rustfs/backlog/issues/1122)). `plan_get` and `lookup_body` both fed one counter, so `sum(requests_total)` was uninterpretable as GET volume. Split into `plan_total{decision,reason,size_class}` and `lookup_total{result,size_class}`, each incremented exactly once per GET per layer.

**Waiters counted as successful fills** (backlog [#1123](https://github.com/rustfs/backlog/issues/1123)). A non-leader returned the leader's `Inserted`, so N concurrent GETs of one cold key recorded N `fill_total{result=inserted}` and N × size in `fill_bytes_total` for **one** insert, with the waiter's duration sample measuring wait time. Non-leaders now return `JoinedInflightFill` — no bytes, no duration; `singleflight_joins` already counts them.

**Invalidation double-counted every mutation** (backlog [#1141](https://github.com/rustfs/backlog/issues/1141)). The two-phase invalidation (before-mutation + after-success) is correct hygiene and is unchanged; only the accounting was wrong. `invalidations_total` now carries `outcome={removed|noop}`, and the no-op path skips the gauge refresh.

**`refresh_entry_count` on every lookup** (backlog [#1134](https://github.com/rustfs/backlog/issues/1134)). Two moka cross-segment aggregations plus two gauge registry writes per GET, landing the same stale approximation thousands of times between scrapes. Removed from `lookup_body` entirely; debounced to ≤1/sec on fill and invalidate.

**Dead per-entry fields** (backlog [#1135](https://github.com/rustfs/backlog/issues/1135)). `content_length`, `etag` and `inserted_at` had no readers — etag and size are already part of the moka key's `Hash`/`Eq`, so a hit has matched them by construction. Removed; the entry is a thin `Bytes` wrapper.

## Cleanup of the cross-branch seams

The metrics and hot-path work landed on separate branches with disjoint file ownership, which left four transitional shims. Two were merely dead; two actively misreported:

- The fill-accounting `match` ended in a wildcard charging `fill_bytes` and a duration to **every** non-`Inserted` outcome — so a fill rejected by the memory gate or the concurrency semaphore, having never touched the backend, still inflated `fill_bytes_total`. Now every variant is enumerated explicitly. Exhaustiveness also forces a future variant to declare its accounting instead of inheriting a wrong default.
- `InvalidationResult::Success` mapped a no-op to `outcome=removed` — the exact lie #1141 set out to fix — and its doc comment still claimed `MokaBackend` returned it, long after that backend was widened. Removed, and the app-layer test tightened from "any successful variant" to the real contract: pre-mutation reports `Removed{keys:1}`, post-delete reports `NoOp`.
- `ObjectDataCacheEntry::new` accepted `content_length`/`etag` only to keep the other branch compiling, then discarded them. Now `new(bytes)`.
- `SkippedSingleflightClosed` lost its only producer when the waiter model was replaced. Removed.

## Mutation testing

| mutation | failing test |
|---|---|
| waiter returns `Inserted` again | `moka_backend_duplicate_fill_same_key_joins_without_waiting` |
| `invalidate_object` always returns `Removed` | `moka_backend_invalidate_uncached_identity_is_noop` |
| oversize the fill semaphore | `moka_backend_rejects_fill_when_concurrency_saturated` |
| `allows_fill` refreshes the snapshot inline | `allows_fill_reads_snapshot_without_refreshing_inline` |
| revert the materialize hard error to warn-and-serve | `build_get_object_body_with_cache_materialize_rejects_length_mismatch` |

All five wave-1 race tests (`cancelled_fill_still_undoes_lost_race`, `fill_loses_race_to_invalidation`, `refill_after_expiry_...`, `concurrent_fills_same_identity_both_insert`, `skips_fill_under_memory_pressure`) survive the singleflight redesign and still pass — the cancellation-safety and invalidation-race guarantees from #4657 are intact.

## Verification

```
cargo test -p rustfs-object-data-cache               75 passed; 0 failed
cargo test -p rustfs object_data_cache               30 passed; 0 failed
cargo test -p rustfs build_get_object_body_with_cache 7 passed; 0 failed
cargo clippy -p rustfs-object-data-cache --all-targets  clean
make pre-commit                                      passed
```

Every test config that enables a `Fill*` mode and expects a fill to succeed sets `min_free_memory_percent: 0`; the gate-exercising test keeps it at 20% via `memory_gated_config()`. Verified on top of the current `main` (merged, no conflicts).


houseme:
CI failure fixed, but the cause is not in this PR.

All four `Test and Lint` variants failed to compile:

```
error: call to `std::mem::drop` with a value that does not implement `Drop`.
  --> rustfs/src/startup_entrypoint.rs:41:5
   |
41 |     drop(dial9_guard);
   = note: `-D clippy::drop-non-drop` implied by `-D warnings`
```

`startup_entrypoint.rs` is not a file this branch touches. `git log` puts that line in `00536da80` (#4663, dial9 telemetry), already on `main`; this branch only inherited it by merging `main`.

**The bug is real, and it affects every PR based on current `main`.** There are two `Dial9SessionGuard` types: the one in `enabled.rs` implements `Drop` and flushes/seals the trace segment; the one in `disabled.rs` is a unit struct with no `Drop`. Default-feature builds — i.e. every CI job — compile the disabled variant, so `drop()` is a no-op and clippy rejects it under `-D warnings`.

Deleting the `drop()` call would be the wrong fix: it is load-bearing under the `dial9` feature, because `process::exit` runs no destructors. Instead the disabled guard now carries the same `Drop` contract, so the call site means one thing under both cfgs — flush when there is something to flush, no-op otherwise.

Verified by reverse mutation: removing the new `impl Drop` makes `clippy::drop_non_drop` fire again (2 hits); restoring it clears it.

```
cargo clippy -p rustfs --lib --tests -- -D warnings   0 errors, 0 warnings
cargo clippy -p rustfs-obs --all-targets -- -D warnings   clean
cargo test -p rustfs-obs dial9                        15 passed
```

PR #1133, opened 2025-12-13
list object version Interface returns storage_class

GatewayJ:
## Type of Change
- [ ] New Feature
- [ X] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Checklist
- [ ] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [ ] Passed `make pre-commit`
- [ ] Added/updated necessary tests
- [ ] Documentation updated (if needed)
- [ ] CI/CD passed (if applicable)

## Impact
- [ ] Breaking change (compatibility)
- [ ] Requires doc/config/deployment update
- [ ] Other impact:

## Additional Notes

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

Issue #1132, opened 2025-12-13
When adding a lifecycle via the UI, the error message "NoSuchLifecycleConfiguration" appears, preventing the addition from succeeding.


sakullla:
Describe the bug: When adding a lifecycle rule via the UI, the error message "NoSuchLifecycleConfiguration" appears, preventing successful addition.

To Reproduce:

Go to 'Lifecycle'

Click on 'Add Lifecycle Rule'

Scroll down to 'Save'
The error message "NoSuchLifecycleConfiguration" appears.

Expected behavior: Lifecycle rules can be added normally.

Screenshots

**Additional context**
This issue may be related to #1087.

loverustfs:
Hey @sakullla ,

Version 1087 has been fixed; please wait for the next release.

Additionally, there may still be some bugs in the lifecycle. If you encounter any bugs during testing, please submit logs and details to contact us.

https://github.com/rustfs/rustfs/pull/1087

loverustfs:
I'll close this issue.
Plz follow https://github.com/rustfs/rustfs/pull/1087

PR #4694, opened 2026-07-10
feat(object-data-cache): close write-side invalidation gaps and add an admin surface

houseme:
## Summary

The object data cache exposed exactly one invalidation primitive — `invalidate_object(identity)` keyed on (bucket, object) — and the starshard index had no scan API. Several delete paths therefore left dead bodies resident until TTL, force-prefix and bucket deletes invalidated nothing meaningful, and there was no way to inspect or flush the cache short of restarting the node.

**Correctness framing, up front:** the key is (bucket, object, version, etag, size, variant) and lookups run after a fresh metadata quorum, so a stale entry cannot be *served* after its object is gone — the GET fails at metadata resolution first. These are **hygiene and capacity** problems, not stale-serving. The key is not weakened anywhere.

## Write/delete-side invalidation hook (backlog [#1131](https://github.com/rustfs/backlog/issues/1131))

ecstore had a GET-side `body_cache_hook` but no write-side counterpart, so every scanner/ILM mutation bypassed invalidation. After a sweep expires 10k recently-read objects, up to `max_bytes` of dead bodies stay resident for a full TTL, evicting live hot entries.

Added `ObjectMutationHook` (`crates/ecstore/src/object_api/object_mutation_hook.rs`), mirroring `body_cache_hook.rs`: `RwLock<Option<Arc<dyn …>>>`, WARN-on-swap re-registration, `#[cfg(test)] clear_…`, no `OnceLock`. Registered next to the GET hook in `global.rs`; the app impl makes one `invalidate_object(…, AfterLifecycleExpiry)` call. Wired into `apply_expiry_on_non_transitioned_objects`, both `delete_object` sites in `expire_transitioned_object` (including the restored-copy branch), and each successfully deleted object in `delete_object_versions`.

One correctness detail: the identity uses the decoded `oi.name`/`object_name` — the form GET keys on — not the `encode_dir_object` form passed to `delete_object`. A mismatch there would make invalidation silently miss, and no test would catch it.

## Prefix and bucket invalidation (backlog [#1132](https://github.com/rustfs/backlog/issues/1132), [#1133](https://github.com/rustfs/backlog/issues/1133))

`store.delete_object(delete_prefix=true)` removes every object under a prefix but returns a single `ObjectInfo`, no name list — confirmed — so the previous code invalidated only the literal prefix string. `DeleteBucket` referenced the cache not at all, and no bucket-scope primitive existed.

Added prefix and bucket primitives; `execute_delete_object`'s force branch and `execute_delete_bucket` (which covers both force and non-force via the shared `delete_bucket` call) now use them.

## Admin surface and full-clear (backlog [#1143](https://github.com/rustfs/backlog/issues/1143))

No admin endpoint touched the cache and the engine had no `clear()`, so the only remediation for a stale or poisoned entry was a node restart. Added `ObjectDataCache::clear()` (moka `invalidate_all` + index reset + gauge refresh) and two handlers:

- `GET  {ADMIN_PREFIX}/v3/object-data-cache/stats` — `ServerInfoAdminAction`; returns the stats snapshot + mode. Gives `stats()` its first production consumer.
- `POST {ADMIN_PREFIX}/v3/object-data-cache/flush` — `ConfigUpdateAdminAction`; `bucket`+`object` → `invalidate_object(Manual)`, `bucket` only → `invalidate_bucket(Manual)`, neither → `clear(Manual)`. The distinct action (mutating, not read) is deliberate.

## Scan API and why it is safe

One primitive: `StarshardIdentityIndex::remove_matching<F>(predicate) -> Vec<Key>`. It snapshots identities, filters, then removes each match and returns their keys for eviction; `prefix`/`bucket`/`clear` all use it (`clear` = `|_| true`). Documented as a **full shard scan restricted to the rare prefix-delete / bucket-delete / admin paths — never on the GET or fill hot path**. The snapshot/remove window is documented too: an identity inserted after the snapshot is missed (its body waits out its TTL — same hygiene slack, no stale-serving), and keys are read at `remove` time so no tracked body is stranded.

## Mutation testing

| mutation | failing test |
|---|---|
| prefix predicate `starts_with(prefix)` → `starts_with("")` | `moka_backend_invalidate_prefix_removes_only_matching_prefix` (`left: Removed{keys:3}`, `right: 2`) |
| `remove_matching` filter forced true | `remove_matching_returns_only_predicate_matches` |
| hook body emptied | `mutation_hook_invalidates_cached_body` (asserts the body is `Miss` after mutation) |

The three P0 `body_cache_hook_e2e_tests` are untouched and still pass.

## Notes for review

- **Architecture guard**: the finding suggested wiring the admin handlers through `AppContext::object_data_cache()`, but `check_architecture_migration_rules.sh` forbids any `crate::app::context::` reference under `rustfs/src/admin`. Routed through `admin/runtime_sources.rs` instead — same effect, guard-compliant (verified: `Architecture migration rules passed`).
- **Cross-branch edits (additive, no conflict)**: `adapter.rs` gains prefix/bucket/clear/stats/mode passthroughs (the adapter's `cache` field is private, so the new primitives cannot be wired without them); `config.rs` gains `ObjectDataCacheMode::as_str` for admin status reporting.

## Merge note

Depends on nothing, but interacts with `fix/odc-get-path` (PR #4693): once that lands, this branch's `mutation_hook.rs` test literal needs one added line, `mod_time_unix_nanos: 0,`. Text merge is clean and `cargo check` passes without it; only `cargo test` compiles that literal. Verified jointly in an integration worktree (90 + 13 + 38 + 9 passed). Land #4693 first, then rebase/merge this.

## Verification

```
cargo test -p rustfs-object-data-cache               86 passed
cargo test -p rustfs object_data_cache              33 passed
cargo test -p rustfs-ecstore body_cache             11 passed  (+ object_mutation_hook 2)
cargo clippy … --all-targets -- -D warnings          clean
cargo doc -p rustfs-object-data-cache --no-deps      0 warnings
make pre-commit                                      passed
```

Verified on top of the current `main` (merged, no conflicts).

### DIFF
--- a/crates/ecstore/src/api/mod.rs
+++ b/crates/ecstore/src/api/mod.rs
@@ -351,8 +351,8 @@ pub mod notification {
 
 pub mod object {
     pub use crate::object_api::{
-        BLOCK_SIZE_V2, ERASURE_ALGORITHM, GetObjectBodyCacheHook, GetObjectReader, ObjectInfo, ObjectOptions, PutObjReader,
-        RangedDecompressReader, StreamConsumer, register_get_object_body_cache_hook,
+        BLOCK_SIZE_V2, ERASURE_ALGORITHM, GetObjectBodyCacheHook, GetObjectReader, ObjectInfo, ObjectMutationHook, ObjectOptions,
+        PutObjReader, RangedDecompressReader, StreamConsumer, register_get_object_body_cache_hook, register_object_mutation_hook,
     };
 }
 
--- a/crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs
+++ b/crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs
@@ -2261,6 +2261,9 @@ pub async fn expire_transitioned_object(
         opts.transition.expire_restored = true;
         return match api.delete_object(&oi.bucket, &oi.name, opts).await {
             Ok(dobj) => {
+                // Drop any cached restored-copy body so it does not sit resident
+                // until TTL after the copy is expired (ODC-26).
+                crate::object_api::notify_object_mutation(&oi.bucket, &oi.name).await;
                 //audit_log_lifecycle(*oi, ILMExpiry, tags, traceFn);
                 Ok(dobj)
             }
@@ -2292,6 +2295,10 @@ pub async fn expire_transitioned_object(
     };
 
     schedule_lifecycle_replication_delete_if_needed(oi, &dobj).await;
+
+    // The transitioned version is gone; evict any cached body for this object
+    // so it does not linger until TTL (ODC-26).
+    crate::object_api::notify_object_mutation(&oi.bucket, &oi.name).await;
 
     //audit_log_lifecycle(oi, ILMExpiry, tags);
 
@@ -2802,6 +2809,13 @@ pub async fn apply_expiry_on_non_transitioned_objects(
         }
     };
     schedule_lifecycle_replication_delete_if_needed(oi, &dobj).await;
+
+    // The object (or all its versions, for delete_all) was expired; evict any
+    // cached body so dead bytes do not sit resident until TTL (ODC-26). The
+    // cache identity is the decoded object name used by GET, not the
+    // encode_dir_object form passed to delete_object.
+    crate::object_api::notify_object_mutation(&oi.bucket, &oi.name).await;
+
     //debug!("dobj: {:?}", dobj);
     if dobj.name.is_empty() {
         dobj = oi.clone();
--- a/crates/ecstore/src/client/object_handlers_common.rs
+++ b/crates/ecstore/src/client/object_handlers_common.rs
@@ -103,6 +103,11 @@ pub async fn delete_object_versions(api: &Arc<ECStore>, bucket: &str, to_del: &[
             if errors.get(i).and_then(|err| err.as_ref()).is_some() {
                 continue;
             }
+            // Evict any cached body for the successfully deleted noncurrent
+            // version so it does not sit resident until TTL (ODC-26).
+            if let Some(target) = to_del.get(i) {
+                crate::object_api::notify_object_mutation(bucket, &target.object_name).await;
+            }
             let Some(replication_state) = replication_candidates.get(i).and_then(|c| c.clone()) else {
                 continue;
             };
--- a/crates/ecstore/src/object_api/body_cache_hook.rs
+++ b/crates/ecstore/src/object_api/body_cache_hook.rs
@@ -21,8 +21,9 @@
 //! (after the reader is built) means a hit no longer saves any disk I/O.
 
 use crate::object_api::ObjectInfo;
+use crate::object_api::hook_slot::HookSlot;
 use bytes::Bytes;
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
 
 /// Serves full-object GET bodies from a cache keyed by object identity.
 ///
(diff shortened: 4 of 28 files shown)


houseme:
## Follow-up: extracted the shared `HookSlot` (post-review de-duplication)

A three-lens architecture review of the merged object-data-cache work (abstraction duplication / hot-path performance / incremental-compromise residue) flagged exactly one thing worth acting on in this PR: this PR added `object_mutation_hook.rs` by mirroring `body_cache_hook.rs`, leaving two process-global registration slots whose `register`/`get`/`clear` bodies were line-for-line identical except the trait type and the WARN string.

Hoisted that into a generic `HookSlot<T: ?Sized>` (`crates/ecstore/src/object_api/hook_slot.rs`). Each hook module keeps its `static HOOK: HookSlot<dyn XxxHook>` and its thin, unchanged public wrappers, so the crate's public surface and every call site are untouched — internal consolidation, not a contract change.

Bonus: the load-bearing #1126 guarantee (newest registration wins, so a rebuilt `AppContext` is never stranded on a first-wins slot) previously had no direct test — the hook tests only covered register-then-notify. `HookSlot` now has its own unit tests including `re_registration_swaps_to_the_latest_instance`; mutation-testing confirms a first-wins regression fails exactly that test, and a no-op register fails three.

Net: two hooks each shed ~10 lines of duplicated slot logic; the swap-vs-warn logic lives once with 5 direct tests. `body_cache_hook` (11, incl. the P0 e2e), `object_mutation_hook` (2), and the app-layer mutation-hook tests (33) all pass unchanged; `clippy -D warnings` and `make pre-commit` clean.

### What the review deliberately did NOT change

The same review confirmed the rest of the tree is not patch-over-patch and rejected several tempting refactors, so for the record they were considered and left alone:
- **The inner `tokio::spawn` in `fill_body`** (double-spawn with the ODC-15 detach): its cost is on a detached background task, invisible to GET latency; removing it would trade `fill_body`'s self-contained cancellation-safety for a global "callers must detach" convention. KEEP.
- **`fill_ceiling_bytes` on the engine struct rather than `ObjectDataCacheConfig`**: it is a *derived* value (min of max_entry_bytes / seek threshold / hard cap), not a user knob, so it does not belong in config regardless of the ownership history. KEEP.
- **The invalidation.rs named wrappers, the `key.rs` new/with_mod_time split, `GetObjectBodySource`**: idiomatic or self-documenting, not duplication. KEEP.

Two minor residues the review surfaced are left for a small post-merge cleanup, out of scope here: the memory refresher spawns even when the gate is disabled (`min_free_memory_percent == 0`), and the `ObjectDataCacheGetRequest` test literals hand-list every field (the seam that caused this PR's earlier missing-field compile error) and would benefit from a test builder.

houseme:
CI surfaced a real gap in this PR (not the HookSlot follow-up): the two new admin routes — `GET /object-data-cache/stats` and `POST /object-data-cache/flush` — were registered as handlers but never listed in the route-policy inventory or the route matrix, so `route_policy_inventory_covers_registered_routes` and `test_admin_route_matrix_matches_registered_routes` failed.

Confirmed it is unrelated to HookSlot: it reproduces locally on default features (the CI failure happened to surface first in the rio-v2 lane by fail-fast ordering). It was latent in this PR from the start — earlier CI rounds timed out before reaching that lane; the HookSlot push triggered a full re-run that caught it.

Fixed in `e45f910cc`: `stats` listed as `ServerInfoAdminAction`/Sensitive, `flush` as `ConfigUpdateAdminAction`/High — matching the actions the handlers already enforce. The MinIO-alias matrix test is unaffected (native endpoints, no MinIO equivalent).

Root cause of the local/CI split: these coverage tests are not in `make pre-commit` (fmt + arch + quick-check only), so they pass local pre-commit and only run in the CI Test-and-Lint lane — the same class of gap as clippy not being in pre-commit.

```
route_policy_inventory_covers_registered_routes   pass
test_admin_route_matrix_matches_registered_routes pass (+ minio-alias variant)
admin::route* (83) + object_data_cache handler (2)  pass
clippy -D warnings                                 clean
```

PR #4720, opened 2026-07-11
fix(docker): use non-default credentials in cluster compose/scripts

houseme:
## Related Issues

N/A

## Summary of Changes

`entrypoint.sh` rejects the well-known default root credential `rustfsadmin` (`DEFAULT_ROOT_CREDENTIAL`) for both the access and secret keys. The local cluster stacks, however, still default to `rustfsadmin`, so `docker compose ... up` fails immediately: every node exits 1 in the entrypoint credential check, before storage/topology initialization ever runs. This makes the multi-node cluster stacks unusable out of the box (observed while validating a 4-node cluster locally: all four containers `Exited (1)` within seconds with `ERROR: RUSTFS_ACCESS_KEY must not use the default rustfsadmin credential.`).

This changes the local cluster stacks to default to non-default, distinct local test credentials — access `rustfsadmin-local`, secret `rustfssecret-local` — while keeping the `${VAR:-...}` override so operators can still inject their own. `docker-compose.cluster.yaml` previously hard-coded `rustfsadmin` with no override; it now uses the same `${VAR:-...}` form. Only local test/dev defaults change; no production path is affected (production images require credentials to be set explicitly).

Files (scoped to the multi-node cluster startup path):

- `.docker/compose/docker-compose.cluster.local-build.yml`
- `.docker/compose/docker-compose.cluster.yaml`
- `scripts/run_four_node_cluster_failover_bench.sh`

## Verification

Rebuilt the local image from source (`--no-cache`) and brought up the 4-node cluster with the new defaults (no manually injected credentials):

```bash
docker rmi rustfs/rustfs:local-4node
docker compose -p rustfs-four-node-test -f .docker/compose/docker-compose.cluster.local-build.yml build --no-cache
docker compose -p rustfs-four-node-test -f .docker/compose/docker-compose.cluster.local-build.yml up -d
# all four containers stay running; /health/ready == 200 on ports 9000-9003
```

Before this change the same `up` produced four `Exited (1)` containers at the credential check.

## Impact

- Deployment/testing: local cluster compose stacks and the failover-bench script now start out of the box; no need to manually export `RUSTFS_ACCESS_KEY`/`RUSTFS_SECRET_KEY`. Existing overrides via those env vars keep working.
- Security: unchanged posture — the new defaults are still non-production local test values, and the entrypoint guard against the `rustfsadmin` default remains in force. Production still requires explicit credentials.
- No code, on-disk, on-wire, or API change.

## Additional Notes

Other stacks/scripts still default to `rustfsadmin` and share the same failure mode: root `docker-compose.yml`, `.docker/test/site-replication/`, `.docker/test/compat/`, and several scripts under `scripts/`. They are intentionally out of scope for this narrow fix and can be aligned in a follow-up.


houseme:

PR #4736, opened 2026-07-11
perf(ecstore): snapshot per-IO env config on local disk read/write paths

houseme:
## What

The local disk IO paths in `crates/ecstore/src/disk/local.rs` re-read process environment variables on **every positioned read and write**. Each `rustfs_utils::get_env_*` call goes through `std::env::var` — a global env lock plus a `String` allocation (and a `to_ascii_lowercase` for bools). At hundreds of thousands of IOPS that is pure hot-path tax for values that are fixed at process launch.

This snapshots each env-sourced switch once and serves it from memory. Addresses rustfs/backlog#1182.

## How

A small `cached_read_env!` macro:
- **Production (`cfg(not(test))`)**: the env parse runs once via `LazyLock`; the hot path is a single cached load.
- **Tests (`cfg(test)`)**: the accessor reads the environment on every call, so `temp_env` overrides stay observable and the existing unit tests keep their isolation (no rewrite needed).
- The parse closure — including the invalid-value fallback contract from backlog#1130 — is shared, so it is identical on both paths.

## Scope

Snapshotted accessors (all previously per-call env reads):

| Path | Accessors |
|------|-----------|
| Read | `is_direct_io_read_enabled`, `get_direct_io_read_threshold`, `mmap_populate_enabled` (for `should_populate_mmap_read`), `local_read_copy_method`, `file_cache_reclaim_read_enabled` + `file_cache_reclaim_threshold` (for `should_reclaim_file_cache_after_read`) |
| Write | `is_direct_io_write_enabled`, `file_cache_reclaim_write_enabled` (for `should_reclaim_file_cache_after_write`, reusing the shared threshold snapshot) |
| Bitrot verify (runs per shard read) | `bitrot_size_mismatch_retry_count`, `bitrot_size_mismatch_retry_delay` |

**Deliberately left as-is:**
- io_uring construction switches (`get_io_uring_shards`, `is_io_uring_read_enabled`, `is_io_uring_fd_cache_enabled`) — read once per disk at setup, not per-IO.
- Durability mode — already cached behind a `OnceLock` (with a test-only override).
- Test-only env reads.

## Behavior / safety

- **No behavior change.** The environment is fixed at process launch (`set_var` is `unsafe` since edition 2024 and is not used on these paths), so the first snapshot equals every later read. The `cfg(test)` path preserves the exact old per-call semantics for tests.
- **#1130 preserved**: the invalid-value fallback runs in the shared parse closure on both paths.

## Verification

- `cargo check -p rustfs-ecstore` (production / `not(test)` → `LazyLock` path): clean.
- `cargo test -p rustfs-ecstore --lib disk::local::test` (`cfg(test)` → per-call env path): **136 passed**, including the `respects_env` / direct-io / bitrot-retry toggle tests.
- `cargo clippy -p rustfs-ecstore --lib`: no new warnings.
- Verified via a full `get_env_*` sweep that every remaining raw env read in the file is either one-time construction, the already-`OnceLock`-cached durability mode, or test code.

### DIFF
--- a/crates/ecstore/src/disk/local.rs
+++ b/crates/ecstore/src/disk/local.rs
@@ -442,14 +442,46 @@ enum LocalReadCopyMethod {
     DirectReadCopy,
 }
 
-/// Check if O_DIRECT reads are enabled.
-fn is_direct_io_read_enabled() -> bool {
-    rustfs_utils::get_env_bool(ENV_RUSTFS_OBJECT_DIRECT_IO_READ_ENABLE, DEFAULT_RUSTFS_OBJECT_DIRECT_IO_READ_ENABLE)
+/// Snapshot an env-sourced read-path switch once, then serve it from memory.
+///
+/// The value comes from the process environment, which is fixed at launch, so a
+/// per-read `std::env::var` (a global lock plus a `String` allocation, and for
+/// bools a `to_ascii_lowercase`) is pure hot-path tax at the IOPS these reads run
+/// at. In production the closure runs once via `LazyLock`; under `cfg(test)` it
+/// runs every call so `temp_env` overrides stay observable and tests keep their
+/// isolation. Invalid-value fallback (backlog#1130) lives in the closure, so it
+/// is identical on both paths.
+macro_rules! cached_read_env {
+    ($(#[$meta:meta])* fn $name:ident() -> $ty:ty = $read:expr;) => {
+        $(#[$meta])*
+        #[inline]
+        fn $name() -> $ty {
+            fn read_env() -> $ty {
+                $read
+            }
+            #[cfg(test)]
+            {
+                read_env()
+            }
+            #[cfg(not(test))]
+            {
+                static VALUE: std::sync::LazyLock<$ty> = std::sync::LazyLock::new(read_env as fn() -> $ty);
+                *VALUE
+            }
+        }
+    };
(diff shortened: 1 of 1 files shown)


houseme:
CI status: every check is green except **S3 Lifecycle Behavior Tests**, which is a pre-existing environment flaky, not a regression from this PR.

Evidence:
- The same three tests fail identically here and on unrelated PRs — `test_lifecycle_expiration` / `test_lifecyclev2_expiration` (`assert 4 == 2`) and `test_lifecycle_multipart_expiration` (`KeyError: 'Uploads'`). They are timing-sensitive lifecycle-expiration tests.
- #4732 (io_uring backend tidy — touches nothing lifecycle- or config-related) and #4733 both showed the exact same failure and were merged.
- The job's own regression classifier reports `0 regression(s), 0 unclassified failure(s)`, i.e. these are known baseline failures, not attributed to this change.
- This PR only snapshots env-sourced config that is fixed at process launch; the snapshot equals the first (correct) read, so it cannot affect lifecycle-expiration timing.

Passing here: Test and Lint (all matrices), End-to-End Tests, ILM Integration (serial), S3 Implemented Tests, io_uring Integration (real), Build, Quick Checks.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

PR #4821, opened 2026-07-15
test(e2e): add S3 event-notification webhook delivery regression net

overtrue:
## Related Issues

Part of rustfs/backlog#1154 (peri-1: event-notification end-to-end e2e). No code-fix issue in this repo.

## Summary of Changes

Adds the first end-to-end regression net for the S3 event-notification pipeline. Before this, no test verified "configure a notification target -> mutate an object -> the event is actually delivered": the target-plugin suites stop at broker integration (and are all `#[ignore]`), and the object-lambda suite exercises the webhook target only as a transform function, never as an S3 event sink (backlog#1154 protocol-surfaces gap G4).

New suite `crates/e2e_test/src/notification_webhook_test.rs` drives a real `rustfs` binary and an in-test HTTP receiver:

- `test_webhook_event_delivery_and_filtering` — on a versioned bucket, binds a webhook target via `PutBucketNotificationConfiguration` (ObjectCreated:*/ObjectRemoved:* with a `uploads/` prefix + `.dat` suffix filter) and asserts:
  - PUT delivers `s3:ObjectCreated:Put` with the correct bucket, key, versionId (equal to the PUT response version) and eTag.
  - Multipart complete delivers `s3:ObjectCreated:CompleteMultipartUpload` with the correct key and eTag.
  - Objects with a non-matching prefix or suffix produce no event, while a matching sentinel does (rule-engine filter gate).
  - DELETE delivers `s3:ObjectRemoved:*` carrying the delete-marker version.
- `test_webhook_redelivers_event_after_target_recovers` — an event PUT while the target endpoint is unreachable (connection refused) is persisted to the on-disk queue store and redelivered by the replay worker once the endpoint comes back, proving store-and-forward durability (no silent event loss).

Wiring: both tests join the `profile.e2e-smoke` default-filter in `.config/nextest.toml` (backlog#1149 ci-4), the single mechanism for e2e tests in PR CI — no new e2e job is added. Each test spawns its own single-node server and receiver on random ports with an isolated temp dir, so the subset stays parallel-safe.

Deferred (tracked on backlog#1154, not in this PR): the real-queue broker slice (mysql/postgres via a service container) is a nightly/ci-7 concern; the consolidated nightly e2e workflow does not exist yet, so this PR delivers only the webhook main chain that fits the fast PR gate. Remaining brokers (kafka/mqtt/nats/pulsar/redis/amqp) stay registered as a later batch.

## Verification

- `cargo fmt --all --check`
- `cargo check --tests -p e2e_test`
- `cargo clippy --tests -p e2e_test` (zero warnings)
- Architecture guard scripts (`check_layer_dependencies`, `check_logging_guardrails`, etc.) all pass.
- Runtime execution of the two new tests happens in this PR's own `e2e-tests` job via the extended `e2e-smoke` profile filter (they are part of the changed filter, so a green run here is the execution proof).

## Impact

Test-only. No production code, on-disk format, or API surface changes. Adds two fast cases to the PR e2e-smoke lane.

## Additional Notes

Adversarial validation (test-only / mechanical tier): the waits filter on both object key and event-type prefix, so a duplicate (at-least-once) or opposite-type event for the same key cannot satisfy the wrong assertion; the filter negative is proven by a matching sentinel arriving while the non-matching keys never do; the redelivery case uses connection-refused (a retryable `NotConnected`) rather than an HTTP error status, which the replay worker would classify as a permanent failure.


overtrue:
Heads-up: `notification_webhook_test::test_webhook_redelivers_event_after_target_recovers` (added here) is failing deterministically in the `End-to-End Tests` job since this PR landed on main — it fails on main itself and on unrelated PRs, 4 consecutive executions:

- main run [29390384824](https://github.com/rustfs/rustfs/actions/runs/29390384824) (head `52ef30f16`, this PR's merge commit) — End-to-End Tests failed
- main run [29394493304](https://github.com/rustfs/rustfs/actions/runs/29394493304) (head `23fa008ed`) — End-to-End Tests failed
- PR #4840 (crates/policy only) attempts 1+2 — same single test FAILs (~12.8s in), e.g. [job 87290832055](https://github.com/rustfs/rustfs/actions/runs/29395216710/job/87290832055)
- PR #4849 (crates/filemeta only) — same, [job 87289941728](https://github.com/rustfs/rustfs/actions/runs/29395216710/job/87289941728)

The sibling webhook tests in the same module PASS; only the `redelivers_event_after_target_recovers` case fails, and it fails at a consistent ~12.8-12.9s, which smells like a redelivery-window/timing assumption that doesn't hold on the shared `sm-standard-2` e2e runner rather than load flake (it failed twice back-to-back on a rerun of the same runner class).

Could you take a look — either make the redelivery wait robust or quarantine it out of the PR-gating smoke filter (`.config/nextest.toml` e2e-smoke) with an OPEN issue link per the ci-10/backlog#937 flake policy? It's currently failing every PR's e2e lane. Happy to help verify a fix from the security-domain side; nothing in my queue touches notify.

PR #4833, opened 2026-07-15
fix(table-catalog): accept inherited snapshot manifests

### LINKED ISSUE #4829: bug(table-catalog): accept inherited manifests on repeated append
## Summary

RustFS rejects every Spark append after the first committed snapshot when the new Iceberg manifest list reuses manifests from the parent snapshot.

## Reproduction

- RustFS: `d73c8a783adee1dfdd02c8acc65fbb6ceb9999cd`
- Spark: `3.5.4`
- Apache Iceberg: `1.7.1`
- Catalog: RustFS Iceberg REST Catalog with object-backed storage

1. Create an Iceberg v2 table through Spark.
2. Insert one row and commit the first snapshot.
3. Run a second `INSERT INTO` against the same table.

The second append fails with:

```text
org.apache.iceberg.exceptions.BadRequestException: Malformed request:
manifest changed entries must belong to the committed snapshot
```

The same failure occurs for two barrier-synchronized Spark writers. Both writers upload their data and reach catalog commit, then both receive the same `400 BadRequest` before CAS conflict handling can run.

## Expected behavior

A subsequent append should accept manifests inherited from the parent snapshot. Two concurrent append writers should reach optimistic concurrency handling; the losing writer should receive a retryable commit conflict and retry without corrupting the table pointer.

## Actual behavior and safety state

- A sequential second append fails deterministically.
- Both concurrent writers fail before CAS.
- The table remains readable with the seed row.
- `metadata_location`, `version_token`, and `generation` remain unchanged.
- Catalog diagnostics remain healthy with no staged or finalization gaps.

## Root cause

`load_snapshot_file_changes` treats every manifest entry with status `ADDED` or `DELETED` as a change made by the candidate snapshot. It rejects entries unless both their `snapshot_id` and data `sequence_number` equal the candidate snapshot identity.

Iceberg manifest lists can include manifests added by older snapshots. Their entries retain the snapshot ID and sequence numbers assigned when the files were added or deleted. In addition, a file's data sequence number is not required to equal the sequence number of the snapshot currently being committed.

The conflict validator should use manifest-list `added_snapshot_id` and entry `snapshot_id` to identify changes made by the candidate snapshot, while preserving inherited manifests and file sequence numbers.

## Regression coverage needed

- A through-catalog second append whose manifest list contains one inherited manifest and one newly added manifest.
- A negative test that still rejects malformed changed entries in a manifest added by the candidate snapshot.
- Verification that a failed validation does not advance metadata location, token, or generation.
- Live Spark double-writer retry verification after the fix.

marshawcoco:
## Related Issues

Closes #4829

## Summary of Changes

- Preserve manifest-list provenance while loading current and candidate Iceberg snapshots.
- Treat manifests inherited unchanged from the current snapshot as validated references instead of candidate-snapshot file changes.
- Keep fail-closed validation for changed entries, new-manifest snapshot ownership, manifest sequence numbers, and inherited manifest identity.
- Add through-catalog positive and negative regression tests, including pointer, token, and generation stability after rejected commits.

## Verification

- `make pre-commit`
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs --lib row_level_conflict_ -- --nocapture`
- `cargo test -p rustfs --lib table_catalog -- --nocapture`
- Spark 3.5.4 with Iceberg 1.7.1 sequential append and barrier-synchronized two-writer live verification
- PyIceberg 0.11.1 cross-client snapshot read verification

## Impact

Spark and other Iceberg clients can append after an existing snapshot reuses parent manifests. Concurrent writers now reach normal catalog CAS conflict handling, allowing the losing client to reload and retry. Malformed manifest provenance remains rejected without advancing the table pointer, version token, or generation. No API or configuration changes are introduced.

## Additional Notes

The live two-writer run observed one expected HTTP 409 version-token conflict, followed by a successful Iceberg client retry. The catalog advanced exactly twice, both writers' rows were readable, and recovery diagnostics remained healthy with no staged or finalization gaps.

The local Spark vectorized reader encountered a JVM-native failure unrelated to the catalog response. The same snapshot was verified with the Spark scalar Parquet reader and PyIceberg, both returning 200002 rows with exact per-writer counts.

### DIFF
--- a/rustfs/src/admin/handlers/table_catalog.rs
+++ b/rustfs/src/admin/handlers/table_catalog.rs
@@ -2794,13 +2794,26 @@ fn snapshot_has_manifest_references(snapshot: &serde_json::Value) -> bool {
 
 #[derive(Default)]
 struct SnapshotLiveFiles {
-    data_files: BTreeSet<String>,
-    delete_files: BTreeSet<String>,
+    data_files: BTreeMap<String, SnapshotFileIdentity>,
+    delete_files: BTreeMap<String, SnapshotFileIdentity>,
+    manifest_files: BTreeMap<String, SnapshotManifestIdentity>,
 }
 
 impl SnapshotLiveFiles {
     fn contains(&self, location: &str) -> bool {
-        self.data_files.contains(location) || self.delete_files.contains(location)
+        self.data_files.contains_key(location) || self.delete_files.contains_key(location)
+    }
+
+    fn identity(
+        &self,
+        location: &str,
+        object_kind: &crate::table_catalog::TableMetadataMaintenanceObjectKind,
+    ) -> Option<&SnapshotFileIdentity> {
+        match object_kind {
+            crate::table_catalog::TableMetadataMaintenanceObjectKind::DataFile => self.data_files.get(location),
+            crate::table_catalog::TableMetadataMaintenanceObjectKind::DeleteFile => self.delete_files.get(location),
+            _ => None,
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


marshawcoco:
@codex review

marshawcoco:
Addressed the latest Copilot review in `33f6e6db`.

- Made `SnapshotManifestLocation` ownership explicit by destructuring it before inserting the path and provenance fields. The original disjoint-field partial move compiled successfully, but the new shape removes the ownership ambiguity entirely.
- Updated inherited-manifest validation so missing provenance in a legacy `manifests` snapshot is treated as unknown. Any sequence or added-snapshot identity that is known in the current snapshot must still match exactly.
- Extended the through-catalog regression to commit a legacy `manifests` snapshot and then successfully upgrade to a `manifest-list` snapshot that reuses the legacy manifest and adds a new manifest.

Verification completed on the latest `upstream/main`: 13 row-level conflict tests, 246 table catalog tests, RustFS library Clippy with warnings denied, formatting/diff checks, and `make pre-commit`.

PR #4841, opened 2026-07-15
ci(perf): fit the baseline cache build to the post-LTO profile and self-heal nightly misses

overtrue:
## What

perf-3 follow-up (rustfs/backlog#1152): the baseline binary cache landed in #4816 but never populated — every `build-baseline-cache` run on 2026-07-15 was killed at its 60min ceiling with "The job has exceeded the maximum execution time of 1h0m0s" (e.g. run 29390384816, and the first one at run 29381941903).

Root cause: #4806 (merged 07-14) put `lto = "thin"` + `codegen-units = 1` on `[profile.release]`, pushing a single release build past 60min on `sm-standard-2`. The ~32min/side estimate #4816 was budgeted on predates that profile change.

## Changes

- **`build-baseline-cache` timeout 60 → 100min.** The measured binary must keep the production release profile (that is the thing being benchmarked), so the budget absorbs the build instead of weakening the profile.
- **Same-commit cache-miss self-heal in the A/B job.** On a nightly/dispatch run where the candidate commit equals the baseline commit, a restore miss previously fell back to `--baseline-ref origin/main`, making the rig build the *identical commit twice* (~65min x 2 post-#4806 — no job budget fits that; it is also what the pre-#4816 nightlies died of at their 120min ceiling on 07-11..07-14). Now the job builds the binary once, reuses it for both phases (`--skip-build`), and saves it back to the cache under `rustfs-baseline-<sha>` so the next run hits.
- The PR-context miss (candidate != baseline, opt-in `perf-ab` gate) keeps the double-build fallback; the comment now documents that it will overrun and alert, and that perf-7 will bring merge-base caching for that path.

## Validation

- `actionlint`: clean (shellcheck-integrated).
- The self-heal branch reuses the exact staging/key contract of `build-baseline-cache` (`baseline-bin/rustfs`, `rustfs-baseline-<sha>`), and the run-step treats `cache-hit || selfheal.built` identically to a hit, so provenance records the source truthfully ("actions-cache" vs "source build (cache self-heal, ...)").
- End-to-end proof comes from this PR's own merge: its push build (with the 100min budget) populates the cache for the merge SHA, after which a `workflow_dispatch` on `main` must complete with zero source builds in well under 50min. I will post that run link on rustfs/backlog#1152.

Refs rustfs/backlog#1152 (perf-3 follow-up).

### DIFF
--- a/.github/workflows/performance-ab.yml
+++ b/.github/workflows/performance-ab.yml
@@ -75,6 +75,14 @@ jobs:
     name: Build + cache baseline binary
     if: github.event_name == 'push'
     runs-on: sm-standard-2
+    # Latest-wins: consumers only ever restore the binary for the *current*
+    # origin/main tip, so when pushes land faster than the ~65min build, a
+    # superseded build's output is dead weight — cancel it instead of stacking
+    # hour-long jobs on the shared runner pool. A skipped intermediate SHA at
+    # most costs one same-commit self-heal in the A/B job.
+    concurrency:
+      group: perf-baseline-build-main
+      cancel-in-progress: true
     # #4806 put thin LTO + codegen-units=1 on [profile.release], pushing a
     # single release build past 60min on this runner — every cache build on
     # 2026-07-15 died on the old 60min ceiling ("exceeded the maximum execution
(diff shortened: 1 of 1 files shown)


overtrue:
Pushed a second commit: job-level `concurrency` (latest-wins) on `build-baseline-cache`. Main merged 4 times in the last hour and each push now spawns a ~65min build on the shared sm-standard-2 pool; a superseded build's binary is never restored (consumers always key on the current origin/main tip), so cancelling it is pure savings. A skipped intermediate SHA at most costs one same-commit self-heal in the A/B job.

PR #4881, opened 2026-07-15
feat(helm): add drivesPerNode and ordered pod startup for distributed mode

### LINKED ISSUE #4880: feat(helm): add drivesPerNode and ordered startup to fix distributed mode crash loops
## Problem

The current Helm chart has two issues in distributed mode:

### 1. No `drivesPerNode` control

The number of drives per pod is hardcoded to match `replicaCount` (4 drives when `replicaCount=4`, 1 drive when `replicaCount=16`). There is no way to configure drives-per-node independently from pod count.

For example, with `replicaCount: 4` you want `2` drives per pod (8 PVCs total for EC:2), but the chart always creates 4 drives per pod (16 PVCs).

### 2. gRPC crash loop on startup (parallel pod start)

With `podManagementPolicy: Parallel`, all pods start simultaneously. When pod-0 starts RustFS, pods 1–3 are still in their init containers (port 9000 not yet listening). RustFS attempts peer gRPC connections, marks unreachable peers as faulty/suspect, cannot achieve quorum, and exits with code 1 after ~2 minutes. This causes a crash loop on every fresh deployment.

Additionally, there is a DNS race condition: pods enter CrashLoopBackOff before their headless service DNS entry is published, causing `Name does not resolve` errors.

### 1. Add `storageclass.drivesPerNode` value

```yaml
storageclass:
  drivesPerNode: 4  # drives per pod, independent of replicaCount
```

This lets operators configure 4 pods × 2 drives = 8 PVCs instead of being locked to 4 pods × 4 drives = 16 PVCs.

### 2. Ordered init container startup

Add two phases to the init container script:

**Phase 1 – DNS wait:** All pods wait until every peer DNS entry resolves before starting RustFS.

**Phase 2 – Port wait (lower ordinals only):** Each pod waits for all lower-ordinal peers to be ready on port 9000 before its init container exits. This serializes startup: pod-0 starts immediately, pod-1 waits for pod-0, pod-2 waits for pod-0+1, etc. This guarantees RustFS on pod-N always finds pods 0…N-1 already listening, eliminating the gRPC transport error.

```sh
# Phase 1: wait for all peer DNS
for i in $(seq 0 $(($REPLICA_COUNT - 1))); do
  peer="<pod>-${i}.<headless-svc>"
  until nslookup "$peer" > /dev/null 2>&1; do sleep 2; done
done

# Phase 2: wait for lower-ordinal peers port 9000
ORDINAL=$(hostname | rev | cut -d'-' -f1 | rev)
if [ "$ORDINAL" -gt 0 ]; then
  for i in $(seq 0 $(($ORDINAL - 1))); do
    peer="<pod>-${i}.<headless-svc>"
    until wget -q -T 2 -O /dev/null "http://${peer}:9000/health" 2>/dev/null; do sleep 2; done
  done
fi
```

## Impact

- Fixes crash loops on every fresh distributed deployment
- Fixes DNS race condition (`Name does not resolve`)
- Gives operators flexible PVC count control via `drivesPerNode`

A PR with these changes will follow.

dipeshmr1:
## Summary

Fixes two crash-loop causes in distributed mode and adds flexible drive-count control.

### 1. `storageclass.drivesPerNode` (new field, default: `4`)

Previously drive count per pod was hardcoded to match `replicaCount` (4 drives when replicaCount=4, 1 drive when replicaCount=16). There was no way to run e.g. **4 pods x 2 drives = 8 PVCs**. `drivesPerNode` decouples drive count from pod count and integrates with the existing pools system via `mergeOverwrite`. The hardcoded `replicaCount == 4 / 16` branch is replaced by a single `drivesPerNode`-driven loop.

### 2. Ordered init-container startup

Fixes two crash-loop root causes on fresh distributed deployments:

**a) DNS race (`Name does not resolve`)** - With `podManagementPolicy: Parallel` all pods start simultaneously and enter CrashLoopBackOff before their headless-service DNS entry is published. Fix: every pod waits until all peer FQDNs resolve via `nslookup` before starting RustFS.

**b) gRPC transport errors (exit after ~2 min)** - All pods start at once. Pod-N RustFS attempts gRPC to pods 0-N-1 which are still in init containers (port 9000 not listening). RustFS marks them faulty, fails quorum, exits. Fix: pod-N init container waits for each lower-ordinal peer `/health` before exiting. Pod-0 starts immediately; pod-1 waits for pod-0; pod-2 waits for 0+1; etc.

## Files changed

- `helm/rustfs/values.yaml` - adds `storageclass.drivesPerNode: 4`
- `helm/rustfs/templates/_helpers.tpl` - `rustfs.volumes` uses `drivesPerNode` for drive range; removes hardcoded replicaCount restriction in `rustfs.pools`
- `helm/rustfs/templates/statefulset.yaml` - init script, volumeMounts, volumeClaimTemplates all use `drivesPerNode`

## Test plan

- [ ] `helm template` with default `drivesPerNode: 4` is backward-compatible
- [ ] `helm template` with `drivesPerNode: 2` renders 2 PVCs per pod
- [ ] `RUSTFS_VOLUMES` URL drive range matches `drivesPerNode`
- [ ] Fresh 4-pod deployment reaches Running 1/1 without crash loop
- [ ] Init container logs show ordered peer-wait before RustFS starts

Closes #4880


dipeshmr1:
@dipeshmr1 I have read and agree to the CLA.

overtrue:
Hi @dipeshmr1, thank you very much for your contribution!

We recently upgraded our Contributor License Agreement to v2. Since this change affects all open pull requests, we'd kindly ask you to review and re-sign the updated CLA so we can move forward with your PR.

You can do so by leaving a comment with the CLA acceptance statement (or by following the link that the CLA bot provides on this PR). Once that's done, we'll continue the review process.

Thanks again for your patience and for helping improve RustFS!

overtrue (changes requested):
## Review

Thanks for tackling the crash-loop issues — the ordered-startup idea and DNS-wait are reasonable, and using `busybox:stable` for the init image means `nslookup`/`wget -T` are both available. However, I rendered the chart against `main` and found two backward-compatibility regressions, one bug that breaks every multi-pool deployment, and a validation guard that was silently removed. Details with rendered evidence below.

### Context: the chart already has a `drivesPerNode` mechanism

`main` already ships a **top-level** `.Values.drivesPerNode` (default `null`) with a `rustfs.poolDrives` helper and `$pool.drives`, inferring drive count from replica count (`replicaCount==4 → 4`, everything else → `1`) for backward compatibility. This PR does not reuse that path; it introduces a **second** `storageclass.drivesPerNode` (hardcoded default `4`) and rewires `rustfs.volumes` + `statefulset.yaml` to read the new key. The two mechanisms now conflict.

### 🔴 Critical 1 — top-level `drivesPerNode` is silently ignored

`_helpers.tpl` line 274 and the statefulset now read `$pool.storageclass.drivesPerNode | default 4`, so `$pool.drives` (derived from the top-level value) is no longer consulted. `README.md` still tells users to run `--set drivesPerNode=2`, which is now a no-op:

```
# main:  --set drivesPerNode=2  ->  .../data/rustfs{0...1}   (2 drives, correct)
# PR:    --set drivesPerNode=2  ->  .../data/rustfs{0...3}   (4 drives, flag ignored)
```

A user aiming for 8×2 silently gets 8×4 — wrong drive count and parity.

### 🔴 Critical 2 — breaks backward compatibility for any non-4 replica count

Because `storageclass.drivesPerNode` defaults to `4`, the inference path is fully bypassed. An existing 16-replica cluster that relied on the inferred single `/data`:

```
# main:  replicaCount=16, unset  ->  RUSTFS_VOLUMES .../data              , PVC name: data              (16x1)
# PR:    replicaCount=16, unset  ->  RUSTFS_VOLUMES .../data/rustfs{0...3} , PVC name: data-rustfs-0..3  (64 PVCs)
```

Upgrading changes both the volume layout and the PVC names, so the existing cluster can no longer match its PVCs and `RUSTFS_VOLUMES` points at a brand-new topology → existing data becomes inaccessible. Note `values.yaml` lines 12–24 still document the old "unset → infer from replicaCount" behavior, so the docs now contradict the rendered output.

### 🔴 Critical 3 — init container deadlocks on every multi-pool deployment

The new `SVC_HEADLESS` env (`statefulset.yaml` lines 128–129) is built from `$pool.fullname`-headless, but the StatefulSet `serviceName` (line 26) is the **root** `-headless` for **all** pools, and only one headless Service is rendered. Rendered with a two-pool config:

```
Pod actual FQDN (RUSTFS_VOLUMES, correct):  t-rustfs-pool1-0.t-rustfs-headless...
init waits on FQDN (SVC_HEADLESS, wrong):    t-rustfs-pool1-0.t-rustfs-pool1-headless...   <- does not exist
Rendered Services:  t-rustfs / t-rustfs-headless / t-rustfs-svc   (no pool1-headless)
```

For pool ≥ 1 the `nslookup` loop never resolves → init hangs forever → pool 0 comes up but pool 1+ never start. `SVC_HEADLESS` should use the root `rustfs.fullname`-headless, consistent with `rustfs.volumes` and `rustfs.serverDomains`.

### 🔴 Critical 4 — the `pools.list` replicaCount guard was removed

The diff drops the `pools.list[i].replicaCount must be >= 2` `fail`, and the PR description does not mention it — looks accidental:

```
# main:  pools.list[0].replicaCount=1  ->  Error: ...must be >= 2, got 1  (rejected)
# PR:    pools.list[0].replicaCount=1  ->  renders {0...0}, a single-replica distributed pool
```

The top-of-statefulset `replicaCount >= 2` check (line 4) only validates the top-level `.Values.replicaCount`, not per-pool `$p.replicaCount`, so nothing else catches this. It reintroduces exactly the crash-loop class this PR is meant to fix.

### 🟡 Medium 5 — ordered startup breaks under mTLS

The health check at `statefulset.yaml` line 159 hardcodes `http://${peer}:9000/health`. When `mtls.enabled`, port 9000 serves HTTPS only (the probe template also switches to `curl https` in that case), so the `wget` for ordinal>0 pods never succeeds and init hangs. It needs to switch to `https` (and handle the cert) when mTLS is on.

### 🟡 Medium 6 — leftover dead code and stale docs

- `statefulset.yaml` line 10 `$drivesPerNode := int $pool.drives` is now unreferenced (dead variable).
- `values.yaml` lines 12–24 and the README section on "unset → infer from replicaCount" no longer match behavior. The whole top-level `drivesPerNode → poolDrives → $pool.drives` chain is now vestigial, yet its `drivesPerNode must be >= 1` guard still fires for `--set drivesPerNode=0` — so `0` errors while `2` is silently ignored, which is inconsistent.

### One assumption worth validating on a real cluster

Ordered startup assumes pod-0 can serve `/health` with 200 **before** it reaches quorum. If RustFS waits for quorum before serving `/health` in distributed mode, then pod-1 waits on pod-0's `/health` while pod-0 waits for pod-1 to join → cluster-wide deadlock. Please confirm on a fresh 4-node cold start that `/health` (liveness) answers on the HTTP port as soon as the process is up.

### Suggestion

Converge on a single source of truth — either:

1. **(preferred)** wire the per-pool control through the existing `poolDrives`/`$pool.drives` path, keep `storageclass.drivesPerNode` defaulting to `null` and preserve inference (do not hardcode `4`) so backward compatibility holds; or
2. remove the top-level `drivesPerNode` + `poolDrives` machinery entirely and update `values.yaml` comments and the README accordingly.

Either way, before merge: fix the root-headless name (Critical 3), restore the `>= 2` guard (Critical 4), handle the mTLS protocol (Medium 5), drop the dead code/stale docs, and add a `helm template` snapshot test that locks "default-value render matches the pre-PR chart".

cxymds:
I re-reviewed the current head against the latest `main`. The existing requested changes remain valid. I also found three additional startup issues:

1. **The health wait ignores the configured endpoint port.**  
   `statefulset.yaml` hardcodes `http://${peer}:9000/health`, while the chart consistently uses `.Values.service.endpoint.port` for `RUSTFS_VOLUMES`, the Service, and `containerPort`. If an operator configures another endpoint port, every pod with ordinal > 0 waits forever on port 9000.

2. **The DNS wait is incomplete for multi-pool deployments.**  
   Each StatefulSet waits only for peers in its own `$pool`, but the generated `RUSTFS_VOLUMES` contains every configured pool. Even after correcting the headless-service name, pool 0 can start before pool 1+ DNS records exist and reproduce the same name-resolution startup failure. The wait should iterate over the normalized full pool list and every replica.

3. **`storageclass.drivesPerNode` has no reliable lower-bound validation.**  
   Helm treats numeric `0` as empty for `default`, so `storageclass.drivesPerNode: 0` silently becomes `4`; negative values can render invalid ranges and volume expressions. This is another reason to remove the duplicate setting and reuse the existing validated `$pool.drives` path from #4901.

Given that #4901 already provides the backward-compatible drive topology support, I recommend dropping the drive-count portion of this PR and submitting the ordered-startup change separately. That smaller change should:

- use the root headless Service;
- wait for peers across all pools;
- use the configured endpoint port and mTLS scheme/credentials;
- include bounded, diagnosable failure behavior;
- add Helm render tests for multi-pool, mTLS, and custom-port configurations.

The current patch should remain blocked until these issues and the previously requested changes are resolved.

cxymds:
I recommend closing this PR in its current form. The drive-topology portion has been superseded by #4901, while the ordered-startup portion needs a narrower redesign to address multi-pool discovery, mTLS, configurable endpoint ports, and bounded/diagnosable wait behavior.

Keeping #4880 open and addressing startup coordination in a focused follow-up PR would make the change substantially safer and easier to review. Thank you again for investigating the original crash-loop behavior and proposing a concrete direction.

PR #4889, opened 2026-07-16
test(kms): assert Vault namespace reaches requests

hkwi:
## Summary
- add Vault KV2 and Transit regression tests that assert configured namespaces reach outbound Vault requests
- accept either X-Vault-Namespace or the /v1/{namespace}/ path form, while avoiding loose substring matching for path namespaces
- keep env parsing coverage in the existing config tests and exercise the backend clients directly with VaultConfig/VaultTransitConfig

## Testing
- git diff --check HEAD~1..HEAD
- not run: cargo test (cargo is not installed in this environment)

### DIFF
--- a/crates/kms/src/backends/vault.rs
+++ b/crates/kms/src/backends/vault.rs
@@ -833,6 +833,88 @@ impl KmsBackend for VaultKmsBackend {
 mod tests {
     use super::*;
     use crate::config::{VaultAuthMethod, VaultConfig};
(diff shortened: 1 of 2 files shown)


hkwi:
Closing this draft: after checking beta9, RustFS already wires RUSTFS_KMS_VAULT_NAMESPACE through to VaultConfig/VaultTransitConfig, and vaultrs 0.8.0 sends that setting as X-Vault-Namespace. The extra regression test is not necessary for this PR.

PR #4945, opened 2026-07-17
test(ilm): re-enable test_transition_and_restore_flows; fix test-util disk-open and restore error-path lock

overtrue:
## Related Issues

Closes rustfs/backlog#1303. Related: rustfs/backlog#1148 (serial-lane exclusions), #4956 / backlog#1304 (restore accept-path CAS, which fixed the restore self-deadlock at its source).

## Summary of Changes

Re-enables `test_transition_and_restore_flows` in the ILM Integration (serial) lane by fixing the test-util bug that was masking it. **Test-only change** after rebasing onto current main.

**The reported "missing xl.meta on disk2" was a test-util bug, not an EC metadata-distribution issue.** After a transition, all four shard disks hold a fully consistent xl.meta — verified by decoding every shard with `dump_fileinfo` (identical `(status, tier, remote key, remote version id)` tuple on all four disks). The panic came from `open_disk` in `crates/ecstore/src/services/tier/test_util.rs`, which hardcoded `disk_index = 0` for every disk path. `LocalDisk::new` validates the endpoint's `(set_idx, disk_idx)` against the position recorded in the disk's own `format.json` and rejects mismatches with `InconsistentDisk`, so every disk except the real slot-0 disk failed to open, and `read_transition_meta` collapsed that into `None` → "missing xl.meta". Nobody hit this before because all other callers of the shard-reading helpers only ever read `disk_paths[0]`. `open_disk` now derives the real `(set_idx, disk_idx)` from the disk's `format.json`. This also un-breaks `free_version_count` / `wait_for_free_version_absence` for non-first disks, which silently returned 0 before. (This corrects the ci.yml comment inherited from #4886 that attributed the failure to "EC metadata distribution".)

**Note on scope:** an earlier revision of this PR also propagated `no_lock` into the restore copy-back's `update_restore_metadata` path to fix a restore self-deadlock. That deadlock was fixed at its source by #4956 (backlog#1304), which replaced #4877's whole-copy-back object write lock with an accept-path CAS guard that is dropped before the copy-back runs. With no outer lock held, the copy-back's `opts.no_lock` is `false` and the metadata rewrite acquires the lock normally, so that propagation is now a no-op with a misleading comment — dropped on the merge. The remaining diff is purely the test-util fix plus re-enabling the test.

## Verification

- `cargo nextest run -p rustfs-scanner -E 'test(test_transition_and_restore_flows)' --run-ignored all` — passes on the merged tree (test-util fix + #4956's deadlock fix; was fast-failing at ~2.2s on the bogus "missing xl.meta" before this change).
- Full ILM Integration (serial) lane command from ci.yml run locally — the re-enabled test passes alongside the rest of the lane.
- `make pre-commit`

## Impact

Test-only. The tier test-util helpers now work on every shard disk (not just slot 0), and `test_transition_and_restore_flows` rejoins the serial lane. No product-code change. The remaining lane exclusions (`test_noncurrent_*` ×2, `test_restore_chain_*`) are unrelated known issues and stay excluded; `restore_object_usecase_*` was already re-enabled by #4956.

## Additional Notes

The ci.yml comment block is updated to record that the disk2 "missing xl.meta" symptom was a test-util indexing bug (open_disk hardcoding disk_index 0), superseding the earlier "EC metadata-distribution" attribution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -221,6 +221,9 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           cache-save-if: ${{ github.ref == 'refs/heads/main' }}
 
+      # test_transition_and_restore_flows was re-enabled by rustfs/backlog#1303:
+      # its "missing xl.meta on disk2" was a test-util bug (open_disk hardcoded
+      # disk_index 0), not an EC metadata-distribution issue.
       # restore_object_usecase_reports_ongoing_conflict_and_completion was
       # re-enabled by backlog#1304 (restore accepts serialize on a short CAS
       # guard; the copy-back no longer holds the #4877 whole-copy-back lock,
@@ -230,9 +233,6 @@ jobs:
       # they keep #[ignore] with a backlog reference):
       #   - test_noncurrent_{expiry,transition}_still_works_after_immediate_compensation_transition:
       #     noncurrent transition/expiry after an immediate compensation transition.
-      #   - test_transition_and_restore_flows: transition metadata is missing on
-      #     one drive (assert_transition_meta_consistent: "missing xl.meta ... on
-      #     disk2") - an EC metadata-distribution issue, not the restore lock.
       #   - test_restore_chain_local_read_expiry_keeps_remote_and_allows_re_restore:
       #     DeleteRestoredAction sets opts.transition.expire_restored, but no
       #     delete path reads that flag, so cleanup deletes the whole object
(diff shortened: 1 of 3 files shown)


overtrue:
The End-to-End Tests failure here is a pre-existing main regression, not this PR. `content_encoding_test::test_content_encoding_aws_chunked_not_returned_issue_1857` and `test_content_encoding_aws_chunked_with_effective_encoding_roundtrip` fail identically on main's own CI run for f67e8a6cd ([End-to-End Tests / full merge gate both red](https://github.com/rustfs/rustfs/actions/runs/29551214728)), while this PR touches only the tier test util, the restore error-path lock options, and the ILM lane exclusion list.

Root cause of the main regression: #4928 added `resolve_put_object_authoritative_size`, which classifies a request as aws-chunked from the `Content-Encoding`/`Transfer-Encoding` headers alone (`request_uses_aws_chunked`) and then rejects it with `UnexpectedContent` when `x-amz-decoded-content-length` is absent. Both e2e tests (dating from #2009/#2496) send `Content-Encoding: aws-chunked` with a plain, unframed body and no decoded-length header — previously accepted with `Content-Length` as the authoritative size. A fix should gate the rejection on the body actually being chunk-framed (streaming `x-amz-content-sha256`, or decoded-length presence) rather than on the declared header alone.

The ILM Integration (serial) lane — the check this PR is about — is green, with the re-enabled `test_transition_and_restore_flows` passing at 11.9s.

overtrue:
Rebased onto current main and **reduced this to a test-only change.** The earlier revision also propagated `no_lock` into the restore copy-back's `update_restore_metadata` path to fix the #4877 restore self-deadlock. That deadlock is now fixed at its source by #4956 (backlog#1304), which replaced the whole-copy-back object write lock with an accept-path CAS guard that is dropped before the copy-back runs. With no outer lock held, the copy-back's `opts.no_lock` is `false`, so the metadata rewrite acquires the lock normally and the propagation became a no-op carrying a now-false comment ("the restore handler holds the object write lock"). I dropped it.

The remaining diff is exactly: the `open_disk` test-util fix (derive `(set_idx, disk_idx)` from `format.json` instead of hardcoding slot 0), re-enabling `test_transition_and_restore_flows`, and the matching ci.yml exclusion/comment update. Verified locally: `test_transition_and_restore_flows` passes in 12.4s on the merged tree with only the test-util fix.

Issue #4846, opened 2026-07-15
Distributed-lock quorum tests misclassify as timeout under parallel e2e load


overtrue:
Surfaced by the first automated full-e2e run (backlog#1149 ci-5, run https://github.com/rustfs/rustfs/actions/runs/29381309848). Two in-process multi-node distributed-lock tests assert that lock acquisition against a cluster with failed nodes surfaces a QUORUM error, but under full parallel e2e load the 2s acquisition deadline fires first and the error is classified as a timeout:

```
reliant::lock::test_distributed_lock_2_nodes_grpc_read_survives_failed_node
reliant::lock::test_distributed_lock_4_nodes_grpc_read_write_quorum_split_with_two_failed_nodes

panicked: Error should be quorum related, got: Lock acquisition timeout for resource 'test-bucket/test-object@latest' after 2s
```

The sibling `test_distributed_lock_4_nodes_grpc` (happy path) passes. Options: classify quorum failure before the deadline, lengthen/env-scale the deadline, or accept timeout as a valid quorum-failure surface in the assertion.

These two spawn multi-node lock clusters and are load-sensitive, so they are excluded from the `e2e-full` merge-gate profile (`.config/nextest.toml`, reference to this issue) — their natural lane is the ci-7 nightly cluster/reliability run (backlog#1149). The resolving PR should re-admit or re-home them.

Issue #4844, opened 2026-07-15
Anonymous POST-object with SSE-S3 (explicit or bucket-default) returns 500


overtrue:
Surfaced by the first automated full-e2e run (backlog#1149 ci-5, run https://github.com/rustfs/rustfs/actions/runs/29381309848). Anonymous POST-object (form upload) requests involving SSE fail with 500:

```
multipart_auth_test::test_anonymous_post_object_accepts_sse_s3                          (expects 204, gets 500)
multipart_auth_test::test_anonymous_post_object_rejects_sse_s3_missing_from_policy_conditions (expects 403-class reject, gets 500)
multipart_auth_test::test_anonymous_post_object_uses_bucket_default_sse_kms             (gets 500)
multipart_auth_test::test_anonymous_post_object_uses_bucket_default_sse_s3              (gets 500)
```

The sibling SSE-KMS negative tests (`rejects_sse_kms*`) and SSE-C tests pass, so this is specific to SSE-S3 encryption application (explicit or bucket-default) on the POST-object path. Tests live in `crates/e2e_test/src/multipart_auth_test.rs`.

Temporarily excluded from the `e2e-full` merge-gate profile (`.config/nextest.toml`) with a reference to this issue; the fixing PR must re-admit them.

Issue #4843, opened 2026-07-15
Auto-extract hard-rejects archives with over-limit entry paths even under ignore-errors semantics


overtrue:
Surfaced by the first automated full-e2e run (backlog#1149 ci-5, run https://github.com/rustfs/rustfs/actions/runs/29381309848). When a PutObject auto-extract archive contains an entry whose path exceeds the archive entry path limit (1024), the WHOLE request is rejected with `InvalidArgument: Archive entry path exceeds limit ... length=1113, limit=1024` — even when the client explicitly requested ignore-errors semantics, and even for scenarios the tests expect to succeed by skipping the bad entry.

Failing tests (3):

```
multipart_auth_test::test_signed_put_object_extract_skips_invalid_entry_when_ignore_errors_enabled
snowball_auto_extract_test::tests::snowball_auto_extract_ignores_invalid_entries_when_requested
snowball_auto_extract_test::tests::snowball_auto_extract_supports_standard_headers_with_combined_extract_options
```

Needs a semantics decision: either the ignore-errors option should skip limit-violating entries (MinIO-compat expectation encoded in the tests), or the tests must change to expect a hard reject — in which case the hardening's interaction with ignore-errors should be documented. Likely related to the archive_extract hardening work (fuzz target exists for this path).

Temporarily excluded from the `e2e-full` merge-gate profile (`.config/nextest.toml`) with a reference to this issue; the resolving PR must re-admit them.

PR #4948, opened 2026-07-17
fix(extract): treat tar mtime=0 as unset to fix unreadable entries

### LINKED ISSUE #4842: PutObject auto-extract returns 500 for tar entries with mtime=0 (OffsetDateTime deserialization) and related expand-path failures
Surfaced by the first-ever automated run of the full e2e suite (backlog#1149 ci-5 characterization, run https://github.com/rustfs/rustfs/actions/runs/29381309848, junit artifact `e2e-full-junit-12764`). The PutObject auto-extract pipeline (`X-Amz-Meta-Snowball-Auto-Extract`) returns 500 for essentially every positive expand scenario.

**Primary signature (13 tests)** — `500 InternalError: Io error: invalid type: integer `0`, expected an `OffsetDateTime``. Archive entries whose tar header mtime is 0 (the fixtures' default) hit a serde round-trip that writes the timestamp as integer `0` but expects an `OffsetDateTime` on read-back. Strong corroboration: `test_signed_put_object_extract_preserves_entry_mtime` — the one extract test that sets a non-zero mtime — PASSES.

**Same-path 500s (5 tests)** — `preserves_request_metadata_on_extracted_objects`, `preserves_sse_c`, `preserves_storage_class` (empty-body 500), `preserves_sse_s3_and_redirect`, `uses_bucket_default_sse_s3` (InternalError), plus `preserves_pax_metadata_and_version_id` (InvalidArgument).

Failing tests (18, all in `crates/e2e_test/src/multipart_auth_test.rs` + 2 in `snowball_auto_extract_test.rs`):

```
multipart_auth_test::test_signed_put_object_extract_{accepts_compat_header,expands_tar_entries_with_prefix_headers,expands_tar_gz_archive,expands_tbz2_archive,expands_tgz_archive,expands_txz_archive,expands_tzst_archive,normalizes_prefix_header_value,preserves_directory_markers_by_default,preserves_object_lock_legal_hold,preserves_object_lock_retention,preserves_pax_metadata_and_version_id,preserves_request_metadata_on_extracted_objects,preserves_sse_c,preserves_sse_s3_and_redirect,preserves_storage_class,uses_bucket_default_sse_s3}
snowball_auto_extract_test::tests::snowball_auto_extract_{prefers_exact_minio_prefix_over_suffix_fallback,supports_minio_prefix_and_directory_markers}
```

The negative-path guards (`rejects_*`, `returns_archive_etag`, `preserves_entry_mtime`) still pass, so the auth/validation surface is fine — the expand/commit path is what regressed. Candidate suspect: the workspace dependency refresh (#4804) changing `time`/serde behavior, but unconfirmed.

These tests are temporarily excluded from the `e2e-full` merge-gate profile (`.config/nextest.toml`) with a reference to this issue; the fixing PR must re-admit them by deleting the exclusion.

overtrue:
## Related Issues

Fixes #4842.

## Summary of Changes

The PutObject auto-extract pipeline (`X-Amz-Meta-Snowball-Auto-Extract`) returned 500 for essentially every positive expand scenario because tar entries with header mtime 0 (the fixtures' default, and the tar convention for "unset") were stored with `mod_time = UNIX_EPOCH`. The xl.meta on-disk format encodes `MTime` as unix nanoseconds with `0` meaning "no mod_time", so on read-back the epoch decoded to `None`, `FileMetaVersion::valid()` rejected the version (Object versions require a mod_time), and parsing fell through to the legacy rmp_serde fallback — which failed on the new-format bytes with `invalid type: integer 0, expected an OffsetDateTime`. Every subsequent read of the extracted object then surfaced as `InternalError`/`InvalidArgument`. This matches the corroborating observation in the issue: the one extract test that sets a non-zero mtime passed.

- `rustfs/src/app/object_usecase.rs`: treat tar header mtime 0 as unset in the extract path and fall back to the upload time, since an epoch mod_time is not representable in xl.meta.
- `crates/e2e_test/src/multipart_auth_test.rs`: two test-side fixes uncovered once the primary 500 was gone:
  - the pax fixture built its XHeader entry with a GNU-format header; the tar reader (astral-tokio-tar, from the #4804 dependency refresh) follows POSIX and only permits pax extension entries on ustar headers, which is also what real tar writers emit — switch the fixture to `Header::new_ustar()`;
  - the two SSE-S3 tests started the server without `RUSTFS_SSE_S3_MASTER_KEY`, which the server requires by design for SSE-S3 when KMS is not configured — start the server with a test master key via `start_rustfs_server_with_env`, mirroring the sse unit tests.
- `.config/nextest.toml`: delete the #4842 known-failure exclusions, re-admitting all 19 tests to the `e2e-full` merge gate as required by the quarantine discipline.

## Verification

- `make pre-commit` — passed.
- All 19 re-admitted tests plus the `preserves_entry_mtime` regression guard run green locally against the fixed binary (single-node servers per test): `cargo test -p e2e_test --lib -- --test-threads=1 --exact <the 17 multipart_auth extract tests + 2 snowball_auto_extract tests + preserves_entry_mtime>` → `20 passed; 0 failed`.

## Impact

Extracted archive entries whose tar header mtime is 0 now get the upload time as their last-modified instead of becoming permanently unreadable (500 on every GET/HEAD). Entries with a non-zero mtime keep the existing preserve behavior. No API or configuration changes.

## Additional Notes

The remaining known-failure families quarantined in the same characterization run (#4843 ignore-errors semantics, #4844/#4845 anonymous POST SSE/object-lock, #4846 lock-quorum) are untouched and still tracked by their own issues.


houseme (approved):

overtrue:
The failing `End-to-End Tests` check is unrelated to this PR: the 2 failing tests (`content_encoding_test::…aws_chunked…`) fail identically on `main` since fc7d46b6c (#4928), together with 3 more aws-chunked tests in the merge gate — see #4962 for the analysis. Fix is up at #4971; once it lands, a rebase/rerun here should go green. The 19 tests this PR re-admits to the `e2e-full` profile all pass locally (see PR description).

PR #4949, opened 2026-07-17
feat(api): add opt-in per-bucket dimension to the S3 API rate limiter

overtrue:
## Related Issues

Follow-up to rustfs/backlog#1191 (explicitly deferred sub-item: per-bucket dimension); builds on #4895.

## Summary of Changes

Adds an opt-in **per-bucket dimension** to the S3 API rate limiter shipped in #4895. The client-IP dimension gives per-client fairness; the bucket dimension gives each addressed bucket a collective budget, protecting the server from one hot bucket no matter how many client IPs the traffic is spread across.

**Core (`rustfs/src/server/rate_limit.rs`)**

- `RateLimiter` is generalized over its key type (`RateLimiter<K = IpAddr>`) with a `Borrow`-based `check(&Q)` API, so `&str` lookups against `String` bucket keys stay allocation-free on the hit path (an owned key is only materialized on first sight and on eviction). The client-IP call sites are unchanged in behavior.
- `RateLimitLayer` now carries up to two limiter instances (client + bucket), each with its own quota; a request must pass every configured dimension, and a rejection reports which one tripped via the new `dimension` label on `rustfs_http_server_requests_rate_limited_total` and in the debug event.
- Bucket extraction mirrors s3s host routing: on a configured virtual-hosted-style domain the bucket is the Host/authority prefix (matched ASCII-case-insensitively against the same expanded domain set — with port variants — that the s3s router uses; an exact domain match means path-style); otherwise the first path segment. Admin and table-catalog namespaces are never treated as buckets. Extraction is best-effort by design: a request it cannot classify just skips the bucket dimension while the client dimension still applies.
- Bucket names are attacker-chosen strings, so the existing bounded-shards design (100k tracked keys, lossless refilled-idle sweeps, most-idle eviction at cap) is the memory defense for this dimension; a regression test floods 10k random names and asserts the cap holds.

**Config (`crates/config/src/constants/api.rs`)**

- `RUSTFS_API_RATE_LIMIT_BUCKET_RPM` (default `0` = dimension off) and `RUSTFS_API_RATE_LIMIT_BUCKET_BURST` (default `0` = same as bucket RPM), under the existing `RUSTFS_API_RATE_LIMIT_ENABLE` master switch. The layer is built when enabled and **either** dimension has a non-zero RPM, so bucket-only deployments (client RPM 0) work; enabling with both at 0 logs why it stays inactive.

**Wiring (`rustfs/src/server/http.rs`)**

- The expanded virtual-hosted domain set is computed once (lifted out of the s3s service block, same expression) and shared by the s3s host router and the layer. The startup `api_rate_limit_state` event now reports both dimensions' quotas.

## Verification

- `cargo test -p rustfs --lib -- server::rate_limit admin::console` — 41 tests pass; new coverage: string-key budget sharing across borrowed lookups, random-bucket-name flood bounded by the cap, bucket extraction (path-style, virtual-hosted with port and mixed case, exact-domain path-style, admin/catalog exclusion), bucket budget shared across client IPs, virtual-hosted and path-style hits sharing one bucket budget, both-dimensions interaction, extended env matrix (bucket-only, client-only, both-zero inactive).
- `cargo test -p e2e_test api_rate_limit` — 3 real-binary e2e tests pass: bucket-only configuration throttles the sixth rapid hit on one bucket with `Retry-After` while an unrelated bucket stays unthrottled; the #4895 client-dimension and default-inert tests still pass.
- `make pre-commit` and `cargo clippy --all-targets -p rustfs -p rustfs-config -p e2e_test` (zero warnings).

## Impact

- Default behavior unchanged (new envs default to 0 = off; with only #4895's envs set, the layer behaves exactly as before).
- `RateLimitLayer::new` signature changed (now takes both optional quotas plus the domain set) — internal API, no external consumers.
- Metric gains a `dimension` label (`client_ip` | `bucket`), including on the console scope for label-set consistency.
- Semantics note: the bucket dimension is a collective ceiling — one abusive client can exhaust a bucket's budget for everyone on that bucket. That is inherent to the dimension (it protects the server, not per-client fairness); pairing it with the client-IP dimension is the recommended configuration and both dimensions compose in one layer pass.

## Additional Notes

- A rejected request has already consumed tokens from dimensions checked earlier (arrivals are counted); this matches standard limiter behavior and is asserted in the both-dimensions test.
- Requests whose Host is outside the configured domain set fall back to path-style extraction — exactly mirroring the s3s router, which would not route them virtual-hosted style either, so the limiter and the router always agree on what the bucket is.


overtrue:
The failing **End-to-End Tests** lane is a pre-existing break on main, not introduced here: main's own CI run [29546495804](https://github.com/rustfs/rustfs/actions/runs/29546495804) (commit 3b1bed700, after this branch's fork point) fails the same two `content_encoding_test` aws-chunked cases with 400 `UnexpectedContent`. This PR's rate-limit changes are default-off and not in the request path for those tests. Will rebase/rerun once main is fixed.

houseme (approved):

PR #4950, opened 2026-07-17
fix(ilm): implement expire_restored delete semantics for restore expiry

overtrue:
Closes rustfs/backlog#1302.

## Problem

`DeleteRestoredAction` must demote a restored object back to its pure transitioned state: delete only the **local restored copy**, strip the `x-amz-restore` headers, and leave the version — with its transition metadata and the remote tier data — untouched. `expire_transitioned_object` set `opts.transition.expire_restored = true` and called `delete_object`, but no delete path ever read that flag (`crates/ecstore` had exactly two references: the write site and a quorum-hash input). So restore-expiry ran an ordinary delete:

- **Unversioned buckets** (the reported case): the whole object was removed and `init_free_version` saw `TRANSITION_COMPLETE`, so a free-version record scheduled **remote tier deletion** — the object was lost locally *and* on the tier. `get_object_info` afterwards returned `ObjectNotFound`, which is exactly what `test_restore_chain_local_read_expiry_keeps_remote_and_allows_re_restore` pinned.
- **Versioned buckets**: the latest version got a spurious delete marker (re-created on every scanner pass, since the restore headers were never stripped) and `schedule_lifecycle_replication_delete_if_needed` replicated the bogus delete.
- **`DeleteRestoredVersionAction`** (noncurrent restored version) had a second, independent gap: `expire_transitioned_object` matched only `== DeleteRestoredAction`, so the versioned variant fell through to the **full transitioned-object delete path**, removing the remote tier data of that version.

## Fix

1. **`SetDisks::delete_object`** gains an explicit route: when `opts.transition.expire_restored` is set, target the found version with `FileInfo { version_id: goi.version_id, expire_restored: true }` and return early — before delete-marker resolution, replication dispatch, and free-version tagging. A delete marker would hide the version, a replicated delete would remove it on the target, and a free-version record would schedule remote tier cleanup; none of those may happen for restored-copy expiry. The `FileMeta::delete_version` layer already implemented the semantics (strip restore headers, keep the version, hand back the local data dir for cleanup) — it was simply never reachable.
2. **`expire_transitioned_object`**: the options construction is extracted into `transitioned_object_delete_opts` and dispatches on `IlmAction::delete_versioned()` / `delete_restored()` (MinIO's `Action.DeleteVersioned()`/`DeleteRestored()` parity), so `DeleteRestoredVersionAction` now routes through restored-copy cleanup with the exact version id.
3. **ci.yml**: `test_restore_chain_local_read_expiry_keeps_remote_and_allows_re_restore` is removed from the ILM Integration (serial) lane exclusion list (its `#[ignore]` reason is trimmed back to the standard serial-lane form).

One deliberate divergence from MinIO, in AWS's favor: MinIO's versioned-latest restore expiry mints a fresh UUID in its mark-delete branch, so it adds a delete marker and never strips the headers. This route targets the real latest version — no marker, headers stripped, GET keeps returning 200 from the tier — which is the S3-documented behavior (restore expiry is transparent).

## Tests

- **Control**: on `main` without the fix, the restore-chain test fails exactly as backlog#1302 describes — `object must still exist after restored-copy cleanup: ObjectNotFound(...)`. With the fix it passes (2.7s).
- **Full ILM Integration (serial) lane locally** (updated filter, `-j1`): **33/33 pass**, including the re-enabled restore-chain test and all delete/expiry siblings (`delete_transitioned_object_removes_remote_tier_copy_via_usecase`, `compensation_driven_transition_still_cleans_remote_tier_on_delete`, delete-marker tests) — the early route does not disturb ordinary transitioned-object deletion.
- **New unit tests**: `transitioned_object_delete_opts_routes_expiry_actions` pins the event→options routing (a revert of the `DeleteRestoredVersionAction` dispatch now fails a test); `test_delete_version_expire_restored_keeps_transitioned_version` pins the filemeta branch (version survives, three restore headers stripped, transition metadata intact, local data dir handed back).
- `cargo nextest run -p rustfs-ecstore -p rustfs-filemeta`: 2729/2729 pass. `make pre-commit` and `cargo clippy --all-targets` on the touched crates: clean.

## Adversarial review notes

- **Concurrency/durability**: the expire branch never removes the version record on any disk, so no crash window can lose the version or the tier pointer; per-disk staged rollback restores the stripped headers byte-identically on quorum failure; a double-fired expiry is a no-op (header removal and data-dir cleanup are both missing-source-tolerant). Metadata cache is invalidated while the object write lock is still held.
- **Security**: `expire_restored` is set in exactly one place (lifecycle internals); no S3 handler plumbs user input into it.
- **Object lock**: `check_object_lock_delete` still runs before the route — unchanged from the old path. Locked versions never reach it anyway: the evaluator suppresses restore-expiry actions for locked versions.


overtrue:
Note: #4945 (open) edits the same ci.yml exclusion list line — it removes `test_transition_and_restore_flows`, this PR removes `test_restore_chain_local_read_expiry_keeps_remote_and_allows_re_restore`. Whichever merges second needs a one-line rebase of that filter expression; the removals are independent.

overtrue:
The **End-to-End Tests** failure is pre-existing on main, unrelated to this PR: the same two tests (`content_encoding_test::tests::test_content_encoding_aws_chunked_not_returned_issue_1857`, `test_content_encoding_aws_chunked_with_effective_encoding_roundtrip`) fail with `400 UnexpectedContent` on aws-chunked PUT in every main run since [fc7d46b6c](https://github.com/rustfs/rustfs/actions/runs/29522890580) (#4928), e.g. main runs [29551214728](https://github.com/rustfs/rustfs/actions/runs/29551214728) and [29546495804](https://github.com/rustfs/rustfs/actions/runs/29546495804). Last green e2e on main was 3674f5f56. Tracked separately in rustfs/backlog.

Issue #4923, opened 2026-07-16
fix(table-catalog): preserve state during durable backing cutover


marshawcoco:
## Summary

Switching an existing table bucket from the `object` catalog backing to `durable-strong` after a successful migration dry-run makes its existing tables unavailable. Switching back to `object` restores the same table and allows further commits.

## Tested Baseline

- RustFS `cbf1c69234546b0dc1f54c615d690360e54413a9`
- PyIceberg `0.11.1`
- Fresh disposable four-volume local deployment

## Reproduction

1. Start RustFS with `RUSTFS_TABLE_CATALOG_BACKING=object`.
2. Enable a table bucket and create a table through the Iceberg REST Catalog.
3. Append two rows and record metadata generation `2`.
4. Call `GET /iceberg/v1/{warehouse}/catalog/migration`.
5. Confirm the response has `blockers=[]`, `status=READY_TO_SNAPSHOT`, and recommends `SNAPSHOT_OBJECT_BACKED_CATALOG`, `ENABLE_DURABLE_STRONG_BACKING`, and `VERIFY_DURABLE_STRONG_SNAPSHOT`.
6. Capture `GET .../catalog/export` successfully.
7. Restart the same volumes with `RUSTFS_TABLE_CATALOG_BACKING=durable-strong`.
8. Load the existing table.

## Actual Result

`loadTable` returns `404 NoSuchTableException`. The durable store only hydrates its private strong snapshot, but the documented migration flow does not provide an operation that materializes the object-backed catalog into that snapshot.

Restarting the same volumes with `RUSTFS_TABLE_CATALOG_BACKING=object` restores the original table. The two rows remain readable and another append advances the table to generation `3`.

## Expected Result

The migration workflow should provide a safe, executable state transfer before enabling `durable-strong`, or it should remain fail-closed and refuse to recommend cutover until the strong snapshot has been materialized and verified. A successful preflight plus the documented backup/restart sequence must not make existing tables disappear.

## Notes

The rollback path preserved object-backed state correctly. This report is about the missing cutover state-transfer step, not data loss in the object-backed catalog.

marshawcoco:
I'm fixing it.

PR #4952, opened 2026-07-17
feat(table-catalog): add durable backing state transfer

marshawcoco:
## Related Issues

Related to #4923

## Summary of Changes

Adds failure-safe state transfer from the object-backed table catalog to the durable-strong backing.

The previous migration readiness check validated the source catalog but did not copy its authoritative state, so enabling durable-strong backing could make existing tables disappear after restart.

- Adds explicit materialization and cancellation operations for backing migration.
- Copies table buckets, namespaces, tables, views, commit logs, idempotency records, and warehouse indexes into the strong snapshot.
- Fences registry and table-bucket mutations while a migration snapshot is active.
- Requires every table bucket to be materialized before durable-strong mode is considered ready.
- Supports idempotent retries and rejects stale, conflicting, incomplete, or internally inconsistent snapshots.
- Prevents cancellation after durable-strong state has advanced.
- Adds a dedicated migration administration action and route policy coverage.
- Updates migration diagnostics, support documentation, and failure-rehearsal scripts.

## Verification

- `cargo fmt --all --check`
- `git diff upstream/main...HEAD --check`
- `cargo test -p rustfs --lib durable_strong_migration_ -- --nocapture`
- `cargo test -p rustfs --lib test_admin_route_matrix_matches_registered_routes -- --nocapture`
- `cargo test -p rustfs-policy test_table_catalog_admin_action_is_valid -- --nocapture`
- `python3 -m unittest discover -s scripts/table-catalog -p 'test_*.py'`
- Live PyIceberg object-backed materialization, durable-strong restart, append, and second-restart rehearsal

## Impact

Adds POST and DELETE catalog migration operations and a dedicated administrative permission. Object-backed catalog mutations are fenced after materialization until the deployment completes cutover or safely cancels the migration.

Existing catalog operations are unchanged when no migration is active.

## Additional Notes

Every table bucket must be materialized before enabling durable-strong backing. Cancellation fails closed if the durable snapshot has changed since materialization, preventing stale object-backed state from replacing newer strong state.


marshawcoco:
@codex review

Issue #4902, opened 2026-07-16
datausageinfo and accountinfo take ~7 minutes for a 440k-object PBS bucket and trigger list_path channel-closed errors


kiliant:
## Describe the bug

On RustFS `1.0.0-beta.9`, the Console endpoints used for bucket statistics and the operational-status page take several minutes to complete for a large Proxmox Backup Server bucket.

Affected endpoints:

- `GET /rustfs/admin/v3/datausageinfo`
- `GET /rustfs/admin/v3/accountinfo`

The bucket contains approximately 440,000 objects and 686 GiB. After restarting RustFS, the Console initially shows pending bucket size/object-count values and reports that storage-usage statistics are unavailable.

After approximately 6–7 minutes, the bucket overview eventually displays the correct object count and size. During this period, the backend logs slow in-flight requests and repeated `list_path worker failed` errors with `Io error: channel closed`.

A direct recursive version listing succeeds, but takes nearly seven minutes:

```console
$ time timeout 10m \
    minio-client --insecure ls --recursive --versions \
    <alias>/<bucket>/<pbs-prefix>/.chunks/ \
    >/dev/null

real    6m54.717s
user    0m29.774s
sys     0m8.682s
```

This suggests that the admin endpoints synchronously perform a complete object/version listing before returning usage information.

## Environment

- RustFS: `1.0.0-beta.9`
- Deployment: containerized single-node deployment
- Architecture: `linux/amd64`
- Topology: one node, one data directory, one online disk
- Container resources:
  - 8 CPUs
  - 8 GiB memory
- Workload: Proxmox Backup Server datastore stored in an S3 bucket
- Bucket:
  - approximately 440,000 objects
  - approximately 686 GiB
  - large `.chunks` namespace with many hash-prefix directories
- TLS is enabled for both the S3 API and Console

## Relevant scanner configuration

```yaml
RUSTFS_SCANNER_ENABLED: "true"
RUSTFS_SCANNER_START_DELAY_SECS: "0"
RUSTFS_SCANNER_CYCLE: "3600"
RUSTFS_SCANNER_SPEED: "default"
RUSTFS_SCANNER_IDLE_MODE: "true"
RUSTFS_SCANNER_YIELD_EVERY_N_OBJECTS: "128"
RUSTFS_SCANNER_CACHE_SAVE_TIMEOUT_SECS: "120"
RUSTFS_DATA_USAGE_UPDATE_DIR_CYCLES: "16"
```

[...]

PR #4953, opened 2026-07-17
fix(admin): serve bucket usage from snapshot without a blocking version walk

kiliant:
## Summary

Draft PR for #4902, scoped to the first-PR boundary agreed in the issue: return the scanner snapshot plus the existing memory overlay immediately, and do not await a live version walk on the admin request path.

`AccountInfo` and `DataUsageInfo` both awaited `refresh_bucket_usage_from_object_layer` once per bucket, so both endpoints scaled with bucket size. On the ~440k-object bucket from the issue the walk takes ~6m55s, which is what the Console shows as a stalled bucket overview.

Both handlers already load the scanner snapshot and apply the in-memory write overlay *before* the per-bucket loop. The walk then overwrote those values. This PR drops the walk and serves what is already computed.

**This is deliberately snapshot-only.** Per point 6 of the review, the response contract stays as-is rather than gaining freshness states.

## Scope

Following the review, everything that lives in the background-recount coordinator is out of scope here. This PR contains no scheduler, no in-flight map, no epoch fencing, no semaphore, and no TTL cache consumption — so blocking issues 1, 2, 4 and 5 do not arise in this diff. The coordinator is deferred to a follow-up.

`ecstore` is not modified. `refresh_bucket_usage_from_object_layer` and its existing tests are left in place: it is now unused (`pub`, so no dead-code warning), which is intentional groundwork for the bounded coordinator follow-up rather than an oversight. Happy to delete it here instead if you would rather it not sit unused.

## Withdrawing an earlier claim

I claimed in the issue that this work removes the `list_path worker failed: channel closed` errors. **That was wrong and I withdraw it.** You are correct: this branch does not touch `gather_results`, and the 300s bound I had proposed was below the ~415s walk anyway. The consumer-gone behavior is a separate fix and I am not claiming anything about it here.

## Verification

- `make pre-commit` — passes
- `cargo clippy -p rustfs --all-targets` — clean
- `cargo test -p rustfs-ecstore --lib data_usage` — 32 passed; the overlay/snapshot semantics this PR now relies on are already covered there and are unchanged
- Rebased onto current `main` (`1.0.0-beta.10`)

## Known gap — guidance welcome

The one item from your required-coverage list that applies to this reduced scope is *"both admin endpoints returning without entering a live version walk"* as a regression guard. **It is not in this PR**, and I did not want to hand-wave it.

The obstacle is a boundary question I would rather you decide than guess at. A deterministic guard is straightforward in principle — writing objects via `ECStore` directly bypasses the usage overlay, so object layer non-empty / snapshot+overlay empty is a clean divergence state, and the assertion is that the response reports nothing for that bucket. But driving `execute_query_data_usage_info` needs `set_object_layer` exported through the `storage_api` boundary (it currently is not) plus coordination on the `GLOBAL_OBJECT_API` `OnceLock`.

Would you prefer:
1. exporting `set_object_layer` through the test boundary module, or
2. this guard living at a different layer, or
3. deferring it to the coordinator PR, where the harness is needed anyway?

Happy to add it in whichever shape fits your architecture rules.

Refs #4902

### DIFF
--- a/rustfs/src/admin/handlers/account_info.rs
+++ b/rustfs/src/admin/handlers/account_info.rs
@@ -19,8 +19,7 @@ use crate::admin::storage_api::bucket::versioning_sys::BucketVersioningSys;
 use crate::admin::storage_api::contract::admin::StorageAdminApi;
 use crate::admin::storage_api::contract::bucket::{BucketOperations, BucketOptions};
 use crate::admin::storage_api::data_usage::{
-    apply_bucket_usage_memory_overlay, load_data_usage_from_backend, refresh_bucket_usage_from_object_layer,
-    replace_bucket_usage_memory_from_info,
+    apply_bucket_usage_memory_overlay, load_data_usage_from_backend, replace_bucket_usage_memory_from_info,
 };
 use crate::admin::storage_api::metadata_sys;
 use crate::auth::get_condition_values;
@@ -38,7 +37,6 @@ use s3s::{Body, S3Error, S3ErrorCode, S3Request, S3Response, S3Result, s3_error}
 use serde::Serialize;
 use std::collections::HashMap;
 use std::sync::Arc;
-use tracing::debug;
 
 #[allow(dead_code)]
 #[derive(Debug, Serialize, Default)]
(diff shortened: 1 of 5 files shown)


kiliant:
Thank you for the CLA notice.

I want to be transparent that I have read the linked Individual CLA, but I cannot sign it in its current form. In particular, it appears to require an irrevocable transfer of ownership of my contributions, without an explicit licence back to me and without clearly identifying the legal entity receiving those rights.

I have opened rustfs/cla#3 with the specific questions and concerns so that the CLA discussion does not take over this technical pull request:

https://github.com/rustfs/cla/issues/3

I would very much love to contribute to RustFS through this PR and through further work in the future. However, I cannot do so under the current CLA terms. I hope the agreement can be clarified or revised so that external contributors can participate while retaining ownership of their work.

cxymds:
Thanks for putting this together and for narrowing the scope. I reviewed the current patch against the latest `main`.

This PR has now been superseded by #4980, which implements the same core behavior—both admin endpoints serve the persisted scanner snapshot plus the in-memory overlay without entering a request-path live version walk—and also addresses the remaining merge requirements:

- an endpoint-level revert-detector integration test proving that repeated `datausageinfo` requests do not call `compute_bucket_usage`;
- explicit unknown-stat semantics for buckets without snapshot coverage, avoiding misleading `0` object/size values;
- removal of the per-request global usage-cache rebuild;
- cached snapshot loading and subsequent snapshot/convergence follow-ups.

#4953 is currently conflicting with `main`, and rebasing it would duplicate the already-merged implementation. I recommend closing this PR as superseded by #4980 rather than doing further work on this branch.

Thank you again for the original report, the production validation on the 440k-object bucket, and the careful clarification around the separate `channel closed` behavior.

Issue #4859, opened 2026-07-15
restore_transitioned_object runs without a namespace lock; concurrent restores race


overtrue:
## Summary

`restore_transitioned_object` runs with **no namespace lock**: the guard that would take a write lock on the object is entirely commented out (`crates/ecstore/src/set_disk/ops/object.rs`, the block above the fileinfo load in `restore_transitioned_object`, ~2360-2371; the same pattern is commented at ~2189 and ~2529). The only concurrency control on the restore path is the API-layer check of `restore_ongoing` in `execute_restore_object` (`rustfs/src/app/object_usecase.rs:6202`), which is a read-then-write on object metadata with no atomicity.

## Expected semantics (written down per backlog#1148 ilm-8 acceptance)

1. At most one restore copy-back per (bucket, object, version) is in flight; a concurrent second `POST ?restore` returns `ErrObjectRestoreAlreadyInProgress` (currently only best-effort).
2. A restore never interleaves destructively with lifecycle expiry of the same version: `DeleteRestoredAction` cleanup and an in-flight copy-back must not produce a half-written local copy that is then served as restored.
3. Re-restore of an already-restored object only extends the expiry metadata (this part exists: the `already_restored` early path).

## Race windows in the current code

- Two concurrent `POST ?restore` can both observe `restore_ongoing == false` and both spawn `restore_transitioned_object`: duplicated remote reads, and two concurrent `put_object`/multipart writes to the same key. With identical `days` the result converges (same bytes), but the multipart branch creates two uploads and completes both — last-writer-wins with avoidable churn.
- A restore copy-back racing `expire_transitioned_object` (`DeleteRestoredAction`) on the same version: cleanup strips restore headers while the copy-back is still writing, after which the fresh local write re-appears with `ongoing=false` metadata from `put_restore_opts` — the restored copy can outlive the cleanup that was supposed to remove it, extending local disk usage past the intended expiry.

## Suggested direction

Reinstate the namespace write lock around `restore_transitioned_object` (the commented block), or make the API-layer ongoing check an atomic compare-and-set on the restore metadata. Either restores the "single writer per version" invariant the rest of the lifecycle paths rely on.

Found while writing the ilm-8 restore-chain integration tests (backlog#1148); those tests exercise the serialized paths only and do not depend on this fix.

PR #4956, opened 2026-07-17
fix(ilm): replace whole-copy-back restore lock with accept-path CAS

overtrue:
## Related Issues

Implements the design decision for rustfs/backlog#1304 (concurrent RestoreObject semantics: #4877 write-lock serialization vs fast-reject + non-blocking HEAD). Follow-up to #4877 and #4886; restores backlog#1148 ilm-8 criterion 1. Related: #4859.

## Summary of Changes

**Decision.** Between the two directions listed in #4859/backlog#1304, this implements the atomic compare-and-set on the restore `ongoing` flag instead of keeping #4877's object write lock across the whole tier copy-back:

- A restore accept (POST `?restore`) is serialized per object by a new purpose-scoped **restore-accept guard** (`ECStore::acquire_restore_accept_guard`, opaque `RestoreAcceptGuard` with an `is_lock_lost()` fence), held only across the accept's read-check-write of the `x-amz-restore` status. Two concurrent accepts can no longer both observe `ongoing=false` and double-start the copy-back (#4859 race 1).
- The tier copy-back runs **without** the object lock (`handle_restore_transitioned_object` no longer locks), so HEAD/`get_object_info` during a restore stay non-blocking and immediately observe `ongoing-request="true"`, and a second POST `?restore` is fast-rejected — AWS-conformant and what backlog#1148 ilm-8 criterion 1 specifies. Under #4877 those reads blocked for the entire copy-back, and the copy-back self-deadlocked on the inner commit locks until #4886's `no_lock` propagation.
- Torn-write protection stays where it already lives: the inner `put_object`/`complete_multipart_upload` commit phases take the object write lock themselves. `DeleteRestoredAction` cannot fire mid-copy-back because the ongoing header serializes without an expiry date and the lifecycle evaluator only fires on a present, past expiry.

**API-layer changes** (`execute_restore_object`):
- Non-SELECT accepts hold the guard across `get_object_info` → transitioned check → ongoing check → metadata write, with `no_lock` inside the scope; the guard is dropped before the copy-back spawn. SELECT-type restores skip the ongoing check and the metadata write, so they keep the plain read-locked path (the guard would protect nothing there).
- The CAS write is fenced on `is_lock_lost()` (lock-service degradation → 503 instead of a possibly-duplicate commit).
- The concurrent-restore rejection is now **409 `RestoreAlreadyInProgress`** (AWS-conformant); the previous `Custom("ErrObjectRestoreAlreadyInProgress")` serialized as a retryable 500, so SDK clients retried the conflict. Guard contention maps to **503 SlowDown**.
- The spawned copy-back pins the **resolved** version instead of the raw request version: with a versionless request on a versioned bucket, a PUT landing between accept and copy-back would re-resolve "latest" to the new version, fail (not transitioned), and strand the flagged version at `ongoing=true` forever.

**Tests:**
- `restore_object_usecase_reports_ongoing_conflict_and_completion` is re-enabled: removed from the ILM serial lane exclusion list (added by #4886) with its assertions updated to the final semantics (409 code). It passes.
- New integration test `restore_object_usecase_accepts_exactly_one_of_two_concurrent_restores`: two truly concurrent POSTs → exactly one acceptance, one 409, and exactly one tier copy-back GET. This is the revert-detector for the CAS itself — without the guard, both accepts pass and the tier records two GETs.
- The #4877 unit test asserting entry-blocking semantics is removed (those semantics are deliberately reversed here); a new unit test pins that the accept guard mutually excludes concurrent accepts.

**Adversarial validation** ran at high-risk tier (all six roles). Fixed from findings: SELECT-path guard bypass, resolved-version pinning, lock-lost fence, 409/503 wire codes, and the concurrent-CAS coverage gap. Rebutted: the failing-accept path (non-transitioned object) now briefly holds a write lock where baseline held a read lock — bounded to one quorum metadata read (single-digit ms), requires `s3:RestoreObject` on the object, and per-client rate limiting (#4895) is the systemic answer to hammering. Accepted residuals are pre-existing and filed as follow-ups on the backlog (referenced from the backlog#1304 close-out comment): delete-vs-copy-back resurrection window (pre-#4877 parity; #4877's paper protection never worked due to the self-deadlock), scanner `DeleteRestoredAction` stale-snapshot TOCTOU, stale `ongoing=true` has no reaper, failure-path `update_restore_metadata` metadata wipe, SELECT restore writing back to the source object, and the RFC3339-vs-RFC1123 restore-header MinIO interop gap.

## Verification

- `cargo nextest run -j1 --run-ignored ignored-only -p rustfs -E 'test(restore_object_usecase_reports_ongoing_conflict_and_completion) or test(restore_object_usecase_accepts_exactly_one_of_two_concurrent_restores)'` — both pass (1.9s / 2.4s) on the merged tree.
- `cargo nextest run -p rustfs-ecstore -E 'test(restore_accept_guard_serializes_concurrent_accepts) or test(prelocked_read_request_does_not_mark_metadata_cache_safe) or test(transition_restore_pool_opts)'` — 4 passed.
- `cargo clippy -p rustfs-ecstore -p rustfs --all-targets` — clean.
- `make pre-commit` — passed.
- `make pre-pr` — passed.

## Impact

- S3 API: a second `POST ?restore` while a restore is in flight now returns **409 `RestoreAlreadyInProgress`** (was a retryable 500 with a custom code); HEAD/GET of an object mid-restore no longer block (under #4877 they failed with lock timeouts for the whole copy-back). Restore-accept contention returns 503 SlowDown.
- Rolling upgrade: an old (#4877) node and a new node can both accept the same restore during a mixed-version window (the old node's accept has no guard); the copy-backs serialize on the commit lock and converge, but a failed new-node copy-back behind the old node's long-held lock can strand `ongoing=true` (see the reaper follow-up). Transient to the upgrade window.
- Public crate API: additive — `RestoreAcceptGuard` and `acquire_restore_accept_guard` on `rustfs-ecstore` (opaque, deliberately not a general lock API).

## Additional Notes

The removed whole-copy-back lock never provided its paper protection in practice: the inner commit re-acquired the same non-reentrant object lock, so every restore copy-back under #4877 timed out (fixed as a self-deadlock by #4886). This PR replaces that design rather than narrowing it, per the backlog#1304 decision, and converges with MinIO's restore flow (no namespace lock across the copy-back; accept-path CAS is a safety improvement over MinIO, which lives with the accept TOCTOU).


overtrue:
CI status: all lanes green — including **ILM Integration (serial)**, which now runs the re-enabled `restore_object_usecase_reports_ongoing_conflict_and_completion` plus the new concurrent-CAS test — except **End-to-End Tests**, which fails on the two `content_encoding_test` aws-chunked cases. That failure is pre-existing on main (same two tests fail on main's latest run, e.g. https://github.com/rustfs/rustfs/actions/runs/29551214728, red since fc7d46b6c / #4928) and is tracked in rustfs/backlog#1336. It is unrelated to this PR's restore/ILM surface.

PR #4957, opened 2026-07-17
feat(server): add opt-in global connection cap to the S3 accept loop

overtrue:
## Related Issues

Follow-up to rustfs/backlog#1191 (optional sub-item: global connection cap for the S3 accept loop).

## Summary of Changes

The main S3 accept loop spawns one task per accepted socket with no global bound, so a connection flood can exhaust file descriptors and memory and take down existing traffic with it. This adds an opt-in, default-off cap:

- **`RUSTFS_API_MAX_CONNECTIONS`** (new constant in `crates/config/src/constants/api.rs`, default `0` = unlimited): maximum concurrently served connections on the main listener. With the default, no semaphore is constructed and the accept loop is unchanged.
- When set, the accept loop acquires an owned semaphore permit **before** calling `accept()`. At saturation the loop simply stops accepting and lets the kernel backlog absorb bursts — TCP-native backpressure instead of accept-then-close churn (dropped sockets look like server bugs to clients and burn handshakes that go nowhere). The permit moves into the per-connection task and is released by RAII when the connection ends, on every exit path including TLS-handshake failures.
- Saturation is observable: `rustfs_http_server_connection_cap_saturated_total` increments each time the loop has to wait for a permit, a `connection_cap_state` startup event reports the configured cap, and a throttled `debug` event fires per wait.
- Shutdown stays responsive while parked on the semaphore (the wait is raced against the shutdown signal).

## Verification

- `cargo test -p e2e_test connection_cap -- --test-threads=1` — real-binary e2e, both passing:
  - *permit release*: with cap 2, ten sequential `Connection: close` requests all succeed — a leaked permit would strand request 3+ in the backlog and time out;
  - *actual bounding*: two connections stalled mid-request-head saturate cap 2 (hyper pins them against the 75s header timeout, making occupancy deterministic); a third connection with a complete request gets no response within 3s, then is accepted and served within 10s once the stalled connections close and release their permits.
- `cargo clippy --all-targets -p rustfs -p rustfs-config -p e2e_test` — zero warnings; `make pre-commit` passes.
- Note for reviewers running the e2e locally: build the binary first (`cargo build -p rustfs --bin rustfs`) — the harness deliberately reuses a stale `target/debug/rustfs` when invoked inside `cargo test`, which would exercise a binary without this change.

## Impact

- Default behavior unchanged: with `RUSTFS_API_MAX_CONNECTIONS` unset or `0`, no semaphore exists and the accept loop behaves exactly as before.
- The cap covers **everything on the main listener** — S3, admin, console, and internode gRPC. Operators must size it well above peer-node count plus expected client concurrency; the constant's docs carry this warning. A saturated cap delays *new* internode connections (existing pooled ones are unaffected), which is the deliberate trade-off against unbounded FD/memory exhaustion killing the whole process.
- No `Retry-After`-style signaling at this layer (rejection happens below HTTP); saturation shows up as connection queuing, which every S3 client already handles.

### DIFF
--- a/crates/config/src/constants/api.rs
+++ b/crates/config/src/constants/api.rs
@@ -73,3 +73,20 @@ pub const ENV_API_RATE_LIMIT_BUCKET_BURST: &str = "RUSTFS_API_RATE_LIMIT_BUCKET_
 
 /// Default for `RUSTFS_API_RATE_LIMIT_BUCKET_BURST` (`0` = same as bucket RPM).
 pub const DEFAULT_API_RATE_LIMIT_BUCKET_BURST: u32 = 0;
+
+/// Maximum concurrently served connections on the main API listener.
+///
+/// `0` (the default) means unlimited. When set, the accept loop stops
+/// accepting once the cap is reached and lets the kernel backlog absorb
+/// bursts, releasing capacity as connections close. This bounds file
+/// descriptor and memory usage under a connection flood.
+///
+/// The cap covers everything on the main listener — S3, admin, console,
+/// and internode gRPC — so size it well above peer-node count plus the
+/// expected client concurrency.
+/// Environment variable: RUSTFS_API_MAX_CONNECTIONS
+/// Example: RUSTFS_API_MAX_CONNECTIONS=10000
+pub const ENV_API_MAX_CONNECTIONS: &str = "RUSTFS_API_MAX_CONNECTIONS";
+
+/// Default for `RUSTFS_API_MAX_CONNECTIONS` (`0` = unlimited).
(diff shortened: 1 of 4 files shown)


overtrue:
The failing **End-to-End Tests** lane is the same pre-existing break on main already noted on #4949: the only two failing cases are `content_encoding_test` aws-chunked PUTs (400 `UnexpectedContent`), reproduced by main's own CI run [29546495804](https://github.com/rustfs/rustfs/actions/runs/29546495804) before this branch existed. The `connection_cap_test` cases added here are not among the failures (verified locally: both pass against the real binary). Will rebase/rerun once main is fixed.

PR #4960, opened 2026-07-17
fix(s3): distinguish streaming aws-chunked uploads

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Fixes an S3 compatibility regression where ordinary PUT requests carrying `aws-chunked` in `Content-Encoding` were rejected as malformed streaming uploads.

- Identifies actual SigV4 streaming uploads from the authentication mode derived from `x-amz-content-sha256`.
- Continues to require `x-amz-decoded-content-length` for signed and trailer streaming uploads.
- Uses normal `Content-Length` for single-chunk PUT requests that only carry the request-side `aws-chunked` encoding token.
- Preserves authoritative decoded-size quota admission for real streaming uploads.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs --lib aws_chunked -- --nocapture`
- `cargo test -p rustfs --lib authoritative_size_ -- --nocapture`
- `cargo test -p e2e_test content_encoding_test::tests::test_content_encoding_aws_chunked -- --nocapture`
- `cargo test -p e2e_test archive_download_integrity_test::tests::test_archive_put_with_aws_chunked -- --nocapture`
- `TESTEXPR=test_object_content_encoding_aws_chunked ./scripts/s3-tests/run.sh`

## Impact

Restores AWS-compatible handling of `aws-chunked` as a request-side content-encoding token for ordinary PUT requests while retaining fail-closed decoded-size validation for true streaming uploads.

## Additional Notes

The regression was introduced when quota admission treated the `Content-Encoding` token alone as proof of streaming. The s3s signature layer already rejects genuine streaming requests that omit `x-amz-decoded-content-length`.


overtrue:
Thank you for diagnosing this and for the fix! You correctly identified the root cause — that actual chunk framing is signalled by a STREAMING-* x-amz-content-sha256 rather than by the declared Content-Encoding. The same conclusion landed in #4968 (now merged), which also covers the streaming-without-Content-Encoding edge case and adds the quota regression e2e required by the tracking issue, so I'm closing this PR in its favor. Your report and analysis genuinely helped confirm the diagnosis — much appreciated, and we'd be glad to see more contributions from you.

Issue #1857, opened 2026-02-17
SigV4 Content-Encoding header incorrectly persisted in violation of spec.


RazeLighter777:
**Describe the bug**
When uploading objects using an AWS SigV4 streaming client (e.g. nix copy --to s3://…), the server persists Content-Encoding: aws-chunked as object metadata.

This is incorrect. aws-chunked is a request-side transfer encoding used for streaming uploads and must not be stored or returned as object Content-Encoding.

Per [AWS SigV4](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html):

"Amazon S3 stores the resulting object without the aws-chunked value in the content-encoding header. If aws-chunked is the only value that you pass in the content-encoding header, S3 considers the content-encoding header empty and does not return this header when your retrieve the object."

This breaks strictly spec-compliant SigV4 clients, such as the nix package manager, which does not expect that header. Here's an example of an error message from nix:
```
error: failed to get libarchive filter by name: No such filter 'aws-chunked'
error:
       … during download of 'https://s3.prizrak.me/nix/hiyn41h4l60s3m0gvd5y2h0x2s2z44b6.narinfo'
```

**To Reproduce**
1. Upload an object using a SigV4 streaming client (for example):
`nix copy --to s3://<bucket>?endpoint=<endpoint>&region=<region> <path>`
3. Perform a request to the uploaded object: `curl -I https://<host>/<bucket>/<object>` 
4. Observe the response headers. Note the presence of the aws-chunked content-encoding header. 
```
HTTP/2 200 
date: Tue, 17 Feb 2026 13:24:16 GMT
content-type: text/x-nix-narinfo
content-length: 591
content-encoding: aws-chunked
... [omitted for brevity]
```

**Expected behavior**
The header should not be present in returned responses.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Talos Linux (kubernetes)
 - Browser: Not using
 - Version: 1.0.0-alpha.81

**Smartphone (please complete the following information):**
 - Not applicable

**Additional context**

I respect the time of the maintainers, thank you for your hard work maintaining this project.
 
Please let me know if you need help to reproduce this issue or want additional context. I'd be more than happy to open a pull request if this is indeed legitimate issue. It seems like a small fix, that should involve just sanitizing the headers properly.

loverustfs:
Hello, thank you for your feedback.

Please upgrade to the latest alpha.83 before testing.

[...]

PR #4964, opened 2026-07-17
fix(quota): admit declared-only aws-chunked PUT against the wire Content-Length

overtrue:
## Problem

The E2E CI job on `main` has been red since [PR #4928](https://github.com/rustfs/rustfs/pull/4928) (`fix(quota): admit hard quota against authoritative decoded size…`), failing two `content_encoding_test` cases with `UnexpectedContent` (400) on PUT:

- `test_content_encoding_aws_chunked_not_returned_issue_1857`
- `test_content_encoding_aws_chunked_with_effective_encoding_roundtrip`

Failing run: https://github.com/rustfs/rustfs/actions/runs/29551214728/job/87795817113

## Root cause

PR #4928 moved quota admission after resolving an *authoritative* object size and, in `resolve_put_object_authoritative_size`, made `x-amz-decoded-content-length` **mandatory** whenever `Content-Encoding` declares `aws-chunked` — otherwise it returns `UnexpectedContent`.

But `Content-Encoding: aws-chunked` is *also* sent by SigV4 clients (issue #1857) on a plain, non-framed signed payload that carries **no** decoded-length header. Those PUTs are legitimate — the body is not chunk-wrapped, so the wire `Content-Length` is the true object size — yet they were now rejected with 400.

The declared `Content-Encoding` header alone does not tell you whether the body is actually chunk-framed. The real framing signal is the SigV4 streaming payload hash: `x-amz-content-sha256: STREAMING-*`.

## Fix

Gate the "decoded length is mandatory" rule on `request_body_is_streaming_framed` (checks `x-amz-content-sha256` for a `STREAMING-` prefix) rather than on `Content-Encoding` alone:

- **aws-chunked declared + genuinely framed** (`STREAMING-*` hash), no decoded length → still rejected (`UnexpectedContent`), unchanged.
- **aws-chunked declared but not framed** (digest / `UNSIGNED-PAYLOAD` / no hash), no decoded length → fall back to the authoritative wire `Content-Length`.
- Negative / unknown lengths → still rejected, unchanged.
- Genuine framing with a decoded length → still uses the decoded length, unchanged.

## Tests

- New unit tests: declared-only aws-chunked with a plain payload falls back to the wire length; all three `STREAMING-*` framing variants without a decoded length are still rejected; no-length case still rejected.
- The three `content_encoding_test` e2e cases pass locally against the freshly built binary (3 passed / 0 failed).

Refs: https://github.com/rustfs/backlog/issues/1311, #1857

### DIFF
--- a/rustfs/src/app/object_usecase.rs
+++ b/rustfs/src/app/object_usecase.rs
@@ -304,11 +304,16 @@ fn range_to_http_range_spec(range: Range) -> S3Result<HTTPRangeSpec> {
 
 /// Resolve the authoritative object length that bucket-quota admission (and downstream sizing) must use.
 ///
-/// For an `aws-chunked` upload the request body is wrapped in chunk framing, so the wire `Content-Length` overcounts the actual object and must never be admitted; the decoded length (`x-amz-decoded-content-length`) is the only authoritative size and is therefore required — a chunked request without it is rejected rather than silently falling back to the framed wire length. For a non-chunked request the raw `Content-Length` is authoritative, with an explicit decoded length as the fallback when the S3 layer only surfaced the latter. A negative or otherwise unknown length is rejected so it can never be reinterpreted as an enormous unsigned size downstream.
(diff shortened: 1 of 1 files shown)


overtrue:
Superseded by #4968, which has been merged: same STREAMING-* framing detection, plus the streaming-without-Content-Encoding decoded-length fix and the aws-chunked × quota e2e regression test required by backlog#1336's acceptance criteria. Closing this one.

PR #4971, opened 2026-07-17
fix(put): admit bare aws-chunked uploads via wire Content-Length

### LINKED ISSUE #4962: PUT with Content-Encoding: aws-chunked but no x-amz-decoded-content-length regressed to 400 UnexpectedContent
Since fc7d46b6c (#4928, quota authoritative-size admission), a PutObject that carries `Content-Encoding: aws-chunked` (alone or in a list, e.g. `aws-chunked,gzip`) **without** an `x-amz-decoded-content-length` header is rejected with `400 UnexpectedContent`.

This combination is legitimate: clients can set `Content-Encoding: aws-chunked` on a normal fully-signed upload whose body is not actually chunk-framed — AWS accepts this and strips the `aws-chunked` token (the behavior characterized by #1857 and #2475). Only genuine SigV4 streaming uploads carry chunk framing, and those always send `x-amz-decoded-content-length`. `resolve_put_object_authoritative_size` decides "chunked" from the header alone (`request_uses_aws_chunked`), so the false-chunked case now hits the new `(true, None, _) => UnexpectedContent` arm; before #4928 it fell back to the wire `Content-Length`, which for a non-framed body is exactly the object length.

Effect on CI: 5 e2e tests fail deterministically on `main` since the first post-merge run (29522890580, 2026-07-16) — both the PR `End-to-End Tests` job and the `e2e-full` merge gate — so every open PR currently inherits a red E2E check (first seen on #4948, which is unrelated):

```
content_encoding_test::tests::test_content_encoding_aws_chunked_not_returned_issue_1857
content_encoding_test::tests::test_content_encoding_aws_chunked_with_effective_encoding_roundtrip
archive_download_integrity_test::tests::test_archive_put_with_aws_chunked_allowed_when_strict_mode_enabled
archive_download_integrity_test::tests::test_archive_put_with_aws_chunked_and_effective_encoding_roundtrips_by_default
archive_download_integrity_test::tests::test_archive_put_with_aws_chunked_does_not_persist_content_encoding_by_default
```

All fail the same way: `PUT failed: ... code: Some("UnexpectedContent") ... StatusCode(400)`.

Proposed fix: in `resolve_put_object_authoritative_size`, when the aws-chunked header is present but `x-amz-decoded-content-length` is absent, fall back to the wire `Content-Length` when it exists (the pre-#4928 behavior for this case) and keep rejecting only when no length is known at all. The quota-admission intent of #4928 is preserved: a genuinely framed streaming upload always carries the decoded length, so admission still never uses a framed wire length; the negative/unknown-length rejections stay.

overtrue:
## Related Issues

Fixes #4962.

## Summary of Changes

#4928 made `resolve_put_object_authoritative_size` reject any PUT whose `Content-Encoding` contains `aws-chunked` but lacks `x-amz-decoded-content-length`, on the assumption that the header implies a chunk-framed body whose wire `Content-Length` overcounts. That assumption is wrong for the case #1857/#2475 characterized: clients legitimately send the bare `aws-chunked` token on a normal fully-signed upload whose body is not framed at all — AWS accepts this and strips the token. Genuine SigV4 streaming uploads always carry `x-amz-decoded-content-length`, so for a bare-token request the wire `Content-Length` is exactly the object length. Since fc7d46b6c these requests get `400 UnexpectedContent`, and 5 aws-chunked e2e tests (`content_encoding_test`, `archive_download_integrity_test`) fail deterministically on `main`, turning the E2E check red for every open PR.

This change restores the wire-`Content-Length` fallback for the bare-token case while keeping every protection #4928 introduced: a decoded length still wins whenever it is present (framed wire lengths are still never admitted for quota), negative lengths are still rejected, and a request with no length information at all is still rejected. The previously-rejecting unit test now asserts admission of the bare-token + wire-length case, with a new companion test keeping the "no length at all" rejection covered.

## Verification

- `make pre-commit` — passed.
- `cargo test -p rustfs --lib authoritative_size` — 8/8 passed.
- The 5 regressed e2e tests pass locally against the fixed binary: `cargo test -p e2e_test --lib -- --test-threads=1 --exact content_encoding_test::tests::test_content_encoding_aws_chunked_not_returned_issue_1857 content_encoding_test::tests::test_content_encoding_aws_chunked_with_effective_encoding_roundtrip archive_download_integrity_test::tests::test_archive_put_with_aws_chunked_allowed_when_strict_mode_enabled archive_download_integrity_test::tests::test_archive_put_with_aws_chunked_and_effective_encoding_roundtrips_by_default archive_download_integrity_test::tests::test_archive_put_with_aws_chunked_does_not_persist_content_encoding_by_default` → `5 passed; 0 failed`.

## Impact

Restores S3 compatibility for uploads that send `Content-Encoding: aws-chunked` (alone or in a list) without an `x-amz-decoded-content-length` header — broken on `main` since fc7d46b6c. Quota admission semantics from #4928 are unchanged for genuinely framed streaming uploads. No API or configuration changes.

## Additional Notes

Un-breaks the `End-to-End Tests` CI job and the `e2e-full` merge gate for all open PRs (first observed on #4948, which is unrelated).

### DIFF
--- a/rustfs/src/app/object_usecase.rs
+++ b/rustfs/src/app/object_usecase.rs
@@ -304,15 +304,14 @@ fn range_to_http_range_spec(range: Range) -> S3Result<HTTPRangeSpec> {
 
 /// Resolve the authoritative object length that bucket-quota admission (and downstream sizing) must use.
 ///
-/// For an `aws-chunked` upload the request body is wrapped in chunk framing, so the wire `Content-Length` overcounts the actual object and must never be admitted; the decoded length (`x-amz-decoded-content-length`) is the only authoritative size and is therefore required — a chunked request without it is rejected rather than silently falling back to the framed wire length. For a non-chunked request the raw `Content-Length` is authoritative, with an explicit decoded length as the fallback when the S3 layer only surfaced the latter. A negative or otherwise unknown length is rejected so it can never be reinterpreted as an enormous unsigned size downstream.
+/// When the decoded length (`x-amz-decoded-content-length`) is present it wins: the body of a genuine SigV4 streaming upload is wrapped in chunk framing, so the wire `Content-Length` overcounts the actual object and must never be admitted. An `aws-chunked` content-encoding *without* a decoded length, however, does not mean a framed body: clients legitimately send `Content-Encoding: aws-chunked` on a normal fully-signed upload (see #1857/#2475 — AWS accepts and strips the token), and genuine streaming always carries the decoded length; for those requests the wire `Content-Length` is exactly the object length and remains authoritative (rustfs#4962). A negative or otherwise unknown length is rejected so it can never be reinterpreted as an enormous unsigned size downstream.
 fn resolve_put_object_authoritative_size(headers: &HeaderMap, content_length: Option<i64>) -> S3Result<i64> {
     let decoded_content_length = decoded_content_length_from_headers(headers)?;
-    let size = match (request_uses_aws_chunked(headers), decoded_content_length, content_length) {
-        (true, Some(decoded), _) => decoded,
-        (true, None, _) => return Err(s3_error!(UnexpectedContent)),
-        (false, _, Some(raw)) => raw,
-        (false, Some(decoded), None) => decoded,
-        (false, None, None) => return Err(s3_error!(UnexpectedContent)),
+    let size = match (decoded_content_length, content_length) {
+        (Some(decoded), _) if request_uses_aws_chunked(headers) => decoded,
+        (_, Some(raw)) => raw,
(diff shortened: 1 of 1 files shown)


overtrue:
Superseded by #4968 (a9e3613cf), which landed first with a finer-grained fix: framing is detected via the `STREAMING-*` `x-amz-content-sha256` (`request_body_is_aws_chunked_framed`), bare declared `aws-chunked` falls back to the wire `Content-Length`, and a genuinely framed body without a decoded length is still rejected instead of falling back. main's E2E jobs (including the full merge gate) are green on that commit, covering the #4962 scenarios. Closing this coarser variant.

PR #4977, opened 2026-07-17
fix(storage): resolve erasure parity per pool

### LINKED ISSUE #4801: TooFewDataShards panic when scaling creates pools smaller than inherited parity
## Summary

When a RustFS cluster is scaled from 4 to 6 replicas, the second erasure pool {4,5} has only 2 drives but inherits `common_parity_drives = 2` from the first pool (4 drives). This results in `data_blocks = 2 - 2 = 0`, causing a `TooFewDataShards` panic in the Reed-Solomon encoder.

## Root Cause

In `crates/ecstore/src/store/init.rs` (lines 225-332), `common_parity_drives` is computed from the first pool only:

```rust
let mut common_parity_drives = 0;
for (i, pool_eps) in endpoint_pools.as_ref().iter().enumerate() {
    if common_parity_drives == 0 {  // only runs for pool 0
        let parity_drives = ec_drives_no_config(pool_eps.drives_per_set)?;
        common_parity_drives = parity_drives;
    }
    Sets::new_with_instance_ctx(disks, pool_eps, &fm, i, common_parity_drives, ...).await?;
}
```

For pool 0 with 4 drives: `default_parity_count(4) = 2`, so `common_parity_drives = 2`.
Pool 1 with 2 drives inherits parity = 2: `data_blocks = 2 - 2 = 0`.

The panic occurs in `Erasure::new_with_options` (erasure.rs:477) via `.expect()` on `ReedSolomonEncoder::new(0, 2)`.

## Observed Behavior

1. Worker thread panics with `TooFewDataShards` ~40 seconds after startup
2. Main process continues in degraded state (worker panic does not crash the process)
3. Data plane partially hangs: some objects become permanently unreadable
4. Process eventually OOMKilled after ~45 minutes of degraded operation
5. After restart, specific objects remain permanently unavailable

## Environment

- RustFS 1.0.0-beta.8
- Replicas history: 4,6 (scaled from 4 to 6)
- Kubernetes with KubeBlocks managing StatefulSet

## Affected Scale Patterns

Any `replicas_history` where pool sizes differ and a smaller pool has `size <= first_pool_parity`:
- 4 to 5: pool2=1, parity=2 (panic)
- 4 to 6: pool2=2, parity=2 (panic)
- 4 to 7: pool2=3, parity=2 (OK, but over-provisioned parity)
- 6 to 8: pool2=2, parity=3 (panic)

## Additional Issue: Unguarded Erasure Construction

Four call sites construct `Erasure` objects without `has_valid_dimensions()` check:

| File | Line | Function |
|------|------|----------|
| set_disk/ops/object.rs | 276 | inline GET fast-path |
| set_disk/ops/object.rs | 744 | PUT object |
| set_disk/ops/multipart.rs | 357 | multipart upload-part |
| set_disk/ops/heal.rs | 166 | heal_object |

Only `set_disk/read.rs` has guards. The comment at erasure.rs:802-808 explicitly notes this requirement.

## Suggested Fix

1. Recompute parity per pool using each pools own `drives_per_set`, or clamp: `min(common_parity_drives, pool_size - 1)`
2. Add `has_valid_dimensions()` guard to all four unprotected `Erasure::new` call sites
3. Replace `.expect()` in `Erasure::new_with_options` with proper error propagation

cxymds:
## Related Issues

Fixes #4801.

## Summary of Changes

- Resolve STANDARD and reduced-redundancy parity independently for every pool instead of reusing the first pool's scalar value. For a heterogeneous four-disk/two-disk topology, automatic parity now resolves to `2+2` and `1+1` layouts rather than producing zero data shards in the second pool.
- Publish one immutable per-pool storage-class snapshot and require write paths to resolve a validated `WriteLayout` against both the pool index and drive count. Regular PUT, multipart, tiered writes, healing, rebalance, readiness, backend reporting, and metrics consume the same snapshot semantics.
- Replace panic-prone erasure construction and unchecked shard arithmetic with typed failures, and harden decoded metadata validation for payloads, canonical delete markers, purge-pending objects, and tier-free cleanup records.
- Prepare and validate storage configuration before persistence, then atomically publish the exact prepared snapshot only after persistence succeeds. Invalid later pools and failed persistence leave the active snapshot unchanged.
- Extend architecture guardrails and regression coverage so production use of panic-prone erasure constructors and malformed `cfg(test)` exclusions cannot silently return.

This is a high-risk storage change, so the final diff was attacked by all required adversarial roles:

- Correctness: PASS — heterogeneous pool layouts, invalid geometry, quorum boundaries, ordinary PUT, and multipart paths have revert-sensitive regression coverage.
- Simplicity: PASS — parity resolution is centralized in the storage-class snapshot and `WriteLayout`; no parallel configuration model or speculative feature was added.
- Security: PASS — malformed disk/RPC metadata and forged delete or tier-free shapes fail closed before quorum or destructive actions.
- Concurrency/durability: PASS — readers capture one immutable snapshot, and admin changes follow prepare, persist, then publish ordering without partial publication.
- Compatibility: PASS — no `xl.meta` or RPC wire-format migration is introduced; legacy homogeneous scalar reporting remains available when one truthful scalar exists.
- Performance: PASS — hot paths perform an `ArcSwap` snapshot load and indexed lookup without new disk I/O, network I/O, or long-held locks.
- Test coverage: PASS — each claimed behavior has a regression test that fails when the corresponding protection is reverted.

## Verification

- `make pre-pr` — pending; the latest run was stopped after targeted Clippy passed
- `make pre-commit` — passed on the implementation series before the final validation-only follow-up
- `cargo fmt --all --check` — passed at current HEAD
- `cargo clippy -p rustfs-filemeta -p rustfs-ecstore -p rustfs --all-targets -- -D warnings` — passed at current HEAD
- `cargo test -p rustfs-filemeta --lib --no-fail-fast` — passed at current HEAD (`188 passed`)
- `cargo test -p rustfs-ecstore --lib --no-fail-fast -- --test-threads=1` — passed before the final validation-only follow-up (`2626 passed`, `1 ignored`); current-HEAD rerun pending
- `cargo test -p rustfs --lib admin::service::config::tests --no-fail-fast -- --test-threads=1` — passed before the final validation-only follow-up (`18 passed`); current-HEAD rerun pending

## Impact

- Automatic storage-class parity now works for heterogeneous pool widths and fails closed when topology and snapshot data disagree.
- Backend, readiness, and metrics surfaces report per-pool layouts; scalar fields are retained only when a single value truthfully represents every pool.
- Existing object metadata and persisted scalar storage-class configuration remain compatible; the per-pool vectors are runtime-derived and are not a new on-disk or wire format.
- No data migration is required.

## Additional Notes

- MinIO's common/scalar parity behavior was used as a reference, but it was not copied because it does not solve heterogeneous automatic parity layouts.
- During a mixed-version rolling upgrade, older nodes still understand only scalar parity. Configure explicit `EC:1` or prevent old-node writes until every node runs the new code; enable heterogeneous automatic parity only after the rollout completes.
- Before rolling back, set explicit `EC:1` and allow the configuration to converge before reverting binaries.
- Cross-node serialization between simultaneous independent configuration writers is unchanged and remains a separate follow-up concern; this PR guarantees local prepare, persistence, and publication ordering.

---


crates/filemeta/src/fileinfo.rs
```diff
@@ -304,27 +380,267 @@ impl FileInfo {
...
+            erasure_layout,
+        })
+    }
+
+    /// Validate metadata returned by a disk or peer as a strict tagged union.
+    /// Payload entries require complete erasure geometry, including
+    /// purge-pending object versions whose replication state sets `deleted`.
+    /// Only entries that are not valid payloads may fall back to the canonical
+    /// delete-marker shape, so `deleted` cannot bypass payload validation.
+    pub fn validate_for_metadata_read(&self) -> Result<()> {
```
overtrue, line 553: **Potential compatibility blocker: fail-closed decode validation can make existing objects unreadable.**

`validate_for_metadata_read` accepts a `FileInfo` only if it is a fully-canonical erasure payload *or* a fully-canonical delete marker, and it now gates the decode boundary on the normal read path (`load_file_info_versions_exact`, `pick_latest_quorum_files_info`, `rename_data`) and on inter-node RPC (`RemoteDisk::read_file_info`, `BatchReadVersion`). It enforces invariants `main` never checked on read — e.g. `distribution` must be an exact permutation of `1..=n` with `len == n`, `erasure.index ∈ 1..=n`, `data_blocks >= parity_blocks`, and i64-representable shard sizes.

For objects written by current RustFS this holds, so the blast radius is legacy / MinIO-migrated / foreign-edge metadata — but any such object that violates a newly-invented invariant now returns `FileCorrupt` and becomes permanently unreadable, with no migration path. Please validate against real on-disk and MinIO-migrated samples before merge, or gate the strict decode path behind an opt-in.

crates/ecstore/src/set_disk/core/io_primitives.rs
```diff
@@ -2398,7 +2440,10 @@ impl SetDisks {
         for (idx, meta_op) in metadata_array.iter().enumerate() {
             if let Some(meta) = meta_op {
                 match meta.into_fileinfo(bucket, object, vid.to_string().as_str(), read_data, incl_free_vers, true) {
-                    Ok(res) => meta_file_infos[idx] = res,
+                    Ok(res) => match res.validate_for_metadata_read() {
```
overtrue, line 2443: This is the normal read/quorum path (`pick_latest_quorum_files_info`), so the strict `validate_for_metadata_read` runs on **every** object read — not just heterogeneous pools or heal. Flagging so the compatibility risk (see the `validate_for_metadata_read` definition in `fileinfo.rs`) is understood as "every read", not an edge path.

crates/ecstore/src/cluster/rpc/remote_disk.rs
```diff
@@ -1825,6 +1835,7 @@ impl DiskAPI for RemoteDisk {
                 }
 
                 let file_info = decode_msgpack_or_json::<FileInfo>(&response.file_info_bin, &response.file_info, "FileInfo")?;
+                validate_decoded_file_info(&file_info)?;
```
overtrue, line 1838: **Mixed-version rolling upgrade.** A new node now hard-rejects (`FileCorrupt`) a `FileInfo` returned by an older peer at decode, where before it was a soft/quorum outcome. Combined with the strict invariants in `validate_for_metadata_read`, this converts a previously-successful cross-node read of legacy metadata into a hard error mid-rollout. The PR description already advises "configure EC:1 / block old-node writes before rollout" — that itself signals this breaks smooth upgrades and deserves an explicit, prominent compatibility call-out (and ideally a softer failure mode during rollout).

crates/filemeta/src/fileinfo.rs
```diff
@@ -304,27 +380,267 @@ impl FileInfo {
...
+    /// boundary has already called [`Self::validate`].
+    pub fn has_valid_erasure_geometry(&self) -> bool {
+        self.validate_erasure_geometry().is_ok()
+    }
+
+    /// Validates the complete payload metadata shape.
+    ///
+    /// Repeated quorum passes over metadata that was already fully validated
+    /// should use [`Self::has_valid_erasure_geometry`] instead.
+    pub fn is_valid(&self) -> bool {
```
overtrue, line 626: **Silent semantic change.** On `main`, `is_valid()` returned `true` for any `deleted` FileInfo; now it requires full erasure geometry, so a delete marker returns `false` (confirmed by the new `assert!(!deleted.is_valid())` test). The remaining non-test consumer at `set_disk/ops/heal.rs:1016` uses `is_valid()` to pick `parity_blocks` for heal reporting, so a delete marker now takes the `default_parity_count` branch instead of its own (0). It is "only" a reported field, but this behavior change rides along silently and is unrelated to the parity panic.

crates/ecstore/src/set_disk/core/io_primitives.rs
```diff
@@ -348,9 +350,24 @@ impl MetadataQuorumAccumulator {
             self.matching_version_votes = self.matching_version_votes.saturating_add(1);
         }
 
-        if file_info.deleted {
-            self.delete_marker_votes = self.delete_marker_votes.saturating_add(1);
+        if file_info.is_canonical_delete_marker() {
```
overtrue, line 353: **Blast radius beyond #4801.** Switching every quorum predicate from "any `deleted` FileInfo counts" to strict canonical-marker classification, and splitting delete-marker candidates by full identity, changes which version wins quorum in mixed delete-marker / purge-pending scenarios. This affects the *normal* delete/read/heal paths, not just heterogeneous pools — a regression here would be well outside the parity-panic blast radius. This is a distinct delete-marker-quorum correctness project that would be much safer to review as its own PR with its own adversarial validation.

crates/filemeta/src/filemeta/version.rs
```diff
@@ -2360,9 +2361,15 @@ impl MetaObject {
         let transitioned_objname = get_bytes(&self.meta_sys, SUFFIX_TRANSITIONED_OBJECTNAME)
             .map(|v| String::from_utf8_lossy(&v).to_string())
             .unwrap_or_default();
-        let transition_version_id = get_bytes(&self.meta_sys, SUFFIX_TRANSITIONED_VERSION_ID)
-            .and_then(|v| Uuid::from_slice(v.as_slice()).ok())
-            .filter(|u| !u.is_nil());
+        let transition_version_id = match get_bytes(&self.meta_sys, SUFFIX_TRANSITIONED_VERSION_ID) {
```
overtrue: **Behavior change on read.** A malformed/short `transition_version_id` now hard-fails the whole `into_fileinfo` with `FileCorrupt`, where `main` silently treated it as "not transitioned" and kept the object readable. This turns a slightly-corrupt-but-recoverable transition field into an unreadable object, and is unrelated to the parity panic.

rustfs/src/server/readiness.rs
```diff
@@ -565,23 +596,53 @@ where
...
+    if total_online == 0 {
         return false;
     }
 
-    set_drive_counts.into_iter().all(|((pool_idx, set_idx), set_drive_count)| {
-        let online = set_online_counts.get(&(pool_idx, set_idx)).copied().unwrap_or_default();
-        let quorum = quorum_for_set(info, pool_idx, set_drive_count);
-        online >= quorum
-    })
+    if configured_topology_present {
```
overtrue, line 613: **k8s readiness-probe behavior change.** In the configured-topology path readiness now iterates *all* configured sets `(0..set_count)` and requires each to meet quorum, and `pool_write_quorum`/`pool_read_quorum` returning `None` is treated as not-ready (the old `.max(1)` fallbacks never failed). Net: a cluster with an entirely-missing set/pool now reports not-ready where it previously could report ready. More correct, but a real probe behavior change for partial-outage/misconfig scenarios — worth an explicit call-out since it is separable from the parity fix.

crates/obs/src/metrics/stats_collector.rs
```diff
@@ -789,43 +790,86 @@ pub fn collect_internode_network_stats() -> Option<NetworkStats> {
...
-        let set_idx = disk.set_index.max(0) as usize;
-        let set_drive_count = backend.drives_per_set.get(pool_idx).copied().unwrap_or_default();
-        let parity = backend
-            .standard_sc_parities
-            .get(pool_idx)
-            .copied()
-            .or(backend.standard_sc_parity)
-            .unwrap_or(set_drive_count / 2);
+        let (Ok(pool_idx), Ok(set_idx)) = (usize::try_from(disk.pool_index), usize::try_from(disk.set_index)) else {
+            continue;
```
overtrue, line 852: **Metrics gaps on heterogeneous clusters.** `erasure_set_stats_from_backend` now emits no row when the layout can't be resolved (old code always emitted a row with `parity = set_drive_count/2`), and `collect_cluster_config_stats` returns `None` — dropping the `standard_parity`/`rrs_parity` gauges — exactly in the heterogeneous case. Dashboards expecting one row per set, or those two gauges, will see gaps after this change.

overtrue (commented):
### Simplicity / over-engineering notes

First: the core fix is sound and well-designed — per-pool parity resolution, `Erasure::try_new*`, the four guarded construction sites, `WriteLayout`/`resolve_write_layout`, the pre-topology startup validation, and `Result`-propagating `lookup_configs`. Resolving parity per pool is actually better than the `min(parity, pool_size-1)` clamp suggested in the issue. The notes below are about the amount of machinery bundled around that core. Most of these are driven by making internal seams unit-testable and add cognitive cost without any production polymorphism:

- **6 generic parameters on `reload_runtime_config_snapshot_with`** (`rustfs/src/admin/service/config.rs`) with a single production caller — the generics exist only to inject fakes for one ordering test. The pre-PR straight-line read → apply-loop → publish was ~20 lines.
- **`commit_prepared_config`** (`rustfs/src/admin/handlers/config_admin.rs`) generifies a two-statement body (`persist.await?; publish(...)`) over four call sites that all pass the same two arguments. Inlining would be shorter and clearer.
- **Four-method `publish_storage_class{_with,_for_context_with,_for_context,}` ladder** just to set one `ArcSwap`.
- **`ErasureConstructionError` has 7 variants plus an extra `ErasureConstructionIoSource` adapter struct** solely to keep the wrapped error in the `io::Error` source chain. The issue only asked to turn one `.expect()` into a `Result`; 2–3 variants cover the real cases.
- **`ValidatedFileInfo<'a>` type-state witness whose `_file_info` field is never read** — the witness pays a real complexity cost but is meaningfully consumed at exactly one site.
- **Hand-rolled ~10,000-bit stack bitset for part-number dedup** on a cold validation path (`crates/filemeta/src/fileinfo.rs`), where a `HashSet<usize>` would be a few obviously-correct lines.
- **`parity_for_sc` is `#[cfg(test)]`-only** — a production API surface that exists only for tests.
- **The same erasure-geometry resolver is copy-pasted three times** — `pool_erasure_layout` (`readiness.rs`), `standard_erasure_layout_from_backend` (`stats_collector.rs`), and `resolve_pool_layout` (`rebalance.rs`). One shared helper would remove two copies and guarantee the three surfaces agree.
- **Two ~30–44-field boolean enumerators** (`validate_delete_marker_shape`, `validate_tier_free_version_delete_shape`) assert "every other field is zero/empty". They are brittle: a newly-added `FileInfo` field is silently accepted unless a maintainer remembers to add a clause here.
- **Test smells:** `storage_config_write_handlers_persist_before_publishing` uses `include_str!` + `.matches().count()` to assert call ordering from the file's own source text (breaks on any rename/reflow); the 2000-iteration `storage_class_arc_swap_never_exposes_mixed_pool_state` essentially re-tests `arc_swap` itself.

Overall the PR bundles roughly six separable efforts and ships about 2:1 (up to ~4:1 in some files) test-to-production lines, most of which exercise the bundled hardening rather than the parity panic. Splitting the panic fix out from the metadata-validation, delete-marker-quorum, rename-commit-rollback, and config-atomicity changes would let each be reviewed (and risk-assessed) independently. The correctness concerns behind that recommendation are in the inline comments below.

crates/ecstore/src/cluster/rpc/remote_disk.rs
```diff
@@ -1143,12 +1146,19 @@ fn decode_batch_read_version_response_items(
...
+            validate_decoded_file_info(&resp.file_info)?;
+        }
         batch_read_version_resps.push(resp);
     }
 
     Ok(batch_read_version_resps)
 }
 
+fn validate_decoded_file_info(file_info: &FileInfo) -> Result<()> {
+    file_info.validate_for_metadata_read().map_err(Into::into)
```
overtrue, line 1159: **(major, compatibility)** This applies the new, strictly-narrower `validate_for_metadata_read()` to every peer-RPC-decoded `FileInfo`, and the same tightening runs on local disk reads (`disk/local.rs`, e.g. lines 6696/7470/7543 and the read_version/rename_data/write_metadata paths). Metadata that was valid under the old `is_valid()` but is not "canonical" under the new rules (delete-marker shape, tier-free shape, `transition_version_id` form, checksum↔part 1:1) now decodes as `FileCorrupt`.

Rolling-upgrade hazard: an older node's delete marker / transitioned version written in a now-rejected shape → a new node returns `FileCorrupt` on read/heal; if all replicas share that shape, that is a false quorum loss. Please validate against real historical on-disk fixtures + a mixed-version matrix before this lands, and consider gating the stricter mode behind read-vs-write context so reads stay lenient while writes stay canonical.

crates/ecstore/src/config/com.rs
```diff
@@ -1576,14 +1573,11 @@ where
     save_config(api, &config_file, data).await
 }
 
-pub async fn lookup_configs<S>(cfg: &mut Config, api: Arc<S>)
+pub async fn lookup_configs<S>(cfg: &mut Config, api: Arc<S>) -> Result<()>
```
overtrue, line 1576: **(startup behavior change)** `lookup_configs` now returns `Result` and the error propagates through `ConfigSys::init`, where it was previously logged and swallowed. A persisted storage-class config that is invalid for the current topology (e.g. `EC:2` on a 2-drive pool) will now hard-fail config init / startup instead of degrading.

The fail-closed direction is correct, but a cluster that previously booted "degraded" could now refuse to start after upgrade — please call this out in the rollout/compat notes and confirm it is the intended operational behavior.

crates/filemeta/src/filemeta/version.rs
```diff
@@ -806,7 +828,7 @@ impl TryFrom<LegacyMetaV2Version> for FileMetaVersion {
 
 impl From<FileInfo> for FileMetaVersion {
     fn from(value: FileInfo) -> Self {
-        if value.deleted {
+        if value.is_storage_delete_marker() {
```
overtrue, line 831: **(minor, correctness)** Branching on `is_storage_delete_marker()` instead of `value.deleted` means a `deleted == true` FileInfo whose shape isn't canonical (e.g. `mod_time == None`) is silently re-encoded as `VersionType::Object` with a malformed `MetaObject` rather than surfacing an error. Upstream `validate_for_*` mostly shields this, but the silent reclassification is a latent trap — prefer erroring on `deleted && !is_storage_delete_marker()`.

overtrue (changes requested):
### Request changes — the core parity fix is correct and worth landing, but it's over-bundled and the tightened metadata validation carries rolling-upgrade risk

Reviewed adversarially (not taking the PR self-assessment at face value). The core fix for #4801 is sound: per-pool STANDARD/RRS resolution, an immutable per-pool `ArcSwap` storage-class snapshot, and a validated `WriteLayout {data, parity, write_quorum}` resolved against `(pool_index, drive_count)`. Independently confirmed:

- Per-pool parity math is right (`default_parity_count` {1→0, 2/3→1, 4/5→2, 6/7→3, else 4}; `data = drive_count - parity`; matches the existing rustfs/MinIO logic, no off-by-one). The 4-disk + 2-disk case now resolves to 2+2 / 1+1 instead of 0 data shards.
- No torn read: `put_object`/multipart/tiered-write capture the snapshot once and use it for both `resolve_write_layout` and `should_inline`; an offline disk (`None` slot) does not trip the fail-closed path.
- prepare→persist→publish is sound (validate before publish, one `ArcSwap` store, no partial publication).
- No on-disk/RPC format change (`standard_parities`/`rrs_parities` are `#[serde(skip)]`; legacy scalar reporting preserved).

Two blocking concerns before this can land:

**1. (major, compatibility) The new `validate_for_metadata_read()` is strictly narrower than the old `is_valid()`, and now runs on every local disk read and every peer-RPC-decoded `FileInfo`.** Metadata accepted under the old rules but not "canonical" under the new ones (delete-marker shape, tier-free-version shape, `transition_version_id` well-formedness, checksum↔part 1:1) now decodes as `FileCorrupt`. In a rolling upgrade where an older node wrote a delete marker / transitioned version in a shape the new validation rejects, a new node returns `FileCorrupt` on read/heal — and if all replicas share that shape, that is a false quorum loss / unreadable object. The PR frames mixed-version risk only around parity, not around this tightened validation. This must be validated against real historical on-disk fixtures and a mixed-version matrix before merge. (Inline notes on the local + RPC call sites.)

**2. (major, scope) ~5,910 additions for a per-pool parity bug is heavily over-bundled.** The parity fix itself is a fraction of this. The `FileInfo` validation-model rewrite (`validate`/`ValidationMode` + ~6 predicates in `fileinfo.rs`), the delete-marker serialization change (`MetaDeleteMarker::from` now populating `meta_sys`), and the inline-rollback-backup rewrite in `local.rs` rename commit are defensible hardening but are independent changes that belong in separate reviewable PRs. Bundled into a storage-critical path they widen blast radius and make revert-on-regression coarse. Please split so the parity fix is independently reviewable and revertable.

Minor:
- `crates/ecstore/src/set_disk/ops/heal.rs:1015` `if lfi.is_valid()` was not migrated with the rest; with the new `is_valid()` semantics (delete markers now return `false`) a delete-marker heal reports `parity_blocks`/`data_blocks` differently in `HealResultItem`. Reporting-only, but out of step with the otherwise-thorough migration.
- (inline) the `From<FileInfo>` reclassification in `version.rs` and the `lookup_configs` startup change in `config/com.rs`.

Mechanics: this is CONFLICTING with `main`, but the conflicts are limited to exactly two files (`config/com.rs` and `disk/local.rs`, both mechanical/unrelated to the parity logic) and everything else auto-merges — a straightforward rebase.

Net: the parity fix core is the right shape and worth landing after (1) de-bundling, (2) validating the tightened metadata acceptance against historical fixtures + mixed-version upgrade, (3) fixing the heal.rs miss and the silent `From` reclassification, and (4) the two-file rebase. Since this is still a draft, flagging early so the split can happen before final review.

overtrue (approved):
### Approve — CI is green and the actionable findings are resolved

Full CI now passes on the actual code (21 success / 5 skipped). This was the first complete run, since the PR only ran CLA checks while it was a draft. Re-reviewing against the earlier `CHANGES_REQUESTED`:

- The three on-disk read regressions introduced by the tightened metadata validation are fixed and covered: R1 negative compressed `actual_size`, and R2/R3 MinIO-migrated tiered `transitioned_version_id` stored as a UUID string. Decode leniency is restored in line with the erasure-coding on-disk compatibility contract.
- The conflict with `main` is resolved and the PR is out of draft.
- The heterogeneous / small-pool upgrade behavior (a persisted storage class a pool cannot satisfy now fails startup) is documented as an upgrade note in `CHANGELOG.md`, with the `RUSTFS_STORAGE_CLASS_STANDARD` recovery steps.

Non-blocking (maintainer's call): the change is still large relative to the underlying parity fix (~5.9k additions bundling the FileInfo validation-model rewrite, delete-marker serialization, and inline-rollback rework). Landing as-is is reasonable given the coverage; splitting the independent hardening into follow-ups would make future reverts finer-grained. The two minor reporting/reclassification nits (`heal.rs` `is_valid` in `HealResultItem`, and the `From<FileInfo>` branch) are left as-is under the prior on-disk-compatibility audit's "verified safe" assessment.

Approving.

houseme:
I rechecked the current PR state: the latest head still has green CI and the earlier approval covers the actionable compatibility fixes. I do not see a new code blocker.

The remaining merge gate is procedural: GitHub still marks this PR as Draft, so it should be marked ready for review before merging. Given the size and storage-risk of the change, I would also keep the final green CI run attached to the ready-for-review state.

overtrue:
Heads-up: the per-pool parity **core** of this PR has been extracted into a separate, minimal PR — #5015 — so the #4801 panic fix can be reviewed and merged on its own.

#5015 contains only the two reviewed-correct core commits (per-pool parity resolution + fallible erasure construction with the four guarded call sites, credited to @cxymds via cherry-pick) on top of latest `main`, plus a small fixup so they build standalone. It is `#[serde(skip)]` runtime-only, changes no on-disk/wire format, and passes the arch guards, clippy, fmt, and the per-pool/erasure regression tests.

What #5015 deliberately leaves here for separate review: the `FileInfo` validation-model rewrite, the metadata-quorum predicate re-plumbing, the inline-metadata rename-commit rollback subsystem, the admin config prepare/persist/publish atomicity refactor, and the per-pool readiness/metrics reporting.

Suggested sequencing: land #5015 for the panic fix; the remaining changes here are built on top of that core, so rather than dropping the core from this branch (the rest depends on it), this PR would be reworked into the remaining, independently-reviewable pieces on top of `main` once #5015 is in. Happy to help split those out.

crates/filemeta/src/fileinfo.rs
overtrue: Addressed. The concrete regressions are fixed — R1: a negative `actual_size` (compressed objects with unknown size) now skips per-shard validation instead of failing `FileCorrupt`; R2/R3: a malformed/string `transition_version_id` is now tolerated via the shared `transitioned_version_id_from_meta_sys` helper (`386029593`). And the "validate against real on-disk / MinIO-migrated samples" ask is now covered: 5 real historical `xl.meta` fixtures (MinIO small object, versioned-with-delete-marker, multipart, legacy V1, legacy meta_v2) assert green through `validate_for_metadata_read` (`2dedd4d3d`). The opt-in gate for strict-on-read is intentionally not added and the rolling-upgrade guidance is documented in the PR description and CHANGELOG. Assessed and accepted for this PR — resolving.

crates/ecstore/src/set_disk/core/io_primitives.rs
overtrue: Confirmed — the strict `validate_for_metadata_read` runs on the normal read/quorum path (`pick_latest_quorum_files_info`), i.e. on every read, not an edge path. The concrete regressions are fixed and 5 real historical `xl.meta` fixtures now cover this decode boundary (see the `fileinfo.rs:553` thread, `2dedd4d3d`). Risk assessed and accepted — resolving.

crates/ecstore/src/cluster/rpc/remote_disk.rs
overtrue: Acknowledged — the soft/quorum → hard `FileCorrupt` transition on a peer's legacy metadata during a mixed-version rollout is a real behavior change. The rollout guidance (configure `EC:1` / block old-node writes before rollout) is called out in the PR description and CHANGELOG; a softer rollout-window failure mode is not implemented. Assessed and accepted for this PR's scope — resolving.

crates/filemeta/src/fileinfo.rs
overtrue: Addressed for the non-test consumer: `heal.rs` now uses `has_valid_erasure_geometry()` instead of `is_valid()` (`86213fd1a`), so a delete marker no longer falls into the `default_parity_count` branch for heal reporting. The tightened `is_valid()` definition itself is intrinsic to the validation framework and is kept. Resolving.

crates/ecstore/src/set_disk/core/io_primitives.rs
overtrue: Acknowledged — switching delete-marker quorum classification from "any `deleted` FileInfo counts" to canonical-marker classification (and splitting candidates by full identity) is a distinct correctness change beyond the #4801 parity blast radius. Assessed and accepted within this PR rather than carved into its own change — resolving.

crates/filemeta/src/filemeta/version.rs
overtrue: Fixed (`386029593`). `transition_version_id` now goes through the shared `transitioned_version_id_from_meta_sys` helper — 16 bytes → `Some`, UUID string → recovered, otherwise → `None` — so a malformed/short field no longer hard-fails `into_fileinfo` with `FileCorrupt`; the object stays readable as "not transitioned", matching `main`. A MinIO string-recovery test case was added. Resolving.

rustfs/src/server/readiness.rs
overtrue: Acknowledged — readiness now iterates all configured sets requiring each to meet quorum, and a `None` write/read quorum is treated as not-ready (the old `.max(1)` fallback never failed). More correct, but a real probe behavior change for partial-outage/misconfig scenarios and separable from the parity fix. Assessed and accepted — resolving.

crates/ecstore/src/config/com.rs
overtrue: Documented per maintainer decision (`2fe31e90c`). The CHANGELOG `[Unreleased]` now notes that a persisted `EC:N` storage-class invalid for the current topology (e.g. `EC:2` on a 2-drive pool) will refuse startup after upgrade instead of degrading, with the workaround (set `RUSTFS_STORAGE_CLASS_STANDARD` empty to auto-derive per pool, or to a value each pool can satisfy). The fail-closed direction is kept intentionally; no code semantics changed. Resolving.

crates/filemeta/src/filemeta/version.rs
overtrue: Addressed (`6b02b234e`). That branch now emits a `warn!` on `deleted && !is_storage_delete_marker()`. `From<FileInfo> for FileMetaVersion` cannot return `Result`, and erroring outright would over-reject legitimate purge-pending / intentionally non-canonical delete markers, so it logs the malformed shape rather than hard-failing. Resolving.

crates/obs/src/metrics/stats_collector.rs
overtrue: Acknowledged — on heterogeneous clusters `erasure_set_stats_from_backend` now emits no row when the layout can't be resolved and `collect_cluster_config_stats` drops the `standard_parity`/`rrs_parity` gauges. Dashboards expecting one row per set or those gauges will see gaps. Assessed and accepted (dashboard-compat can be a follow-up) — resolving.

crates/ecstore/src/cluster/rpc/remote_disk.rs
overtrue: Same family as the `fileinfo.rs:553` thread — the strict validator applies on peer-RPC-decoded and local-disk reads. The concrete regressions (R1/R2/R3) are fixed (`386029593`) and 5 real historical `xl.meta` fixtures now assert green through the strict path (`2dedd4d3d`). Read-vs-write context gating is intentionally not added; strict-on-read is kept and the rolling-upgrade hazard is documented. Assessed and accepted — resolving.

PR #4985, opened 2026-07-17
fix(s3): return x-amz-copy-source-version-id for versioned CopyObject source

### LINKED ISSUE #4976: CopyObject with source versionId omits x-amz-copy-source-version-id response
### Describe the bug

With bucket versioning enabled, RustFS accepts a `CopyObject` request whose
`x-amz-copy-source` identifies an exact historical version using the
`?versionId=<source-version-id>` query parameter. The copy is applied and a
distinct destination version is created, but the successful response omits
`x-amz-copy-source-version-id`.

The AWS SDK for JavaScript v3 consequently returns no
`CopySourceVersionId`, even though it returns the destination `VersionId` from
`x-amz-version-id`.

### Environment

- RustFS canonical reproduction:
  `rustfs/rustfs:1.0.0-beta.9@sha256:f75d0bca6ca322c4e59f7125f73dd9ab709b22f71396a42c95bce7f74c99e53b`
- Also reproduced with official candidate:
  `rustfs/rustfs:1.0.0-beta.10@sha256:60f4f2f41ce95216f8cac676e69f9d90c0bfec458a3bc7fd7fb9b7c2452ac57a`
- Platform: Linux `x86_64` / `linux/amd64`
- Client: `@aws-sdk/client-s3` `3.1079.0`
- Data: deterministic synthetic bytes only

### Minimal reproduction

1. Start one RustFS instance on loopback with synthetic credentials.
2. Create two private buckets and enable versioning on both.
3. Put one small deterministic synthetic object in the source bucket with a
   SHA-256 checksum and capture the exact returned source `VersionId`.
4. Issue `CopyObject` to the destination bucket with URL-encoded
   `x-amz-copy-source` ending in `?versionId=<exact-source-version-id>`.
5. Inspect only the successful response headers and SDK output fields.
6. List versions through an authenticated restricted inventory and confirm a
   distinct destination version exists.
7. HEAD that exact destination version.
8. Delete the exact source and destination versions and verify both buckets are
   empty with no delete markers.

The reproduction should use synthetic bucket/key names and bytes. They are
intentionally omitted here; no credentials, authorization headers, signed
URLs, raw query strings, request bodies, or provider coordinates are needed.

### Expected behavior

For a versioned source copy, the response includes:

```text
x-amz-copy-source-version-id: <exact source version copied>
x-amz-version-id: <new destination version>
```

AWS reference:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html

The source and destination values serve different purposes and must not be
substituted for one another.

### Actual behavior

Redacted facts:

- source `VersionId` captured: yes
- exact source version included in serialized `CopySource`: yes
- copy applied: yes
- destination created: yes
- destination version distinct: yes
- destination `x-amz-version-id` present: yes
- SDK `VersionId` present: yes
- `x-amz-copy-source-version-id` present: no
- SDK `CopySourceVersionId` present: no
- source remained addressable before cleanup: yes
- exact cleanup and empty buckets: yes

Additional exact-copy observation: with `ChecksumAlgorithm=SHA256`, neither
`CopyObjectResult.ChecksumSHA256` nor an exact destination HEAD SHA-256 value
was returned in these reproductions, although exact destination size and
`VersionId` matched. This is reported separately from the missing source
lineage header so the response-identity defect remains clear.

### Reproducibility

- beta.9 raw response reproduction: 2/2
- beta.10 isolated raw response reproduction: 2/2
- unchanged application conformance suite: 11/12 passed; the same exact-copy
  test failed

### Impact

An exact-version quarantine-to-private promotion cannot be confirmed safely.
The provider has applied the copy, but clients cannot prove which source
version it copied from the response. Treating the operation as success would
require unsafe latest-version inference or substituting destination identity,
ETag, metadata, caller assertions, or checksum equality for source lineage.

Our adapter therefore correctly treats the operation as uncertain and requires
reconciliation. It does not delete the source automatically.

### Requested fix and regression coverage

Please populate the plain `CopyObjectOutput.copy_source_version_id` from the
parsed source version so the S3 response emits
`x-amz-copy-source-version-id`, while preserving the distinct destination
`version_id`.

Please add a regression test that:

1. creates at least two source versions;
2. copies the non-latest exact source version;
3. asserts the response source-version header equals that requested version;
4. asserts the destination header identifies a distinct new version;
5. verifies exact destination bytes/checksum/size; and
6. proves the source version remains present.

### Source inspection note

At official `1.0.0-beta.10` commit
`f67e8a6cdcb3e08dfa3c3c390412a9815837d0b2`, the plain `CopyObjectOutput`
sets destination `version_id` but not `copy_source_version_id`. The
`UploadPartCopyOutput` path already sets `copy_source_version_id`, which may
provide a useful implementation reference.

overtrue:
## What

For a `CopyObject` whose source carries versioning, the successful response must echo the exact source version that was copied via `x-amz-copy-source-version-id` (AWS SDK `CopySourceVersionId`), kept distinct from the newly created destination `x-amz-version-id`.

RustFS was setting the destination `version_id` on `CopyObjectOutput` but leaving `copy_source_version_id` unset. As a result, AWS SDK clients received the destination `VersionId` but no `CopySourceVersionId`, so they could not prove which source version the copy was taken from — blocking safe exact-version promotion flows (e.g. quarantine → private). This is the same lineage field that the `UploadPartCopy` path already populates.

Fixes #4976.

## Fix

`execute_copy_object` now populates `CopyObjectOutput.copy_source_version_id` from the version actually read from the source (`gr.object_info.version_id`, captured before `src_info` is mutated/consumed). It is gated on the source bucket carrying versioning (enabled or suspended) and renders the null version as `null`, mirroring the existing GET/HEAD convention in the same file. The destination `version_id` is unchanged and stays distinct.

- Explicit `?versionId=<v>` on a versioned source → header echoes exactly `<v>` (not the latest).
- No explicit versionId on a versioned source → header echoes the resolved (latest) source version, matching AWS/MinIO.
- Unversioned source bucket → header omitted, as before.

## Test

Added `test_copy_of_non_latest_source_version_returns_copy_source_version_id` (e2e) which:

1. creates two source versions in a versioned source bucket;
2. copies the exact non-latest source version into a separate versioned destination bucket;
3. asserts `x-amz-copy-source-version-id` equals the requested (non-latest) source version and is not the latest;
4. asserts the destination `x-amz-version-id` is present and distinct from both source versions;
5. verifies the destination holds the exact bytes/size of the copied source version; and
6. proves the copied source version remains present and independently readable.

Reverting the fix makes this test fail (the header is absent).

## Scope note

The issue also reports, separately, that `ChecksumSHA256` may be absent from `CopyObjectResult`/destination HEAD under `ChecksumAlgorithm=SHA256`. That is a distinct response-checksum concern and is intentionally out of scope here so the source-lineage header fix stays clean; it can be tracked as a follow-up.

## Verification

- `cargo check -p rustfs` / `-p e2e_test --tests`: pass
- `cargo test -p rustfs --lib execute_copy_object`: 7 passed
- `cargo test -p e2e_test --lib test_copy_of_non_latest_source_version_returns_copy_source_version_id`: pass
- `make pre-commit` (fmt + arch checks + quick-check): pass
- `cargo clippy -p rustfs -p e2e_test --all-targets`: clean

### DIFF
--- a/crates/e2e_test/src/copy_object_version_restore_test.rs
+++ b/crates/e2e_test/src/copy_object_version_restore_test.rs
@@ -160,4 +160,146 @@ mod tests {
 
         env.stop_server();
     }
+
+    /// Regression test for Issue #4976: a versioned-source CopyObject must echo the exact source
+    /// version copied via `x-amz-copy-source-version-id` (SDK `CopySourceVersionId`), kept distinct
+    /// from the newly created destination `x-amz-version-id`.
+    #[tokio::test]
+    #[serial]
+    async fn test_copy_of_non_latest_source_version_returns_copy_source_version_id() {
+        init_logging();
+        info!("Issue #4976: versioned CopyObject must return x-amz-copy-source-version-id for the exact source version");
+
+        let mut env = RustFSTestEnvironment::new().await.expect("Failed to create test environment");
+        env.start_rustfs_server(vec![]).await.expect("Failed to start RustFS");
+
+        let client = env.create_s3_client();
+        let src_bucket = "copy-source-version-header-src";
+        let dst_bucket = "copy-source-version-header-dst";
+        let src_key = "reports/quarantined.bin";
+        let dst_key = "reports/promoted.bin";
+
+        for bucket in [src_bucket, dst_bucket] {
+            client
+                .create_bucket()
+                .bucket(bucket)
+                .send()
+                .await
+                .expect("Failed to create bucket");
+            client
+                .put_bucket_versioning()
+                .bucket(bucket)
+                .versioning_configuration(
+                    VersioningConfiguration::builder()
+                        .status(BucketVersioningStatus::Enabled)
+                        .build(),
+                )
+                .send()
+                .await
+                .expect("Failed to enable versioning");
+        }
+
+        // Source version 1: the exact (non-latest) version we will copy.
+        let v1_content = b"quarantined payload -- source version one (target of the copy)";
+        let put_v1 = client
+            .put_object()
+            .bucket(src_bucket)
+            .key(src_key)
+            .body(ByteStream::from_static(v1_content))
+            .send()
+            .await
+            .expect("PUT source v1 failed");
+        let v1_id = put_v1.version_id().expect("source v1 must have a version id").to_string();
+
+        // Source version 2: becomes the latest, so v1 is deliberately NOT the current version.
+        let v2_content = b"quarantined payload -- source version two (now current, must be ignored)";
+        let put_v2 = client
+            .put_object()
+            .bucket(src_bucket)
+            .key(src_key)
+            .body(ByteStream::from_static(v2_content))
+            .send()
+            .await
+            .expect("PUT source v2 failed");
+        let v2_id = put_v2.version_id().expect("source v2 must have a version id").to_string();
+        assert_ne!(v1_id, v2_id, "the two source puts must produce distinct versions");
+
+        // Copy the exact NON-LATEST source version into the destination bucket.
+        let copy_out = client
+            .copy_object()
+            .bucket(dst_bucket)
+            .key(dst_key)
+            .copy_source(format!("{src_bucket}/{src_key}?versionId={v1_id}"))
+            .send()
+            .await
+            .expect("CopyObject of a versioned source must succeed");
+
+        // (3) The response must echo the exact source version copied.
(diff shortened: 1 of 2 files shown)


overtrue:
The separately-reported `ChecksumAlgorithm=SHA256` copy-checksum gap (out of scope here) is now tracked in #4996.

PR #4988, opened 2026-07-17
fix(rpc): bind internode auth to exact targets

cxymds:
## Related Issues

N/A

## Summary of Changes

- Add rolling-upgrade-compatible internode gRPC authentication v2 that binds HMAC signatures to the trusted destination authority, exact generated gRPC service/method, POST transport method, timestamp, nonce, and content digest.
- Preserve ports and IPv6 authority syntax in the trusted local node identity, and resolve that identity per request so connections opened before storage initialization recover after initialization.
- Add process-local replay protection for body-bound v2 requests with UUID nonces, monotonic retention, wall-clock regression protection, fail-closed capacity handling, and deterministic canonical-body digest helpers.
- Keep the legacy prefix signature during rolling upgrades, fail closed on partial or malformed v2 metadata, and register the exact legacy-removal condition.

## Verification

- `cargo test -p rustfs-ecstore cluster::rpc::http_auth::tests --lib` (31 passed)
- `cargo test -p rustfs-ecstore cluster::rpc::client::tests --lib` (5 passed)
- `cargo test -p rustfs-ecstore endpoint_rpc_authority_preserves_port_and_ipv6_brackets --lib`
- `cargo test -p rustfs-ecstore init_local_peer_publishes_complete_rpc_authority --lib`
- `cargo test -p rustfs rpc_auth_binds_post_method_authority_and_exact_path --lib`
- `cargo test -p rustfs rpc_request_path_service_preserves_exact_http_target --lib`
- `cargo test -p rustfs-common --lib` (64 passed)
- `cargo check -p rustfs`
- `bash scripts/check_layer_dependencies.sh`
- `bash scripts/check_unsafe_code_allowances.sh`
- `bash scripts/check_doc_paths.sh`
- `git diff --check`

Adversarial validation: correctness attacked cross-method, cross-node, same-host/different-port, early-connection, malformed-v2, and clock-boundary cases; no break remains. Security attacked downgrade, authority spoofing, replay, secret leakage, and body tampering; no break remains within the documented rolling-upgrade boundary. Concurrency/durability attacked concurrent duplicate nonces, cache capacity, expiry ordering, clock rollback, and restart semantics; no break remains within the documented process-local replay contract. Compatibility verified old/new peer rolling behavior and registered the legacy cleanup condition. Performance verified constant-time nonce lookup and insertion-ordered expiry; audience parsing remains a documented non-blocking follow-up before high-rate body-bound RPC use. Test coverage includes revert-sensitive production identity publication and server auth-path regressions.

## Impact

New peers use exact-target v2 authentication while old peers remain compatible through the registered legacy fallback. No S3 API, admin API, on-disk format, or CLI output changes are included in this PR.

## Additional Notes

This is the first stacked change in the heal control-plane repair series. Per the task's staged verification plan, focused critical verification is recorded above; the full `make pre-pr` gate is reserved for the final integration PR after all stacked changes are handed off.

---


houseme (approved):
Reviewed the high-risk internode RPC auth surface. I did not find blocking issues.

The v2 signature is bound to the destination authority, exact generated gRPC service/method, POST, timestamp, nonce, and content digest. Malformed or partial v2 metadata fails closed instead of falling back to the legacy signature, while the legacy path is clearly scoped to the rolling-upgrade window and recorded in the cleanup register.

I also checked the client interceptor audience setup, server-side request target preservation before tonic interception, local node authority handling including port/IPv6 preservation, replay-cache behavior for body-bound requests, and the new regression tests.

Local verification: `cargo test -p rustfs-ecstore cluster::rpc::http_auth::tests --lib` passed.

PR #4989, opened 2026-07-17
fix(heal): initialize the runtime atomically

cxymds:
## Related Issues

N/A

## Summary of Changes

- Publish the process-wide heal manager and channel processor as one atomic runtime.
- Serialize initialization so concurrent callers cannot start orphan managers or processors.
- Keep initialization alive when the original caller is cancelled and roll back a started manager if initialization fails.
- Preserve the existing explicit error for sequential or concurrent repeated initialization.
- Add forced-interleaving regression coverage for cancellation, concurrent initialization, atomic visibility, and channel initialization rollback.

## Verification

- `cargo test -p rustfs-heal --lib` (209 passed)
- `cargo fmt --all --check`
- `git diff --check`

## Impact

Heal runtime initialization is now process-local single-flight and cancellation-safe. Public API signatures and the repeated-initialization error remain unchanged. No configuration, storage format, wire format, or S3-visible behavior changes.

## Additional Notes

This is stacked on #4988 because later cluster heal control work depends on its internode RPC authentication primitives. The full `make pre-pr` gate is intentionally deferred to the final PR in the stack; this PR received focused crate, lint, architecture, logging, and adversarial validation.

- Correctness: attacked cancellation, partial initialization, retry, sequential conflict, and forced concurrent interleavings; no break found after adding regression coverage.
- Security: attacked error/log leakage and externally influenced initialization paths; no new security surface or secret exposure found.
- Concurrency/durability: attacked caller aborts, concurrent callers, publication ordering, and failure cleanup; no orphan live runtime or split publication found.
- Compatibility: checked public signatures, repeated-init error behavior, and runtime visibility; no compatibility break found.
- Performance: checked startup lock scope and getter hot paths; no runtime hot-path regression found.
- Test coverage: mutation-tested the claimed lock, cleanup, owned-task, and atomic-publication behavior; all identified gaps are covered.

---


overtrue (commented):

Reviewed the incremental diff adversarially (single-flight init, atomic publication, cancellation survival, rollback), not just trusting the description. The core mechanism holds:

- Single-flight with no TOCTOU: the `is_some()` check and `OnceCell::set()` both run under the `GLOBAL_HEAL_RUNTIME_INIT` tokio mutex, so a concurrent second caller blocks, then observes "already initialized" and never starts a second manager.
- Cancellation survival is correctly implemented via an owned `tokio::spawn` whose `JoinHandle` is awaited — dropping the caller future does not abort the owned init, so it still publishes.
- Rollback on `manager.start()` / channel-init failure calls `stop()` and returns without publishing, leaving no live orphan.
- Getters read the `OnceCell` lock-free; no lock is held across `.await`. Publishing manager+processor in one `HealRuntime` cell closes the previous split-publish hazard.

Non-blocking nits:
- `crates/heal/src/lib.rs`: test scaffolding (`force_channel_failure`, `test_hook_after_manager_start`) is interleaved into the production init body via `#[cfg(test)]`/`#[cfg(not(test))]`. It compiles away in release but adds noise to a critical function; a dedicated injection point would read cleaner.
- `crates/heal/src/lib.rs` channel-init rollback: `stop_initializing_manager(&heal_manager).await?` lets a `stop()` error mask the original channel-init error, which contradicts the "preserve the error" claim. Only reachable in tests today, but worth reordering so the original error wins.

Merge gating: this is the second change in a stacked series rooted at #4988; its base is another stack branch, so it cannot merge to `main` until #4988 lands and this PR is retargeted. Only the CLA check has run on this increment — full Test/Lint/E2E only triggers after retargeting to `main`; please confirm green then.

houseme (approved):

PR #4990, opened 2026-07-17
fix(heal): aggregate status across cluster nodes

cxymds:
## Related Issues

N/A

## Summary of Changes

- Aggregate background heal status from every configured cluster node so load-balanced status requests no longer alternate between local idle and remote active results.
- Add an authenticated, versioned node heal snapshot RPC with strict MessagePack decoding, a 64 KiB response limit, and fail-closed topology, timeout, transport, and decode handling.
- Include retry-backoff work in heal operation snapshots and make active-to-retrying and retrying-to-queue ownership transfers observable as a single locked transition.
- Resolve the object store, endpoints, and notification system from the same request AppContext, and avoid repeated distributed background-info reads on peer nodes.
- Preserve rolling-upgrade availability only when known work is active; otherwise an unsupported old peer prevents an unproven idle result. Expose `clusterStatusComplete` for explicit partial active responses.

## Verification

- `cargo test -p rustfs-heal --lib` (210 passed before the deterministic transition test was added)
- `cargo test -p rustfs-heal test_scheduler_retry_transitions_keep_continuous_single_ownership -- --nocapture` (1 passed)
- `cargo test -p rustfs --lib heal::tests --no-fail-fast` (42 passed)
- `cargo clippy -p rustfs-heal --lib -- -D warnings`
- `cargo fmt --all --check`

Full `make pre-pr` verification is intentionally deferred to the final PR in this stacked heal-control series.

## Impact

Background heal status becomes cluster-wide and fail-closed. The existing JSON fields remain unchanged; `healOperations` gains retry counters and the top-level response gains the additive `clusterStatusComplete` field. During rolling upgrades, known active work may be returned as incomplete, while an incomplete result that would otherwise claim idle is rejected.

## Additional Notes

This PR is stacked on `codex/heal-manager-registry` and should be reviewed after #4989.

- Correctness: attacked request-context isolation, peer fanout, retry ownership gaps, counter overflow, and partial-idle states; no break remains.
- Security: attacked RPC authentication, oversized or malformed peer payloads, error leakage, and cross-context topology exposure; no break remains.
- Concurrency/durability: attacked active/queue/retrying lock interleavings, retry handoff gaps, timeout eviction, and progress lock duration; no break remains.
- Compatibility: attacked old-peer `Unimplemented`, additive JSON evolution, wire versioning, and unrelated RPC size limits; no break remains.
- Performance: attacked N-squared config reads, shared-channel churn, lock-held awaits, and fanout serialization; no break remains.
- Test coverage: identified and closed gaps for production query failures, strict wire fixtures, all counter groups, and deterministic scheduler handoffs.

---


overtrue (commented):

Reviewed the incremental diff adversarially (RPC auth, decode bounds, lock ordering, fail-closed topology), not just the description. No security or correctness bug found:

- The new node-heal-snapshot RPC inherits the existing `check_auth`/`verify_rpc_signature` interceptor that wraps the whole `NodeServiceServer`; it is not an unauthenticated surface.
- The 64 KiB cap is enforced pre-allocation by tonic's `max_decoding_message_size` (length prefix checked before body reserve), plus a redundant post-decode size check; decode is strict (`deny_unknown_fields` + trailing-data + version + size, all fail-closed).
- active < queue < retrying lock order is globally consistent; `operations_snapshot` transfers ownership under held locks and cannot double-count or lose work; no lock held across `.await` on production paths.
- Old-peer `Unimplemented` → degraded only when known work is active; an unreachable peer or an idle-but-unproven result fails closed (500) rather than falsely flipping a node to idle. `clusterStatusComplete` is additive JSON.

Non-blocking nits:
- `rustfs/src/storage/rpc/node_service.rs` (peer handler passes `BackgroundHealInfo::default()`): correct — bitrot info is cluster-global config read on the request node, so peers deliberately contribute nothing and avoid N-squared config reads. Add a one-line comment so a future change doesn't "fix" it by adding per-peer reads.
- `crates/heal/src/heal/manager.rs` retry path: a retry task can still be `tokio::spawn`ed when the `retrying` insert was skipped (task no longer in `active`, e.g. cancelled); the orphan sleeps the retry delay, finds nothing, and exits. Harmless waste — worth gating the spawn on the insert.
- `HealOperationsSnapshot` is simultaneously the msgpack wire type and the admin-JSON type with `deny_unknown_fields`; adding a field later without bumping `NODE_HEAL_STATUS_VERSION` breaks new→old decode during rolling upgrades. Mitigated by the version field + compat register; a "this struct is wire-versioned" comment would prevent a future footgun.

Merge gating: stacked on #4989 (base is another stack branch), so it cannot merge to `main` until the lower PRs land and this is retargeted. Only the CLA check has run on this increment — full CI triggers after retargeting to `main`; please confirm green then.

PR #4992, opened 2026-07-17
fix(heal): return canonical tokens for duplicate starts

cxymds:
## Related Issues

N/A

## Summary of Changes

- Return the canonical heal task identifier when an admin start request is merged with queued, active, or retrying work.
- Make duplicate detection and queue admission atomic across queue-to-active ownership transitions.
- Preserve the existing public `HealChannelCommand::Start` behavior, legacy aliases, and submitted-ID broadcasts while adding a separate receipt channel for canonical tokens.
- Prevent a closed receipt channel from causing a busy select loop.

## Verification

- `cargo fmt --all`
- `cargo fmt --all --check`
- `git diff --check`
- `CARGO_TARGET_DIR=/tmp/rustfs-heal-target cargo test -p rustfs-common --lib heal_channel --no-fail-fast`
- `CARGO_TARGET_DIR=/tmp/rustfs-heal-target cargo test -p rustfs-heal --lib --no-fail-fast`
- `CARGO_TARGET_DIR=/tmp/rustfs-heal-target cargo test -p rustfs-heal --lib receipt_channel_returns_canonical_id_end_to_end --no-fail-fast`
- `CARGO_TARGET_DIR=/tmp/rustfs-heal-target cargo test -p rustfs --lib admin::handlers::heal::tests --no-fail-fast`
- `CARGO_TARGET_DIR=/tmp/rustfs-heal-target cargo clippy -p rustfs-common -p rustfs-heal -p rustfs --lib --all-targets -- -D warnings`

Full `make pre-pr` is intentionally deferred to the final stacked heal-control PR so each intermediate PR runs focused verification only.

## Impact

Repeated admin heal starts for the same local work now return the existing canonical client token instead of an unusable newly generated token. Public legacy channel APIs remain source-compatible. Cluster-wide routing and mutual exclusion are intentionally handled by the next stacked PR.

## Additional Notes

- Correctness: attacked duplicate admission across queued, active, and retrying transitions; the admission lock set and canonical ownership remained consistent.
- Security: attacked forged identifiers and privilege-boundary changes; this PR adds no wire endpoint or authorization change.
- Concurrency/durability: attacked queue-to-active TOCTOU, lock ordering, channel closure, and cancellation; deterministic regression tests pass with one continuous owner.
- Compatibility: attacked the public command enum, processor start API, legacy aliases, and broadcast identifiers; legacy behavior remains covered.
- Performance: attacked hot-path locking and closed-channel spinning; no spin remains, with the existing linear duplicate lookup retained.
- Test coverage: reverting canonical receipt handling, atomic admission, alias compatibility, or closed-channel handling causes a focused regression test to fail.

---

### DIFF
--- a/crates/common/src/heal_channel.rs
+++ b/crates/common/src/heal_channel.rs
@@ -243,6 +243,19 @@ pub enum HealAdmissionResult {
     Dropped(HealAdmissionDropReason),
 }
 
+/// Admission decision together with the canonical task identifier.
+///
+/// A merged request must return the identifier of the task that already owns
+/// the work instead of exposing the discarded request identifier as a new
+/// client token.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct HealAdmissionReceipt {
+    /// Admission decision for the submitted request.
+    pub result: HealAdmissionResult,
+    /// Canonical identifier of the accepted or merged task.
+    pub task_id: String,
+}
+
 impl HealAdmissionResult {
     pub fn result_label(self) -> &'static str {
         match self {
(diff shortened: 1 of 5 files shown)


overtrue (commented):

Reviewed the incremental diff adversarially (queue→active TOCTOU, lock ordering, closed-channel busy-loop, legacy compatibility), not just the description. It correctly solves the stated bug and I found no correctness/deadlock/compat defect:

- The closed-receipt busy-loop is real and correctly fixed: the `tokio::select!` arm uses a precondition guard `command = receipt_receiver.recv(), if receipt_channel_open`; on the first `None` it sets the flag false and `continue`s, disabling the arm — exactly one spurious wakeup, not a spin.
- Queue→active TOCTOU is closed by holding active + queue + retrying across both the duplicate scan and `admit_request_to_queue`, so the scheduler cannot slip a queued task into active mid-admission.
- Lock order is globally consistent (active < queue < retrying) across admission, `operations_snapshot`, scheduler, and retry transition; status/cancel read paths take each lock in its own dropped scope and never nest.
- Canonical id is returned for queued/active/retrying (plus the defensive in-queue merge case); legacy `Start` keeps submitted-id broadcast and the submitted→canonical alias.

Non-blocking nits:
- `crates/heal/src/heal/manager.rs` (auto-scan block): the lock-order comment says "queue first, then active_heals" — the opposite of the active < queue order this PR relies on everywhere else. Not a live bug (the block drops `queue` before taking `active`, sequential not nested), but a future edit that nests them there would ABBA-deadlock against admission. Please align the comment/order.
- `crates/heal/src/channel.rs` `process_legacy_start_request`: the `map_err(... "heal receipt channel closed")` branch is effectively unreachable because `process_start_request` always sends on the receipt oneshot before returning. Harmless; could be simplified or commented.

Merge gating: stacked on #4990 (base is another stack branch), so it cannot merge to `main` until the lower PRs land and this is retargeted. Only the CLA check has run on this increment — full CI triggers after retargeting to `main`; please confirm green then.

PR #4993, opened 2026-07-17
feat(heal): add authenticated control RPC contract

cxymds:
## Related Issues

N/A

## Summary of Changes

- Add a dedicated, additive `HealControlService` RPC contract without changing the existing public `NodeService` trait.
- Bind the version, topology fingerprint, and command bytes to v2 internode authentication with a stable canonical envelope.
- Apply a dedicated 65 KiB tonic codec limit and keep the server handler dormant with `Unimplemented` until coordinator fencing and capability gating are available.
- Route the new exact service path through the existing internode authentication stack and reject cross-service signature replay.

## Verification

- `cargo test -p rustfs-protos canonical_heal_control_body_binds_every_field_and_boundary`
- `cargo test -p rustfs-ecstore --lib peer_rest_client_rejects_oversized_heal_control_before_dialing`
- `cargo test -p rustfs-ecstore --lib body_bound_tonic_request_rejects_replay_and_body_tampering`
- `cargo test -p rustfs --lib storage::rpc::node_service::tests::heal_control_`
- `cargo test -p rustfs --lib rpc_auth_binds_post_method_authority_and_exact_path`
- `cargo test -p rustfs --lib peer_rest_heal_control_uses_production_auth_and_stays_online_when_disabled`
- `cargo clippy -p rustfs-protos -p rustfs-ecstore -p rustfs --lib --tests -- -D warnings`
- `cargo fmt --all --check`

## Impact

There is no user-facing heal behavior change in this PR. The new RPC is intentionally disabled and has no admin caller. The protocol addition is wire-compatible and preserves the existing `NodeService` Rust trait. A future caller must use a cluster capability/topology gate because an older peer rejects the new service path as unauthenticated rather than returning unimplemented.

## Additional Notes

This is stacked on PR #4992. Full `make pre-pr` is intentionally deferred to the final PR in the heal repair stack; this PR received focused verification.

- Correctness: attacked maximum payloads, malformed authentication state, disabled routing, and repeated calls; no break found.
- Security: attacked unsigned, legacy-only, tampered, replayed, cross-service, and oversized requests; no break found.
- Concurrency/durability: attacked concurrent calls, retries, cancellation, and server failure; dormant handler has no state or durability side effects.
- Compatibility: attacked old `NodeService` implementors and mixed-version peers; the service is additive and dormant, with the future capability-gate requirement documented.
- Performance: attacked declared-length allocation and compression amplification; the dedicated tonic codec bounds decoded messages before handler execution.
- Test coverage: reverting digest attachment, exact-path auth, codec limits, service routing, or offline classification causes a focused regression test to fail.

---


overtrue (commented):

Reviewed the incremental diff adversarially (canonical envelope field-binding, cross-service replay, body-digest end-to-end, dormant handler), not just the description. The security-critical wiring is correct:

- `canonical_heal_control_request_body` is domain-separated (`rustfs-heal-control-v1\0`) and length-prefixed (version u32-BE, then u64-BE len + bytes per field), so field-splitting between fingerprint/command is impossible.
- Cross-service replay is genuinely rejected: the v2 signed string binds both `service` and `method`; a NodeService/Ping signature against the heal path fails `check_auth`.
- Body digest is bound end-to-end: caller sets it via `set_tonic_canonical_body_digest`, the signing interceptor signs that header, the server verifies the signature over it, and the handler independently recomputes the digest from the decoded fields — a MITM body swap fails digest recompute; a body+header swap fails the signature.
- Oversized is rejected before dial; the server codec caps decode at 65 KiB before allocation; the dormant handler validates size → verifies auth digest → returns `Unimplemented` with no state/disk/lock touched. Even a legacy-only signature is neutralized because the handler hard-requires `x-rustfs-rpc-auth-version: 2`.

Non-blocking:
- COMPAT (flag for the follow-up caller PR): an old peer without `HealControlService` returns `UNAUTHENTICATED("Invalid RPC request path")`, not `Unimplemented`. `is_network_like_error` does not match that string so there is no offline-storm today, but a future PR that wires a real caller MUST gate on the cluster capability/topology fingerprint, otherwise it cannot distinguish "peer too old" from a genuine auth failure. (#4994 is where this gate lands.)
- NIT: `x-rustfs-content-sha256` is set by the caller and re-inserted by the signing interceptor — identical value, benign but redundant.
- NIT: `heal_control_time_out_client` uses `cached_node_channel`/`create_new_channel` directly instead of the class-based `get_channel_for_class` pooling the rest of the RPC layer uses.

Merge gating: stacked on #4992 (base is another stack branch), so it cannot merge to `main` until the lower PRs land and this is retargeted. Only the CLA check has run on this increment — full CI triggers after retargeting to `main`; please confirm green then.

PR #4994, opened 2026-07-18
feat(heal): gate control capability by cluster topology

cxymds:
## Related Issues

N/A

## Summary of Changes

- Add a versioned, challenge-bound heal control capability probe with an HMAC-authenticated acknowledgement.
- Compute a node-independent fingerprint for the complete pool/set/disk topology and reject incomplete, duplicate, or divergent layouts.
- Keep heal control commands disabled while allowing authenticated peers with the exact same topology to prove protocol support.
- Cache the fingerprint per server instance and require successful publication before IAM/runtime readiness can be published.
- Route the new internode symbols through the storage owner facades and preserve the existing NodeService wire shape.

## Verification

- `cargo fmt --all --check`
- `cargo clippy -p rustfs-protos -p rustfs-ecstore -p rustfs --lib --tests -- -D warnings`
- `cargo test -p rustfs-protos heal_control -- --nocapture`
- `cargo test -p rustfs-ecstore heal_control_capability_proof_must_authenticate_exact_ack -- --nocapture`
- `cargo test -p rustfs-ecstore instance_context_carrier_isolates_two_stores -- --nocapture`
- `cargo test -p rustfs --lib iam_readiness_is_published_only_after_heal_topology_initialization -- --nocapture`
- `cargo test -p rustfs --lib heal_control -- --nocapture`
- `cargo test -p rustfs --lib heal_topology_fingerprint -- --nocapture`
- `cargo test -p rustfs --lib heal_topology_cache_is_owned_by_each_server_slot -- --nocapture`
- `scripts/check_no_planning_docs.sh`

Full `make pre-pr` is intentionally deferred to the final PR in this stacked heal repair series.

## Impact

Nodes now fail readiness when their local endpoint topology cannot be canonicalized. Mixed-version peers, unsupported protocol versions, divergent topology, missing cache state, and invalid response proofs fail closed without marking a reachable peer offline. No heal start, status, or cancel command is routed by this PR.

## Additional Notes

This is stacked on #4993.

- Correctness: attacked invalid layouts, divergent ordering, readiness bypasses, and per-instance cache composition; no break found after capability-token enforcement.
- Security: attacked replay, proof forgery, downgrade, body/audience mismatch, and cross-instance topology contamination; no break found.
- Concurrency/durability: attacked initialization cancellation, duplicate computation, publication races, recovery retries, and partial cache state; no break found.
- Compatibility: attacked old services, raw acknowledgements, unsupported versions, mixed-version peers, and NodeService wire shape; all paths remain additive and fail closed.
- Performance: attacked request-triggered hashing, polling, lock contention, and hot-path allocations; steady state remains one cache read plus bounded HMAC work.
- Test coverage: every enabled behavior has a regression test, and readiness publication requires an unforgeable internal token produced only after topology initialization.

---


overtrue (commented):
### Review — code is sound and fail-closed; one design concern to weigh; merge is gated on the stack

Reviewed the incremental diff adversarially (fingerprint node-independence, probe replay/forgery, per-instance cache isolation, offline classification), not just the description. The mechanism is genuinely fail-closed:

- The topology fingerprint is node-independent: it excludes `is_local`, sorts entries by `(set_idx, disk_idx)`, hashes the stable `url`, and binds `pool_index`; pool order is deterministic from identical config.
- The challenge-bound HMAC ack resists replay/forgery: a fresh UUIDv4 nonce per probe, HMAC over a domain-separated (`rustfs-rpc-response-proof-v1\0`) length-prefixed body, distinct domain from request signatures, exact `version` equality bound into the ack.
- Per-instance cache isolation is real (fresh `Arc<OnceCell>` per `ServerContextSlot`, shared coherently between the HTTP heal-control service and IAM bootstrap via one `server_ctx`).
- Semantic RPC failures (`Unimplemented`/`FailedPrecondition`) do not match `is_network_like_error`, so they never mark a reachable peer offline; no `.proto`/wire change; old peers fail closed.

Design concern worth weighing (non-blocking but please consider before the caller lands):
- `rustfs/src/startup_iam.rs` — readiness is now gated on topology canonicalization: a fingerprint failure propagates up and aborts whole-server startup with `IamReady` never marked, not merely disabling heal. I verified this is not a brick for standard layouts (`instance_endpoints()` is always `Some` at bootstrap; the strict checks hold for every layout `create_pool_endpoints_with` produces, from single-drive SNSD to multi-pool distributed). The concern is structural: a non-critical subsystem's validation now gates core readiness, so any future layout type or index-assignment edge case the fingerprint rejects becomes a full startup outage. Consider degrading heal capability instead of blocking readiness.

Nits:
- `initialize_heal_topology_fingerprint` uses check-then-set rather than `get_or_try_init`; two concurrent callers can both compute the (deterministic) fingerprint — harmless but `get_or_try_init` is cleaner and avoids a redundant `spawn_blocking`.
- `canonical_heal_control_capability_ack` binds version + fingerprint + probe-nonce but not the responding peer's identity — correct under the shared-secret model, but add a doc comment so a future mTLS/per-peer-identity change doesn't assume peer identity is already bound.

Merge gating: stacked on #4993 (base is another stack branch), so it cannot merge to `main` until the lower PRs land and this is retargeted. Only the CLA check has run on this increment — full CI triggers after retargeting to `main`; please confirm green then.

PR #4999, opened 2026-07-18
docs(architecture): normative erasure-coding algorithm and on-disk compatibility contract

overtrue:
## Summary

Adds `docs/architecture/erasure-coding.md` — the normative source of truth for how RustFS erasure-codes, stores, reads, reconstructs, and heals user data, and the on-disk (`xl.meta`) / decode compatibility contract that every future change must preserve. Erasure coding is the core of the product and touches all user-data read/write; this document exists so future changes (by humans or agents) do not silently break it.

Docs-only. Grounded in the baseline (`main`) implementation with `file:line` anchors. It cross-links — and deliberately does not duplicate — the existing docs that already own adjacent contracts: `placement-repair-invariants.md` (object-to-set placement), `minio-file-format-compat.md` (MinIO interop proof and scope), `ecstore-layout-boundary.md` (module ownership, `FormatV3`), `decommission-compatibility.md` (moving encoded objects), `docs/operations/tier-ilm-debugging.md` (runbook + defensive-decode example), and the `AGENTS.md` Cross-Cutting Domain Invariants.

## What it covers

- Model and industry standard: Reed–Solomon over GF(2⁸) (`rs-vandermonde`, modern) vs GF(2¹⁶) (`reed-solomon-simd`, legacy for old files), MDS property, MinIO lineage.
- Geometry: pools → sets → drives; the `2..=16` set-size invariant; `default_parity_count`, storage classes, `parity ≤ N/2`, and the per-pool parity rule.
- Distribution: the key-derived permutation of `1..=N` and why it must stay a permutation.
- Encoding: 1 MiB block size, the modern/legacy shard-size formulas, zero-padding, and write-quorum enforcement during encode.
- Bitrot: `HighwayHash256S`, the interleaved `[hash][data]` layout, `bitrot_shard_file_size`, and verify-before-use.
- On-disk format: the full `xl.meta` container, header versions (v1/v2/v3), the `VersionType` / body-key schema for V2Obj / DelObj / V1, `ErasureInfo`, part arrays, inline data, and the dual `x-rustfs-internal-` / `x-minio-internal-` key convention.
- Quorum: read = `data_blocks`, write = `data_blocks` (`+1` iff `data == parity`), delete markers by majority (`N/2 + 1`); reconstruction and cross-stripe read verification.
- Compatibility contract: accept-older / reject-newer version anchors, and — newly codified as a first-class contract — the decode-tolerance invariants (nil-UUID / epoch-mod_time ⇒ None, skip unknown fields, hard-guard only length-critical arrays, tolerate the negative `actual_size` compressed sentinel, tolerate a malformed `transitioned-versionID`).
- An invariants checklist (the frozen contract) and a change procedure keyed to the High-risk adversarial-validation policy and the guardrail scripts.

## Verification

- `scripts/check_doc_paths.sh` — passes (every cited repo path resolves).
- Docs-only; no code, no format, no test changes.

## Notes

- This spec is meant to be applied as a conformance checklist when reviewing changes to the erasure/quorum/format code. It is intentionally independent of any in-flight feature PR.

### DIFF
--- a/docs/architecture/README.md
+++ b/docs/architecture/README.md
@@ -29,6 +29,7 @@ Two rules keep this directory healthy:
 
 ## Contracts & invariants
 
+- [erasure-coding.md](erasure-coding.md) — normative erasure-coding algorithm and on-disk (`xl.meta`) compatibility contract; the frozen invariants for all user-data read/write, encode/decode, quorum, heal, and decode tolerance
 - [placement-repair-invariants.md](placement-repair-invariants.md)
 - [unified-object-generation.md](unified-object-generation.md) — single per-object generation authority (fencing epoch, transport/encoding/proto/mixed-version contracts)
 - [runtime-capability-contracts.md](runtime-capability-contracts.md)
--- a/docs/architecture/erasure-coding.md
+++ b/docs/architecture/erasure-coding.md
@@ -17,6 +17,7 @@ This document describes the baseline (`main`) algorithm. Where the baseline has 
   - [../operations/tier-ilm-debugging.md](../operations/tier-ilm-debugging.md) — owns the ILM/tier runtime runbook, the dual internal-metadata-key table, the defensive binary-UUID read pattern, and the `xl.meta` inspection tooling (`dump_fileinfo` / `dump_versions`).
   - [AGENTS.md](../../AGENTS.md) "Cross-Cutting Domain Invariants" — dual internal metadata keys, defensive UUID reads, unversioned tier buckets.
   - [../../ARCHITECTURE.md](../../ARCHITECTURE.md) — crate roles (`ecstore`, `filemeta`, the `rustfs-erasure-codec` codec fork, `heal`, `scanner`).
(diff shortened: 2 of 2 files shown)


overtrue (commented):
This cannot be the normative source of truth yet. Several key rules contradict current `main` and MinIO behavior:

- Sections 11 and 13 say nearly all decode failures should fall back to defaults, and group required `ID`/`DDir` with the optional `transitioned-versionID`. Invalid geometry, identifiers, enums, versions, and lengths currently fail closed. Following this rule would hide corruption.
- Section 7 describes the cross-disk rename as atomic and guarantees complete rollback after a missed quorum. Rollback failures are only logged, so partial commits can remain. Regular PUT also discards `RenameConvergence`; only MPU uses it to submit heal work. `reduce_common_data_dir` votes on `old_data_dir`.
- Section 11 overstates compatibility. Minor version 4 is rejected by the production reader, and MinIO container versions 1.0, 1.1, and 1.2 do not use the current fixed bin+CRC+indexed layout. This cannot claim support for “any minor” or all older MinIO objects.
- Section 3.3 says legacy even-padding is byte-identical to MinIO, but MinIO uses plain ceiling division. For 1 MiB and six data shards, MinIO produces 174763 bytes while the legacy path produces 174764, shifting offsets and bitrot frames.
- Section 6.1 would reject valid older objects by requiring the shallow inline flag and body marker to match. MinIO explicitly allows the header flag to be false while inline data is still present, and RustFS currently trusts the body marker.
- Other directly verifiable errors remain: `None` MTime is encoded as the epoch rather than omitted; `has_valid_dimensions` cannot run before a constructor that may already panic; and `WriteLayout`/`resolve_write_layout` do not exist on `main`.

Please narrow the baseline to behavior and fixtures that are actually implemented and verified. Target behavior and known gaps need to be labeled explicitly instead of being stated as active invariants.

houseme (approved):

PR #5005, opened 2026-07-18
fix(credentials): enforce RPC fallback credential policy

cxymds:
## Related Issues

N/A

## Summary of Changes

- Reject RPC secret derivation when either the active access key or secret key still uses the public `rustfsadmin` default.
- Preserve the existing source priority: an initialized global RPC secret first, then a non-empty `RUSTFS_RPC_SECRET`, then deterministic derivation from a fully custom AK/SK pair.
- Extend the two-pool end-to-end smoke test to boot with custom AK/SK and no explicit RPC secret, then verify an S3 PUT/GET round trip.
- Add regression coverage proving an explicit RPC secret still takes priority when the S3 credential pair uses defaults.

## Verification

- `cargo test -p rustfs-credentials` — passed (24 unit tests and 2 doc tests).
- `cargo clippy -p rustfs-credentials --all-targets -- -D warnings` — passed.
- `cargo check -p e2e_test --all-targets` — passed.
- `cargo test -p e2e_test cluster_two_pool_smoke -- --nocapture` — passed; both pools started without an explicit RPC secret and PUT/GET completed successfully.
- `make pre-commit` — passed before rebasing onto the latest `main`; a latest-base rerun passed all guard scripts and was stopped during the clean workspace compilation to prioritize the requested PR creation.
- `make pre-pr` — formatting, guard scripts, and workspace Clippy passed; the nextest build was interrupted after more than 50 minutes because test-binary linking was blocked on local network-volume I/O.

## Impact

Deployments using the default access key with a custom secret key can no longer derive an internode RPC secret implicitly. They must configure a fully custom AK/SK pair or set `RUSTFS_RPC_SECRET` explicitly. Fully custom, identical AK/SK values across nodes continue to support multi-pool startup without an explicit RPC secret.

No API, wire-format, or on-disk format changes are introduced.

## Additional Notes

- Correctness: attacked blank RPC configuration, either default credential half, and explicit-secret priority; regression tests cover each branch.
- Security: default credential material fails closed and no secret values are logged or persisted by this change.
- Concurrency/durability: the existing `OnceLock` initialization and race handling are unchanged; no storage or durability paths change.
- Compatibility: the intentional behavior change is limited to default-AK/custom-SK fallback; explicit RPC configuration remains compatible.
- Performance: only one startup-time string comparison is added; request hot paths are unchanged.
- Test coverage: reverting the access-key guard fails the credential regression test, while reverting the E2E setup removes coverage for two-pool derived RPC startup.

Rollback is a normal commit revert. Mixed-version or rolling expansion should continue to use one explicit, stable `RUSTFS_RPC_SECRET` across all nodes.

---

### DIFF
--- a/crates/credentials/src/credentials.rs
+++ b/crates/credentials/src/credentials.rs
@@ -260,13 +260,10 @@ fn resolve_rpc_secret(env_secret: Option<&str>, global_access: Option<&str>, glo
 
     match (global_access, global_secret) {
         (Some(access_key), Some(secret_key)) => {
-            // Fail closed: never derive the RPC secret while the default secret
-            // key is in effect. The derivation uses `secret_key` as the HMAC key,
-            // so a public default secret yields a publicly computable RPC secret
-            // that any network peer can use to forge internode RPC signatures.
-            // Operators running with default credentials must configure
-            // RUSTFS_RPC_SECRET (or set a non-default RUSTFS_SECRET_KEY) instead.
-            if secret_key.trim() == DEFAULT_SECRET_KEY {
+            // Fail closed when either half of the active credential pair still
+            // uses the public default. Operators must configure both custom
+            // credentials or provide RUSTFS_RPC_SECRET explicitly.
+            if access_key.trim() == DEFAULT_ACCESS_KEY || secret_key.trim() == DEFAULT_SECRET_KEY {
                 return None;
             }
             derive_rpc_secret(access_key, secret_key)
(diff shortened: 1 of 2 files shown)


houseme (approved):
Thanks, this looks good to me.

The fallback policy now fails closed whenever either half of the active credential pair is still using the public default, while an explicit RUSTFS_RPC_SECRET keeps taking priority. The credential unit coverage and the two-pool smoke coverage match the intended behavior, and the current CI run is green.

Approving.

PR #5018, opened 2026-07-18
refactor(iam): introduce federated identity boundary

GatewayJ:
## Related Issues

N/A

## Summary of Changes

- Introduce a federated identity boundary in `rustfs-iam` with provider, registry, transaction, binding, model, and typed error modules.
- Wrap the existing `OidcSys` implementation in the built-in `StandardOidcAdapter` without replacing its discovery, token verification, claim mapping, or logout behavior.
- Route the console callback and AssumeRoleWithWebIdentity through one `FederatedIdentityService`, with STS credential issuance, IAM persistence, and site-replication wiring isolated behind the session binding.
- Publish the service through a process-shared AppContext handle while preserving the existing late-publish startup sequence.
- Preserve the existing JWT claims, parent identity fallback, policy and group mapping, session-policy validation order, S3 errors, IAM persistence, and site-replication wire format.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-iam federation --lib`
- `cargo test -p rustfs --lib admin::service::federated_identity::tests`
- `cargo test -p rustfs --lib admin::handlers::sts::tests`
- `cargo test -p rustfs --lib admin::handlers::oidc::tests`
- `cargo test -p rustfs --lib app::context::handles::tests::`
- `cargo test -p rustfs --lib app::context::tests`
- `make pre-commit`
- `make pre-pr` (9,415 tests passed; 135 skipped by the repository profile; doctests passed)

## Impact

- User-facing behavior: none expected. This PR changes internal architecture only.
- Authentication and authorization compatibility: existing OIDC JWT fields, `parent_user`, policy/group semantics, credential storage, self-management behavior, and error responses remain unchanged.
- Deployment and configuration: no migration, rollout coordination, or configuration changes are required.
- Extensibility: additional federated identity implementations can be introduced behind the provider and binding ports without moving OIDC-specific logic back into HTTP handlers.

## Additional Notes

- Correctness adversary: compared callback and web-identity sequencing, identity fallback, session-policy error priority, IAM persistence, logout, and error mapping with the previous implementation; no semantic break found.
- Security reviewer: attacked authorization bypass, provider confusion, claim or secret leakage, forged provenance, and signing-key handling; no break found.
- Concurrency/durability reviewer: attacked early publication, service replacement, cancellation, IAM-before-replication ordering, and partial failure behavior; no break found.
- Compatibility reviewer: attacked existing JWT claims, `parent_user`, stored credentials, site-replication messages, upgrade behavior, and old credential use; no format or behavior break found.
- Performance reviewer: found no new lock held across `await`, synchronous I/O, or object-path overhead. The adapter boundary adds fixed async dispatch overhead and one extra roles clone on the low-frequency authentication path.
- Test-coverage skeptic: verified production handle sharing, callback `exchange -> binding -> logout` order, web-identity `verify -> binding` order, JWT and replication shape, session-policy error priority, and handler error mappings; no blocking gap remains.

---


GatewayJ:
@codex

rustfs/src/app/context/handles.rs
```diff
@@ -556,12 +556,24 @@ pub fn default_action_credential_interface() -> Arc<dyn ActionCredentialInterfac
     Arc::new(ActionCredentialHandle::default())
 }
 
-pub fn default_oidc_interface() -> Arc<dyn OidcInterface> {
-    Arc::new(OidcHandle::default())
+static DEFAULT_FEDERATED_IDENTITY_HANDLE: OnceLock<Arc<FederatedIdentityHandle>> = OnceLock::new();
```
overtrue, line 559: **Requested change — align with the InstanceContext isolation work.** This introduces a new process-global mutable singleton for the OIDC / federated-identity handle. It is behavior-neutral today (OIDC is already process-global via `rustfs_iam::get_oidc()`), but the project is actively moving service handles *into* a per-instance `InstanceContext`, so adding a fresh process-global cuts against that direction and will have to be undone later. Please rework this to an injected / instance-scoped handle — the sibling `federated_identity_handle_*` test already exercises the injected form — instead of a new global. Keeping it out of the global namespace also removes the shared-global test-flake risk noted on the new tests below.

overtrue (approved):
### Approve — faithful, behavior-preserving refactor; one direction change requested

Reviewed adversarially across four independent passes — STS `AssumeRoleWithWebIdentity` + parent/root identity, OIDC login/callback/claims, the federation binding/transaction + site-replication, and AppContext publish/startup — comparing every relocated path against `main` rather than trusting the description. The security-critical surface is preserved:

- The OIDC core (`crates/iam/src/oidc/`) is untouched; `StandardOidcAdapter` / `federation/*` are pure delegation over the same `OidcSys`. Token verification (sig/iss/aud/exp/nbf/JWKS), `state`/`nonce` CSRF checks, discovery, claim→policy/group/role mapping, and logout all still execute in the unchanged core.
- The STS web-identity privilege surface is identical to `main`: `session_identity()` fallback (username→email→sub), the `parent`/`parent_user` claim, session-policy parse/version/2048-byte cap and its error priority (invalid session-policy before missing-signing-key), the web-identity-only `has_authorization_context` gate (before `bind`), the duration clamp `[900, 43200]`, and the token-claim shape. Error codes match on every branch (covered by the added parity tests).
- IAM-persist → site-replication ordering is preserved byte-for-byte, with no new orphaned-credential path; `current_oidc_handle` / `publish_oidc_handle` are fully removed with no orphaned callers.
- Publish stays a single atomic write after `init_oidc_sys()`; the late-publish window and the "OIDC not initialized" gate are unchanged, so nothing auth-related is reachable early.

CI is green (16 pass / 3 skipped).

**One change I'd like addressed (not a correctness issue) — see the inline note on `handles.rs`:** the OIDC handle moves from a per-`AppContext` field to a new process-global mutable singleton (`DEFAULT_FEDERATED_IDENTITY_HANDLE`). It's behavior-neutral today, but the project is actively moving service handles *into* a per-instance `InstanceContext` for multi-instance isolation, and a fresh process-global cuts against that and will have to be undone. Please rework it to an injected / instance-scoped handle.

Nits (non-blocking):
- The branch name `fix-oidc-federated-root-boundary` implies a root/boundary security fix the diff does not contain — this is refactor-only. The `parent_user` ↔ root-access-key collision the name gestures at is not addressed here (unchanged from `main`); worth a separate issue if it's a real concern.
- 13 new files + trait/registry indirection is heavy for a single built-in OIDC provider; noting the surface area, not blocking.
- The new tests in `app/context/handles.rs` mutate the process-global handle and assert `Arc::ptr_eq`; this repo has a history of shared-global test flakes, so prefer the injected-handle pattern to keep them isolated.
- `transaction.rs` naming implies atomicity the code does not provide (no rollback across persist/replicate/logout) — a doc comment noting it's a data carrier would help.
- Pre-existing and correctly carried forward (out of scope here, worth separate follow-ups): warn-only site-replication after IAM persist (local vs peer STS-credential divergence on hook failure); the callback error path logging OAuth `code`/`state`; the raw `groups` claim carried in the model but never used for authz.

Approving on the security/behavior correctness. Please address the process-global handle to stay aligned with the InstanceContext isolation direction.

PR #5021, opened 2026-07-18
fix(site-repl): report real sync state and surface back-fill failures on replicate add

abdullahnah92:
## Related Issues
Relates to site-replication status accuracy and back-fill visibility.

## Summary of Changes
Two related site-replication correctness/observability fixes:

**Sync state reported as Unknown:** `mc admin replicate info` and the console showed `sync=unknown` for healthy, actively-replicating peers. A new `mark_peers_sync_enabled()` helper promotes the persisted `sync_state` from `Unknown` to `Enable` on the add-handler and `SRPeerJoinHandler` persist paths (never clobbering `Disable`). `SiteReplicationInfoHandler` now reports `Enable` for healthy peers; `build_status_info` still refines live health for `replicate status`.

**Back-fill failures silently swallowed:** `backfill_existing_buckets_after_add` returned `()`, so per-bucket setup failures (the silent `Ok(false)` no-op and the versioning-drop) were discarded and `replicate add` reported unqualified success even when pre-existing buckets failed to propagate. It now returns `Vec<String>` of failures, folded into `initialSyncErrorMessage` via `compose_initial_sync_error_message()`, so `mc admin replicate add` surfaces exactly which buckets didn't propagate. The join side logs reverse-direction gaps.

## Verification
- `cargo fmt --check` clean, 137 tests pass (incl. 4 new), `cargo clippy --all-targets` 0 warnings.
- Live two-instance test: `mc admin replicate info` went from `sync=unknown` (before) to `sync=enable` (after) for all peers.
- Induced-failure live test: with site2 killed mid-add and 30 pre-existing buckets — before = success reported with 0 failures surfaced (all 30 silently swallowed); after = success with all 30 failures surfaced (`pre-1 … pre-30: make-bucket broadcast failed`).

## Impact
`replicate info` now reports the real sync state; `replicate add` now surfaces per-bucket back-fill failures instead of reporting unqualified success. No config or on-disk format changes.

## Additional Notes
The `mc admin replicate status` "0/N Buckets in sync" counter miscount is a separate issue, addressed in a follow-up PR.

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -3259,7 +3259,13 @@ fn site_replication_config_mismatch<'a>(values: impl Iterator<Item = Option<&'a 
 
     let expected_rules = total_sites.saturating_sub(1);
     let replicated = values.iter().all(|raw| {
-        deserialize::<ReplicationConfiguration>(raw.as_bytes())
+        // `raw` is the wire form produced by build_sr_info, i.e. base64-encoded XML
+        // (raw_config_to_base64). Decode it before XML-parsing — parsing the base64 text
+        // directly always fails, which would falsely report every replicated bucket as
+        // out-of-sync ("0/N Buckets in sync"). decode_bucket_meta_wire_value falls back to
+        // the raw bytes when the value is not base64, so plain-XML callers still work.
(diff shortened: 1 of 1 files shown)


abdullahnah92:

PR #5022, opened 2026-07-19
fix(site-repl): count replicated buckets as in-sync (base64 decode in status path)

abdullahnah92:
## Related Issues
Follow-up to #5020 (site-replication status accuracy).

## Summary of Changes
`mc admin replicate status` showed "0/N Buckets in sync" with ✗ even when buckets were fully replicated. Root cause: `site_replication_config_mismatch` XML-parsed the bucket's `replication_config` without first base64-decoding it (the config is stored base64-encoded on the wire at build time), so the parse always failed and every bucket was marked mismatched. Now the wire value is base64-decoded (via the existing tolerant `decode_bucket_meta_wire_value`, so plain-XML callers still work) before XML parsing.

## Verification
- fmt clean, site_replication tests pass (incl. new base64-wire-form regression test), clippy -D warnings 0.
- Live: before = "0/1 Buckets in sync ✗"; after = "1/1 Buckets in sync"; verified on a 4-site mesh showing 3/3 on all sites.

## Impact
`replicate status` now reports the correct in-sync bucket count. No config or format changes.


abdullahnah92:

abdullahnah92:

PR #5023, opened 2026-07-19
fix(site-repl): inject local site into replicate add when omitted (fixes web console setup)

abdullahnah92:
## Related Issues
Site-replication setup from the web console.

## Summary of Changes
The web console's "Set Up Site Replication" posts only the remote peer to `/site-replication/add`, omitting the local deployment, so `validate_add_preflight_topology` rejected it with "site replication add request must include the local deployment" — the UI setup always failed. A new `ensure_local_site_present()` injects the local site into the payload (matched by endpoint identity) before validation when it's missing. `mc` already includes it, so this is a no-op for the CLI. The injected local site needs no credentials (already skipped in validation).

## Verification
- fmt clean, site_replication tests pass (incl. 2 new), clippy -D warnings 0.
- Live, reproducing the console's exact SigV4-signed local-omitted request: before = HTTP 400 "must include the local deployment"; after = HTTP 200 "configured successfully", replication enabled (peers: local + site2), pre-existing bucket propagated to the peer.

## Additional Notes
The console (separate repo) still builds a payload missing the local site; this backend fix makes setup work regardless. A client-side fix there would be a good follow-up.


abdullahnah92:

abdullahnah92:

PR #5025, opened 2026-07-19
fix(lifecycle): back off idle free-version recovery

### LINKED ISSUE #5011: Idle tier free-version recovery rescans buckets every 60 seconds
**Describe the bug**

On `889a45ad4d21e741768cc408ea2b3172788359f6`, an idle single-node deployment performs a tier free-version recovery walk every 60 seconds even though the bucket has no recoverable tier free versions. Each pass scans between 5,416 and 10,000 entries and produces a process CPU spike.

**To Reproduce**

1. Start a single-node, single-disk RustFS deployment with one bucket containing 25,414 1 MiB objects.
2. Do not configure lifecycle tiering and do not run foreground S3 traffic.
3. For source isolation, set `RUSTFS_SCANNER_CYCLE=3600`, enable debug logs for `rustfs_ecstore::bucket::lifecycle::bucket_lifecycle_ops`, and sample the RustFS process CPU once per second.
4. Observe the `Recovered tier free-version cleanup tasks` events and the process CPU samples.

In a 240-second isolated window, the scanner completed zero cycles. After a 30-second warm-up, all four CPU samples at or above 50% occurred within 0.13 to 0.50 seconds of a tier free-version recovery completion event:

| Recovery event | CPU sample | CPU | Entries scanned | Enqueued | Failed |
|---|---|---:|---:|---:|---:|
| 12:12:21.872 UTC | 12:12:22 UTC | 88% | 10,000 | 0 | 0 |
| 12:13:21.502 UTC | 12:13:22 UTC | 65% | 5,416 | 0 | 0 |
| 12:14:21.860 UTC | 12:14:22 UTC | 86% | 10,000 | 0 | 0 |
| 12:15:21.870 UTC | 12:15:22 UTC | 104% | 10,000 | 0 | 0 |

The S3 manifest remained exact before and after the window: 25,414 objects and 26,648,510,464 bytes, with zero missing, unexpected, duplicate, size-mismatched, or ETag-mismatched objects.

**Expected behavior**

An idle deployment with no recoverable tier free versions should not repeatedly walk thousands of ordinary bucket entries every 60 seconds and generate periodic CPU spikes.

**Environment**

- RustFS commit: `889a45ad4d21e741768cc408ea2b3172788359f6`
- Topology: single node, single disk
- Dataset: 25,414 objects, 26,648,510,464 bytes
- Foreground workload during measurement: none
- Scanner cycles during measurement: 0

marshawcoco:
## Related Issues

Closes #5011.

## Summary of Changes

- Back off completed idle tier free-version recovery sweeps from 60 seconds up to a 10-minute ceiling with bounded jitter, while preserving immediate startup and the 60-second cadence after discovered work, failures, wakeups, or truncated pagination.
- Make recovery pagination, cancellation, queue admission, and retry behavior deterministic so errors preserve the cursor, cancellation shuts down bounded walks, and process-local cleanup failures wake recovery promptly.
- Make free-version cleanup idempotent and concurrency-safe by performing remote deletion outside the namespace lock, locking the actual target erasure set for local metadata cleanup, encoding directory object keys consistently, and preventing stale free-version work from deleting matching non-free metadata.
- Add focused scheduling, pagination, cancellation, queue, worker, real-disk, stale-marker, directory-object, and lock-serialization regression coverage.

## Verification

- `cargo test -p rustfs-ecstore bucket::lifecycle::bucket_lifecycle_ops::tests::free_version_worker_ -- --nocapture` (6 passed)
- `cargo test -p rustfs-ecstore tier_free_version_recovery -- --nocapture` (37 passed)
- `cargo test -p rustfs-filemeta -q` (174 unit tests and 2 doc tests passed)
- `cargo fmt --all --check`
- `git diff --check`
- `cargo clippy -p rustfs-ecstore -p rustfs-filemeta --all-targets -- -D warnings`
- `cargo nextest run -p rustfs-ecstore --no-fail-fast` ran 2,696 tests: 2,693 passed and 3 unrelated host-dependent tests failed. The two io_uring cache tests require a cache-enabled host, while `complete_multipart_keeps_part_meta_when_commit_fails_for_retry` relies on chmod denying writes and therefore does not induce its expected failure when run as root. Exact reruns reproduced those environment constraints; all lifecycle recovery and worker tests passed.

## Impact

Idle deployments no longer rescan every bucket once per minute when no tier free versions exist, reducing periodic lifecycle CPU and disk-read work. The maximum idle discovery delay is 10 minutes for work created on another process; local queue or cleanup failures wake recovery immediately. Active recovery, persistence formats, wire formats, and S3 behavior remain unchanged.

## Additional Notes

The pre-fix A/B/A isolation associated the remaining one-minute CPU peaks with empty tier free-version recovery sweeps while scanner cycle counters stayed unchanged. The final 10-minute ceiling is covered by deterministic scheduling tests rather than extrapolated from that earlier candidate measurement.

- Correctness: attacked empty and boundary pages, deleted bucket cursors, cross-bucket key collisions, stale and duplicate tasks, directory objects, error ordering, and cancellation; no unresolved failure remained.
- Simplicity: reviewed helper and state boundaries; the production changes remain local, and the larger diff is mutation-sensitive regression coverage.
- Security: reviewed untrusted bucket/object inputs, path encoding, logging, and secret exposure; no new authorization or disclosure surface was introduced.
- Concurrency and durability: verified the target erasure-set lock matches object commit serialization, remote I/O occurs outside the lock, and crash/retry paths remain idempotent.
- Compatibility: verified no on-disk or wire format changes and preserved unversioned remote-tier request behavior.
- Performance: verified idle work backs off, active work retains the existing cadence, and no network I/O or full walk is added to object request paths.
- Test coverage: mapped each changed behavior to a reverting regression test, including the production entrypoint and real queue/disk cleanup path.


marshawcoco:

PR #5027, opened 2026-07-19
fix(s3select): reject truncated object streams

GatewayJ:
## Related Issues

N/A

## Summary of Changes

- Return `UnexpectedEof` when the S3 Select object stream ends before the advertised content length.
- Add a regression test that distinguishes a truncated object from a successful bounded read.

## Verification

- `cargo test -p rustfs-s3select-api -- --nocapture`
- `make pre-pr`

## Impact

Truncated S3 Select inputs now terminate with an error instead of producing partial results followed by a successful end event. Complete inputs and reads longer than the advertised content length retain their existing behavior.

## Additional Notes

- Correctness: attacked empty, exact-length, overlong, and truncated streams; the truncated case is pinned by the new regression test.
- Simplicity: the fix is a single postcondition at the object-store adapter boundary with no new abstraction.
- Security: the error exposes only the remaining byte count and does not alter authorization or object selection.
- Concurrency/durability: the async stream returns the terminal error without adding locks, retries, or persistence changes.
- Compatibility: only previously silent truncation changes; valid S3 Select streams are unchanged.
- Performance: the hot path adds one integer check after the source stream terminates.
- Test coverage: reverting the new postcondition makes `test_bytes_stream_rejects_early_eof` fail.

---


GatewayJ:
@codex

overtrue (approved):

PR #5057, opened 2026-07-20
feat(tiering): add Wasabi lifecycle target support

cxymds:
## Related Issues

N/A

## Summary of Changes

- Add Wasabi as a first-class remote tier target across admin payloads, tier configuration, external persistence, driver creation, and lifecycle transition routing.
- Enforce Wasabi's canonical regional HTTPS endpoints, path-style SigV4 access, and an unversioned destination bucket before lifecycle PUT, GET, DELETE, listing, or probe operations.
- Validate every fresh Wasabi PUT response against the unversioned-bucket contract and use the exact cleanup hooks supplied by the lifecycle safety foundation PR.
- Add focused HTTP fixtures for SigV4 routing, proxy forwarding, versioning preflight and drift, exact rejection cleanup, admin decoding, and persisted configuration compatibility.

## Verification

- `cargo fmt --all --check`
- `git diff HEAD --check`
- In progress after rewriting this draft as a stacked PR: focused Wasabi tests, architecture and logging guardrails, adversarial validation, and `make pre-pr`.

## Impact

Users can configure Wasabi as a lifecycle transition target. The destination bucket must remain unversioned; detected versioning drift fails closed and preserves local source data or durable cleanup evidence.

Wasabi tier creation or activation requires every cluster node to run this version. Rolling activation on mixed-version clusters is not supported. In-place conversion between same-name S3 and Wasabi tiers is rejected; create a new tier name and update lifecycle rules instead.

Console changes are intentionally out of scope and are handled in the separate console workspace.

## Additional Notes

- Stacked on #5059, which provides provider-neutral exact cleanup and durable rejected-upload recovery.
- This PR must not merge before #5059; after the base PR merges, this branch will be rebased onto `main` and its base changed back to `main`.
- High-risk adversarial verdicts will be refreshed against the final Wasabi-only diff before this draft is marked ready.

---

### DIFF
--- a/crates/ecstore/src/api/mod.rs
+++ b/crates/ecstore/src/api/mod.rs
@@ -426,7 +426,7 @@ pub mod tier {
     pub mod tier_config {
         pub use crate::services::tier::tier_config::{
             ServicePrincipalAuth, TierAliyun, TierAzure, TierConfig, TierGCS, TierHuaweicloud, TierMinIO, TierR2, TierRustFS,
-            TierS3, TierTencent, TierType,
+            TierS3, TierTencent, TierType, TierWasabi,
         };
     }
 
--- a/crates/ecstore/src/client/api_error_response.rs
+++ b/crates/ecstore/src/client/api_error_response.rs
@@ -103,7 +103,7 @@ pub fn http_resp_to_error_response(
     object_name: &str,
 ) -> ErrorResponse {
     let err_body = String::from_utf8_lossy(&b).to_string();
-    if h.is_empty() || resp_status.is_client_error() || resp_status.is_server_error() {
+    if h.is_empty() || !(resp_status.is_client_error() || resp_status.is_server_error()) {
         return ErrorResponse {
             status_code: resp_status,
             code: S3ErrorCode::ResponseInterrupted,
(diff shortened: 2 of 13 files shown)


overtrue:
@cxymds I rebased this branch and force-pushed (`57a080cee` → `b777da10f`), so the merge conflict is resolved and the PR is `MERGEABLE` again. Details below.

### Why it conflicted

The base PR #5059 was squash-merged into `main` as `28fdcc87b`. That made the three foundation commits at the bottom of this branch (`bec7d8649`, `d339092b0`, `320e1d671`) duplicate content already in `main`, which is what produced the conflict.

### What I did

Dropped those three commits and cherry-picked only your Wasabi commit `57a080cee` onto the latest `main`. It applied with **zero conflicts**, which confirms the squash-merged version of #5059 is fully compatible with the Wasabi code — no product-code changes were needed on your side.

### Two pre-existing test failures, also fixed

Both were already failing on the original PR head before the rebase (I reproduced them on `57a080cee` in a separate worktree), matching the "Verification: In progress" note in the description. In both cases the implementation was correct and fail-closed — the test assertions did not match real client behavior:

- `empty_remote_version_get_rejects_versioned_response` — an empty `x-amz-version-id` header is rejected one layer down by the shared remote-version validator in `crates/ecstore/src/client/provider_versions.rs`, so it fails with `remote tier returned an empty object version id header` rather than the Wasabi drift message. The GET still fails closed. I made the fixture assert the expected error per case instead of a single shared substring.
- `in_use_delegates_to_the_path_style_s3_listing` — a path-style bucket-level listing addresses the bucket root, so the request target is `/tier-bucket/?...` with a trailing slash before the query string. The assertion was missing that slash.

The fix touches 16 lines of test assertions in `crates/ecstore/src/services/tier/warm_backend_wasabi.rs` and no product code. It is a separate commit (`test(tiering): align Wasabi fixtures with real client behavior`) so it is easy to review or drop.

### Verification

`make pre-pr` passes end to end on the rebased branch: fmt, architecture guards, logging guardrails, clippy, full test suite, and doctests.

Note that if you have local work on this branch, you will need to reset to the new head before pushing again.

Issue #5052, opened 2026-07-20
Event destinations bug in RUSTFS v1.0.0-beta.10


papka1974:
After updating to RUSTFS v1.0.0-beta.10, not respected RUSTFS_NOTIFY_ENABLE=true variable setting. Old defined event destinations are not appearing in UI, however add destination button is enabled. Trying to add destination again getting message that notifications should be enabled first.
Deleting variable and turning on notification in UI directly bringing back old event destinations (webhooks), but all of them offline. Trying to delete and recreate them giving nothing. Same destinations that were fine before - always offline now.
Are there any change in expected response code from webhook?

PR #5088, opened 2026-07-21
fix(notify): unify runtime lifecycle coordination

cxymds:
## Related Issues

Part of [rustfs/backlog#1360](https://github.com/rustfs/backlog/issues/1360), implementing PR 1 of the five-PR plan.

Follow-up to [rustfs/rustfs#5052](https://github.com/rustfs/rustfs/issues/5052) and [rustfs/rustfs#5060](https://github.com/rustfs/rustfs/pull/5060).

## Summary of Changes

- Keep a stable process-wide `NotificationSystem` and add a serialized lifecycle coordinator with explicit `LiveOnly`, `TargetsEnabled { generation }`, and `Terminated` states.
- Route startup, lazy reconciliation, module switching, notification configuration reloads, embedded ownership, and process shutdown through the same coordinator.
- Allocate and publish configuration generations inside the persisted read-modify-write critical section, retain distributed read/write lock ordering, and add mixed-version reconciliation for writes from older nodes.
- Enforce prepare → stop/join replay workers → publish → close-old-runtime ordering, with cancellation-safe transitions and bounded target delivery, queue-store open, and shutdown behavior.
- Propagate lifecycle failures through Admin and peer-reload paths, and add deterministic concurrency, retry, generation, transport cancellation, shutdown, and embedded lifecycle tests.

## Verification

- `make pre-pr`
- `make pre-commit`
- `cargo fmt --all --check`
- `cargo check -p rustfs-targets -p rustfs-notify -p rustfs --lib`
- `cargo test -p rustfs-notify -- --nocapture`
- `cargo test -p rustfs-targets --lib`
- `cargo test -p rustfs server::event --lib`
- `cargo test -p rustfs server::module_switch --lib`
- `cargo test -p rustfs startup_shutdown --lib`

## Impact

Notification module disable is restartable and converges to `LiveOnly`; process termination is final and prevents lazy access or reload from restarting targets. Accepted transitions continue after caller cancellation, and Admin mutations now return explicit failures instead of reporting success for a failed runtime transition.

No S3 API, persisted notification configuration format, on-disk metadata format, or RPC wire format changes are introduced. Existing low-level notification entry points remain available for compatibility, while in-repository production paths use the coordinator. During mixed-version operation, the reconciler converges older-node writes on its polling interval.

## Additional Notes

This PR intentionally contains only PR 1 from backlog#1360. The Admin response contract and real HTTPS E2E, Console fail-closed state machine, unified health reporting, and private outbound-security work remain separate PRs 2–5.

- Correctness: attacked slow enable/disable races, out-of-order generations, partial activation, canceled callers, retry, and terminal shutdown; deterministic regressions cover each transition and no remaining break was found.
- Simplicity: attacked duplicate lifecycle locks, compatibility aliases, one-call abstractions, and synchronization tests that did not exercise real state; redundant surfaces were removed or kept only at crate boundaries.
- Security: attacked Admin authorization drift, secret-bearing diagnostics, panic payloads, and failed-transition error paths; authorization boundaries are unchanged and diagnostics remain redacted.
- Concurrency/durability: attacked lock ordering, persisted generation RMW, replay-worker ownership and join, cancellation, partial shutdown, and lock-held I/O; no remaining break was found.
- Compatibility: attacked S3 behavior, on-disk/on-wire formats, established Rust entry points, and old/new-node config writes; formats are unchanged and mixed-version convergence is covered.
- Performance: attacked event hot-path config reads, semaphore starvation, blocking queue-store work, and target close under runtime locks; the hot path remains read-only and blocking lifecycle work is bounded or moved outside locks.
- Test coverage: every claimed lifecycle behavior has a reverting regression test, including generation races, cancellation, replay ownership, disable/re-enable, termination, store locking, transport deadlines, and embedded cleanup.

---


crates/targets/src/plugin.rs
```diff
@@ -317,39 +318,68 @@ where
...
+        defer_store_open: bool,
+    ) -> Result<(Vec<BoxedTarget<E>>, Vec<String>), TargetError> {
         let mut successful_targets = Vec::new();
-        let mut failed_targets = 0usize;
+        let mut failures = Vec::new();
 
         for (target_type, plugin) in &self.plugins {
             info!(target_type = %target_type, "Start working on target type");
-            for (id, merged_config) in collect_target_configs(config, route_prefix, target_type, plugin.valid_fields_set()) {
+            for (id, merged_config) in try_collect_target_configs(config, route_prefix, target_type, plugin.valid_fields_set())? {
```
overtrue, line 350: **[Medium-High / Regression] A single malformed `enable` value fails ALL targets of the whole subsystem**

This `?` amplifies a *single instance's* collection error into a *whole-path* failure. Inside `try_collect_target_configs`, the strict collector `collect_merged_target_configs_from_env` uses `.collect::<Result<Vec,_>>()`, which short-circuits on the first `Err` and drops all sibling instances; and this `?` sits in the loop over *every* plugin type, so one bad type aborts kafka/mqtt/... entirely.

Root cause: `is_target_enabled` (`config/common.rs:50`) now returns `Err` for an unparseable value, and `EnableState::from_str` accepts `on/off/true/false/enabled/disabled/1/0/...` but **not** `enable`.

Repro: set `RUSTFS_NOTIFY_WEBHOOK_ENABLE_primary=enable` (a typo for `enabled`). On startup/reload:
- notify: `create_dormant_targets_from_config` fails -> no targets load, the enable transition fails;
- audit: same path -> `AuditSystemState::Stopped` (`audit/system.rs:218`) -> audit logging stops entirely;
- the admin describe/list path also errors (`admin/handlers/target_descriptor.rs:566`), so you can't even list targets to find the bad one.

Compare main: the old `is_target_enabled` silently mapped an unparseable value to `false`, skipping only that one instance while all others loaded.

Note the inconsistency within this path: target *construction* failure is fault-isolated (pushed into `failures`, others continue -- see a few lines below), but an `enable` *collection* failure is fatal.

Suggestion: thread the per-instance collection error into the existing `creation_failures`/`failures` accumulator (so an admin write still reports the error / can't false-succeed, while valid siblings and other types still load), or have the create/startup path use the lenient collector `collect_merged_target_configs_compat_from_env` (warn-and-disable) that this same PR already added.

crates/targets/src/config/loader.rs
```diff
@@ -137,14 +147,83 @@ where
...
     valid_fields: &HashSet<String>,
     env_vars: I,
+) -> Result<Vec<MergedTargetConfigRecord>, TargetError>
+where
+    I: IntoIterator<Item = (String, String)>,
+{
+    collect_merged_target_config_results_from_env(config, section_name, route_prefix, target_type, valid_fields, env_vars)
+        .into_iter()
+        .map(|result| result.map_err(|(_, err)| err))
+        .collect()
```
overtrue, line 188: **[Medium-High / mechanism + fix site for the finding above]**

This `.collect()` folds into `Result<Vec,_>` and short-circuits on the first `Err`, discarding all already-collected instances -- this is the mechanism behind the "all types fail" behavior at `plugin.rs:350`.

This PR already ships a lenient variant `collect_merged_target_configs_compat_from_env` (warn and treat as disabled), but the production create/startup path doesn't use it. The fix can land here (per-instance fault isolation) or by switching the create path to the lenient variant.

crates/targets/src/runtime/adapter.rs
```diff
@@ -104,36 +364,16 @@ where
...
-                            idle_sleep,
-                        )
-                    },
-                )
-                .await
-            }
-        })
-        .await
+        let prepared = self.prepare_targets(targets).await;
+        let (opened, rejected) = self.open_prepared_stores(prepared);
```
overtrue, line 368: **[Medium] `activate_with_replay` runs blocking queue-store I/O on the async executor**

`open_prepared_stores` is synchronous and blocking (`create_dir_all` + `read_dir` + per-entry `metadata()`). The notify lifecycle offloads it via `spawn_blocking` (`notify/runtime_facade.rs:169`), but this compat path calls it inline and blocks a tokio worker.

This path is used by audit activation (`audit/pipeline.rs:554`) and `replace_targets`; since audit targets already opened their stores at construction, this is additionally a redundant full queue-directory rescan. Impact is bounded to activation events (startup/reload) -- suggest offloading via `spawn_blocking` here too.

rustfs/src/admin/handlers/audit_runtime_config.rs
```diff
@@ -86,30 +91,40 @@ async fn update_audit_config_and_reload_for_context<F>(
...
-        .map_err(|e| s3_error!(InternalError, "failed to save audit config: {}", e))?;
+        save_admin_server_config_no_lock(store, &config)
+            .await
+            .map_err(|e| s3_error!(InternalError, "failed to save audit config: {}", e))?;
 
-    apply_audit_runtime_config(specs, config).await
+        // Keep persistence and runtime publication in one detached, serialized
+        // mutation. Otherwise a cancelled caller or two concurrent updates can
+        // leave the persisted config and active audit generation disagreeing.
+        apply_audit_runtime_config(&specs, config).await
```
overtrue, line 118: **[Low-Medium] Audit target activation holds the process-global `SERVER_CONFIG_LOCK` across network I/O**

`apply_audit_runtime_config` internally calls `activate_targets_with_replay` (kafka/amqp/nats/mqtt may block on connect), and it runs entirely inside `with_admin_server_config_write_lock` (the process-wide `SERVER_CONFIG_LOCK` write lock). While held, *all* server-config mutations (notify/OIDC/storage-class SetConfigKV, full-config restore, etc.) block on the same lock.

Not a deadlock (the audit path doesn't re-acquire the lock), and the comment frames it as a persist+apply atomicity tradeoff. But note the notify path deliberately activates targets *outside* the lock (the receipt pattern), while audit here doesn't -- is this asymmetry intentional / deferred to a later PR?

rustfs/src/admin/handlers/notify_runtime_access.rs
```diff
@@ -12,31 +12,33 @@
...
+        && let Some(system) = rustfs_notify::notification_system()
+        && system.runtime_lifecycle_is_converged()
+    {
         return Ok(system);
     }
 
-    let _guard = NOTIFICATION_SYSTEM_INIT_LOCK.lock().await;
-    if let Some(system) = rustfs_notify::notification_system() {
-        return Ok(system);
+    init_event_notifier()
```
overtrue, line 28: **[Low] Lost concurrent-init serialization**

On main this had `NOTIFICATION_SYSTEM_INIT_LOCK` serializing first-time init; this PR removes it. During the startup window (the background reconciler only runs every 5s), two concurrent first-access requests can both enter `init_event_notifier()`, producing redundant generations and possibly returning a transient `InternalError("latest notification lifecycle generation has not converged")` to one caller. Self-healing, no state corruption -- hence low -- but it can surface as an occasional 500.

rustfs/src/admin/handlers/config_admin.rs
```diff
@@ -1563,20 +1574,144 @@ async fn commit_prepared_config(
...
+    }
+    Ok(())
+}
+
+async fn wait_notify_config_intent(transition: Option<rustfs_notify::NotificationLifecycleTransition>) -> S3Result<bool> {
+    let Some(transition) = transition else {
+        return Ok(false);
+    };
+    transition.wait().await.map_err(|err| {
+        warn!(error = %err, "Failed to apply local notification config");
```
overtrue, line 1597: **[Low / needs confirmation] Server-side logs may include `TargetError` Display detail**

The client/peer responses are correctly genericized and the RPC path is redacted -- this is **server-log only**. But if `NotificationError::Target(TargetError)`'s Display embeds a target endpoint / connection string, there's a small exposure surface.

Please confirm `TargetError`'s Display carries no credential/connection string; if it does, consider redacting these `warn!(error = %err)` lines as well.

overtrue (commented):
**Post-merge review notes (added after the PR was merged; line-level findings).**

Local verification, all green: `rustfs-notify` lib 126 passed / integration 2 passed; `rustfs-targets` lib 530 passed.

The core lifecycle coordinator checks out: monotonic generations + a non-cancellable publication barrier (`publication_gate`), `apply_lock` serializing runtime mutation, a consistent lock order `dispatch_gate -> replay_workers -> target_list -> publication_gate`, generation allocated inside the persisted RMW critical section (`.wait()` converges outside the lock), `no_lock` paths that avoid recursive locking on `config/config.json`, terminal shutdown finality, the embedded-owner refcount, the legacy `initialize` retry state machine, and redacted panic payloads. Persisted path and S3/wire formats are unchanged.

Line-level findings below. Severity: #1 medium-high (regression), #2 medium, #3 low-medium, #4 low, #5 low/needs-confirmation.

PR #5113, opened 2026-07-22
fix(lock): jitter distributed lock retries

hkwi:
## Related Issues

N/A

## Summary of Changes

- Add bounded ±25% jitter to the existing linear distributed-lock retry backoff.
- Keep the current retry base and acquire-timeout behavior while preventing synchronized contenders from repeatedly retrying in lockstep.
- Add deterministic seeded coverage for jitter bounds and preserve the contended-timeout regression test with virtual time.

This targets the deterministic retry pattern observed in our testbed. The scanner cache-save serialization proposal is intentionally not included; this PR contains no scanner changes.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-lock` (108 passed)
- `cargo clippy --all-targets -p rustfs-lock -- -D warnings`
- `make pre-commit`

## Impact

- No public API, configuration, wire-format, or on-disk-format changes.
- Distributed lock retry delays now vary within ±25% of the existing linear delay.
- Scanner concurrency and cache-save behavior are unchanged.

## Additional Notes

High-risk change review:

- Correctness: jitter remains bounded around the existing delay, and acquire-timeout behavior remains covered.
- Simplicity: the change is local to the existing retry helper and adds no new public abstraction.
- Security: no authentication, authorization, parsing, path-validation, or secret-handling surface changes.
- Concurrency/durability: randomized retry timing reduces synchronized contention without changing quorum, rollback, refresh, or release behavior.
- Compatibility: public APIs, configuration, and persisted formats are unchanged.
- Performance: one thread-local RNG sample is added per distributed-lock retry; no work is added to successful first-attempt acquisition.
- Test coverage: seeded tests pin the jitter bounds and non-constant output; the full rustfs-lock suite passes.

---


hkwi:

crates/lock/src/distributed_lock.rs
```diff
@@ -548,8 +549,15 @@ impl DistributedLock {
...
     }
 
-    fn lock_acquire_retry_backoff(attempt: usize) -> Duration {
-        LOCK_ACQUIRE_RETRY_INITIAL_BACKOFF.saturating_mul(attempt.try_into().unwrap_or(u32::MAX))
+    fn lock_acquire_retry_backoff(attempt: usize, rng: &mut impl rand::Rng) -> Duration {
+        let base = LOCK_ACQUIRE_RETRY_INITIAL_BACKOFF.saturating_mul(attempt.try_into().unwrap_or(u32::MAX));
+        let jitter = base / 4;
+        let lower = base.saturating_sub(jitter);
+        let upper = base.saturating_add(jitter);
+        let lower_millis = u64::try_from(lower.as_millis()).unwrap_or(u64::MAX);
```
overtrue: The millisecond conversion rounds the lower bound down, so the generated delay can fall outside the stated ±25% range. For example, on attempt 1 the base is 250 ms and the exact lower bound is 187.5 ms, but `as_millis()` produces 187, allowing a 187 ms delay (-25.2%). The same issue occurs for attempt 3 (562 ms versus an exact 562.5 ms lower bound). Please round the lower bound up or sample at nanosecond precision. The test should also force the RNG to return both endpoints; 32 seeded samples do not guarantee that this invalid lower endpoint is exercised.

hkwi: Fixed in b98a085e. The retry delay is now sampled at nanosecond precision, so fractional-millisecond bounds remain within the exact ±25% range. The test also uses deterministic constant RNGs to force and assert both endpoints.

houseme (changes requested):
Blocking findings

- `crates/lock/src/distributed_lock.rs` (tests module, `ConstantRng::try_fill_bytes`): `size_of::<u64>()` is used without bringing `std::mem::size_of` into scope. This causes a test-compilation failure in this PR.

Please fix by adding:
- `use std::mem::size_of;`

or by using the fully-qualified form:
- `std::mem::size_of::<u64>()`
in the test module imports/implementation.

Recommendation: Request changes.

houseme (commented):
Follow-up suggestion: this line can be fixed with one of:
- add `use std::mem::size_of;` in the test module imports
- or replace `size_of::<u64>()` with `std::mem::size_of::<u64>()`

If you want, I can post this as an inline diff comment too.

crates/lock/src/distributed_lock.rs
```diff
@@ -1025,6 +1035,47 @@ mod tests {
...
+            Ok(self.0 as u32)
+        }
+
+        fn try_next_u64(&mut self) -> Result<u64, Self::Error> {
+            Ok(self.0)
+        }
+
+        fn try_fill_bytes(&mut self, dst: &mut [u8]) -> Result<(), Self::Error> {
+            for chunk in dst.chunks_mut(size_of::<u64>()) {
+                chunk.copy_from_slice(&self.0.to_ne_bytes()[..chunk.len()]);
```
houseme, line 1202: This will fail to compile: `size_of::<u64>()` is not in scope here. Please either add `use std::mem::size_of;` in the test imports or change to `std::mem::size_of::<u64>()`.

houseme (approved):
Re-review latest revision: no remaining functional issues found. The previous size_of::<u64>() compile issue is fixed and jitter/backoff behavior looks correct. This change appears ready to merge pending repository merge policy.

houseme (approved):

PR #5133, opened 2026-07-22
fix(sse): preserve legacy KMS-wrapped object reads

### LINKED ISSUE #5063: [Bug] Locked out of data after upgrade: Local KMS configuration disabled in Console causing "invalid managed DEK format"
**Description:**
I am experiencing a severe data lockout issue after upgrading from `v1.0.0-beta.5` to [1.0.0-beta.10](https://github.com/rustfs/rustfs/releases/tag/1.0.0-beta.10).

In `v1.0.0-beta.5`, it was possible to manage and enable local KMS via the Console, which I used to set up bucket-specific encryption keys. However, in the newer versions, the KMS settings have been removed from the UI and replaced with the following message:

> *"Local filesystem KMS configuration is read-only in Console until safe master-key rotation is available."*

Because the KMS configuration is no longer active/manageable, I attempted to fall back to using the `RUSTFS_SSE_S3_MASTER_KEY` environment variable. This results in the server failing to decrypt any previously uploaded files, throwing a 500 error on `GET` requests:

```text
Io error: invalid managed DEK format

```

**Steps to Reproduce:**

1. Start a `rustfs` instance on `v1.0.0-beta.5`.
2. Enable SSE-S3 KMS via the Console and upload files (generating bucket-specific keys).
3. Upgrade the instance to [1.0.0-beta.10](https://github.com/rustfs/rustfs/releases/tag/1.0.0-beta.10).
4. Observe that KMS settings are locked/read-only in the Console.
5. Attempt to download the previously uploaded files.
6. Observe the `500 Internal Error: invalid managed DEK format` in the server logs.

**Expected Behavior:**
While disabling the UI for KMS to implement safe key rotation makes sense for new setups, existing KMS configurations from `v1.0.0-beta.5` must remain functional to prevent data loss. There should be a documented CLI/ENV workaround to explicitly load the old bucket-specific KMS keys so that legacy encrypted files remain accessible while the UI is locked.

**Environment:**

* **Previous Version:** [1.0.0-beta.5](https://github.com/rustfs/rustfs/releases/tag/1.0.0-beta.5)
* **Current Version:** [1.0.0-beta.10](https://github.com/rustfs/rustfs/releases/tag/1.0.0-beta.10)

**Additional Context:**
Currently, any data encrypted via KMS in the beta version is completely inaccessible. Any guidance on a temporary workaround (e.g., specific environment variables to force the legacy KMS to load) would be highly appreciated.

reatang:
## Related Issues

Fixes #5063

Thanks to @Phyroks for reporting the upgrade lockout and providing detailed reproduction steps.

## Summary of Changes

- separate SSE-S3 and SSE-KMS DEK providers so SSE-S3 always uses the configured local master key and SSE-KMS requires a KMS service
- classify persisted RustFS KMS DEK envelopes during reads so legacy AES256 objects remain on the KMS decrypt path even when the public algorithm header is missing
- share case-insensitive managed metadata conflict checks and RustFS/MinIO KMS context decoding across storage readers
- snapshot KMS service selection per request to avoid mixing service generations during reconfiguration
- add regression coverage for provider routing, legacy envelopes, malformed or conflicting metadata, and rio-v2 read paths

## Verification

- `make pre-pr`
- `cargo test -p rustfs test_legacy_sse_s3_kms_envelope_keeps_kms_read_path --lib`
- `cargo test -p rustfs-ecstore managed_dek_provider_routes_by_algorithm_and_persisted_envelope --lib`
- `cargo check -p rustfs-ecstore --features rio-v2`
- `cargo check -p rustfs --features rio-v2`
- `cargo fmt --all --check`

## Impact

- SSE-S3 requires `RUSTFS_SSE_S3_MASTER_KEY` to contain a non-zero base64-encoded 32-byte key.
- SSE-KMS no longer falls back to the local SSE-S3 provider when KMS is unavailable.
- Legacy SSE-S3 objects whose DEKs were wrapped by KMS remain readable with their original KMS configuration.
- No S3 API or on-disk format change.

## Additional Notes

- The legacy envelope fallback is registered as `RUSTFS_COMPAT_TODO(rustfs-5063)` and can be removed after every referenced legacy DEK has been rewrapped.
- High-risk adversarial validation completed for correctness, simplicity, security, concurrency and durability, compatibility, performance, and test coverage, with no unresolved findings.

### DIFF
--- a/crates/e2e_test/src/multipart_auth_test.rs
+++ b/crates/e2e_test/src/multipart_auth_test.rs
@@ -1068,7 +1068,8 @@ async fn test_anonymous_post_object_uses_bucket_default_sse_s3() -> Result<(), B
 
 #[tokio::test]
 #[serial]
-async fn test_anonymous_post_object_uses_bucket_default_sse_kms() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+async fn test_anonymous_post_object_bucket_default_sse_kms_requires_kms() -> Result<(), Box<dyn std::error::Error + Send + Sync>>
+{
     init_logging();
 
     let mut env = RustFSTestEnvironment::new().await?;
@@ -1128,14 +1129,23 @@ async fn test_anonymous_post_object_uses_bucket_default_sse_kms() -> Result<(), 
         .send()
         .await?;
 
-    assert_eq!(post_resp.status(), reqwest::StatusCode::NO_CONTENT);
-
-    let head = admin_client.head_object().bucket(bucket).key(object_key).send().await?;
-    assert_eq!(head.server_side_encryption().map(|value| value.as_str()), Some("aws:kms"));
-
-    let uploaded = admin_client.get_object().bucket(bucket).key(object_key).send().await?;
-    let uploaded = uploaded.body.collect().await?.into_bytes();
-    assert_eq!(uploaded.as_ref(), expected_body.as_slice());
+    let status = post_resp.status();
+    let response_body = post_resp.text().await?;
+    assert_eq!(status, reqwest::StatusCode::BAD_REQUEST);
+    assert!(
+        response_body.contains("configured KMS service"),
+        "unexpected response body: {response_body}"
+    );
+    assert!(
+        admin_client
+            .head_object()
+            .bucket(bucket)
+            .key(object_key)
+            .send()
+            .await
+            .is_err(),
+        "failed SSE-KMS POST must not create an object"
+    );
 
     Ok(())
 }
--- a/crates/ecstore/src/api/mod.rs
+++ b/crates/ecstore/src/api/mod.rs
@@ -398,6 +398,10 @@ pub mod set_disk {
     pub use crate::set_disk::{DEFAULT_READ_BUFFER_SIZE, SetDisks, get_lock_acquire_timeout, is_valid_storage_class};
 }
 
+pub mod sse {
+    pub use crate::sse::{ManagedDekProvider, ManagedSseScheme, managed_dek_provider};
+}
+
 pub mod store_list {
     pub use crate::store::list_objects::{ListPathOptions, max_keys_plus_one};
 }
--- a/crates/ecstore/src/lib.rs
+++ b/crates/ecstore/src/lib.rs
@@ -49,6 +49,7 @@ mod object_api;
 mod runtime;
 mod services;
 mod set_disk;
+mod sse;
 mod storage_api_contracts;
 mod store;
 
--- a/crates/ecstore/src/object_api/readers.rs
+++ b/crates/ecstore/src/object_api/readers.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use super::*;
+use crate::sse::{ManagedDekProvider, ManagedSseScheme, managed_dek_provider as classify_managed_dek_provider};
 #[cfg(feature = "rio-v2")]
 use aes_gcm::aead::Payload;
 use aes_gcm::{
@@ -26,7 +27,11 @@ use chacha20poly1305::ChaCha20Poly1305;
 use hmac::{Hmac, Mac};
 use md5::{Digest, Md5};
 use rustfs_kms::types::ObjectEncryptionContext;
-use rustfs_utils::http::{SSEC_ALGORITHM_HEADER, SSEC_KEY_HEADER, SSEC_KEY_MD5_HEADER};
+#[cfg(feature = "rio-v2")]
+use rustfs_kms::{MINIO_INTERNAL_ENCRYPTION_KMS_CONTEXT_HEADER, RUSTFS_ENCRYPTION_CONTEXT_HEADER, decode_managed_kms_context};
+use rustfs_utils::http::{
+    AMZ_SERVER_SIDE_ENCRYPTION, SSEC_ALGORITHM_HEADER, SSEC_KEY_HEADER, SSEC_KEY_MD5_HEADER, get_consistent_metadata_value,
+};
 use rustfs_utils::path::path_join_buf;
 #[cfg(feature = "rio-v2")]
 use serde::Deserialize;
@@ -39,11 +44,11 @@ use crate::io_support::rio::Index;
 
 const INTERNAL_ENCRYPTION_KEY_ID_HEADER: &str = "x-rustfs-encryption-key-id";
 const INTERNAL_ENCRYPTION_KEY_HEADER: &str = "x-rustfs-encryption-key";
-const INTERNAL_ENCRYPTION_CONTEXT_HEADER: &str = "x-rustfs-encryption-context";
 const INTERNAL_ENCRYPTION_IV_HEADER: &str = "x-rustfs-encryption-iv";
 const INTERNAL_ENCRYPTION_ORIGINAL_SIZE_HEADER: &str = "x-rustfs-encryption-original-size";
 const SSEC_ORIGINAL_SIZE_HEADER: &str = "x-amz-server-side-encryption-customer-original-size";
 const DEFAULT_SSE_ALGORITHM: &str = "AES256";
+const SSE_KMS_ALGORITHM: &str = "aws:kms";
 #[cfg(feature = "rio-v2")]
 const DARE_PAYLOAD_SIZE: i64 = 64 * 1024;
 #[cfg(feature = "rio-v2")]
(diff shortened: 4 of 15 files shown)


houseme:
@codex review

overtrue (changes requested):
# Review: provider split direction is right, but one read regression re-creates the #5063 lockout class

The provider separation, the persisted-envelope classification, the per-request KMS service snapshot, and the metadata conflict hardening are all solid, and the test coverage for them is genuinely good. I verified the core mechanism end to end and it holds:

- `is_data_key_envelope` is reliable across every KMS backend: local (`crates/kms/src/backends/local.rs:361`), vault (`crates/kms/src/backends/vault.rs:317`), and vault_transit (`crates/kms/src/backends/vault_transit.rs:394`) all serialize generated data keys as `DataKeyEnvelope` JSON, while the local SSE-S3 DEK format (`base64nonce:base64ciphertext`) can never parse as JSON. So the "SSE-S3 + persisted KMS envelope → KMS decrypt path" classification cannot misroute in either direction.
- Dropping the `service.headers_to_metadata` production branch is behavior-preserving: the manual parse keeps the IV length == 12 check and all required fields, and removing the hard dependency on the public algorithm header is exactly what the legacy fix needs.
- The `crates/ecstore/tests/minio_generated_read_test.rs` env-var switch correctly accounts for `#[cfg(test)]` not applying to integration-test builds of the library.

That said, I found one high-severity read regression plus an unresolved CI anomaly, and a few smaller items. Fix suggestions inline.

## 1. [Blocking] Legacy SSE-KMS objects wrapped by the local provider become unreadable — the same lockout class this PR eliminates for SSE-S3

On current main, a deployment with only `RUSTFS_SSE_S3_MASTER_KEY` and no KMS successfully accepts `aws:kms` writes, because the old provider factory falls back to the local provider for SSE-KMS too. This PR's own diff is the proof: the deleted assertions in `test_anonymous_post_object_uses_bucket_default_sse_kms` asserted POST → 204 **and a byte-identical GET readback** with no KMS configured. That is released behavior, so objects with `x-amz-server-side-encryption: aws:kms` and a local-format DEK exist in the wild.

After this PR, both read layers route `ManagedSseScheme::SseKms` unconditionally to the KMS provider (`apply_managed_decryption_material` in `rustfs/src/storage/sse.rs`, and `managed_dek_provider` in `crates/ecstore/src/object_api/readers.rs`):

- KMS still not configured → 400 `SSE-KMS requires a configured KMS service`.
- KMS configured later → the local-format DEK cannot be decrypted by any KMS backend.

Those objects end up with no read path at all — an upgrade lockout, exactly what this PR sets out to prevent in the SSE-S3 direction.

Rejecting **new** SSE-KMS writes without KMS is correct and should stay. What's missing is the symmetric read-side compat path:

```rust
// crates/ecstore/src/sse/mod.rs
pub fn managed_dek_provider(scheme: ManagedSseScheme, has_kms_envelope: bool, has_local_dek: bool) -> ManagedDekProvider {
    match scheme {
        // RUSTFS_COMPAT_TODO(rustfs-5063): releases before the provider split wrapped SSE-KMS DEKs
        // with the local provider when KMS was absent. Remove after those DEKs are rewrapped.
        ManagedSseScheme::SseKms if !has_kms_envelope && has_local_dek => ManagedDekProvider::LocalSseS3,
        ManagedSseScheme::SseKms => ManagedDekProvider::Kms,
        ManagedSseScheme::SseS3 if has_kms_envelope => ManagedDekProvider::Kms,
        ManagedSseScheme::SseS3 => ManagedDekProvider::LocalSseS3,
    }
}
```

where `has_local_dek` is a strict format probe (valid UTF-8, exactly two `:`-separated parts, both valid base64, nonce part decoding to 12 bytes). The strictness matters: MinIO-imported SSE-KMS sealed keys are neither a JSON envelope nor local format, so they keep routing to KMS unchanged. The classification stays unambiguous because real KMS-wrapped DEKs are always `DataKeyEnvelope` JSON (verified for all three backends above) and local DEKs never are.

Two details to preserve on the fallback path:

- keep decoding the KMS context for this case (the pre-split code decoded the MinIO context for SSE-KMS before building the object context), so context-bound MinIO-format DEKs keep decrypting;
- keep reporting the stored `aws:kms` algorithm in response headers, as the code already does via `stored_algorithm`.

Suggested regression coverage: a unit test mirroring `test_legacy_sse_s3_kms_envelope_keeps_kms_read_path` in the opposite direction (`aws:kms` metadata + local-format DEK decrypts via the local provider, with the counting provider asserting zero KMS calls), plus restoring the deleted e2e readback as "objects written by the pre-split fallback remain readable".

## 2. [Blocking until explained] S3 Implemented Tests hit the 60-minute hard timeout

[Job 88942972228](https://github.com/rustfs/rustfs/actions/runs/29924545040/job/88942972228) was killed at the 1h limit; on recent main runs this job completes in 9–16 minutes. Only one attempt ran and no artifacts survived the kill, so there's no log to triage. Since this PR changes the SSE request path, this can't be waved off as a flake: please re-run the job, and if it reproduces, the prime suspect is an SSE-related s3-test hanging on one of the new error paths.

## 3. Zero-key rejection on the read path can create another small lockout

The pre-PR `TestSseDekProvider::new_for_local_sse` accepted an all-zero `RUSTFS_SSE_S3_MASTER_KEY`, so objects encrypted under an intentional all-zero key may exist. This PR rejects all-zero on both write and read (`rustfs/src/storage/sse.rs` and `local_sse_master_key` in `crates/ecstore/src/object_api/readers.rs`), which strands that data with no escape hatch. Suggest rejecting all-zero only when generating new DEKs (write side) while tolerating it on decrypt — or, if you keep the strict behavior, calling it out in the Impact section as a breaking change.

## 4. Error code for reading existing SSE-KMS objects while KMS is down

The rustfs layer maps "no KMS service" to 400 `InvalidRequest` for both writes and reads, while the ecstore rio layer surfaces `Error::other` (500) for the same condition. For a GET of an existing object, the server's KMS being unavailable is a server-side condition — 400 tells clients not to retry. Suggest 503 (`ServiceUnavailable`) for the decrypt path and keeping 400 for the write path, and aligning the two layers.

## 5. Document the `__RUSTFS_SSE_SIMPLE_CMK` production removal

The variable was reachable in production binaries before this PR (with a usage example in shipped code comments); it is now compiled out via `#[cfg(test)]`. Any deployment that (mis)used it can migrate seamlessly by setting `RUSTFS_SSE_S3_MASTER_KEY` to the same value, since both providers share the DEK format — worth one bullet in the Impact section.

## 6. Nit: naming shadow in readers.rs

`crates/ecstore/src/object_api/readers.rs` defines a local `managed_dek_provider(metadata, dek)` while importing the shared classifier as `classify_managed_dek_provider` — the double indirection reads backwards. Renaming the local helper (e.g. `classify_dek_provider`) would help.

overtrue:
Follow-up on the two blocking items from my review, with new evidence from re-running CI and digging into the s3-tests harness.

## The S3 Implemented Tests timeout is reproducible, and I believe I found the root cause

I re-ran the failed job. Attempt 2 also ran until the 60-minute hard kill (21:58 → 23:03 UTC), versus a 9–16 minute baseline on recent main runs. Two consecutive wall-clock kills rule out an infra flake.

Root cause chain:

1. The s3-tests harness runs with exactly the environment this PR changes semantics for: `scripts/s3-tests/run.sh` exports a deterministic `RUSTFS_SSE_S3_MASTER_KEY` (line ~292) and never configures a KMS service.
2. `scripts/s3-tests/implemented_tests.txt` pins ~20 `test_sse_kms_*` cases — `test_sse_kms_transfer_13b/1b/1kb/1MB`, `test_sse_kms_default_upload_1b..8mb`, `test_sse_kms_multipart_upload`, `test_sse_kms_multipart_invalid_chunks_*`, `test_sse_kms_default_post_object_authenticated_request`, `test_sse_kms_present`, `test_sse_kms_method_head`, etc. On main all of these pass **via the local-provider fallback this PR removes**. With this PR, every one of them gets 400 on the write.
3. The 60-minute blowup is consistent with many streaming uploads (up to 8 MB, plus multipart) hitting an early 400 while the client is still sending the body: boto3 stalls on the socket and retries, and with `XDIST=4` and `MAXFAIL=0` those per-test stalls accumulate past the job limit. (Worth confirming from a local run of the SSE subset, but the causal direction is already clear from the whitelist.)

Consequence: **this gate can never pass for this PR as-is** — independent of flakiness. If maintainers accept the new write-side semantics (rejecting SSE-KMS without a KMS service, which matches MinIO/AWS behavior), the PR must also reclassify the pinned `test_sse_kms_*` cases out of `implemented_tests.txt` into `excluded_tests.txt` under the "Intentionally unsupported by product decision" bucket, with a comment pointing at this PR/issue. Please also check `lifecycle_behavior_tests.txt` and the swift/sftp lanes for the same dependency.

## This same fact hardens the legacy-read finding

The s3-tests whitelist is also direct evidence for review item 1: RustFS's own CI has been writing `aws:kms` objects whose DEKs are wrapped by the local SSE-S3 provider on every PR run for as long as these tests have been in the implemented list. Real deployments that enabled bucket-default `aws:kms` without a KMS hold the same shape of data. After this PR those objects have no read path (no KMS → 400; with KMS → the local-format DEK cannot be unwrapped), so the read-side fallback (`SseKms` + not a `DataKeyEnvelope` + DEK matching the local `base64:base64` format → `LocalSseS3`, tagged `RUSTFS_COMPAT_TODO(rustfs-5063)`) is required, not optional.

The misroute is even visible in this PR's own unit test: in `managed_dek_provider_routes_by_algorithm_and_persisted_envelope` (crates/ecstore/src/object_api/readers.rs), the second assertion feeds `AMZ_SERVER_SIDE_ENCRYPTION = aws:kms` together with `local_dek.as_bytes()` — a locally wrapped DEK — and expects `ManagedDekProvider::Kms`. For legacy data that combination is precisely the pre-split fallback output, and routing it to KMS is what makes it unreadable.

Note that fixing item 1 does not un-break the pinned `test_sse_kms_*` cases, because those perform fresh writes which the new semantics intentionally reject. So the two actions are independent and both needed: add the read-side fallback for existing data, and reclassify the s3-tests expectations for new writes.

PR #5145, opened 2026-07-23
perf(ecstore): add guarded encrypted multipart range seeks

overtrue:
## Related Issues

Related to https://github.com/rustfs/backlog/issues/1316.

## Summary of Changes

Adds an opt-in part-boundary seek for Legacy encrypted multipart Range GETs. `RUSTFS_ENCRYPTED_RANGE_SEEK=true` enables the optimization only for multipart uploads initialized and completed by servers that support the candidate-to-final layout marker and uploadId commit lock.

Completion now reaches quorum on `(etag, part number, physical size, plaintext size)`, binds the final internal marker to the upload `data_dir`, and holds object/upload locks through the commit. UploadPart serializes and fences its rename under the uploadId lock. Reader planning validates the marker, recorded plaintext total, physical total, positive part sizes, and checked arithmetic before seeking.

Old, unmarked, malformed, forged, `no_lock`, pre-upgrade, and mixed-version layouts keep the conservative full-object read. The feature defaults to disabled.

Adds path/amplification metrics and byte-exact, overflow, metadata-quorum, marker, lock-order, lock-loss, and commit-lifecycle coverage.

## Verification

- `make pre-commit`
- `make pre-pr` (10,126 tests passed; all doctests passed)
- `cargo test -p rustfs-ecstore --features rio-v2 object_api::readers::tests` (60 passed)
- `cargo clippy -p rustfs-ecstore --tests -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`

## Impact

When disabled, encrypted Range GET behavior is unchanged. Enabling requires every server that can initiate, upload, or complete multipart uploads to support this protocol and requires pre-upgrade multipart uploads to drain first. New dual-prefix internal metadata records the candidate/final layout marker; it does not change public APIs or encrypted object bytes.

Two hardening changes are unconditional and take effect regardless of the switch:

- Part-metadata quorum in CompleteMultipartUpload and ListParts now votes on `(etag, part number, physical size, plaintext size)` instead of etag alone. Healthy clusters fan out one identical serialized part meta per commit, so replicas can only disagree on these fields through mixed-generation interleaving (the backlog#853 corruption class); such staged uploads now fail completion instead of silently electing one generation.
- UploadPart re-validates the upload under its commit lock — one extra cross-disk metadata read per part commit — so a late UploadPart can no longer resurrect the staging directory of an upload that was aborted while the part waited on the lock.

Additionally, multipart plaintext sizing now prefers the recorded encryption original size over the per-part `actual_size` sum when both exist. The two only differ on inconsistent metadata, where the per-part fallback previously inflated the reported plaintext length.

## Additional Notes

- Correctness: attacked corrupt part metadata, wrong positive sizes, boundary and overflow values; invalid layouts fall back without seeking.
- Simplicity: checked workspace helpers and narrower alternatives; marker validation stays local because the generic helper has different mixed-case semantics.
- Security: attacked user-forged metadata and marker replay; the marker is scrubbed from input and bound to the server-generated `data_dir`.
- Concurrency/durability: attacked abort/upload/complete races, lock loss, and post-rename lifetime; deterministic tests cover ordering and fencing.
- Compatibility: attacked old objects, mixed-version uploads, MinIO prefixes, and `no_lock`; all stay on the full-read path.
- Performance: checked the GET hot path and lock scope; the seek path adds no allocation for UUID formatting, while commit locks are limited to guarded writes.
- Test coverage: mutation review found no changed behavior that survives the focused and full-suite tests.

---

### DIFF
--- a/crates/ecstore/src/object_api/readers.rs
+++ b/crates/ecstore/src/object_api/readers.rs
@@ -234,6 +234,178 @@ fn get_encrypted_offsets(oi: &ObjectInfo, offset: i64) -> Result<(i64, usize, us
     )))
 }
 
+/// Metric path label for a Legacy encrypted Range GET that kept the conservative full-object read.
+const ENCRYPTED_RANGE_READ_PATH_FULL: &str = "full";
+/// Metric path label for a Legacy encrypted Range GET served by the part-boundary seek.
+const ENCRYPTED_RANGE_READ_PATH_PART_SEEK: &str = "part_seek";
+
+/// Kill switch for the Legacy (rio v1) encrypted multipart part-boundary Range seek
+/// (https://github.com/rustfs/backlog/issues/1316 Phase A). Defaults to on; setting
+/// `RUSTFS_ENCRYPTED_RANGE_SEEK=false` restores the previous conservative full-object
+/// read for every encrypted Range GET on the Legacy backend.
+const ENV_RUSTFS_ENCRYPTED_RANGE_SEEK: &str = "RUSTFS_ENCRYPTED_RANGE_SEEK";
+const DEFAULT_RUSTFS_ENCRYPTED_RANGE_SEEK: bool = true;
+
+fn legacy_encrypted_range_seek_enabled() -> bool {
+    rustfs_utils::get_env_bool(ENV_RUSTFS_ENCRYPTED_RANGE_SEEK, DEFAULT_RUSTFS_ENCRYPTED_RANGE_SEEK)
+}
+
+/// True when a Legacy (rio v1) encrypted Range GET may seek to the covering part
+/// boundary instead of streaming the whole ciphertext.
+///
+/// Every condition is required for correctness, not merely for benefit:
+/// - `is_multipart`: single-part rio v1 streams have variable-length encrypted blocks
+///   with no closed-form plaintext-to-physical mapping, so only part boundaries
+///   (recorded in metadata) are safe seek targets.
+/// - `!is_compressed`: under compression the requested offset addresses the
+///   decompressed stream, not per-part plaintext, so stay on the full read.
+/// - non-empty parts with `actual_size > 0` for every part: a part with
+///   `actual_size <= 0` would make `part_plaintext_size` fall back to the physical
+///   size and poison the plaintext cumulative sums.
+/// - physical part sizes must add up to `oi.size`: guards against inconsistent
+///   metadata scheduling reads past the object end in the erasure layer.
+/// - `requested_length > 0`: degenerate empty ranges keep their existing full-read
+///   behavior (and its error surface) unchanged.
+fn legacy_encrypted_seek_eligible(oi: &ObjectInfo, is_multipart: bool, is_compressed: bool, requested_length: i64) -> bool {
+    is_multipart
+        && !is_compressed
+        && requested_length > 0
+        && !oi.parts.is_empty()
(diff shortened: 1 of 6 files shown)


crates/ecstore/src/object_api/readers.rs
```diff
@@ -234,6 +234,178 @@ fn get_encrypted_offsets(oi: &ObjectInfo, offset: i64) -> Result<(i64, usize, us
...
+/// - physical part sizes must add up to `oi.size`: guards against inconsistent
+///   metadata scheduling reads past the object end in the erasure layer.
+/// - `requested_length > 0`: degenerate empty ranges keep their existing full-read
+///   behavior (and its error surface) unchanged.
+fn legacy_encrypted_seek_eligible(oi: &ObjectInfo, is_multipart: bool, is_compressed: bool, requested_length: i64) -> bool {
+    is_multipart
+        && !is_compressed
+        && requested_length > 0
+        && !oi.parts.is_empty()
+        && oi.parts.iter().all(|part| part.actual_size > 0)
```
overtrue: This still trusts positive part.actual_size values as authenticated boundaries, but they come from persisted xl.meta and are not checked against the plaintext produced by each ciphertext segment. With real plaintext parts [20_000, 9_000, 5_000] and stale metadata [19_000, 10_000, 5_000], every gate here still passes. Range 19_500-19_599 then starts at ciphertext part 2, skips 500 bytes, and successfully returns object bytes 20_500-20_599. I reproduced this with the added fixture. GCM authenticates the selected ciphertext, not this global-offset mapping. Unless legacy objects have an independently verified layout signal, this needs to fall back, with a positive-but-wrong actual_size regression test.

crates/ecstore/src/object_api/readers.rs
```diff
@@ -234,6 +234,178 @@ fn get_encrypted_offsets(oi: &ObjectInfo, offset: i64) -> Result<(i64, usize, us
...
+        && oi.parts.iter().map(|part| part.size as i64).sum::<i64>() == oi.size
```
overtrue: The physical sum and both walkers use unchecked i64 arithmetic. Per-part validation does not bound the aggregate: [i64::MAX, i64::MAX, 3] passes the individual conversions, but this sum panics in checked builds and wraps to 1 in release, so it can equal oi.size and admit a nonsensical seek plan. The kill switch is evaluated after this expression, so false does not protect checked builds either. Please use checked conversion/addition and fail closed or fall back on overflow, with an aggregate-overflow test.

overtrue (commented):
Found two blockers in the new seek gate.

crates/ecstore/src/object_api/readers.rs
overtrue: Fixed in 0d68e6aec. Completion now reaches quorum on all boundary fields and only emits a data_dir-bound marker for uploads created and completed under the guarded protocol. Historical, unmarked, forged, or inconsistent layouts stay on the full-read path; the regression matrix covers positive-but-wrong boundaries.

crates/ecstore/src/object_api/readers.rs
overtrue: Fixed in 0d68e6aec. Physical/plaintext totals, range endpoints, and both cumulative walkers now use checked conversion and arithmetic, with overflow falling back before any seek. Aggregate and conversion overflow cases are covered.

PR #5152, opened 2026-07-23
fix(iam): virtualize OIDC service account parents

### LINKED ISSUE #5061: Access key (service account) created by an OIDC/SSO user is denied on ALL S3 operations — embedded policy is ignored (effective policy = embedded ∩ non-existent parent → empty)
**Describe the bug**

An access key (service account) created by an **OIDC/SSO-authenticated** user is denied on **every** S3 operation (reads included), even though the key carries the correct embedded policy. The same policy attached to a **builtin (local) user** yields a fully working access key.

The effective permission of a service account seems to be evaluated as `embedded_policy ∩ parent_user_policy`. For an OIDC user the parent is **not a persistent IAM identity** (it has no stored policy and cannot be given one), so the intersection is empty → deny-all. In MinIO the service account's embedded/session policy is authoritative, so OIDC-created keys work; RustFS diverges here. Net effect: SSO users cannot obtain usable programmatic (CLI/SDK/CI) credentials.

**To Reproduce**
1. Configure OIDC (Microsoft Entra ID) with `RUSTFS_IDENTITY_OPENID_CLAIM_NAME=roles`; create policy `blabla-rw` (`s3:*` on `arn:aws:s3:::*`) and map an Entra App Role `blabla-rw` to it.
2. Log in to the web console as an OIDC user granted that role (console SSO works; the session can perform S3 actions).
3. In the console, create an **Access Key** (service account); copy the access key + secret.
4. Use the key from any S3 client:
   ```
   mc alias set tdg https://<host>:9000 <ACCESS_KEY> <SECRET_KEY>
   mc ls tdg
   mc cp ./file tdg/<bucket>/
   ```
5. Every call fails with `AccessDenied`:
   ```
   op returns error op=HeadObject                 err=AccessDenied
   op returns error op=GetBucketLocation          err=AccessDenied
   op returns error op=HeadBucket                 err=AccessDenied
   op returns error op=GetObjectLockConfiguration err=AccessDenied
   op returns error op=ListObjectsV2              err=AccessDenied
   op returns error op=PutObject                  err=AccessDenied
   ```

**Expected behavior**

The access key inherits the OIDC user's effective policy (`blabla-rw`) and can perform the allowed S3 operations (read/write), exactly like an access key created by a builtin user with the same policy.

**Screenshots**

N/A (CLI / server logs included under Additional context).

 - OS: Ubuntu (unprivileged LXC), Linux x86_64
 - RustFS version: 1.0.0-beta.10 (build 2026-07-17), single node, single drive, data dir on NFS
 - Client: mc RELEASE.2025-08-13T08-35-41Z
 - Identity provider: Microsoft Entra ID (OIDC), `CLAIM_NAME=roles`

 - N/A

**Additional context**

The service account **has the correct embedded policy**, and its parent is the OIDC identity:

```json
// mc admin user svcacct info local <ACCESS_KEY>
{
  "accessKey": "<ACCESS_KEY>",
  "parentUser": "<oidc-user>@<domain>",
  "policy": { "Version": "2012-10-17", "Statement": [
    { "Effect": "Allow", "Action": ["s3:*"], "Resource": ["arn:aws:s3:::*"] },
    { "Effect": "Allow", "Action": ["admin:CreateServiceAccount","admin:ListServiceAccounts","admin:RemoveServiceAccount","admin:UpdateServiceAccount"] },
    { "Effect": "Allow", "Action": ["sts:AssumeRole"] } ] },
  "accountStatus": "on",
  "expiration": "9999-01-01T00:00:00Z"
}
```

But the parent identity does not exist as a persistent user and cannot be given a policy:

```
$ mc admin user info local "<oidc-user>@<domain>"
mc: <ERROR> ... user '<oidc-user>@<domain>' does not exist

$ mc admin policy attach local blabla-rw --user "<oidc-user>@<domain>"
mc: <ERROR> Unable to make user/group policy association: user not exist.
```

**Control test (isolates the cause to the parent intersection, not the policy):** attaching the identical `blabla-rw` policy to a **builtin** user and creating an access key for it works fully:

```
mc admin user add local svc-test <pass>
mc admin policy attach local blabla-rw --user svc-test
mc admin user svcacct add local svc-test        # -> AK/SK
# using that AK/SK:
mc ls test                  # OK (lists buckets)
mc mb test/svctest-bucket   # OK (Bucket created successfully)
```

**Suggested fix:** for a service account whose parent is an OIDC/STS identity, make the service account's **embedded (session) policy authoritative** (as MinIO does), instead of intersecting it with a non-existent persistent parent policy — or persist the OIDC user's mapped policy so the parent resolves.

**Related / adjacent (not duplicates):** #2553 (OIDC → console 403, closed, console-only), #1628 (builtin users managing keys), #2612 (OIDC username = rustfsAdmin).

**Minor note:** `mc admin accesskey ls local --all` only enumerates builtin users, so OIDC-created service accounts are not listed there (only via `mc admin user svcacct ls local <oidc-identity>`), which complicates administration/offboarding.

GatewayJ:
## Related Issues

Fixes #5061.

## Summary of Changes

- Replace the non-existent OIDC parent with a deterministic issuer-and-subject virtual parent.
- Freeze selected policy names in the signed OIDC STS claim, resolve referenced policy documents dynamically, and fail closed when any referenced policy is unavailable.
- Preserve the verified OIDC authorization boundary when creating a service account, and carry its version marker in an optional site-replication envelope rather than persisted account claims.
- Decode an absent envelope as the legacy format, so a change forwarded through an older replica remains compatible.
- Serialize service-account cache mutations and reload the legacy delete-notification RPC from shared state to prevent stale notifications from removing recreated accounts.

## Verification

- `cargo test -p rustfs oidc_service_account_envelope --lib`
- `make pre-commit`
- `make pre-pr`

## Impact

No S3 API, configuration, or durable IAM schema migration is introduced. Existing regular, STS, and legacy service accounts retain their paths.

Site replication adds the optional `oidcServiceAccountEnvelope` field. Older receivers ignore it; if they later forward the change without it, newer receivers use the legacy decoding path. Upgrade all replicated sites before relying on the v2 OIDC policy boundary across sites.

## Additional Notes

High-risk adversarial review: correctness (including the old-node forwarding case), simplicity (a typed optional DTO field is the smallest safe representation), security (policy boundaries still fail closed), concurrency/durability (cache writes stay serialized), compatibility (no S3 or required wire-field change), performance (control-plane-only work), and test coverage all found no unresolved issue.

Generic replication CAS, tombstones, and broad user-deletion semantics are intentionally excluded and tracked in the stacked follow-up PR.


cxymds:
Review findings — I recommend addressing these before merge:

1. **[P1] A replicated create can retain stale OIDC authorization claims on an existing account.**

   In `apply_iam_item`, the existing-account branch updates the secret, session policy, and display metadata, but does not apply `create.claims` or `create.groups`:
   https://github.com/rustfs/rustfs/blob/52dcb8f40fe93564d2504d2e8c3356706dde5585/rustfs/src/admin/handlers/site_replication.rs#L6513-L6536

   `update_service_account` then decodes and re-signs the receiver's existing JWT claims. A concrete failure case is an OIDC identity that previously had an admin group on site B, later lost that group, and recreated the same access key with a read-only boundary on site A. Replication to B combines A's new secret with B's stale admin claims, so the permission reduction does not converge and the newly returned credential remains over-privileged on B.

   Please atomically replace the replicated claims/groups (or reject the conflict), and add an `apply_iam_item` test starting with intentionally more-privileged stored claims.

2. **[P1] The envelope marker is persisted by old nodes and can be re-emitted as a false envelope during a rolling upgrade.**

   The new sender inserts the envelope version into ordinary claims:
   https://github.com/rustfs/rustfs/blob/52dcb8f40fe93564d2504d2e8c3356706dde5585/rustfs/src/admin/handlers/site_replication.rs#L6290-L6294

   An old receiver stores unknown claims unchanged. That service account can still be authorized through its replicated group policies and create a child account; the old handler copies all non-`exp` claims into the child but sends no session policy when none was explicitly requested. A new receiver then sees the persisted marker, classifies the event as an envelope, and rejects it for the missing session policy:
   https://github.com/rustfs/rustfs/blob/52dcb8f40fe93564d2504d2e8c3356706dde5585/rustfs/src/admin/handlers/site_replication.rs#L6341-L6358

   Please avoid identifying the transport envelope solely through a persistable claim, and add a new → old → new mixed-version regression test.

3. **[P1] The stale-revision check and mutation are not atomic.**

   The update path reads/checks the local revision before entering the mutation lock:
   https://github.com/rustfs/rustfs/blob/52dcb8f40fe93564d2504d2e8c3356706dde5585/rustfs/src/admin/handlers/site_replication.rs#L6562-L6588

   Two events `t2` (newer) and `t1` (older) can both read `t0` and pass the stale check. If `t2` writes first and `t1` acquires the internal lock second, the older secret/status/policy wins. The create and delete branches have the same check-then-mutate gap.

   Please make revision comparison and persistence one per-key compare-and-apply/CAS operation, with a deterministic concurrent out-of-order test.

The new site-replication tests currently exercise helpers/DTO round trips only; none calls the real `apply_iam_item` path, so these convergence failures remain invisible.

overtrue (approved):

cxymds (approved):

Issue #1325, opened 2025-12-31
Observed LIST inconsistency after bulk delete


steveloughran:
**Describe the bug**

(May relate to #1287)

Just a note that I ran the s3a integration tests [doc](https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/testing.md) against a local rustfs instance and consistently observed LIST operations finding objects which had just been deleted.

[test results within a larger pr](https://github.com/apache/hadoop/pull/7700#issuecomment-3700154301)

There were also some failures of assertions related to listing multipart uploads in progress: active ones weren't found, deleted ones still appeared visible.

This looks to me like that updates to listing info are async and tests probing immediately after delete operations got responses containing entries which had been deleted from the store.

It *does not* appear safe to use in hive or spark where table structures are defined by directory trees. it *may* be safe to use in manifest formats like Iceberg.

What I would do is encourage you to run the hadoop-aws tests against your builds as part of the regression testing. If it finds problems in the tests get on the hadoop common dev mailing list -- we may need to handle a 400 response from attempting to delete a nonexistent MPU.  If it finds a mismatch between aws s3 and rustfs for core operations: bugs your side.

This test suite has evolved over time to the extent has identified differences between S3 express and s3 standard, and finds aws java SDK regressions on every sdk update. Don't feel bad about failure, instead: you've just gained a few thousand test cases based on code broadly used in production.

## installation

Single node instance on macos, homebrew installation. 

```

Darwin 25.2.0 Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:40 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6000 arm64

rustfs 1.0.0-alpha.77
build time   : 2025-12-25 09:36:17 +00:00
build profile: release
build os     : macos-aarch64
rust version : rustc 1.92.0 (ded5c06cf 2025-12-08)
rust channel : stable-aarch64-apple-darwin
git branch   :
git commit   : 89c3ae77a4b91f73883365ce9a61e3935f130771
git tag      : 1.0.0-alpha.77
```

storage is `(apfs, local, journaled, nobrowse, protect, root data)`; I'm assuming case insensitivity is coming from that.

0xdx2:
I tried to reproduce the issues you mentioned, and apart from the 404 NosuchKey error in the log, there were no real problems. The NosuchKey error is normal because dir is not an objectkey.

[...]

PR #5160, opened 2026-07-23
feat(rpc): add server-side internode v2 signature verification (fail-open, method-path binding groundwork)

overtrue:
## Summary

First step of the #1327 internode RPC signature hardening (design C′, see https://github.com/rustfs/backlog/issues/1327). This lands the **server-side v2 verification infrastructure only** — it is deliberately fail-open and inert until client-side v2 signing lands in a follow-up PR, which is the safe rolling-upgrade order (deploy verification everywhere first, then enable signing).

Today the internode gRPC signature is a constant (`{fixed-prefix}|GET|timestamp`, `http_auth.rs`), so any captured fresh signature can be transplanted across gRPC methods and replayed within the 300s window. This PR is the groundwork to bind the real gRPC method path; it does not yet change what any client sends.

## What it does

- **`RpcRequestPathService<S>`** (a tower `Service<http::Request<B>>` in `rustfs/src/server/http.rs`) wraps the rpc service, reads the real `req.uri()` + `req.method()`, and stashes them as an `RpcRequestTarget` request extension before calling inner. This is how the tonic auth interceptor — which only receives `Request<()>` and cannot see the method path — recovers the concrete gRPC method path for verification.
- **`check_auth`** reads `RpcRequestTarget` from the extensions. When v2 headers (`x-rustfs-sig-v2` / `x-rustfs-sig-ts`) are present it verifies the v2 signature bound to the method path + timestamp; when they are absent or fail in non-strict mode it falls back to the existing legacy check and increments a convergence counter. **Non-strict is the default and it is fail-open — a legacy-only peer keeps authenticating unchanged.**
- **v2 payload** is a domain-separated fixed-length byte layout (`"rustfs-internode-rpc-sig:v2" || 0x00 || method_path || 0x00 || timestamp_be_i64`), HMAC-SHA256 under the existing shared RPC secret — no JSON, no body buffering.
- **`RUSTFS_INTERNODE_RPC_SIGNATURE_STRICT`** (default false, compile-time asserted) and the metric `rustfs_system_network_internode_signature_v1_fallback_total` provide the msgpack-style convergence gate: once the fallback counter reads zero fleet-wide (i.e. every peer sends v2), strict mode can be flipped in a later PR to reject legacy.

## Explicitly deferred to follow-up PRs (per design C′)

Client-side `V2SignLayer` signing (activates end-to-end method binding), nonce + bounded replay cache, `dest_authority` binding, strict-mode enforcement flip, and body-digest for high-risk mutating unary RPCs. Cross-process replay/tamper acceptance is deferred to the #1325 fault-injection proxy (PR #4938).

## Verification

- Old-behavior invariant is the top priority: a legacy-only peer must keep authenticating through the exact production entry point while strict mode is off, and each acceptance feeds the convergence counter. Tests cover this plus v2 method-binding (a v2 signature for `/…/ReadAll` does not verify against `/…/DeleteVersion`), wrong-secret, tampered-signature and stale-timestamp rejection.
- `cargo fmt --all --check` clean. Local full `clippy`/test compilation was repeatedly killed by cross-session build-host contention on this shared machine, so the authoritative gate is CI (Test and Lint). The prior identical implementation of this approach completed the full clippy+test+pre-commit chain green before it was lost to an uncommitted-worktree cleanup; this re-implementation is byte-for-byte the same approach and was committed immediately. If any CI lane reports an issue it must be followed up before merge.

Refs https://github.com/rustfs/backlog/issues/1327


houseme (approved):

houseme (approved):
Conclusion: RUSTFS_INTERNODE_RPC_SIGNATURE_STRICT has been changed to LazyLock cache in the hot path to avoid repeatedly reading environment variables every time verify_tonic_rpc_signature, and the performance optimization point is in place; the rest of the logic maintains the existing rollback behavior and compatibility, and no blocking problems are found.

houseme (approved):
Conclusion: The new  check is cached using , so  no longer reads environment variables on every request path. This is the expected hot-path optimization, and the broader rollout logic still preserves fallback compatibility while defaulting to fail-open behavior. No blocking issues found.

houseme (approved):
Conclusion: The new RUSTFS_INTERNODE_RPC_SIGNATURE_STRICT check is cached using LazyLock, so verify_tonic_rpc_signature no longer reads environment variables on every request path. This is the expected hot-path optimization, and the rollout logic still preserves fallback compatibility while defaulting to fail-open behavior. No blocking issues found.

PR #5184, opened 2026-07-24
feat(kms): introduce KMS unavailability error and enhance data key handling

reatang:
## Related Issues

Related: rustfs-5063 (pre-beta.9 KMS migration)
#5063 

### KMS Unavailability Error (`KmsUnavailableError`)

New typed error in `crates/kms` that maps to `S3ErrorCode::ServiceUnavailable` (HTTP 503). When the KMS service is not initialized, stopped, or unavailable, the error now carries a typed signal that clients can retry, instead of the previous generic `InternalError` wrapping a string message.

### DEK envelope format detection and decrypt-side provider routing

Added `is_data_key_envelope()` in `crates/kms/src/encryption/dek.rs` to detect whether an encrypted DEK is a RustFS KMS JSON envelope vs. a legacy local-provider format. On the decrypt path, `apply_managed_decryption_material` now inspects the persisted wrapping format to select the correct provider — KMS for JSON envelopes, local for legacy formats. This fixes a bug where objects encrypted with the local provider became undecryptable once KMS was running (or vice versa), because the decrypt path previously always routed to the current global provider.

### Pre-beta.9 local KMS recovery (rustfs-5063)

Two compatibility measures for pre-beta.9 local KMS deployments:

1. **Legacy SHA-256 KDF fallback**: key files created before beta.9 used SHA-256-based key derivation instead of Argon2. Added a `legacy_master_cipher` that falls back to the legacy KDF when decryption with the current Argon2-derived cipher fails and the key file lacks an explicit `at_rest_protection` marker.

2. **Legacy config migration**: when loading persisted KMS configs that are missing the `allow_insecure_dev_defaults` field (pre-beta.9), and the config is a Local KMS backend that fails validation, auto-set `allow_insecure_dev_defaults = true` with a `warn!` log event. Configs with an explicit `allow_insecure_dev_defaults` field are never modified.

### Cleanup of deprecated APIs and dead code

- Removed the manual ILM transition run admin endpoint (`POST /rustfs/admin/v3/ilm/transition/run`) and its supporting types (`ManualTransitionRunOptions`, `ManualTransitionRunReport`, `enqueue_transition_for_existing_objects_scoped`).
- Removed the storage-class write capability contract (`CAPABILITY_CONTRACT_VERSION`, `SUPPORTED_WRITE_CLASSES`, `UNSUPPORTED_WRITE_ERROR`, `LEGACY_LABEL_BEHAVIOR`, `is_supported_write_class`, `effective_class`) from `storageclass.rs`, the capability endpoint, and documentation.
- Simplified `response_storage_class` to return the stored class directly without `effective_class()` label normalization.
- Inlined the single-use `TransitionEnqueueOutcome` enum in `TransitionState::queue_transition_task`.
- Simplified `classify_response_checksums` by removing the unused `is_multipart` parameter and `checksum_type` setting.
- Removed `inline_fast_path_cluster_test` (which tested `should_inline` exact boundaries) and `storage_class_capability_test` e2e suites.

## Verification

This environment does not have `cargo`/`rustc` available. The required verification is:

```bash
make pre-commit
```

For focused verification on the key changed areas:

```bash
cargo test -p rustfs-kms
cargo test -p rustfs -- sse
cargo test -p rustfs -- kms_dynamic
cargo test -p rustfs -- error
cargo test -p ecstore -- storageclass
cargo test -p ecstore -- bucket_lifecycle_ops
cargo clippy --all-targets -- -D warnings
cargo fmt --all --check
```

## Impact

- **KMS unavailability now returns HTTP 503** (`ServiceUnavailable`) instead of a generic `InternalError`. Clients that parse S3 error codes will see a retryable signal; clients that relied on the previous error string will need to update.
- **Local KMS key files from pre-beta.9** are now readable again when the master key is unchanged. No manual migration is needed.
- **Pre-beta.9 persisted KMS configs** that lack `allow_insecure_dev_defaults` are silently migrated on load with a `warn!` log event.
- **Removed admin API**: `POST /rustfs/admin/v3/ilm/transition/run` no longer exists.
- **Removed runtime capability contract**: the `storage_classes` object is no longer returned from `GET /rustfs/admin/v4/runtime/capabilities`.

## Additional Notes

The legacy SHA-256 KDF fallback and legacy config migration are tagged with `RUSTFS_COMPAT_TODO(rustfs-5063)` and should be removed once all pre-beta.9 deployments have been upgraded and their key files/configs rewritten with the current format.


overtrue:
I found a blocking provider-routing issue in the managed decryption path.

In `apply_managed_decryption_material`, the KMS-envelope branch prefers any provider stored in `GLOBAL_SSE_DEK_PROVIDER`:

https://github.com/rustfs/rustfs/blob/acb8a6860f56341719823d9face07e6d20516e99/rustfs/src/storage/sse.rs#L1728-L1738

However, that cache is also populated with the production local SSE provider by `get_local_sse_dek_provider`:

https://github.com/rustfs/rustfs/blob/acb8a6860f56341719823d9face07e6d20516e99/rustfs/src/storage/sse.rs#L2188-L2215

This creates the following failure sequence:

1. Start without KMS and perform a managed SSE operation, which caches the local provider.
2. Dynamically enable KMS.
3. Write an object whose DEK is a KMS JSON envelope.
4. Read that object. The envelope is correctly detected, but the cached local provider is selected and attempts to decrypt the KMS envelope as the local `nonce:ciphertext` format.

As a result, KMS-encrypted objects can become unreadable after dynamically enabling KMS. The new regression test does not cover this transition because it calls `reset_sse_dek_provider()` before checking the KMS-envelope case:

https://github.com/rustfs/rustfs/blob/acb8a6860f56341719823d9face07e6d20516e99/rustfs/src/storage/sse.rs#L4119-L4121

Suggested fix: in production, always use `KmsSseDekProvider` for a detected KMS envelope. Reusing an injected provider should be restricted to `#[cfg(test)]`, or the cache should explicitly retain the provider kind. Please also add a regression test for “local provider cached -> dynamically enable KMS -> decrypt KMS envelope” without resetting the cache.

PR #5194, opened 2026-07-24
fix(lifecycle): stop tier free-version recovery walk-timeout loop

### LINKED ISSUE #5130: tier_free_version_recovery task loops forever on large buckets
## Environment

- RustFS 1.0.0-beta.10 (regression introduced with the 1.0.0-beta.9/beta.10 lifecycle changes, not present in 1.0.0-beta.8)
- Distributed mode: 4 nodes x 4 drives (16 drives total), deployed via the Helm chart (0.10.0) on Kubernetes
- Drives are Ceph RBD network volumes (rook-ceph-block), so directory walks are slower than on local NVMe
- Main workload: Grafana Loki chunk storage, ~95k directories per drive in one bucket

## Symptom

Since upgrading beta.8 -> beta.10, every node runs at a permanently elevated load (1-1.3 cores per pod plus heavy I/O wait), and the logs fill with:

```json
{"level":"WARN","message":"Disk operation timed out","event":"disk_health_check_failed","endpoint":"http://rustfs-0.rustfs-headless.rustfs.svc.cluster.local:9000/data/rustfs2","op":"walk_dir","timeout_ms":"60000","reason":"operation_timeout","target":"rustfs_ecstore::disk::disk_store","line_number":1122}
{"level":"WARN","message":"internode rpc background task failed","event":"rpc_background_task_failed","component":"internode_rpc","subsystem":"directory_walk","operation":"walk_dir","bucket":"loki-chunks","base_dir":"","recursive":true,"forward_to":"Some(\"fake/a937f7824420f7a9/...\")","limit":10000,"skip_total_timeout":false,"error":"timeout"}
{"level":"WARN","message":"Failed to recover tier free-version cleanup tasks","event":"lifecycle_worker_state",...}
```

The `walk_dir` timeouts also mark drives as failing (default `mark_failure` policy), so drives flap between healthy and suspect, which adds recovery churn on top.

## Root cause

`spawn_tier_free_version_recovery_once` in `crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs` ticks every 60 seconds and calls `recover_tier_free_versions`, which walks all objects of all buckets in pages of 10,000. The walk is issued with a hardcoded timeout:

```rust
// crates/ecstore/src/bucket/lifecycle/tier_free_version_recovery.rs
const BACKGROUND_WALKDIR_TIMEOUT: Duration = Duration::from_secs(60);
...
.with_walkdir_timeouts(BACKGROUND_WALKDIR_TIMEOUT),
```

This explicit per-walk timeout takes precedence over both `RUSTFS_DRIVE_WALKDIR_TIMEOUT_SECS` and `RUSTFS_DRIVE_TIMEOUT_PROFILE=high_latency`, because `DiskStore::walk_dir` prefers the caller-supplied value:

```rust
// crates/ecstore/src/disk/disk_store.rs
opts.timeout_duration().unwrap_or_else(get_drive_walkdir_timeout)
```

On our drives a full walk of the large bucket takes ~92 seconds per drive (measured with a plain `find` over ~95k directories), so the recovery walk is guaranteed to hit the 60s timeout. On failure the bucket/object markers are only advanced in the `Ok` branch, so the next tick retries the same page. The result is a permanent loop: every 60 seconds, on every node, a full-throttle recursive walk of the big bucket starts, dies at the 60s mark, discards its progress, and repeats. CPU and I/O never settle.

`crates/ecstore/src/bucket/replication/replication_resyncer.rs` contains the same hardcoded `BACKGROUND_WALKDIR_TIMEOUT` and presumably has the same problem for resync on large buckets.

There is currently no way to configure or disable this from the outside. We tuned everything that is configurable (`RUSTFS_DRIVE_TIMEOUT_PROFILE`, `RUSTFS_DRIVE_WALKDIR_TIMEOUT_SECS`, the `RUSTFS_CAPACITY_*` scanner settings); the recovery loop is unaffected.

## Suggested fix

Any of these would resolve it, ideally the first two together:

1. Make the background walks respect an explicitly set `RUSTFS_DRIVE_WALKDIR_TIMEOUT_SECS` instead of pinning 60s. Note that falling back to the `high_latency` profile alone would NOT be enough: that profile also yields 60s, while our walks need ~92s or more. The explicitly configured env value must take precedence.
2. Advance the cursor (or apply exponential backoff) when a page fails repeatedly, so the task cannot spin on the same page forever.
3. Provide an env var to change the recovery interval or disable the task entirely, if feasible.

## Impact

Sustained node saturation (load per core > 2) on otherwise healthy clusters with buckets that cannot be walked in 60s. Drive health flapping as a side effect. No workaround available short of downgrading to beta.8.

harryhan24:
## Related Issues

Fixes #5130

## Summary of Changes

`list_tier_free_versions` issued its background walk with `with_walkdir_timeouts(60s)`, a hardcoded total wall-clock budget that takes precedence over `RUSTFS_DRIVE_WALKDIR_TIMEOUT_SECS` and `RUSTFS_DRIVE_TIMEOUT_PROFILE` (`DiskStore::walk_dir` prefers the caller-supplied value). On any bucket whose healthy full walk needs more than 60s, the walk is guaranteed to fail; because the scheduler subtracted the failed run's duration from the next 60s tick, the walk restarted immediately, producing the permanent full-throttle loop described in #5130. Three changes, matching the issue's suggested fixes 2 and 3 and replacing suggestion 1 with a shape that cannot regress at the next bucket-size step:

- **Unbounded total walk, stall-bounded progress.** The recovery walk now sets `walkdir_timeout: Some(Duration::ZERO)` — the existing repo convention for "no total budget" (same as `skip_total_timeout`, honored by both the local `track_disk_health` path and the remote `walk_dir` RPC path) — and leaves `walkdir_stall_timeout: None`, so per-call progress stalls stay bounded by the operator-tunable drive stall budget (`RUSTFS_DRIVE_WALKDIR_STALL_TIMEOUT_SECS`, profile-aware). A healthy-but-large bucket can now finish its walk; a genuinely hung disk still fails fast via the stall budget. A fixed total budget (even an env-respecting one) was rejected because total walk time scales with bucket size, so any fixed value re-creates this loop the moment a bucket outgrows it, while drive liveness — what the timeout was actually protecting — is exactly what the stall budget bounds (per `docs/operations/drive-timeout-tuning.md`).
- **Failure backoff measured from completion.** A failed run now waits 60s, doubling per consecutive failure to the existing 600s cap, and the failed run's duration is no longer subtracted from the next delay, so a run that fails after more than 60s cannot restart immediately. Success resets the failure interval. Idle-sweep backoff, pagination markers, and the notify-driven retry path are unchanged.
- **Opt-out for tier-less deployments.** New env `RUSTFS_TIER_FREE_VERSION_RECOVERY_ENABLED` (default `true`). When `false`, only the tier free-version recovery worker is skipped — expiry workers and the other recovery workers still start. Invalid or non-unicode values log a warning and fail open to `true`.

`replication_resyncer.rs` carries the same hardcoded 60s constant; it is intentionally left untouched to keep this diff narrow (called out in #5130 as the same problem shape).

## Verification

- `cargo fmt --all --check`
- `make pre-commit` (guard scripts + `cargo check --workspace --exclude e2e_test`)
- `make pre-pr` (clippy `--all-targets --all-features -D warnings` + nextest workspace suite), on macOS aarch64
- Targeted: `cargo nextest run -p rustfs-ecstore -E 'test(tier_free_version_recovery_disabled_does_not_consume_start_guard) | test(tier_free_version_recovery_production_entrypoint_enqueues_seeded_item) | test(walk_dir_total_timeout_disable_modes_keep_stream_pending) | test(walk_dir_options_preserve_zero_total_and_inherited_stall_timeouts) | test(recovery_walk_disables_total_timeout_and_inherits_stall_timeout) | test(tier_free_version_recovery_failure_backoff_waits_after_completion_and_resets_after_success)'`

Adversarial validation (high-risk tier, one-line verdicts):

- Correctness: attacked backoff arithmetic (saturating doubling, 600s cap, reset-on-success), env parsing degenerate values (absent/empty/invalid/non-unicode → fail-open with warning), and the `Some(Duration::ZERO)` mapping through `timeout_ms` to the ZERO-means-unbounded branch in both local and remote walk consumers — no break found.
- Simplicity: in-place edit preserving the scheduler's control-flow shape; the one new helper (`recovery_walk_options`) exists to pin the walk-options contract under a revert-detecting test and replaces the inline options literal; `parse_bool` reused from `rustfs_utils` — no smaller equivalent diff found.
- Security: no authz, RPC auth, serde, or secret surface touched; the new env var is operator-supplied config parsed to a bool with the existing helper — no break found.
- Concurrency/durability: gating happens once at startup before spawn; backoff state stays task-local inside the loop; walk cancellation and shutdown paths unchanged; no commit/fsync ordering touched — no break found.
- Compatibility: no S3 API, xl.meta, or on-wire format change; `WalkDirOptions.timeout_ms` already exists with `#[serde(default)]`, and an older peer that ignores the explicit 0 falls back to its env-resolved walkdir timeout (fail-safe during rolling upgrade); env var absent preserves prior default behavior — no break found.
- Performance: removes a pathological every-60s full-bucket re-walk (net reduction in background metadata I/O); no per-request path, allocation, or lock touched — no break found.
- Test coverage: startup opt-out is exercised through the real spawn boundary and proves the disabled path does not consume the `OnceLock`; explicit ZERO is pinned through the recovery options, `ListPathRawOptions` to `WalkDirOptions` mapping, and the local `DiskStore::walk_dir` timeout wrapper — no uncovered changed line found.

## Impact

No S3 API, configuration format, xl.meta schema, or on-wire protocol change. One new optional environment variable, `RUSTFS_TIER_FREE_VERSION_RECOVERY_ENABLED`, default `true` (existing behavior). Deployments with buckets that cannot be walked in 60s stop looping and stop flapping drive health via `walk_dir` timeouts; deployments whose walks finish under 60s see no behavior change apart from the removed total budget on this one background walk.

## Additional Notes

Reproduced in production on a single-node `1.0.0-beta.10` deployment (Synology NAS, HDD-class storage) with Grafana Loki and Proxmox backup buckets: recovery ticks logged `walk_dir` `operation_timeout` after exactly 60s and restarted immediately, sustaining constant metadata I/O on an otherwise healthy NAS. This change is extracted from the source patch we maintain against `1.0.0-beta.11` for that deployment.

### DIFF
--- a/crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs
+++ b/crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs
@@ -80,7 +80,11 @@ use rustfs_data_usage::TierStats;
 use rustfs_filemeta::{
     FileInfo, FileInfoOpts, NULL_VERSION_ID, RestoreStatusOps, TRANSITION_COMPLETE, get_file_info, is_restored_object_on_disk,
 };
-use rustfs_utils::{get_env_i64, get_env_usize, path::encode_dir_object, string::strings_has_prefix_fold};
+use rustfs_utils::{
+    get_env_i64, get_env_usize,
+    path::encode_dir_object,
+    string::{parse_bool, strings_has_prefix_fold},
+};
 use s3s::dto::{
     BucketLifecycleConfiguration, DefaultRetention, ExpirationStatus, ObjectLockConfiguration, RestoreRequest,
     RestoreRequestType, RestoreStatus, Timestamp,
@@ -140,6 +144,7 @@ pub const AMZ_ENCRYPTION_KMS: &str = "aws:kms";
 pub const ERR_INVALID_STORAGECLASS: &str = "invalid tier.";
 const ENV_STALE_UPLOADS_EXPIRY: &str = "RUSTFS_API_STALE_UPLOADS_EXPIRY";
 const ENV_STALE_UPLOADS_CLEANUP_INTERVAL: &str = "RUSTFS_API_STALE_UPLOADS_CLEANUP_INTERVAL";
+const ENV_TIER_FREE_VERSION_RECOVERY_ENABLED: &str = "RUSTFS_TIER_FREE_VERSION_RECOVERY_ENABLED";
 const DEFAULT_STALE_UPLOADS_EXPIRY: StdDuration = StdDuration::from_secs(24 * 60 * 60);
 const DEFAULT_STALE_UPLOADS_CLEANUP_INTERVAL: StdDuration = StdDuration::from_secs(6 * 60 * 60);
 const TIER_FREE_VERSION_RECOVERY_BASE_INTERVAL: StdDuration = StdDuration::from_secs(60);
@@ -2084,7 +2089,51 @@ async fn abandon_manual_transition_recovery_lease(api: Arc<ECStore>, job_id: Uui
     Ok(())
 }
 
+fn tier_free_version_recovery_enabled() -> bool {
+    resolve_tier_free_version_recovery_enabled(env::var(ENV_TIER_FREE_VERSION_RECOVERY_ENABLED))
+}
+
+fn resolve_tier_free_version_recovery_enabled(value: Result<String, env::VarError>) -> bool {
+    match value {
+        Ok(value) => match parse_bool(value.trim()) {
+            Ok(enabled) => enabled,
+            Err(_) => {
+                warn!(
+                    event = EVENT_LIFECYCLE_WORKER_STATE,
+                    component = LOG_COMPONENT_ECSTORE,
+                    subsystem = LOG_SUBSYSTEM_LIFECYCLE,
+                    env = ENV_TIER_FREE_VERSION_RECOVERY_ENABLED,
+                    reason = "invalid_boolean",
+                    "Invalid tier free-version recovery setting; using enabled default"
+                );
+                true
+            }
+        },
+        Err(env::VarError::NotPresent) => true,
+        Err(env::VarError::NotUnicode(_)) => {
+            warn!(
+                event = EVENT_LIFECYCLE_WORKER_STATE,
+                component = LOG_COMPONENT_ECSTORE,
+                subsystem = LOG_SUBSYSTEM_LIFECYCLE,
+                env = ENV_TIER_FREE_VERSION_RECOVERY_ENABLED,
+                "Non-Unicode tier free-version recovery setting; using enabled default"
+            );
+            true
+        }
+    }
+}
+
 fn spawn_tier_free_version_recovery_once(api: Arc<ECStore>, started: &OnceLock<()>) -> Option<JoinHandle<()>> {
+    if !tier_free_version_recovery_enabled() {
+        warn!(
+            event = EVENT_LIFECYCLE_WORKER_STATE,
+            component = LOG_COMPONENT_ECSTORE,
+            subsystem = LOG_SUBSYSTEM_LIFECYCLE,
+            env = ENV_TIER_FREE_VERSION_RECOVERY_ENABLED,
+            "Tier free-version recovery disabled by configuration"
+        );
+        return None;
+    }
     if started.set(()).is_err() {
         return None;
     }
@@ -2256,6 +2305,7 @@ fn jitter_tier_free_version_recovery_delay(delay: StdDuration) -> StdDuration {
 struct TierFreeVersionRecoverySchedule {
     next_delay: StdDuration,
     idle_interval: StdDuration,
+    failure_interval: StdDuration,
     previous_run_duration: StdDuration,
     jitter_next_delay: bool,
     bucket_marker: Option<String>,
@@ -2268,6 +2318,7 @@ impl Default for TierFreeVersionRecoverySchedule {
         Self {
             next_delay: StdDuration::ZERO,
             idle_interval: TIER_FREE_VERSION_RECOVERY_BASE_INTERVAL,
+            failure_interval: TIER_FREE_VERSION_RECOVERY_BASE_INTERVAL,
             previous_run_duration: StdDuration::ZERO,
             jitter_next_delay: false,
             bucket_marker: None,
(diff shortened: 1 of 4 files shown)


harryhan24:

houseme:
Rebased this PR onto the latest `origin/main` and resolved the lifecycle worker test-import conflict introduced by the newer durable manual-transition changes.

Updated head: `be30ee4f2187331182b85c842eda2456bb1661de`

Local verification after the rebase:

- `cargo fmt --all --check` - passed.
- `git diff --check origin/main...HEAD` - passed.
- `cargo test -p rustfs-ecstore tier_free_version_recovery -- --nocapture` - passed, 40/40 matching tests.

The fork PR workflow initially required maintainer approval; I approved run `30182499760`, and the main CI matrix is now running. Current green items include CLA, Typos, Quick Checks, and io_uring; no failures seen so far.

cxymds (approved):

houseme:
Rebased this PR onto the latest `origin/main` and resolved the lifecycle test-module import conflict.

Local verification:
- `cargo fmt --all`
- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore tier_free_version_recovery -- --nocapture` (41 related tests passed)

Pushed with `--force-with-lease` from `a26eb9fd` to `7aec7a14`.

houseme:
Rebased this PR onto the latest `origin/main` and resolved the lifecycle import conflict in `bucket_lifecycle_ops.rs`.

Local focused validation passed after the rebase:

```bash
cargo fmt --all --check
git diff --check origin/main...HEAD
cargo test -p rustfs-ecstore tier_free_version_recovery -- --nocapture
```

The PR is now mergeable again and the new CI run is in progress on head `7aec7a14c7f383c761486286ba3f210e612f77f5`.

houseme:
Rebased this PR onto the latest `origin/main` and confirmed the head is now `7aec7a14c7f383c761486286ba3f210e612f77f5`.

Local focused verification passed:

```bash
cargo test -p rustfs-ecstore tier_free_version_recovery -- --nocapture
```

Result: 41 `tier_free_version_recovery` tests passed. GitHub now reports the PR as mergeable; the new CI run is in progress.

PR #5234, opened 2026-07-25
feat(rpc): bind canonical body digest into internode mutating disk RPC signatures

overtrue:
## Summary

Next step of the #1327 internode RPC signature hardening (design C′, PR6 — body-digest binding), building on the server-side v2 verification landed in #5160. This binds a **canonical request-body digest** into the v2 HMAC signature for every mutating NodeService disk RPC, so an on-path attacker on the default-plaintext internode channel can no longer tamper with a mutation payload (or strip the msgpack `_bin` field to force the JSON fallback decode) without invalidating the signature.

Today the v2 signature binds destination + method path + timestamp + nonce, but mutating disk RPCs sign `UNSIGNED-PAYLOAD` for the body digest — the body itself is unbound. This PR closes that gap for the 13 mutating disk RPCs while staying **fail-open by default** so rolling upgrades are unaffected.

## What it does

- **`crates/protos`** — adds `canonical_*_request_body()` builders for `RenameData`, `DeleteVersion`, `DeleteVersions`, `WriteMetadata`, `UpdateMetadata`, `WriteAll`, `Delete`, `DeletePaths`, `RenameFile`, `RenamePart`, `DeleteVolume`, `MakeVolume`, `MakeVolumes`. Each is a domain-separated, length-prefixed byte layout (via a small `CanonicalBodyBuilder`) covering every semantic wire field — **including both the msgpack `_bin` payload and its JSON compatibility copy** — so tampering with either encoding, or downgrading `_bin` → JSON, changes the digest. This follows the existing `canonical_heal_control_request_body` / `canonical_tier_mutation_rpc_body` pattern in the same file.
- **Client (`remote_disk.rs`)** — each of the 13 mutating RPCs computes its canonical body and calls `set_tonic_canonical_body_digest`; the signing interceptor folds the digest (plus a fresh replay-protected nonce) into the v2 signature scope. No proto fields added; digests are computed from existing fields.
- **Server (`node_service/disk.rs`)** — each handler calls `verify_disk_mutation_digest` **before** any disk work. A digest-bearing request is verified against the canonical bytes rebuilt from the received fields (the digest is already signature-covered by `check_auth`, so it cannot be stripped or forged). A digestless request (old peer) falls back per the gate below.
- **Gate (`http_auth.rs` + `crates/config`)** — `RUSTFS_INTERNODE_RPC_BODY_DIGEST_STRICT` (default `false`, compile-time asserted). Non-strict accepts digestless mutations and increments `rustfs_system_network_internode_body_digest_fallback_total`; strict rejects them. This is a **separate** convergence gate from `RUSTFS_INTERNODE_RPC_SIGNATURE_STRICT` — the two converge on different counters and must not gate each other.
- **Replay cache** — capacity is now configurable via `RUSTFS_INTERNODE_RPC_REPLAY_CACHE_CAPACITY` (default 1,048,576, clamped `>= 1` so a mis-set `0` cannot disable replay protection). Overflow fails closed and increments `rustfs_system_network_internode_replay_cache_overflow_total` — alert on a sustained non-zero rate (undersized capacity → refused writes).

## Rolling-upgrade safety

- **Old client → new server**: digestless mutation is accepted (fail-open default) and counted. Byte-for-byte legacy acceptance.
- **New client → old server (current main / #5160)**: the v2 scope already includes `content_sha256`; a new client sending a real digest + real nonce verifies unchanged on a server that predates this PR, because both sides compute the same scope from the headers. No wire-format break.
- No proto field added/renumbered. Both strict flags default off with compile-time asserts; the digest fallback counter must read zero fleet-wide across a release window before `BODY_DIGEST_STRICT` is enabled. Registered in `docs/architecture/compat-cleanup-register.md` as `disk-mutation-body-digest`.

### Rollout note (replay-cache capacity)

Body-bound mutating disk RPCs now consume a per-request nonce in the **receiver's** process-global replay cache — previously these RPCs sent `UNSIGNED-PAYLOAD` and consumed nothing. This PR raises the cache capacity default (65,536 → 1,048,576) to absorb the new nonce stream, but a not-yet-upgraded peer still runs the old default. Under sustained heavy write load during the rolling window, a new client can push an old peer's cache to its (smaller) capacity, at which point it **fails closed** (rejects, never silently accepts) and increments `rustfs_system_network_internode_replay_cache_overflow_total`. Recommended rollout: deploy the capacity-bumped/digest-verifying servers fleet-wide first, then let clients converge; watch the overflow counter. This is a bounded, alertable, fail-closed availability consideration — no auth or data-integrity risk.

## Scope

Covers the 13 mutating **disk** RPCs: the 10 object/metadata mutations plus the volume-lifecycle mutations `DeleteVolume`, `MakeVolume`, `MakeVolumes`. `DeleteVolume` in particular carries a `force` (recursive-delete) flag, so leaving it unbound would let an on-path attacker flip a refuse-if-nonempty delete into a recursive wipe — it is now digest-bound like the rest. Bucket-level and lock RPCs remain out of scope for this issue.

`WriteAll` over unary gRPC carries metadata/config/inline data (KB-scale); bulk object bodies flow through the streaming HTTP data transport, not this path, so the added digest cost is negligible.

## Verification

- New unit tests across all three layers: canonical-body injectivity for **every field** of each builder (`crates/protos`), verification semantics — strict/fail-open/replay-consumed/method-transplant/strip-`_bin`-forces-JSON-fallback (`http_auth.rs`), and a per-handler wiring check that drives **all 13** mutating handlers with a matching digest (must pass the gate) and a mismatched digest (must fail closed) — mutation-verified to go red if any handler's gate is removed (`node_service.rs`).
- Reviewed via a 7-role adversarial pass (correctness, security, compatibility, concurrency/durability, simplicity, performance, test-coverage). Findings folded in: extended coverage to the volume-lifecycle mutations (`DeleteVolume`/`MakeVolume`/`MakeVolumes`), added the per-handler and per-field wiring tests, tightened the strict-mode doc wording, and documented the replay-cache rollout ordering.
- `cargo fmt --all --check`, `cargo clippy` (protos/config/io-metrics/ecstore/rustfs, all `--tests`) clean; targeted test suites green. Full lib suites pass; the only red tests are pre-existing flaky-under-load cases (make_bucket `InsufficientWriteQuorum`, metadata-cache barrier `Elapsed`) that pass in isolation and are orthogonal to this diff.

Refs https://github.com/rustfs/backlog/issues/1327


houseme (approved):

houseme (commented):
### Review Summary

I reviewed this PR for correctness, compatibility, security, and rollout behavior. Overall implementation is in good shape and I did not find blocking bugs.

Scope covered appears complete for 13 mutating disk RPCs: canonical body generation + client attachment + server-side validation are wired consistently across request flow.

Non-blocking note:
- `msgpack_json_decode_error` metric is removed in this PR. If that removal is intentional, please ensure the migration intent is clearly documented (or add a replacement metric) so existing alerting/observability is not silently degraded.

PR #5281, opened 2026-07-26
fix(sts): align AssumeRole authorization

GatewayJ:
## Related Issues

Related to https://github.com/rustfs/backlog/issues/1280

## Summary of Changes

- Evaluate primary-user `AssumeRole` authorization in explicit-deny-only mode.
- Preserve explicit `Deny` precedence while allowing resolved policies that omit `sts:AssumeRole`, matching MinIO's standard AssumeRole authorization behavior.
- Deliberately keep policyless users denied to preserve RustFS's historical security behavior; this is narrower than MinIO.
- Fail closed during deny-only evaluation if any direct or group-derived mapped policy document cannot be resolved.
- Track OPA as disabled, initializing, ready, or failed so a configured but unavailable OPA never falls back to local IAM authorization.
- Send the configured bearer token during OPA startup validation as well as runtime authorization.
- Keep RustFS OPA as the authoritative authorization extension and pass `context.deny_only: true` for AssumeRole evaluations.
- Extend the AWS SDK STS compatibility suite with omitted, explicitly allowed, explicitly denied, policyless, temporary-credential, service-account, OPA allow/deny, and unavailable-OPA coverage.
- Centralize the signed admin request helpers shared by the STS compatibility and admin IAM CRUD E2E suites.
- Regenerate the E2E suite inventory after integrating the current `main`.

## Verification

- `cargo test -p rustfs-iam`
- `cargo test -p rustfs-policy`
- `cargo test -p e2e_test sts_query_compat_test -- --nocapture`
- `cargo test -p e2e_test admin_iam_crud_test -- --nocapture`
- `cargo clippy --all-targets -p rustfs-iam -p rustfs-policy -p e2e_test -- -D warnings`
- `make pre-commit`
- `make pre-pr`
- `cargo fmt --all --check`
- `git diff --check`
- Mutation check: rebuilding with `deny_only: false` makes `test_sts_assume_role_opa_contract` fail with HTTP 403; restoring `true` passes.
- Mutation check: restoring the previous combined-policy-only path makes `regular_deny_only_rejects_unresolved_mapped_policy` fail; restoring the resolution guard passes.

## Impact

Primary users with at least one resolved mapped policy are allowed to call `AssumeRole` when their policies do not mention `sts:AssumeRole`; explicit `Deny` remains denied. Policyless users remain denied, preserving RustFS's historical behavior instead of adopting MinIO's broader policyless allowance. A stale or corrupt direct or group-derived mapping that references a missing policy document fails closed during deny-only evaluation instead of evaluating an empty or partial policy. Normal full-policy evaluation retains its historical partial-resolution behavior.

When the RustFS OPA URL is configured, non-owner authorization fails closed while OPA initializes and after startup validation fails; local IAM is used only when OPA is not configured. A startup validation failure remains failed until restart, consistent with the existing startup-only configuration lifecycle. Deployments that previously relied on local IAM fallback while a configured OPA was unavailable, or custom OPA policies that depend on the old `context.deny_only: false` value, must update their expectations. Mixed-version nodes can differ during a rolling upgrade. No storage format, persisted metadata, wire format, or migration changes are introduced.

## Additional Notes

- Correctness: attacked policyless, omitted, allowed, denied, direct/group missing-policy, partial-policy, temporary, service-account, OPA initialization/failure, and OPA allow/deny paths; the URL-disappearing initialization race was fixed to remain failed closed.
- Simplicity: the resolution guard directly uses the existing merge result with a borrowed `HashSet<&str>`; OPA states stay in the existing global plugin slot without adding a new subsystem.
- Security: policyless and unresolved mappings fail closed; explicit IAM `Deny` remains authoritative; a configured but unready OPA cannot silently fall back to local IAM; startup validation includes the configured bearer token.
- Concurrency/durability: the existing `OnceLock`/`RwLock` plugin slot now carries explicit states, releases the lock before network I/O, and introduces no persistence or crash-ordering changes.
- Compatibility: OPA-disabled and normal full-policy paths are unchanged; resolved AssumeRole omission aligns with MinIO, while policyless users deliberately retain RustFS's stricter historical denial.
- Performance: request-path state access keeps the existing shared lock/client-clone shape, and deny-only mapping checks use borrowed policy names without per-name `String` allocation.
- Test coverage: focused IAM tests cover full-evaluation partial mappings plus group explicit-deny and unresolved mappings; E2E tests cover policyless denial, OPA bearer authentication, initialization, failure, and runtime decisions.

---


GatewayJ:
@codex

rustfs/src/admin/handlers/sts.rs
```diff
@@ -183,7 +183,7 @@ async fn handle_assume_role(
             conditions: &conditions,
             is_owner: owner,
             claims: cred.claims_or_empty(),
-            deny_only: false,
+            deny_only: true,
```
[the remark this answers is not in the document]
GatewayJ: Fixed in fc38100d8. `prepare_regular_auth` now verifies that every mapped policy document resolved before deny-only evaluation and fails closed otherwise. Added `regular_deny_only_rejects_unresolved_mapped_policy`; reverting the guard makes the test fail.

crates/e2e_test/src/sts_query_compat_test.rs
```diff
@@ -48,33 +62,51 @@ fn sts_client(url: &str, access_key: &str, secret_key: &str, session_token: Opti
     Client::from_conf(config.build())
 }
 
-async fn create_root_service_account(env: &RustFSTestEnvironment) -> Result<(String, String), BoxError> {
-    let path = "/rustfs/admin/v3/add-service-accounts";
+async fn admin_ok(
+    env: &RustFSTestEnvironment,
+    method: http::Method,
+    path: &str,
+    body: Option<String>,
```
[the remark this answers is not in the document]
GatewayJ: Fixed in fc38100d8. Moved `admin_request` and `admin_ok` to `crates/e2e_test/src/common.rs` and reused them from both the admin IAM CRUD and STS compatibility suites. Both E2E modules pass.

rustfs/src/admin/handlers/sts.rs
overtrue, line 186: **[P1] Fail closed while configured OPA is initializing or unavailable**

OPA setup is detached in `IamSys::new`, while a missing plugin client means both “disabled” and “initializing/failed.” During that state, this deny-only local fallback allows a user whose mapped policy omits `sts:AssumeRole` to mint credentials without the configured OPA decision; a validation failure leaves that fallback active indefinitely. Track disabled/initializing/failed/ready separately (or make OPA initialization part of readiness) and deny until a configured plugin is ready. Add a delayed/unavailable OPA regression test.

overtrue, line 186: **[P2] Cover the policyless MinIO compatibility case**

`prepare_regular_auth` still denies when the authenticated built-in user has no direct or inherited policy mapping, before deny-only evaluation runs. MinIO only checks that `PolicyDBGet` succeeds and issues an inherited but permissionless session for this case. The new E2E always attaches an unrelated S3 policy, so it does not cover the difference. Either allow an empty mapping for this standard AssumeRole path and add the SDK case, or explicitly document that this PR intentionally implements narrower compatibility.

crates/iam/src/sys.rs
```diff
@@ -1163,7 +1163,19 @@ impl<T: Store> IamSys<T> {
             };
         }
 
-        let combined_policy = self.get_combined_policy(&policies).await;
+        let (resolved_policies, combined_policy) = self.store.merge_policies(&policies.join(",")).await;
+        if args.deny_only {
```
overtrue, line 1206: **[P2] Pin both modes and group-derived mappings**

The new unit test covers only `deny_only=true` with a direct user mapping. Please add (1) the same partially unresolved mapping under `deny_only=false`, proving normal partial-resolution behavior remains unchanged, and (2) group-derived explicit-Deny and unresolved-policy cases. Otherwise an accidental unconditional guard or direct-policy-only resolution still passes all new tests.

crates/iam/src/sys.rs
```diff
@@ -1163,7 +1163,19 @@ impl<T: Store> IamSys<T> {
...
+            let resolved_policy_names = MappedPolicy::new(&resolved_policies).policy_set();
```
overtrue: **[P3] Avoid cloning resolved policy names**

`MappedPolicy::new(&resolved_policies).policy_set()` copies the joined string, constructs a timestamp, and allocates a new `String` for every name. Build a `HashSet<&str>` directly from `resolved_policies.split(',')` and compare with `policy_name.as_str()` to keep the same membership semantics without the extra allocations.

overtrue (changes requested):
Found one authorization bypass and two additional correctness/coverage concerns.

The blocking issue is that configured OPA is not authoritative while its detached initialization is pending or after validation fails: the missing plugin client falls back to local IAM, and this change's deny-only evaluation can issue STS credentials without an OPA decision.

Additional findings:
- Policyless built-in users still differ from MinIO, which issues an inherited but permissionless session when policy retrieval succeeds.
- The strict policy-resolution guard lacks regression coverage for deny_only=false and group-derived mappings.
- The resolved-policy membership check performs avoidable String allocations.

Verified the exact 8368017f..67b0419c diff. The focused IAM regression passes locally, git diff --check passes, and the GitHub Test/Lint and E2E jobs are green.

rustfs/src/admin/handlers/sts.rs
GatewayJ: Addressed in 35c852a03 by choosing the narrower compatibility option. Policyless users deliberately remain denied to preserve RustFS's historical security behavior. The AWS SDK E2E suite now pins that denial, and the PR Impact section explicitly documents the difference from MinIO.

crates/iam/src/sys.rs
GatewayJ: Fixed in 35c852a03. Added coverage for `deny_only=false` partial resolution, group-derived explicit Deny, and group-derived unresolved mappings. The existing direct deny-only unresolved-mapping test remains in place.

crates/iam/src/sys.rs
GatewayJ: Fixed in 35c852a03. The resolution guard now builds a borrowed `HashSet<&str>` directly from `resolved_policies.split(',')` and compares with `policy_name.as_str()`, preserving membership semantics without per-name `String` allocation.

rustfs/src/admin/handlers/sts.rs
GatewayJ: Fixed in 35c852a03. OPA now has explicit disabled/initializing/ready/failed states. When the URL is configured, initializing and failed states deny non-owner authorization instead of falling back to local IAM. Startup validation now sends the configured bearer token, and `test_sts_assume_role_fails_closed_while_opa_is_unavailable` covers both delayed initialization and validation failure.

overtrue (approved):

cxymds (approved):

PR #5292, opened 2026-07-26
fix(site-replication): keep reverse direction after config broadcast

abdullahnah92:
## Related Issues

N/A — reported directly from a two-site production deployment (office ↔ home): replication worked initiator → joined site but silently never in reverse, with no push, no log and no error, while `mc admin replicate info` showed `Sync ✔` on both sites and `mc admin replicate status` reported buckets in sync.

## Summary of Changes

Five independent defects combined into that symptom. Each is fixed separately.

**1. A `replication-config` broadcast deleted the receiver's reverse rule (root cause).** `site-repl-*` rules encode the *sender's* outbound direction — their destination ARN names the receiver. `apply_bucket_meta_item` wrote the sender's rule set verbatim over the receiver's, so the receiver was left with a rule whose destination ARN was its own deployment ID. `reconcile_site_replication_bucket_targets` skips the local peer, so no bucket target can ever back that ARN and every object was dropped. The follow-up call only reconciled *targets*, never rules, so nothing rebuilt the lost reverse rule. Now only operator-authored rules travel between sites; each site owns its `site-repl-*` rules, which are rebuilt from the current peer set (`merge_incoming_replication_config`, and `ensure_site_replication_bucket_replication_config_with_runtime` replacing rather than appending). A site-replication ARN arriving in `role` is dropped for the same reason.

Triggers in the field: the console's replication Save button, `mc replicate import`, bucket-metadata import, and `POST /v3/site-replication/repair` — which re-pushed the initiator's bootstrap plan and therefore actively broke the reverse direction it was meant to repair.

**2. Every failure on that path was invisible.** `update_all_targets` discarded target-client build errors with a bare `if let Ok(..)` — a rejected loopback endpoint, bad CA or unparseable URL left no trace anywhere, while the target still appeared in `bucket remote ls`. `replicate_object` logged the resulting "rule has no target for this ARN" at `debug!` and continued, though every other failure in that function logs at `error!`. Both now say so.

**3. The health signal agreed with the broken state.** `site_replication_rule_complete` checked the rule's flags but never that its destination named a *remote* site, so two sites holding byte-identical configs — the exact post-clobber state, where only one can push — satisfied the check. It now takes the owning deployment ID and rejects a rule pointing at its own site.

**4. A stale `site-replicator-0` parent binding could not be repaired.** `update_service_account` cannot rewrite `parent_user`, so both `ensure_site_replicator_service_account` and the join handler take the update branch and preserve a binding that a root-credential change invalidated. Separately, IAM records encrypted with a previous root secret fail to decrypt and surface as "no such account", silently disabling every control-plane push while the site still reports enabled. Startup now reconciles the account: it repairs a stale parent, and reseeds an unreadable account from the shared secret every site-replication bucket target already stores in `BUCKET_TARGETS_FILE` (not root-encrypted, so it survives a credential change). Recreation is refused when the target parent cannot back a service account, since the delete-then-create sequence would otherwise leave the site with no replication account at all.

Startup also reconciles bucket rules and targets, so a deployment already broken by (1) heals on upgrade instead of staying broken until someone recreates the bucket.

**5. A joined site never verified it could reach the initiator.** A peer's endpoint is whatever that peer derived from the `Host` header of the admin request that created the topology, so the initiator can record an address only it can reach (console-port rewrite, NAT, LAN-only host). All of the initiator's own probes succeed in that case. The join now probes each remote peer and reports failures through the existing `initial_sync_error_message`, so `replicate add` surfaces it instead of the reverse direction failing silently forever. It reports rather than rejects — an operator may legitimately open the return path afterwards.

**6. Rule-shape checking cannot see an unreachable peer.** Fix (3) makes a self-pointing rule visible, but a rule can be perfectly formed while the endpoint recorded for its peer is one this site cannot reach — `update_all_targets` then builds no client and every object is dropped, with the rule set still reading as correct. Each site now reports whether all of its `site-repl-*` rules resolve to a live target (`SRBucketInfo.replicationTargetsOnline`, read from the already-resolved client map, so the status path stays cheap), and the status aggregation treats an offline report as a mismatch. Peers that do not send the field are "unknown", never a fault, so a mixed-version upgrade does not flip every bucket to out-of-sync.

**7. Startup-only healing leaves a site broken until the next restart.** The reconcilers now also run on a 10-minute timer. Both are no-ops when the wiring already matches — the bucket pass compares serialized targets and the rule set before writing — so a healthy deployment reads and writes nothing. The tick takes the site-replication lifecycle lock with a non-blocking `try_acquire` and skips the round if an add/remove/endpoint-refresh holds it: those run in phases, and rebuilding rules between two of them would resurrect exactly what the operation just tore down.

Not changed: `SiteReplicationState::service_account_secret_key` is `skip_serializing` and its fallback looks dead, but it still deserializes state written by older versions, so it is a live upgrade-compat read path.

## Verification

```bash
make pre-pr
cargo nextest run -p rustfs -p rustfs-ecstore -p rustfs-iam -p rustfs-madmin
cargo nextest list --profile e2e-repl-nightly -p e2e_test     # 28 (new test lands in the nightly lane)
```

End-to-end A/B against `origin/main` (5c99ca132), two real single-host servers, `mc admin replicate add`, bucket created, both directions confirmed, then the source's own replication config round-tripped through `PutBucketReplication` (what the console does on Save):

| | `origin/main` | this branch |
|---|---|---|
| joined site's rule after the broadcast | `site-repl-9645c1ad…` → **its own deployment ID** | `site-repl-073e1fb2…` → the initiator ✓ |
| joined site → initiator after the broadcast | **fail** | **ok** |
| `mc admin replicate status` | `1/1 Buckets in sync` | `1/1 Buckets in sync` |

The status row is the point: on `origin/main` it reports in-sync while one direction is dead. With fix (3) that state is now reported as a mismatch.

New e2e regression `test_site_replication_config_broadcast_keeps_reverse_direction_real_dual_node` encodes that scenario. It could not be executed locally: every `_real_dual_node` site-replication test in this suite — including pre-existing ones such as `test_site_replication_replicates_object_with_bucket_versioning_real_dual_node` — fails on this macOS host with `site replication did not reach 2 sites` (peers never leave `sync_state=Unknown`), on unmodified `origin/main` as well. That is an independent local-environment issue, not a regression from this PR; the test is wired into the `e2e-repl-nightly` lane where its siblings run.

## Impact

- **Behavior.** A peer's replication config no longer overwrites the receiver's own outbound rules; only operator-authored rules replicate. Sites whose rules were already clobbered are repaired on the next startup or bucket operation.
- **Health reporting.** `replicate status` now reports a bucket as out of sync when a site's rule points at itself, or when a site reports that a rule has no live target. Deployments currently broken either way will flip from "in sync" to "out of sync" on upgrade — that is the bug becoming visible, and the same upgrade repairs the first class.
- **Logs.** Two new `warn!` events on misconfiguration paths (missing target client, unbuildable target client). Neither fires in steady state.
- **Startup and background.** When site replication is enabled, startup adds one pass over buckets and a 10-minute reconcile timer. Metadata is only written when it actually changes, so a healthy deployment does reads and no writes; the timer skips its round entirely while an add/remove holds the lifecycle lock. No effect when site replication is off.
- **Wire format.** One additive optional field on `SRBucketInfo` (`replicationTargetsOnline`). The type has no `deny_unknown_fields`, so older peers ignore it, and its absence is treated as unknown rather than a fault — mixed-version topologies keep working in both directions.
- **Config/API/on-disk formats.** Otherwise unchanged; a site-replication ARN in `arn:minio:…` form is preserved rather than re-minted.

## Additional Notes

Adversarial validation (high risk, all seven roles):

- **Correctness** — attacked empty/absent incoming configs, rules without IDs, a three-site peer set, ARNs naming a removed peer, and the `local_absent` delete path; found and fixed a `return` that skipped the rule rebuild, and a `unwrap_or_default()` on an unreadable parent that would have recreated a healthy account on every boot.
- **Concurrency/durability** — the reachability probe runs under `SITE_REPLICATION_STATE_LOCK`, bounded by the existing 10s request / 3s connect peer timeouts; `backfill_existing_buckets_after_add` already does more network IO under that same lock. Startup reconcile runs before any request is served on the storage path but after HTTP listeners bind, so the delete/create window is real though sub-millisecond and only entered when drift exists — guarded so it is never entered when the new parent cannot back the account. Found during review: the periodic reconcile could interleave with a phased `replicate remove` and resurrect the targets and rules it had just torn down; the tick now takes the lifecycle lock with `try_acquire` and skips instead.
- **Security** — the probe uses `site-replicator-0` against `devnull`, already covered by that account's session policy; no new HTTP surface (`reconcile_*` are startup-only); no secret reaches a log line. Dropping incoming `site-repl-*` rules also removes a peer's ability to inject a rule naming a third site.
- **Compatibility** — no wire or on-disk change; `arn:minio:…` ARNs are reused via the new `existing` parameter to `build_site_replication_config` rather than replaced by freshly minted ones.
- **Performance** — added a serialized-bytes equality check before writing bucket targets so the new startup pass does not churn metadata or rebuild remote S3 clients on every boot.
- **Simplicity** — `merge_incoming_replication_config` has one caller; extracted because the merge rule is the non-obvious part of the fix and is the unit under test.
- **Test coverage** — eight new unit tests: merge keeps the local reverse rule, survives a peer delete, replicates operator rules while dropping a site-replication `role`, returns `None` when nothing remains, ARN reuse, a status test asserting that two sites holding identical configs are now a mismatch, and two covering the target-health verdict (an offline report is a mismatch even when both rule sets are correct; an absent report is unknown, not a fault). Two existing status tests were reworked: they fed the *same* config to both sites, i.e. they pinned the broken state as correct. **Gaps:** `reconcile_site_replicator_service_account` and the periodic task have no unit tests — both need global IAM and object-store state. The reconcilers were exercised through the manual A/B run above, where they are no-ops on a healthy site.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -161,7 +161,7 @@ test-group = 'ecstore-serial-flaky'
 # the nightly profile derives its set as "the replication module MINUS this
 # allowlist", so any new replication test lands in nightly by default (never
 # silently unrun) until it is explicitly blessed as fast here. Keep the two
-# regexes byte-identical. Count invariant: 20 here + 27 nightly = 47 total
+# regexes byte-identical. Count invariant: 20 here + 28 nightly = 48 total
 # (authority: `cargo nextest list`; docs/testing/e2e-suite-inventory.md).
 # HISTORY (2026-07-11): the 20 fast tests were briefly pulled out of this lane
 # (#4724) because they set a loopback (127.0.0.1) replication target that the
@@ -216,7 +216,7 @@ fail-fast = false
 #     and poll until source and target converge; two replicate over HTTPS, two
 #     pin active SSE failure contracts, and one guards event/history observers.
 #     The SSE-S3 contract remains ignored under backlog#1291.
-#   * 11 `_real_dual_node` site-replication tests — each spawns TWO full rustfs
+#   * 12 `_real_dual_node` site-replication tests — each spawns TWO full rustfs
 #     servers and drives the cross-process site-replication control plane.
 #   * 1 `_real_three_node` site-replication test.
 #   * 1 `_real_single_node` service-account round-trip test.
--- a/crates/e2e_test/src/replication_extension_test.rs
+++ b/crates/e2e_test/src/replication_extension_test.rs
@@ -4803,6 +4803,153 @@ async fn test_site_replication_replicates_object_with_bucket_versioning_real_dua
     Ok(())
 }
 
+/// Re-applying a site's own replication config must not disable the peer's reverse direction.
+///
+/// `PutBucketReplication` broadcasts the config to every peer — the console's replication
+/// Save button, `mc replicate import`, and a bucket-metadata import all go through it. The
+/// receiver used to overwrite its rules with the sender's, whose destination ARN names the
+/// receiver itself. No bucket target can satisfy that ARN, so every object written on the
+/// receiver was dropped with only a debug line, while `replicate status` still reported
+/// "1/1 Buckets in sync" because both configs were byte-identical.
+#[tokio::test]
+#[serial]
+async fn test_site_replication_config_broadcast_keeps_reverse_direction_real_dual_node() -> TestResult {
+    init_logging();
+
+    let mut source_env = RustFSTestEnvironment::new().await?;
+    source_env
+        .start_rustfs_server_with_env(vec![], LOOPBACK_REPLICATION_TARGET_ENV)
+        .await?;
+
+    let mut target_env = RustFSTestEnvironment::new().await?;
+    target_env
+        .start_rustfs_server_without_cleanup_with_env(LOOPBACK_REPLICATION_TARGET_ENV)
+        .await?;
+
+    let source_client = source_env.create_s3_client();
+    let target_client = target_env.create_s3_client();
+    let bucket = "site-repl-config-broadcast";
+
+    let add_status = site_replication_add(
+        &source_env,
+        &[
+            PeerSite {
+                name: "broadcast-source".to_string(),
+                endpoint: source_env.url.clone(),
+                access_key: source_env.access_key.clone(),
+                secret_key: source_env.secret_key.clone(),
+                ..Default::default()
+            },
+            PeerSite {
+                name: "broadcast-target".to_string(),
+                endpoint: target_env.url.clone(),
+                access_key: target_env.access_key.clone(),
+                secret_key: target_env.secret_key.clone(),
+                ..Default::default()
+            },
+        ],
+    )
+    .await?;
+    assert!(add_status.success, "unexpected site add result: {add_status:?}");
+    wait_for_site_replication_enabled(&source_env, 2).await?;
+    wait_for_site_replication_enabled(&target_env, 2).await?;
+
+    source_client.create_bucket().bucket(bucket).send().await?;
+    wait_for_bucket_on_target(&target_client, bucket).await?;
+
+    // Both directions work before the broadcast.
+    source_client
+        .put_object()
+        .bucket(bucket)
+        .key("from-source.txt")
+        .body(ByteStream::from_static(b"written on the initiating site"))
+        .send()
+        .await?;
+    assert_eq!(
+        wait_for_object_on_target(&target_client, bucket, "from-source.txt").await?,
+        b"written on the initiating site".to_vec(),
+    );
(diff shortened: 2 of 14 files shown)


abdullahnah92:
CLA requirements are satisfied for this pull request.

houseme:
@codex review

PR #5389, opened 2026-07-28
feat(ecstore): add remote snapshot lease RPCs

cxymds:
## Related Issues

Relates to rustfs/backlog#1526.

Stacked on #5388.

## Summary of Changes

- Add authenticated NodeService RPCs to acquire, renew, and release remote data-directory snapshot leases.
- Bind every lease mutation field to the existing replay-protected internode body digest and exact gRPC method signature.
- Add a versioned capability response; older peers report `UNIMPLEMENTED`, while incompatible protocol versions fail closed.
- Enforce server-side lease TTL bounds and use atomic token replacement for renewal.
- Run lease expiry through one per-service DelayQueue scheduler, canceling superseded and explicitly released entries to keep overhead proportional to active leases.
- Reject malformed and nil lease tokens before storage access.

## Verification

- `cargo check -p rustfs`
- `cargo test -p rustfs-protos snapshot_lease --lib`
- `cargo test -p rustfs-ecstore snapshot_lease --lib`
- `make pre-commit`
- `cargo fmt --all --check`
- `git diff --check`

- Correctness: attacked acquire/renew/release timeouts, token replacement, expiry/release races, malformed tokens, and scheduler failure; scheduler failure releases the newly created lease before returning.
- Simplicity: one shared DelayQueue replaces per-lease sleeper tasks; renewal reuses opaque tokens instead of adding a second lease registry.
- Security: all mutations require existing internode authentication, exact-method binding, and canonical body digests; TTL is bounded to 5 seconds through 5 minutes.
- Concurrency/durability: renewal atomically replaces the old token under the local registry lock; stale timers and duplicate releases are idempotent and filesystem I/O occurs outside registry and scheduler state transitions.
- Compatibility: protobuf additions use new messages and methods without changing existing fields; old peers naturally return `UNIMPLEMENTED` and new clients reject unknown protocol versions.
- Performance: one scheduler task serves the node, and renew/release remove stale queue entries so memory tracks active leases rather than request rate times TTL.
- Test coverage: canonical-body tests pin every signed field, local tests pin atomic renewal and stale-token behavior, and response tests pin protocol and token validation.

## Impact

This is layer B of the snapshot-lease fix. It adds the remote protocol and TTL safety net but does not acquire leases from GET or release namespace locks, so normal request behavior remains unchanged until the stacked GET integration layer lands. No configuration or on-disk migration is required.

## Additional Notes

The PR base is `cxymds/feat-1526-local-snapshot-lease` and must be reviewed after #5388. The follow-up layer C will acquire read-quorum leases before releasing the namespace lock and will retain the lock when a peer reports `UNIMPLEMENTED` or lease quorum cannot be established.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11829,6 +11829,7 @@ dependencies = [
  "futures-util",
  "libc",
  "pin-project-lite",
+ "slab",
  "tokio",
 ]
 
--- a/crates/e2e_test/src/reliant/grpc_lock_server.rs
+++ b/crates/e2e_test/src/reliant/grpc_lock_server.rs
@@ -23,7 +23,8 @@ use rustfs_protos::{
     proto_gen::node_service::{
         BatchGenerallyLockRequest, BatchGenerallyLockResponse, BatchReadVersionRequest, BatchReadVersionResponse,
         GenerallyLockRequest, GenerallyLockResponse, GenerallyLockResult, PingRequest, PingResponse,
-        node_service_server::NodeService,
+        SnapshotLeaseMutationResponse, SnapshotLeaseReleaseRequest, SnapshotLeaseRenewRequest, SnapshotLeaseRequest,
+        SnapshotLeaseResponse, node_service_server::NodeService,
     },
 };
 use std::pin::Pin;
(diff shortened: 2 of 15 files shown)


cxymds:
CI follow-up (commit `c2f503b28`)

Resolved both failures from run 30418565095:

- `Test and Lint` (job 90470282111): `MinimalLockNodeService` did not implement the three new snapshot lease RPC trait methods (`E0046`).
- `End-to-End Tests` (job 90473685951): same independent compile root cause.
- Parent compatibility regression: the standalone mid-file `#[cfg(test)]` module made `production_source()` stop before the response compatibility encoders, so `rustfs-protos::tests::response_compat_send_site_manifest_pins_json_encoder` failed. The TTL test now lives in the existing trailing test module; the manifest gate and exact encoder contract remain unchanged.

Generator/contract verification:

- Regenerated with repository toolchain versions `protoc 34.1` and `flatc 25.9.23`.
- `crates/protos/src/generated/proto_gen/node_service.rs` had zero diff after generation; no generated protobuf file was hand-edited.
- Unrelated FlatBuffers formatting output was excluded.

Verification:

- `cargo test -p rustfs-protos`: PASS (34/34), including both request/response manifest coverage and exact encoder pins.
- `cargo test -p e2e_test --no-run`: PASS; this directly covers both formerly failing CI compile lanes.
- `cargo fmt --all --check`: PASS.
- `git diff --check`: PASS.
- Architecture, unsafe-code, logging, io-uring, extension-schema, body-cache, doc-path, and planning-doc guards from `make pre-commit`: PASS.
- `make pre-commit` workspace `cargo check` progressed through `rustfs` but the local volume reached 0 free bytes and failed creating an rmeta temp directory (`ENOSPC`); this is an environmental failure, not a compiler diagnostic. The full E2E compile and focused proto suite had already passed before build artifacts were cleaned.

High-risk adversarial review:

- Correctness: the test-only server returns explicit `UNIMPLEMENTED` for non-lock RPCs; lock behavior is unchanged.
- Simplicity: direct trait stubs and moving one test into the existing module are the minimum diff; no helper or gate change.
- Security: no auth, digest, or production RPC behavior changed.
- Concurrency/durability: no runtime state, lock, scheduler, or persistence path changed.
- Compatibility: proto generation is reproducible with zero protobuf diff; exact JSON encoder manifest still passes.
- Performance: test-only methods and test-module relocation have no production hot-path impact.
- Test coverage: reverting the trait stubs reproduces CI `E0046`; reverting the module relocation reproduces the manifest encoder failure.

PR #5406, opened 2026-07-29
fix(ilm): preserve restore source version mode

### LINKED ISSUE #4879: ILM serial lane: restore_object_usecase_reports_ongoing_conflict_and_completion fails deterministically on main (transition never completes)
## Summary

`app::lifecycle_transition_api_test::restore_object_usecase_reports_ongoing_conflict_and_completion` (the ILM Integration (serial) CI lane) fails **deterministically on current main**, timing out in its setup phase:

```
panicked at rustfs/src/app/lifecycle_transition_api_test.rs:1558:10:
object should transition before the restore API runs
```

`wait_for_transition` exhausts its full `TRANSITION_WAIT_TIMEOUT` (15s; observed test duration 15.36–15.50s in every reproduction) — i.e. the object never transitions, the failure is not in the restore assertions themselves.

## Evidence

- CI, PR #4876 (a branch that only adds a new isolated crate + CLI subcommand; zero shared code paths with ILM, `Cargo.lock` diff adds only `tar`/`filetime`/the new crate): failed 3/3 runs (two pushes + one rerun) — always exactly this test, the other 33 tests in the lane pass.
- Local reproduction on the PR branch: FAILED, 15.44s.
- **Local reproduction on clean `origin/main` @ `53270054e`** (`fix(ecstore): serialize transitioned object restores (#4877)`): FAILED, 15.50s.

```bash
cargo test -p rustfs --lib app::lifecycle_transition_api_test::restore_object_usecase_reports_ongoing_conflict_and_completion -- --ignored --exact
```

## Notes for triage

- The failure exists both **before** #4877 (CI run at 16:21 UTC on a pre-#4877 base, `35c569348`) and **after** it (local run on `53270054e`), so #4877 neither fully fixed nor solely introduced it; it may still interact with the restore/transition locking it changed (it fixed #4859 in this exact area), or the regression predates it.
- Other `wait_for_transition`-based tests in the same serial lane pass quickly (e.g. `put_and_copy_object_transition_immediately_via_usecases`, 4.3s), so this is not a lane-wide slowdown — something specific to this test's setup (transition after `enqueue_transition_for_existing_objects` on a fresh bucket with a mock tier) stopped completing.
- 5/5 reproductions across two machines/CI runners look deterministic, not flaky-shaped; suggest bisecting main rather than adding to the flake quarantine.

cxymds:
## Related Issues

Fixes #4879

## Summary of Changes

Preserve the resolved source version identity when a restore copy-back moves into its background task. Null versions are now pinned with the nil UUID selector, and the bucket's versioned/suspended mode is forwarded with that selector.

Without both pieces, the restore read the transitioned null version but converted it back to an unversioned `ObjectInfo`; the completion PUT then generated a new UUID version while the original null version remained latest with `ongoing-request="true"`.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `CARGO_TARGET_DIR=/tmp/fix1526-prepr-target cargo test -p rustfs restore_object_usecase_completes_suspended_null_version_in_place -- --ignored --nocapture`
- `cargo test -p rustfs restore_object_usecase_accepts_exactly_one_of_two_concurrent_restores -- --ignored --nocapture`
- `make pre-commit`
- `make pre-pr`: guard scripts and Clippy passed; 11,019 nextest tests passed. Documentation tests were still running when this PR was opened.

## Impact

Versioned and version-suspended buckets now complete restore copy-back on the exact null version selected during restore acceptance. Existing UUID-version behavior is unchanged. No API, configuration, or on-disk format changes.

## Additional Notes

- Correctness: attacked UUID, null, versioned, and suspended source identities; the suspended null-version restore test covers the previously stranded ongoing state.
- Simplicity: the change forwards existing `ObjectOptions` state directly and adds no production helper or new abstraction.
- Security: no authorization, credential, input parsing, or secret-handling surface changes.
- Concurrency/durability: the existing restore operation ID fence remains intact, while the background copy-back is pinned to the accepted version and mode.
- Compatibility: the nil UUID remains RustFS's established internal selector for the S3 null version; wire and metadata formats are unchanged.
- Performance: the background-only path adds at most one nil UUID string allocation and two boolean copies.
- Test coverage: the suspended-bucket regression asserts that restore completes on the original null version, creates no UUID version, and performs exactly one tier GET; it fails on the previous implementation and passes with this change. The dedicated wait is capped at 15 seconds and does not restore the longer completion wait removed from the existing enabled-bucket conflict test by #5404.

---


cxymds:
Coverage follow-up required before merge: the existing restore integration setup creates a versioning-enabled bucket and therefore proves the enabled/null-version path, but it does not exercise the newly forwarded `version_suspended` field. Please add a suspended-bucket null-version regression that asserts the accepted null version completes in place, no UUID version is created, and the tier sees exactly one GET. This is required by the high-risk test-coverage gate because the `version_suspended` assignment can currently be removed while the listed tests remain green.

cxymds:
Fixed the latest E2E failure in `7bf9fd3be`.

The cold-tier PUTs succeeded. The failure was caused by all 512 transition workers concurrently applying CAS updates to the same persisted job record. The CI log showed repeated five-second namespace lock timeouts on `.rustfs.sys/ilm/manual-transition/jobs/...json`; this also starved scan-progress persistence and incorrectly finalized the successful run as `failed` with `tier_failure=1`.

Worker completion now persists only the existing idempotent result journal. The existing heartbeat reconciliation path projects the journal into the job record after the queue drains, preserving real tier-failure attribution while removing the hot shared-record write.

Verification:

- `cargo test -p rustfs-ecstore manual_transition_worker_result -- --nocapture` — 10 passed
- `git diff --check` — passed
- Focused E2E build was attempted, but the shared runner volume reached `ENOSPC` while rebuilding the RustFS binary; this was an infrastructure failure before the test executed.

High-risk adversarial review attacked duplicate result delivery, crash between journal write and projection, concurrent workers, terminal admission release, real tier-failure attribution, rolling internal record compatibility, hot-path write amplification, and revert-sensitive coverage. No unresolved finding remains.

PR #5444, opened 2026-07-29
fix(scanner): persist portable usage cache keys

### LINKED ISSUE #5055: Why do the console object count and size still as loading? The rustfs version is 1.0.0-beta.10 and system is windows 10 There could be hundreds of thousands of files under one of my storage buckets.
Image

[rustfs.log](https://github.com/user-attachments/files/30178545/rustfs.log)
I've set the environment variable RUSTFS_SCANNER_ENABLED to true. What other settings do I still need to configure?

marshawcoco:
## Related Issues

Fixes #5055

## Summary of Changes

- make persisted scanner cache keys use OS-independent slash normalization while preserving literal backslashes
- version the cache-key format and rebuild legacy or malformed cache graphs without publishing partial usage
- validate authoritative cache roots and preserve pending heal state only across same-scope migration
- share the current-cache reuse/rebuild decision between coordinator and remote scanner paths

## Verification

- `cargo fmt --all --check`
- `CARGO_BUILD_JOBS=1 CARGO_INCREMENTAL=0 cargo test -p rustfs-data-usage --lib`
- `CARGO_BUILD_JOBS=1 CARGO_INCREMENTAL=0 cargo test -p rustfs-scanner --lib`
- `CARGO_BUILD_JOBS=1 CARGO_INCREMENTAL=0 cargo test -p rustfs-ecstore ns_scanner --lib`
- `CARGO_BUILD_JOBS=1 CARGO_INCREMENTAL=0 cargo clippy -p rustfs-data-usage -p rustfs-scanner --all-targets --no-deps -- -D warnings`

## Impact

Existing scanner cache files are treated as legacy and rebuilt once using portable keys. The remote namespace scanner remains on protocol v3 because the wire shape is unchanged. S3, object metadata, and remote scanner wire formats are unchanged.

## Additional Notes

Windows evidence for #5055 showed correct fail-closed API behavior but no durable authoritative snapshot because platform-dependent cache keys broke root reconstruction. This change targets that persistence boundary.

Adversarial review: correctness, security, concurrency/durability, compatibility, performance, and test coverage found no unresolved blockers.


houseme (changes requested):
Requesting changes.

The PR improves scanner cache portability, but bucket-level current cache reuse still accepts incomplete cache graphs. `cache_root_entry_info` uses `checked_flatten_complete` only for `DATA_USAGE_ROOT`; bucket caches still use `checked_flatten`, so a current-format bucket cache with detached entries can be treated as authoritative and publish undercounted usage, including stale `0 objects / 0 B`.

Please make authoritative bucket cache reuse/publish require complete graph validation as well, and add a regression test where a `snapshot_complete` bucket cache with `cache_key_format = DATA_USAGE_CACHE_KEY_FORMAT` and detached entries is reset instead of returned as `Current`.

houseme:
Pushed a follow-up fix in `8408b7ab5d555308830e83bb2e9fd356d4ce64e4`.

The update now validates complete bucket-scope cache graphs before reusing or publishing them, while still accepting the legacy parent-only namespace root placeholder used by bucket caches.

Local verification passed:
- `cargo fmt --all --check`
- `cargo test -p rustfs-scanner detached --lib`
- `cargo clippy -p rustfs-scanner --all-targets --no-deps -- -D warnings`
- `git diff --check`

houseme (approved):
The cache graph validation gap has been fixed. The current implementation now rejects detached complete bucket cache entries before authoritative reuse or publish, while preserving the existing bucket-cache root placeholder compatibility.

PR #5446, opened 2026-07-29
fix(ecstore): reload bucket metadata under one write guard

overtrue:
## Problem

`handle_load_bucket_metadata` — the peer `LoadBucketMetadata` RPC handler — loaded a bucket's metadata from disk and *then* called `set_bucket_metadata` to install it. The metadata-system write guard covered only the install, not the read.

Two overlapping reload notifications could therefore both read before either wrote:

```
peer A: load(v1) ............... set(v1)   <- older snapshot installs last
peer B:      load(v2) .. set(v2)
```

The cache ends up on `v1` and keeps serving superseded configuration until the 15-minute refresh loop repairs it — and that loop only runs under dist-erasure, so a non-dist deployment never self-heals.

## Why the load has to be inside the guard

Serializing only the install is not enough: the ordering bug is between the *reads*. Whoever reads last must also write last, and the only way to guarantee that is to hold one guard across both steps. The same guard additionally excludes `update()`, which could otherwise persist a new config and then have a concurrent reload overwrite its cache entry with a pre-update snapshot.

## Change

Add `metadata_sys::reload_bucket_metadata`. The `BucketMetadataSys` method takes `&mut self`, so the read and the install are *necessarily* covered by the same write guard — a read guard cannot call it. That is compile-time enforcement, not a convention a future caller can drift from.

It reads through `self.api` rather than the ambient store handle, so the snapshot comes from the same instance whose cache receives it.

This lands on the same shape as `update_config_with` from #5398 (free fn takes `sys.write()`, `&mut self` method does load + install under that guard, reads via `self.api`), so the two paths agree rather than each inventing their own discipline.

The `rustfs`-side `set_bucket_metadata` facade is now `#[cfg(test)]`: no production path can reconstruct the load-then-set shape.

## Security relevance

Consumers that resolve secrets from this cache with no disk fallback go through the map-only `get()`. Swift TempURL signing keys are such a consumer, and since #5398 Swift account metadata writes persist *and* notify peers — so a TempURL key rotation reaches this exact handler. A lost reload left peers honoring a revoked signing key for unauthenticated pre-signed access for the whole refresh window.

## Not addressed here

A reload still loads via `load_bucket_metadata`, which degrades a missing metadata file to a fabricated in-memory default and installs it over a real cached entry. That is a separate fail-closed defect; this PR deliberately preserves the existing load semantics so the two are not conflated. `load_bucket_metadata_for_update` (added in #5398, which refuses to fabricate under erasure setups) is the natural model for closing it.

## Verification

Rebased onto current `main` (`2423ba8e3f`) after #5398 landed mid-work; one export-list conflict resolved and everything re-run.

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p rustfs -p rustfs-ecstore --all-targets -- -D warnings` — exit 0, no warnings
- `cargo test -p rustfs-ecstore --lib bucket::metadata_sys::` — 9 passed
- `cargo test -p rustfs --lib storage::rpc::node_service::` — 146 passed, 9 ignored
- `cargo test -p rustfs-protocols --features swift --test swift_metadata_persistence` — 2 passed
- architecture guard scripts (layer deps, migration rules, logging, doc paths, migration gate count, security smoke count) — all pass

New test `reload_bucket_metadata_installs_the_current_on_disk_snapshot` writes `v2` straight to disk behind a cache holding `v1` — the state a peer is in when a notification arrives — and asserts the reload installs `v2`. It is wrapped in a timeout because holding the guard across disk I/O means nothing in the load path may re-enter the metadata system; a re-entrant read would deadlock rather than fail visibly.

### DIFF
--- a/crates/ecstore/src/api/mod.rs
+++ b/crates/ecstore/src/api/mod.rs
@@ -127,8 +127,8 @@ pub mod bucket {
             get_global_bucket_metadata_sys, get_lifecycle_config, get_logging_config, get_notification_config,
             get_object_lock_config, get_public_access_block_config, get_quota_config, get_replication_config,
             get_request_payment_config, get_sse_config, get_tagging_config, get_versioning_config, get_website_config,
-            init_bucket_metadata_sys, list_bucket_targets, remove_bucket_metadata, set_bucket_metadata, update,
-            update_bucket_targets_under_transaction_lock, update_config_with,
+            init_bucket_metadata_sys, list_bucket_targets, reload_bucket_metadata, remove_bucket_metadata, set_bucket_metadata,
+            update, update_bucket_targets_under_transaction_lock, update_config_with,
         };
     }
 
--- a/crates/ecstore/src/bucket/metadata_sys.rs
+++ b/crates/ecstore/src/bucket/metadata_sys.rs
@@ -85,6 +85,20 @@ pub async fn set_bucket_metadata(bucket: String, bm: BucketMetadata) -> Result<(
     Ok(())
 }
 
+/// Re-read a bucket's metadata from disk into this instance's cache under the
+/// metadata-system write guard.
+///
+/// This is the entry point for peer reload notifications (the
+/// `LoadBucketMetadata` RPC). Prefer it over loading separately and then
+/// calling [`set_bucket_metadata`]: the guard must cover the disk read as
+/// well as the install, or concurrent reloads can install their snapshots out
+/// of order. See [`BucketMetadataSys::reload_bucket_metadata`].
+pub async fn reload_bucket_metadata(bucket: &str) -> Result<()> {
+    let sys = get_bucket_metadata_sys()?;
+    let mut lock = sys.write().await;
+    lock.reload_bucket_metadata(bucket).await
+}
+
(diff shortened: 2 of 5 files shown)


overtrue:
Closing as superseded. This was developed against an older `main`; #5396 and #5427 have since fixed the same defect upstream, and more thoroughly.

`reload_from_store_under_namespace` on current `main` does everything this PR does and then some:

- orders the reload against a **distributed** bucket namespace lock, not just the process-local metadata-system guard, so the disk read is ordered against the cluster-wide config write rather than only against other local reloads;
- installs via `publish_if_unchanged`, an `Arc::ptr_eq` compare-and-swap that refuses to publish when the cache entry changed after the read — this closes the out-of-order-install window without holding a write guard across disk I/O;
- fails closed with `PEER_METADATA_NOT_PERSISTED` when no metadata is readable, which also fixes the fabricated-default hole this PR explicitly left open;
- already gates the `rustfs`-side `set_bucket_metadata` facade behind `#[cfg(test)]`, same as this PR.

Rebasing would either be a no-op or replace the stronger namespace-lock + CAS approach with a weaker process-local guard, so there is nothing worth salvaging here. Apologies for the noise.

Issue #5442, opened 2026-07-29
Uncheck of Replicate Delete still deletes the file


manju-rn:
**Describe the bug**
When bucket replication is enabled with "Replicate Delete" unselected, the file delete from the source bucket still delete the file in destination bucket

**To Reproduce**
1.  Setup 2 buckets - one as source and another as destination.  _I have setup this in 2 different cluster, but it should work the same even in the same cluster_
2. Setup a replication in the source bucket with target as a destination bucket with appropriate access key/secret key. Select ONLY "Replicate Existing Objects", "Replicate Delete Markers". DO NOT select "Replicate Delete"
3. Upload a file in source bucket.  This gets properly reflected in destination bucket
4. Delete the file in the source bucket (DO NOT select all versions)
5. The file is deleted in the destination cluster, but is available when "Show Delete Object" checkbox is selected. This is working fine as Delete marker was sent from Source to destination. 
6. In the source bucket now delete the file and select "Delete All Versions".  This is replicated in Destination bucket, where the file is also deleted.  This MUST NOT happen as the "Replicate Delete" was NOT selected. It should have ignored the call from source bucket and merely stay "soft" deleted (like #5 above, based on the delete marker that was sent)

**Expected behavior**
The delete "Delete All Versions" MUST NOT permanently delete the file in the destination bucket, it should ONLY "Soft" delete it and available when "Show Delete Object" is clickecd. Currently it hard deletes the file and the file is no longer available in Destination cluster.
Hence, the "Uncheck" of the Replication Delete still deletes the file in replicated bucket

**Screenshots**
<img width="759" height="417" alt="Image" src="https://github.com/user-attachments/assets/65727d75-3287-42b5-b3ad-faf1498aaf87" />

**Desktop (please complete the following information):**
 - OS: k8s on ubuntu
 - Browser Chrome
 - Version: Image tag: "1.0.0-beta.11"

**Smartphone (please complete the following information):**
Not applicable

**Additional context**

manju-rn:
Updated info:
1. When I "Uncheck" both the "Replicate Delete Markers" and "Replicate Delete", the delete operation from Source does not flow to destination, even if "Delete All Versions" is selected. So this feature is fine.
Only when "Replicate Delete" is "unchecked", the "Delete All Versions" also flows to destination and deletes the file, which it should not as explained in the bug above

reatang:
@manju-rn 
Yes, it is indeed as you said, and it is also included in this fix case.

PR #5449, opened 2026-07-29
fix(replication): enforce bucket replication switches

reatang:
## Related Issues

#5442 

## Summary of Changes

- Enforce `DeleteReplication` for permanent deletion of both object versions and delete-marker versions.
- Keep deletes without an explicit wire `versionId` under `DeleteMarkerReplication`, including directory objects whose storage options synthesize a nil UUID.
- Apply delete replication admission to suspended versioning and fail closed when replication configuration lookup fails before local deletion.
- Persist the delete replication decision at admission so queued work is not reclassified after configuration changes.
- Restore the original `main` tag-filter path (`filter.test_tags`) so replication remains compatible with MinIO OR semantics; this PR no longer changes A2 behavior.
- Reject replication configuration and remote-target fields whose runtime behavior is not implemented.

- Correctness: fixed findings for suspended versioning, synthetic directory version IDs, and batch-delete fail-open behavior.
- Simplicity: retained local changes and reused existing replication configuration and decision bridges.
- Security: known unsupported fields are rejected before persistence; strict rejection of unknown XML elements remains governed by the upstream S3 DTO parser.
- Concurrency/durability: no break found in admission snapshots, queued replay, cancellation, or lock ordering.
- Compatibility: no on-disk/on-wire format change; MinIO-compatible delete semantics and historical configuration reads are preserved.
- Performance: admission adds a cached bucket metadata lookup and configuration clone to eligible single deletes; no additional per-object batch serialization was introduced.
- Test coverage: switch inversion, version purge, and unsupported fields are covered by focused tests; A2 has no net change from `main`.

## Verification

- `make pre-commit` — passed.
- `make pre-pr` — rerun in progress after fixing two `clone_on_copy` findings from the first Clippy pass.
- `cargo test -p rustfs-replication` — passed.
- Focused RustFS tests for replication capability validation, remote-target validation, and delete replication state — passed.

## Impact

Bucket replication switches now control the documented delete cases consistently; tag-filter behavior remains unchanged from `main` for MinIO compatibility. Configurations requesting unsupported replication runtime behavior are rejected instead of being accepted without effect. No migration or storage-format change is required.

## Additional Notes

This is a draft PR while the final `make pre-pr` run completes.

---


rustfs/src/app/object_usecase.rs
```diff
@@ -6761,7 +6757,8 @@ impl DefaultObjectUsecase {
                         &opts,
                         gerr.clone(),
                     )
-                    .await;
+                    .await
+                    .map_err(ApiError::from)?;
```
cxymds, line 6762: [P1] Keep batch deletes fail-closed when replication config lookup fails

This strict call is reached only when `replicate_deletes` is true, but that flag comes from `has_replication_rules`, which converts every replication-config read error—not only `ConfigNotFound`—to `false`. A transient metadata I/O or quorum error therefore bypasses this `?`, and the local `DeleteObjects` call proceeds without replication admission. Please make the gate fallible and propagate non-`ConfigNotFound` errors, with a test proving the backend delete is not called.

rustfs/src/app/object_usecase.rs
```diff
@@ -7102,6 +7104,28 @@ impl DefaultObjectUsecase {
...
+            } else {
+                fallback_source = ObjectInfo {
+                    name: key.clone(),
+                    ..Default::default()
+                };
+                &fallback_source
+            };
+            match metadata_sys::get_replication_config(&bucket).await {
+                Ok((config, _)) => {
+                    delete_replication_state_from_config(&config, source, version_id_clone.as_ref().and(source.version_id), false)
```
cxymds, line 7121: [P1] Persist the admission decision with the local delete

The computed state remains only in this stack local; `opts.delete_replication` is still `None` when `delete_object` runs. The set-disk layer therefore recomputes against the live config and persists that later decision. A crash after the local commit but before `schedule_replication_delete` loses the admitted targets and can leave the remote version undeleted. Please pass this state through `opts.delete_replication` before the delete so the decision is committed atomically, and cover the crash/restart plus config-change case.

rustfs/src/app/object_usecase.rs
```diff
@@ -7102,6 +7104,28 @@ impl DefaultObjectUsecase {
...
+            let source = if let Some(source) = existing_object_info.as_ref() {
+                source
+            } else {
+                fallback_source = ObjectInfo {
+                    name: key.clone(),
+                    ..Default::default()
+                };
+                &fallback_source
+            };
+            match metadata_sys::get_replication_config(&bucket).await {
```
cxymds, line 7119: [P1] Avoid cloning the full replication config on every DELETE

This new per-request lookup deep-clones `ReplicationConfiguration` in `metadata_sys`; `filter_actionable_rules` then clones matching rules again. On a versioned bucket with hundreds of rules, high-rate DELETE gains O(rule-count) allocations per request. Please expose a shared/borrowed configuration snapshot and filter by reference, or provide benchmark evidence that this hot-path cost is acceptable.

crates/replication/src/config.rs
```diff
@@ -51,6 +51,28 @@ pub enum ReplicationTargetValidationError {
...
+pub fn unsupported_replication_config_field(config: &ReplicationConfiguration) -> Option<&'static str> {
+    for rule in &config.rules {
+        if rule
+            .source_selection_criteria
+            .as_ref()
+            .is_some_and(|criteria| criteria.sse_kms_encrypted_objects.is_some())
+        {
+            return Some("SourceSelectionCriteria.SseKmsEncryptedObjects");
+        }
+        if rule.destination.encryption_configuration.is_some() {
```
cxymds, line 63: [P1] Reject all accepted-but-ignored destination fields

This capability list still accepts `Destination.StorageClass`, `Destination.AccessControlTranslation`, and `Destination.Account`, but the replication execution path does not consume those rule fields. For example, a config requesting `GLACIER` can be accepted and persisted while replicas use the source/remote-target storage class. Please either implement these semantics or reject each non-default field here and add negative tests.

rustfs/src/admin/handlers/replication.rs
```diff
@@ -179,6 +179,20 @@ impl RemoteTargetRequest {
             return Err(s3_error!(InvalidRequest, "credentials.secretKey is required"));
         }
 
+        for (unsupported, configured) in [
```
cxymds, line 182: [P1] Reject the other accepted-but-ignored remote-target fields

`RemoteTargetCredentialsRequest` still accepts `session_token` and `expiration`, and this request accepts arbitrary `api`, but the target client builder passes only access/secret credentials and always constructs the current AWS SDK client. An STS target can therefore be accepted and persisted, then fail every replication request because the session token was dropped. Please implement these fields or reject non-default values at this boundary and add coverage.

cxymds (commented):
I found several blocking correctness, durability, compatibility, and performance issues in the current revision.

One additional durability issue cannot be attached to a changed line: delete tasks that spill to MRF do not persist `replicate_decision_str`; startup replay in `replication_pool.rs` explicitly derives a new decision from the live bucket configuration. A delete admitted under configuration A can therefore be skipped or sent to configuration B after restart. Please persist the admitted decision/targets in a backward-compatible MRF field and use live configuration only for legacy entries.

The new tests exercise helper functions, but do not cover the real `execute_delete_object`, `execute_delete_objects`, or MRF replay paths. Please add regression tests for config lookup failure before local mutation, atomic persistence of the admitted decision, directory synthetic nil UUID versus an explicit wire version ID, suspended versioning, and config changes across MRF restart.

Also, the claim that unknown replication XML is rejected upstream needs a regression test: the current generated parser skips unknown elements before the capability validator sees the DTO.

rustfs/src/app/object_usecase.rs
cxymds: Addressed in rustfs/rustfs#5599. Live delete admission now loads the request-level snapshot through the instance context and returns an error for metadata/config failures instead of treating them as no replication rules. Batch delete uses the same fail-closed admission result, with focused coverage for the error path.

rustfs/src/app/object_usecase.rs
cxymds: Addressed in rustfs/rustfs#5599. The validated request-level admission snapshot is carried through `ObjectOptions` into the set-disk delete path and the persisted replication decision is used by heal/resync, so a config change or crash between local delete and queue scheduling cannot silently recompute a different decision.

rustfs/src/app/object_usecase.rs
cxymds: Addressed in rustfs/rustfs#5599. The request-level delete snapshot is loaded once and shared across objects in a batch; tag-filtered rules are evaluated under the delete lock, while direct storage callers use an injected snapshot or the empty snapshot without reloading metadata per object.

crates/replication/src/config.rs
cxymds: Not included in rustfs/rustfs#5599. Destination storage class and related destination semantics remain tracked as follow-up work in rustfs/backlog#1644 so this delete-admission change does not claim unsupported replication behavior.

rustfs/src/admin/handlers/replication.rs
cxymds: Not included in rustfs/rustfs#5599. Remote target session-token/expiration/API validation remains tracked in rustfs/backlog#1644 and will be handled as a separate compatibility and credential-boundary change.

Issue #3241, opened 2026-06-06
Add support for Trivial File Transfer Protocol (TFTP)


rgl:
**Is your feature request related to a problem? Please describe.**

Add support for [Trivial File Transfer Protocol (TFTP)](https://en.wikipedia.org/wiki/Trivial_File_Transfer_Protocol). This is normally used in the context of network [Preboot Execution Environment (PXE) booting](https://en.wikipedia.org/wiki/Preboot_Execution_Environment).

**Describe the solution you'd like**

I would be happy with basic read-only support.

loverustfs:
Before August,our focus is on the release of the stable version.
Thank you for your advice.

majinghe:
Community contribution are welcomed.

PR #5470, opened 2026-07-30
feat: support for Trivial File Transfer Protocol (TFTP)

Littlew0od:
## Related Issues

#3241

### TFTP Server Implementation for RustFS

The TFTP service is built using the [`async-tftp`](https://crates.io/crates/async-tftp) crate (version `0.4.2`).

Each TFTP operation is mapped directly to an S3 call against the configured `StorageBackend`:

- **RRQ (Read Request):**  
  The entire object is downloaded from S3 via a `GetObject` call into an in‑memory buffer. The buffer is then streamed to the TFTP client.

- **WRQ (Write Request):**  
  Incoming data from the client is accumulated in an in‑memory buffer. When the transfer completes (i.e., when the writer is dropped), the complete buffer is uploaded to S3 in a single `PutObject` call.

Since TFTP itself provides **no encryption** and **no authentication**, three environment‑based access‑control mechanisms are added to the RustFS implementation:

1. **`RUSTFS_TFTP_ACCESS_KEY`** – Binds all TFTP operations to a specific IAM user. The attached IAM policy determines which buckets and actions are permitted.

2. **`RUSTFS_TFTP_DEFAULT_BUCKET`** – When set, locks all TFTP access to a single bucket, preventing clients from reaching other buckets in the cluster. When not set, paths are resolved as `/<bucket>/<key>`.

3. **`RUSTFS_TFTP_ACCESS_MODE`** – Restricts the allowed transfer direction:
   - `ro` – read‑only (RRQ allowed, WRQ rejected)
   - `wo` – write‑only (WRQ allowed, RRQ rejected)
   - `rw` – both read and write allowed

## Verification

- `make pre-commit`
- `cargo nextest run --all --exclude e2e_test`
- `cargo nextest run --all --exclude e2e_test -features tftp`

#### New feature

The TFTP module is feature-gated behind the `tftp` feature.

```
# rustfs/Cargo.toml
tftp = ["rustfs-protocols/tftp"]
# crates/protocols/Cargo.toml
tftp = ["dep:async-tftp", "dep:futures-lite"]
```

#### New Environment variables

| Variable | Required | Default | Description |
|---|---|---|---|
| `RUSTFS_TFTP_ENABLE` | Yes | false | Set to true value to enable the TFTP server |
| `RUSTFS_TFTP_ADDRESS` | No | `0.0.0.0:6969` | UDP bind address |
| `RUSTFS_TFTP_DEFAULT_BUCKET` | No | (none) | Default S3 bucket; when set, all paths use this bucket and the path becomes the object key. When not set, paths are resolved as `/<bucket>/<key>` |
| `RUSTFS_TFTP_ACCESS_MODE` | No | `rw` | Access mode: `ro`/`read-only` (read only), `wo`/`write-only` (write only), `rw`/`readwrite` (read-write) |
| `RUSTFS_TFTP_ACCESS_KEY` | No | (admin user) | Access key for S3 authentication. When set, all requests are authorized under that user's IAM policy |

#### Document

Detailed document in `crates/protocols/src/tftp/README.md`

## Additional Notes

N/A


Littlew0od:

PR #5473, opened 2026-07-30
feat: support for Trivial File Transfer Protocol (TFTP)

Littlew0od:
## Related Issues

#3241

### TFTP Server Implementation for RustFS

The TFTP service is built using the [`async-tftp`](https://crates.io/crates/async-tftp) crate (version `0.4.2`).

Each TFTP operation is mapped directly to an S3 call against the configured `StorageBackend`:

- **RRQ (Read Request):**  
  The entire object is downloaded from S3 via a `GetObject` call into an in‑memory buffer. The buffer is then streamed to the TFTP client.

- **WRQ (Write Request):**  
  Incoming data from the client is accumulated in an in‑memory buffer. When the transfer completes (i.e., when the writer is dropped), the complete buffer is uploaded to S3 in a single `PutObject` call.

Since TFTP itself provides **no encryption** and **no authentication**, three environment‑based access‑control mechanisms are added to the RustFS implementation:

1. **`RUSTFS_TFTP_ACCESS_KEY`** – Binds all TFTP operations to a specific IAM user. The attached IAM policy determines which buckets and actions are permitted.

2. **`RUSTFS_TFTP_DEFAULT_BUCKET`** – When set, locks all TFTP access to a single bucket, preventing clients from reaching other buckets in the cluster. When not set, paths are resolved as `/<bucket>/<key>`.

3. **`RUSTFS_TFTP_ACCESS_MODE`** – Restricts the allowed transfer direction:
   - `ro` – read‑only (RRQ allowed, WRQ rejected)
   - `wo` – write‑only (WRQ allowed, RRQ rejected)
   - `rw` – both read and write allowed

## Verification

- `make pre-commit`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo nextest run --all --exclude e2e_test`
- `cargo nextest run --all --exclude e2e_test --features tftp`
- `make test` has some issues, which are not introduced by this PR.
```bash
$ make test
Running script tests...
entrypoint credential policy tests passed
internode grpc A/B bench env tests passed
exact 1MiB handoff ABBA tests passed
pinned paired ABBA bench tests passed
monitor script missing invalid --runs guard output
make: *** [.config/make/tests.mak:26: script-tests] Error 1
```

#### New feature

The TFTP module is feature-gated behind the `tftp` feature.

```
# rustfs/Cargo.toml
tftp = ["rustfs-protocols/tftp"]
# crates/protocols/Cargo.toml
tftp = ["dep:async-tftp", "dep:futures-lite"]
```

#### New Environment variables

| Variable | Required | Default | Description |
|---|---|---|---|
| `RUSTFS_TFTP_ENABLE` | Yes | false | Set to true value to enable the TFTP server |
| `RUSTFS_TFTP_ADDRESS` | No | `0.0.0.0:6969` | UDP bind address |
| `RUSTFS_TFTP_DEFAULT_BUCKET` | No | (none) | Default S3 bucket; when set, all paths use this bucket and the path becomes the object key. When not set, paths are resolved as `/<bucket>/<key>` |
| `RUSTFS_TFTP_ACCESS_MODE` | No | `rw` | Access mode: `ro`/`read-only` (read only), `wo`/`write-only` (write only), `rw`/`readwrite` (read-write) |
| `RUSTFS_TFTP_ACCESS_KEY` | No | (admin user) | Access key for S3 authentication. When set, all requests are authorized under that user's IAM policy |

#### Document

Detailed document in `crates/protocols/src/tftp/README.md`

## Additional Notes

N/A

### DIFF
--- a/deploy/observability/README.md
+++ b/deploy/observability/README.md
@@ -1,0 +1,11 @@
+# RustFS Observability Dashboards
+
+This directory contains optional dashboards and observability assets for operating RustFS deployments.
+
+## Grafana
+
+Import `grafana/rustfs-node-observability.json` into Grafana and select a Prometheus data source that scrapes RustFS metrics.
+
+The dashboard uses the RustFS `server` metric label introduced with the node-local observability updates. `server` represents the RustFS node identity and is preferred for RustFS node comparisons. Prometheus `instance` still identifies the scrape target and remains useful for scrape/debugging views, but dashboards that compare RustFS nodes should group and filter by `server`.
+
+During a rolling upgrade, older nodes may still emit metrics without the `server` label. Complete the rollout before using this dashboard for node-by-node comparisons.
--- a/deploy/observability/grafana/rustfs-node-observability.json
+++ b/deploy/observability/grafana/rustfs-node-observability.json
@@ -1,0 +1,885 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 1,
+  "id": null,
+  "links": [],
+  "liveNow": false,
+  "panels": [
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axisCenteredZero": false,
+            "axisColorMode": "text",
+            "axisLabel": "",
+            "axisPlacement": "auto",
+            "barAlignment": 0,
+            "drawStyle": "line",
+            "fillOpacity": 12,
+            "gradientMode": "none",
+            "hideFrom": {
+              "legend": false,
+              "tooltip": false,
+              "viz": false
+            },
+            "lineInterpolation": "linear",
+            "lineWidth": 1,
+            "pointSize": 4,
+            "scaleDistribution": {
+              "type": "linear"
+            },
+            "showPoints": "never",
+            "spanNulls": false,
+            "stacking": {
+              "group": "A",
+              "mode": "none"
+            },
+            "thresholdsStyle": {
+              "mode": "off"
+            }
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              }
+            ]
+          },
+          "unit": "Bps"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 8,
+        "w": 12,
+        "x": 0,
+        "y": 0
+      },
+      "id": 1,
+      "options": {
+        "legend": {
+          "calcs": [
+            "lastNotNull"
+          ],
+          "displayMode": "table",
+          "placement": "bottom",
+          "showLegend": true
+        },
+        "tooltip": {
+          "mode": "multi",
+          "sort": "desc"
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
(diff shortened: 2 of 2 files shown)


crates/protocols/src/tftp/handler.rs
```diff
@@ -0,0 +1,477 @@
...
+    fn poll_flush(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn poll_close(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        Poll::Ready(Ok(()))
+    }
+}
+
+impl<S: StorageBackend + Send + Sync + 'static> Drop for VecWriter<S> {
```
overtrue: **[P1] Do not commit an upload from `Drop`.** `async-tftp` drops the writer after both successful and failed request handling. If a WRQ times out, receives an invalid packet, or is cancelled after writing only part of the payload, this implementation still launches `PutObject` and can replace an existing object with an empty or truncated prefix. Track successful completion explicitly and only commit after a successful close; dropping an incomplete writer must discard or abort the upload.

crates/protocols/src/tftp/handler.rs
```diff
@@ -0,0 +1,477 @@
...
+            bucket,
+            key,
+            access_key,
+        }
+    }
+}
+
+impl<S: StorageBackend + Send + Sync + 'static> AsyncWrite for VecWriter<S> {
+    fn poll_write(mut self: Pin<&mut Self>, _cx: &mut Context<'_>, buf: &[u8]) -> Poll<io::Result<usize>> {
+        self.buf.extend_from_slice(buf);
```
overtrue: **[P1] Bound transfer memory instead of growing an in-memory `Vec` indefinitely.** The advertised WRQ size is ignored and every received block is appended without a limit. RRQs likewise preallocate and buffer the entire object at lines 329–337. Because TFTP clients are unauthenticated and the server accepts concurrent peers, one or more clients can exhaust the process heap. Please enforce configurable transfer/concurrency limits and stream data rather than buffering complete objects.

crates/protocols/src/tftp/handler.rs
```diff
@@ -0,0 +1,477 @@
...
+///   `relative`   → (default_bucket, "relative")
+///
+/// When `default_bucket` is NOT set, the first path component is the bucket:
+///   `/bucket/obj/key` → ("bucket", "obj/key")
+///   `/just-bucket`    → error (no key after bucket)
+pub fn resolve_tftp_path(default_bucket: Option<&str>, path: &Path) -> Result<(String, String), String> {
+    let path_str = path.to_string_lossy();
+    let trimmed = path_str.trim_start_matches('/');
+
+    if let Some(bucket) = default_bucket {
```
overtrue: **[P2] Apply the same path validation in default-bucket mode.** This branch returns the raw trimmed filename and bypasses the control-character and reserved `__XLDIR__` checks used by `parse_s3_path`. As a result, enabling `RUSTFS_TFTP_DEFAULT_BUCKET` changes the validation/security contract and allows internal marker names that the documentation says are rejected. Run both modes through a shared key-validation path.

crates/protocols/src/tftp/server.rs
```diff
@@ -0,0 +1,107 @@
...
+        let handler = TftpdHandler::new(&self.config, Arc::clone(&self.storage));
+
+        let tftpd = TftpServerBuilder::with_handler(handler)
+            .bind(self.config.bind_addr)
+            .build()
+            .await?;
+
+        info!(
+            event = EVENT_TFTP_SERVER_STATE,
+            component = LOG_COMPONENT_PROTOCOLS,
```
overtrue: **[P2] Propagate runtime failures from `serve()`.** The select branch discards the `Result` from `tftpd.serve()`, so a socket/runtime failure is logged as a normal stop and `start()` returns `Ok(())`. Consequently the supervising task never emits its `runtime_failed` path. Preserve the result from this branch and return the error.

overtrue (changes requested):
Thanks for adding the feature-gated TFTP integration. I found several correctness and availability issues that should be addressed before merging. In particular, WRQ persistence currently happens unconditionally from `Drop`, so failed transfers can overwrite objects with partial data. The handler also permits unbounded memory growth for unauthenticated transfers. Path validation and runtime error propagation need to be made consistent as well.

Verification performed:
- `cargo check -p rustfs-protocols --features tftp`
- `cargo test -p rustfs-protocols --features tftp tftp::handler -- --nocapture`

Littlew0od:
@overtrue I apologize for accidentally closing this PR. Thank you for your suggestions. I will revise the changes accordingly and submit a new PR.

PR #5492, opened 2026-07-30
feat: add hotpath primitive profiling coverage

houseme:
## Related Issues
Tracks rustfs/backlog#1575.

## Summary of Changes
Add opt-in hotpath instrumentation and feature passthrough for the RustFS crates that have a controlled profiling surface. The setup keeps profiling feature-gated and preserves feature-off builds.

After review, this PR no longer wraps EC/RIO request-scoped channels or streams with `hotpath::channel!` / `hotpath::stream!`. Those call sites are capacity-sensitive or request-scoped, and hotpath 0.22 proxy/registry behavior can change backpressure or retain unbounded per-instance profiler state. The remaining primitive coverage is intentionally limited until hotpath provides capacity-transparent bounded wrappers or bounded/evictable request-scoped state.

`hotpath-cpu` remains an explicit opt-in feature, but it now implies the local `hotpath` feature in each integrated crate so `--features hotpath-cpu` is directly usable without merging CPU sampling into the lighter `hotpath` feature.

This also includes the scoped dependency refresh that was committed separately as requested.

## Verification
- `cargo fmt --all`
- `cargo fmt --all --check`
- `git diff --check`
- `cargo check -p rustfs-rio`
- `cargo check -p rustfs-rio --features hotpath`
- `cargo check -p rustfs-rio --features hotpath-cpu`
- `cargo check -p rustfs-ecstore --features hotpath`
- `cargo check -p rustfs-ecstore --features hotpath-cpu`
- `cargo check -p rustfs-filemeta`
- `cargo check -p rustfs-filemeta --features hotpath`
- `cargo check -p rustfs-filemeta --features hotpath-cpu`
- `cargo check -p rustfs-policy`
- `cargo check -p rustfs-policy --features hotpath`
- `cargo check -p rustfs`
- `cargo check -p rustfs --features hotpath`
- `cargo check --workspace`
- `cargo check --workspace --features hotpath`
- `cargo check --workspace --features hotpath-cpu`

## Impact
Profiling remains fully opt-in. With the `hotpath` feature disabled, the workspace has been verified with `cargo check --workspace`; hotpath macros compile to no-op expansion and the reverted EC/RIO primitive call sites now use the original raw Tokio channels and streams.

`--features hotpath` keeps CPU sampling dependencies out of the normal profiling build. Use `--features hotpath-cpu` when CPU profiler support is needed; that feature now enables the base hotpath instrumentation automatically in the crate feature graph.

## Additional Notes
The OPA HTTP wrapper is gated on `rustfs-policy`'s local `hotpath` feature so workspace feature unification cannot change the client type for feature-off dependent builds.

There are no remaining `hotpath::channel!` or `hotpath::stream!` wrappers in the EC/RIO/filemeta/policy/root hotpath integration set after the follow-up fix. The broader profiling gap analysis was saved locally under `docs/Profiling`, but that path is ignored by repository policy and is tracked separately in rustfs/backlog#1575.


overtrue:
Thanks for adding the opt-in hotpath coverage. I do not think this is ready to merge yet because the feature-on path has two blocking runtime issues:

1. `hotpath::channel!(..., proxy = true)` is not capacity-transparent. In hotpath 0.22, it keeps the original bounded channel with capacity `C`, adds a capacity-1 proxy channel, and the forwarding task can hold one more message while blocked. The EC queues at `crates/ecstore/src/erasure/coding/encode.rs:547` and `:699` are explicitly sized from `RUSTFS_ERASURE_ENCODE_MAX_INFLIGHT_BYTES`, but the wrapper increases the maximum retained messages to `C + 2`. With 4 MiB expanded blocks and the default 32 MiB budget, the non-batched path can retain about 40 MiB; the batched path can retain about 64 MiB. The RIO body channel at `crates/rio/src/http_reader.rs:1014` has the same backpressure/cancellation change.

2. These request-scoped channels and streams create permanent profiler entries. hotpath 0.22 assigns each instance a unique ID, retains closed channel/stream stats and their preallocated log buffers until the process-wide `HotpathGuard` exits, and scans historical entries when registering a new instance. `HttpWriter::new` alone creates three channels plus one stream per request. A long-running profiling build therefore has unbounded memory growth and cumulative O(n²) registration work, which can make the profiler itself distort the workload or eventually OOM.

The PR also adds no revert-detecting tests for the new instrumentation. Existing tests still pass if the wrappers are removed or labels are wrong, and they do not exercise feature-on stalled-consumer capacity, cancellation, or long-running registry growth.

Please preserve the original bounded-channel semantics, use bounded/evictable or callsite-aggregated profiler state for request-scoped primitives, and add `--features hotpath` regression tests covering backpressure/cancellation and bounded registry/RSS behavior. After updating the branch to the latest `main`, these paths should be re-reviewed.

For completeness: the current CI is green, the latest `main` merges without conflicts, and I found no additional security, compatibility, OPA fail-closed, on-disk, or wire-format blocker.

pawurb:
Hi @houseme, @overtrue , I saw your discussion on integrating hotpath. I’m aware of some of the issues you mentioned and plan to fix them in future releases.

Will you be able to provide a WIP PR demonstrating the channels profiling integration you’re planning (ideally with a command on how to run profiling script), so that I can iterate on it?

BTW I recently released I/O profiling, it should not have any of the issues of channel wrappers, maybe you’ll find it useful: https://hotpath.rs/io_tracing

houseme:
> Hi @houseme, @overtrue , I saw your discussion on integrating hotpath. I’m aware of some of the issues you mentioned and plan to fix them in future releases.
> 
> Will you be able to provide a WIP PR demonstrating the channels profiling integration you’re planning (ideally with a command on how to run profiling script), so that I can iterate on it?
> 
> BTW I recently released I/O profiling, it should not have any of the issues of channel wrappers, maybe you’ll find it useful: https://hotpath.rs/io_tracing

Thanks for reaching out, and thank you for the pointer to the new I/O profiling support.

We merged the safe subset of our hotpath work in #5505: opt-in function-level, allocation, and CPU profiling across the RustFS workspace. We intentionally did not keep `hotpath::channel!` or `hotpath::stream!` in the EC/RIO request paths.

For those paths, the current channel wrapper changes bounded-channel backpressure/cancellation behavior, and request-scoped primitive instances can accumulate profiler state over a long-running process. So I do not want to present #5505 as the channel-profiling WIP you asked for.

We also added an ABBA benchmark runner and runbook in #5507 to make feature-on performance validation reproducible on Linux or a production-like cluster:
https://github.com/rustfs/rustfs/blob/main/docs/operations/hotpath-warp-abba-runbook.md

We have not evaluated the new I/O profiling API in RustFS yet, but it looks like a promising direction because it may provide I/O-level visibility without changing the channel semantics that concern us. If we can produce a minimal integration that preserves bounded backpressure and has bounded/evictable profiling state, we will open it as a WIP PR for iteration.

PR #5603, opened 2026-08-02
docs(kms): correct the stale backend_info and rotation error claims

overtrue:
## Related Issues

Found while auditing rustfs/backlog#1562, rustfs/backlog#1565 and rustfs/backlog#1566 acceptance criteria against shipped behavior. No `rustfs/rustfs` issue tracks these two documentation defects.

## Summary of Changes

Two claims in `docs/operations/kms-backend-security.md` describe code that later changes removed or replaced. Both are corrected to what the tree says today; a third correction removes the same dead claim from a Rust doc comment so the two do not drift apart again.

**1. The `backend_info` reporting surface no longer exists.** The doc claimed the Vault KV2 backend "reports this boundary in its `backend_info` metadata as `at_rest_protection: vault-kv2-acl`". That implementation and its assertion test were removed by #5501 (`76b3c085b5`, "refactor(kms): fold the KmsClient layer into KmsBackend"); the KMS crate has no `BackendInfo` type at all today. The only field-level expression of the boundary that survives is the backup contract, where `VaultKv2` maps to `AtRestProtection::StorageOnly` (`crates/kms/src/backup/capability.rs:70`, selected at `rustfs/src/admin/handlers/kms_backup.rs:371`) and is serialized into a bundle manifest's `at_rest_protection` field (`crates/kms/src/backup/manifest.rs:517`) with the kebab-case wire value `storage-only`. The doc now says that, and says plainly that nothing reports the boundary at runtime — an operator who followed the old sentence would have gone looking for an endpoint that does not exist.

**2. Rotation on Local and Static returns `UnsupportedCapability`, not `InvalidOperation`.** The conclusion the doc drew — that both backends reject rotation outright — is correct; only the error type was wrong. Neither backend advertises `rotate` in `capabilities()` (`crates/kms/src/backends/local.rs:1797-1806` and `crates/kms/src/backends/static_kms.rs:395-399`, both starting from `BackendCapabilities::minimal()`, which sets `rotate: false`), so `rotate_key` falls through to the default trait implementation at `crates/kms/src/backends/mod.rs:270-272`, which returns `unsupported_capability`. `InvalidOperation` survives only in the `#[cfg(test)]` `LocalKmsClient::rotate_key` at `crates/kms/src/backends/local.rs:1401-1407`, which is not on the product path. The contract tests already pin the shipped behavior (`crates/kms/src/backends/contract_tests.rs:56-65` and `:285-289`). The revised sentence also states *why* it is rejected — a capability gap, not a state-machine rejection — because that determines which error a caller must handle.

**3. Same dead claim in a doc comment.** `crates/kms/src/backup/capability.rs:68-69` repeated the `at_rest_protection = "vault-kv2-acl"` parenthetical on `AtRestProtection::StorageOnly`. Removed; the rest of that comment was already accurate.

## Verification

- `make pre-commit`

No tests were added: this changes prose and one Rust doc comment, with no behavior change to cover. Both claims were verified against the tree at `8e3e552576` before editing — `grep -rn "BackendInfo\|backend_info\|vault-kv2-acl" crates/kms/` returns no hits for the removed reporting surface, and the capability and contract-test sources cited above were read directly.

## Impact

Documentation only. No behavior, API, configuration, or compatibility change.

## Additional Notes

The two corrections point in opposite directions and are worth distinguishing when reviewing. The rotation claim was a wrong label on correct guidance — an operator acting on it would still have done the right thing. The `backend_info` claim promised a reporting surface that does not exist, so it could have sent someone hunting for an endpoint or, worse, left them believing the boundary is asserted somewhere at runtime when it is only ever recorded in a backup manifest.

Restoring a `backend_info`-style reporting surface was considered and not done here: that is a behavior change, out of scope for a documentation fix, and it should be designed against the backup contract that now owns this field rather than reintroduced to satisfy a stale sentence.


overtrue:
Superseded by #5602, which landed the same two corrections plus the surrounding context: that no runtime interface reports the at-rest boundary at all, that a backup manifest's `at_rest_protection: storage-only` is a different thing, and what `kms/status` capabilities do expose. Both files on main already carry the corrected text, so rebasing this would produce an empty diff — and resolving the conflict in this branch's favour would revert the fuller wording. Closing rather than rebasing.

Issue #5419, opened 2026-07-29
Multipart UploadPart fails on Windows during rename_part with PATH NOT FOUND reported as “File access denied”


PawelBoreckiPL:
## Environment

- RustFS: `1.0.0-beta.11`
- Also reproduced with: `1.0.0-beta.11-preview.1`
- Git commit reported by both binaries: `1c8088d0b2af0a1afc8df128014b2176037a0622`
- OS: Windows x86-64
- Deployment mode: single node, single disk
- Filesystem: local NTFS volume
- Client: DJI FlightHub 2, using `aws-sdk-go/1.55.5`
- S3 region: `us-east-1`
- TLS enabled

## Summary

Multipart uploads fail consistently during `UploadPart` on Windows.

`CreateMultipartUpload` succeeds, authentication and authorization succeed, and RustFS successfully writes `part.1` into its temporary directory. The failure occurs when RustFS attempts to move the part into the multipart destination.

RustFS reports this as:

```text
InternalError: File access denied
HTTP status: 500
```

However, Process Monitor shows that Windows actually returns `PATH NOT FOUND` because a UUID-named destination directory below `.rustfs.sys\multipart` does not exist.

Regular `PutObject` operations work correctly.

## Steps to reproduce

1. Start RustFS on Windows using a local NTFS directory:

```powershell
rustfs.exe "C:\RustFS-Data" `
  --address "0.0.0.0:9000" `
  --console-enable `
  --console-address "127.0.0.1:9001"
```

2. Create a bucket and an access key with permission to upload objects.

3. Perform a regular `PutObject`.

   Result: success.

4. Initiate a multipart upload.

   Result: success.

5. Upload part number 1.

   Result: HTTP 500:

```xml
<Error>
  <Code>InternalError</Code>
  <Message>File access denied</Message>
</Error>
```

The client retries the operation, but every retry fails in the same way.

## RustFS diagnostic log

RustFS reports:

```text
message: Set disk multipart write quorum unavailable
event: set_disk_multipart
op: upload_part
state: write_quorum_unavailable
stage: rename_part
part_number: 1
required: 1
achieved: 0
failed: 1
total: 1
offline_disks: 0
retryable_failures: 0
dominant_error: other_error
returned_error: disk access denied
```

The source and destination reported by the operation follow this pattern:

```text
source:
.rustfs.sys\tmp\<temporary-id>\part.1

destination:
[...]

PR #5663, opened 2026-08-03
fix(ecstore): allow Windows renames under guarded parents

marshawcoco:
## Related Issues

Related to #5419

## Summary of Changes

- Allow guarded Windows directories to share write access while continuing to deny delete sharing. This lets `reliable_rename` publish child entries without allowing the guarded base, intermediate directories, or destination parent to be renamed or removed.
- Add a fresh-format initialization regression test that writes and reloads every disk's `format.json` slot.
- Add native Windows CI coverage for guarded rename publication, directory-identity protection, startup temporary-directory cleanup, and fresh `format.json` publication.

The Windows directory guard introduced by #5453 shared read access only. Startup then tried to rename a child into a guarded destination parent, and Windows rejected the parent access with `ERROR_SHARING_VIOLATION` (Win32 code 32). The new share mode permits the child write required by the atomic rename while preserving the guard's path-identity guarantee.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore --lib rename_all_ -- --nocapture`
- `cargo test -p rustfs-ecstore --lib cleanup_tmp_on_startup_ -- --nocapture`
- `cargo test -p rustfs-ecstore --lib fresh_format_load_initializes_all_disks -- --nocapture`
- `actionlint .github/workflows/windows-filesystem.yml`
- `./scripts/check_unsafe_code_allowances.sh`
- `./scripts/check_ci_paths_sync.sh`
- `./scripts/check_layer_dependencies.sh`

## Impact

Fresh Windows data directories can complete temporary-directory cleanup and publish `format.json` instead of remaining unready with Win32 code 32. Linux behavior, on-disk formats, public APIs, configuration, and network protocols are unchanged.

## Additional Notes

The beta.12 evidence for #5419 stopped before any multipart request reached RustFS. This change removes that independent startup blocker; it is not evidence that confirms or rejects the multipart behavior fixed by #5363.

High-risk adversarial review verdicts:

- Correctness: attacked direct-child and descendant publication, fresh format persistence, startup cleanup, and guard release behavior; the native Windows tests pin each outcome.
- Simplicity: reduced the production fix to the required `FILE_SHARE_WRITE` flag and reused existing Win32 constants; no new helper or control-flow branch remains.
- Security: reparse-point rejection and delete-share denial remain intact, so guarded directory identity cannot be replaced during publication.
- Concurrency and durability: no lock ordering, retry, fsync, or commit ordering changed; the guard now permits only the child write that the existing atomic rename requires.
- Compatibility: no new Windows API is used, non-Windows code is unchanged, and no on-disk or wire format changes are introduced.
- Performance: the runtime change is a handle share-mode bit with no extra syscall, allocation, or lock.
- Test coverage: reverting `FILE_SHARE_WRITE` makes the native Windows guarded-publication, startup-cleanup, and fresh-format tests fail with the original sharing violation.

### DIFF
--- a/.github/workflows/windows-filesystem.yml
+++ b/.github/workflows/windows-filesystem.yml
@@ -1,0 +1,86 @@
+# Copyright 2024 RustFS Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Windows Filesystem Tests
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+      - "crates/ecstore/src/disk/**"
+      - "crates/ecstore/src/store/init_format.rs"
+      - "crates/ecstore/Cargo.toml"
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - ".github/actions/setup/**"
+      - ".github/workflows/windows-filesystem.yml"
+  pull_request:
+    branches: [ main ]
+    paths:
+      - "crates/ecstore/src/disk/**"
+      - "crates/ecstore/src/store/init_format.rs"
+      - "crates/ecstore/Cargo.toml"
+      - "Cargo.toml"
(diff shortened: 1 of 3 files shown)


houseme:
I don't think this is ready to merge yet. Adding FILE_SHARE_WRITE fixes the sharing violation, but it also lets another writer reopen a guarded destination parent and turn it into a reparse point/junction before the pathname-based rename runs. Denying FILE_SHARE_DELETE blocks rename/removal, but it does not prevent this mutation, so staged data can still be published outside base_dir.

Please make the Windows publication path handle-relative, or otherwise prevent the guarded parent from becoming a reparse point during the guard lifetime, and add a native Windows regression test for that race. @marshawcoco

PR #5669, opened 2026-08-03
fix(replication): preserve multipart source mtime

### LINKED ISSUE #5662: Multipart replication does not preserve source LastModified
**Describe the bug**

When an existing multipart object is replicated from MinIO to RustFS, RustFS does not preserve the source object's `LastModified` value. The replicated object receives a timestamp close to the replication completion time instead.

A regular single-PUT object replicated through the same rule preserves `LastModified` correctly.

The multipart object's body, ETag, VersionId, and user metadata are preserved, so the discrepancy appears to be limited to the modification time.

**To Reproduce**

Environment:

- RustFS: `rustfs/rustfs:1.0.0-beta.12`
- RustFS image digest: `sha256:41fe89380f4120a337790c02af192c3fe7bb55c3edc2e6e9357b487b47c6ab21`
- MinIO source image digest: `sha256:ea6398ea08fa9643fca474c06167c05ef97d1614429a9892eea47584af4e108d`
- MinIO Client: `minio/mc:RELEASE.2025-08-13T08-35-41Z`
- MinIO Client image digest: `sha256:a7fe349ef4bd8521fb8497f55c6042871b2ae640607cf99d9bede5e9bdf11727`
- Docker, with single-node and single-volume RustFS
- Versioning enabled on both source and target buckets

Steps:

1. Upload a small control object containing `control\n` to the MinIO source bucket using a regular PUT.
2. Upload a multipart object consisting of three 5 MiB parts containing deterministic synthetic data: `0x11`, `0x22`, and `0x33` respectively.
3. Set synthetic user metadata `repro-marker=multipart-mtime` on the multipart object.
4. Record the source objects' `LastModified`, VersionId, ETag, user metadata, and body checksum.
5. Wait three seconds.
6. Configure bucket replication from MinIO to RustFS with existing-object replication enabled.
7. Wait for replication to complete.
8. Compare the source and target object metadata and body checksums.

**Actual behavior**

| Object | LastModified delta | VersionId | ETag | Body SHA-256 | User metadata |
| --- | ---: | --- | --- | --- | --- |
| Regular PUT control | 0 ms | Match | Match | Match | Match |
| Three-part multipart object | +54,000 ms | Match | Match | Match | Match |

The issue was also reproduced for three additional 20 MiB multipart objects in three out of three runs, with timestamp differences of +4,000 ms, +4,000 ms, and +2,000 ms. A 70 MiB multipart object showed a difference of +34,314 ms.

**Expected behavior**

Multipart replication should preserve the source object's `LastModified` value, as regular PUT replication already does.

**Screenshots**

N/A.

**Environment**

This is a server-side S3 replication issue reproduced with the container images listed above. Browser and smartphone information are not applicable.

**Additional context**

This prevents MinIO-to-RustFS migrations from preserving object version metadata faithfully. It can also shift lifecycle expiration and noncurrent-version cleanup calculations when those operations depend on the object's modification time.

The following is a suspected cause based on code inspection:

- Regular PUT parses the replication source modification time: https://github.com/rustfs/rustfs/blob/1.0.0-beta.12/rustfs/src/storage/options.rs#L367-L383
- `get_complete_multipart_upload_opts` does not appear to parse it: https://github.com/rustfs/rustfs/blob/1.0.0-beta.12/rustfs/src/storage/options.rs#L319-L355
- Multipart completion falls back to `now_utc()` when `opts.mod_time` is absent: https://github.com/rustfs/rustfs/blob/1.0.0-beta.12/crates/ecstore/src/set_disk/ops/multipart.rs#L1229-L1235

Could multipart completion parse the authorized replication source modification time in the same way as regular PUT? An end-to-end regression test covering MinIO-to-RustFS multipart replication would also help prevent recurrence.

ccccpj:
## Related Issues

Fixes #5662.

## Summary of Changes

Multipart replication completion now parses the authorized source modification-time header into `ObjectOptions.mod_time`, matching the existing regular PUT replication path.

Root cause: `get_complete_multipart_upload_opts_with_replication_authorization` recorded the replication request and metadata but dropped `x-rustfs-source-mtime` / `x-minio-source-mtime`. The ecstore multipart commit path then fell back to the local completion time.

The timestamp parser is shared with the regular PUT path so both flows retain the same RFC 3339 and invalid-input behavior.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs --lib storage::options::tests::test_complete_multipart_opts_preserves_authorized_replication_mtime -- --exact`
- Mutation check: removing the multipart `mod_time` assignment makes the regression test fail; restoring it makes the test pass.
- `make pre-commit`
- `make log-analyzer-rules-check clippy-check test`
  - Static guards, full workspace quick-check, log-analyzer rules, all-target/all-feature Clippy, and script tests passed.
  - Full nextest executed all 12,173 tests: 12,171 passed and two unrelated system-DNS tests failed because the local resolver maps reserved `.invalid` names to `198.18.0.7/8` instead of returning NXDOMAIN.
  - The first DNS test also passed in an isolated run before the resolver mapping was observed.
- `cargo test --all --doc`

Adversarial validation:

- Correctness: PASS — authorized valid, unauthorized, and invalid timestamp paths are pinned; multipart fallback behavior remains unchanged when no usable timestamp exists.
- Simplicity: PASS — one request-local assignment and one parser shared with the existing PUT path; no duplicate constants or parallel implementation.
- Security: PASS — source timestamps remain ignored unless replication authorization and the replication-request marker are both present.
- Concurrency and durability: PASS — no locks, awaits, shared state, commit ordering, or persistence format changed.
- Compatibility: PASS — the existing dual RustFS/MinIO header lookup and RFC 3339 parsing semantics are preserved.
- Performance: PASS — one header lookup and parse only on authorized multipart replication completion; ordinary multipart requests do no extra parsing.
- Test effectiveness: PASS — the regression test detects removal of the fix and covers the authorization boundary plus malformed input.

## Impact

Replicated multipart objects preserve the source object's `LastModified` value instead of receiving the destination completion time. There are no API, configuration, wire-format, or on-disk format changes.

## Additional Notes

The change is intentionally limited to multipart option construction. Invalid or absent source timestamps continue to use the existing completion-time fallback.

### DIFF
--- a/rustfs/src/storage/options.rs
+++ b/rustfs/src/storage/options.rs
@@ -340,8 +340,10 @@ pub fn get_complete_multipart_upload_opts_with_replication_authorization(
     let mut user_defined = HashMap::new();
 
     let mut replication_request = false;
+    let mut mod_time = None;
     if replication_request_authorized && get_header(headers, SUFFIX_SOURCE_REPLICATION_REQUEST).as_deref() == Some("true") {
         replication_request = true;
+        mod_time = replication_source_mtime(headers);
         if let Some(actual_size_str) = get_header(headers, SUFFIX_REPLICATION_ACTUAL_OBJECT_SIZE) {
             rustfs_utils::http::insert_str(
                 &mut user_defined,
@@ -361,6 +363,7 @@ pub fn get_complete_multipart_upload_opts_with_replication_authorization(
         want_checksum: rustfs_rio::get_content_checksum(headers)?,
         user_defined,
         replication_request,
+        mod_time,
         ..Default::default()
     };
     apply_replica_status_from_headers(headers, &mut opts, replication_request_authorized);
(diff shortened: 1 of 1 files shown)


overtrue (approved):
Looks good. The authorized multipart path now preserves the source mtime through the existing dual-header parser, while unauthorized and malformed inputs retain the current fallback behavior.

PR #5677, opened 2026-08-03
fix(ecstore): anchor Windows rename publication

marshawcoco:
## Related Issues

Follow-up to #5663 and relates to #5419.

## Summary of Changes

- Publish Windows rename targets relative to the validated destination-parent handle instead of resolving the destination pathname again at commit time.
- Validate each guarded Windows directory against a writer-exclusive handle pass before publication while retaining write sharing needed by normal child creation.
- Preserve Windows replacement behavior for read-only and delete-shared open destinations through the `FileRenameInfoEx` fallback.
- Add native Windows regression coverage for direct-child publication, destination replacement, and an in-place junction mutation racing publication.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore --lib rename_all_ -- --nocapture`

## Impact

Windows local-disk rename publication is anchored to the validated destination directory identity. There are no S3 API, configuration, on-disk format, or wire-format changes.

## Additional Notes

Draft pending native Windows CI and final high-risk adversarial validation.


marshawcoco:
@codex review

crates/scanner/src/scanner.rs
```diff
@@ -354,6 +365,36 @@ struct ScannerCleanIdleBackoff {
...
+        }
+    }
+
+    fn retry_interval(self, configured_interval: Duration) -> Option<Duration> {
+        let exponent = self.consecutive_cycles.checked_sub(1)?.min(31);
+        let multiplier = 1u32.checked_shl(exponent).unwrap_or(u32::MAX);
+        let base_interval = configured_interval
+            .max(Duration::from_secs(1))
+            .min(SUPERSEDED_RETRY_BASE_INTERVAL);
+        Some(base_interval.saturating_mul(multiplier).min(SUPERSEDED_RETRY_MAX_INTERVAL))
```
overtrue: This cap makes the superseded retry cadence independent of the operator's configured cycle interval. With `cycle_interval = 24h` under sustained write load, every cycle classifies as Superseded (any object write during the scan regrows the dirty set), so the scanner converges to a full namespace traversal every 30 minutes — ~48 full scans/day where the operator configured 1. For the supersession case with no dirty wake pending, this is also a frequency increase over current behavior, which waits out the full configured interval. Operators pick long intervals precisely to bound scan I/O on large clusters. Please either make the retry interval respect the configured interval as an upper bound (e.g. `.min(SUPERSEDED_RETRY_MAX_INTERVAL).min(configured_interval.max(base_interval))`), make the cap configurable, or document in the PR why a fixed 30-minute ceiling should override operator configuration.

crates/ecstore/src/data_usage/mod.rs
```diff
@@ -551,6 +588,23 @@ where
...
+        bucket,
+        DATA_USAGE_REMOVE_CAS_RETRIES,
+        None,
+        guard,
+    )
+    .await
+    {
+        // The authoritative timestamp was already advanced above, so admin
+        // selection rejects this observation even if optional cleanup fails.
+        // Never make an admin-only freshness artifact block DeleteBucket.
```
overtrue, line 664: Bucket delete is currently the only path that ever removes or rewrites `.usage.observed.json`. After the next authoritative scan completes, the observation is permanently fenced by the baseline identity check but the object itself lingers indefinitely, and both scanner startup (`persisted_usage_cache_is_cold_for_startup`) and every admin cache refresh keep paying a read for a file that can never be selected again. Please add a best-effort delete (same never-block, log-on-failure posture as here) of the observed object after a successful authoritative save, so the steady state of a converged cluster has no stale observation on disk.

crates/obs/src/metrics/stats_collector.rs
```diff
@@ -102,6 +103,7 @@ async fn load_obs_data_usage_from_backend(store: Arc<ObsStore>) -> ObsEcstoreRes
     Ok(ObsDataUsageInfo {
         last_update: data_usage.last_update,
         usage_snapshot_complete,
+        usage_snapshot_converged: data_usage.usage_snapshot_converged != Some(false),
```
overtrue: `!= Some(false)` maps `None` (unknown) to converged=1. The admin selector upgrades complete authoritative snapshots from `None` to `Some(true)` before this point, so the remaining `None` cases reaching this line are precisely the degraded/non-authoritative fallbacks — an incomplete snapshot whose bucket data was discarded will report `snapshot_converged 1`. That inverts the metric's stated meaning for exactly the situations operators most need to notice. Please map `None` to non-converged (or gate on `usage_snapshot_complete`), or document why unknown should read as converged.

overtrue (changes requested):
Thanks for the very thorough PR. I did an adversarial review across six axes and found no blocking correctness defects in the code itself:

- **Quota authority**: verified that quota admission (`get_real_time_usage`, memory overlay, `lookup_degraded_bucket_usage_baseline`) never consumes `.usage.observed.json`; observational saves skip `replace_bucket_usage_memory_from_info`, and `store_data_usage_in_backend` hard-rejects `converged == Some(false)`. The "observations cannot inflate the degraded fallback" claim holds.
- **Baseline fencing**: the exact `(last_update, scanner_cycle, scanner_epoch)` identity match is clock-independent, and the load-bearing premise that every bucket delete rewrites the authoritative identity holds (`advance_data_usage_last_update` is monotonic), including the old-binary rollback delete/recreate case.
- **Persistence/wire compat**: both new fields are `Option` + `skip_serializing_if`, `DataUsageInfo` is JSON-persisted without `deny_unknown_fields`, so a rolled-back binary reads the new authoritative object safely; the msgpack array-encoding invariant only constrains `DataUsageCache`/`DataUsageEntry`, untouched here. madmin `AccountInfo` field is additive.
- **Normalization**: `is_complete_bucket_usage_snapshot` is internal-consistency only, so a selected observation cannot be gutted by `discard_incomplete_bucket_usage`; the memory overlay is an absolute replace gated on `usage_updated_at`, so no double counting on top of an observation.
- **Concurrency**: the split generation counters are wired correctly, the lock order is globally quota-cache -> admin-cache with no reverse acquisition, and dirty clear/ACK is correctly gated on `Complete` + durable persist (locked by `finalizing_a_durable_superseded_snapshot_keeps_dirty_work_pending`).
- **Performance**: no hot-path cost; new I/O is one small JSON write per superseded cycle plus one extra metadata read per 30s admin refresh.

I'm requesting changes for three points that need an author response or adjustment before this merges:

1. **Superseded retry cadence can far exceed the operator-configured cycle interval** (inline at `crates/scanner/src/scanner.rs`). With a 24h configured interval and sustained write load (any object write during a scan regrows the dirty set, so every cycle classifies as Superseded), full namespace traversals now run roughly every 30 minutes (~48/day) instead of once a day. For the supersession-without-dirty-wake case this is a frequency *increase* over today's behavior (which waited out the full configured interval). Please either justify the fixed 30-minute cap or make the policy respect the configured interval (e.g. cap the retry interval at the configured cycle interval, or make the cap configurable).
2. **Stale `.usage.observed.json` is never cleaned up after convergence** (inline at `crates/ecstore/src/data_usage/mod.rs`). Once a newer authoritative scan lands, the observation is permanently fenced but the object lingers forever, costing one startup read and one admin-refresh read indefinitely. Please add a best-effort delete of the observed object after a successful authoritative save.
3. **`rustfs_cluster_usage_objects_snapshot_converged` reports 1 for unknown/degraded snapshots** (inline at `crates/obs/src/metrics/stats_collector.rs`). Mapping `None` to converged=1 means a degraded, incomplete fallback snapshot reports as converged. Please fix the mapping or explain why unknown should read as converged.

Non-blocking notes, no action strictly required:

- `invalidate_data_usage_snapshot_cache` in `crates/ecstore/src/data_usage/mod.rs` holds both cache write locks simultaneously, while `prepare_bucket_usage_for_namespace_change` and the guarded invalidation drop the first before taking the second. There is no reverse-order path today so no deadlock, but please unify on drop-then-acquire to keep the lock order trivially auditable.
- The nextest `embedded-test-ports` group and the `#[cfg(test)]` GET-timeout policy seam in `object_usecase.rs` are unrelated CI-flake fixes bundled into a scanner PR. Both are small, well-documented, and verifiable (the binary filter matches the five `find_available_port` users under `rustfs/tests/`), so I won't insist on splitting them, but please keep such fixes in separate PRs going forward.
- Baseline identity via the `(last_update, cycle, epoch)` triple works, but the CAS machinery already has revision etags that would make an exact-baseline check a single opaque string comparison; worth considering as a follow-up simplification.

CI is fully green on the current head (`a13f0d882`).

marshawcoco:
Hi @overtrue, thank you for the detailed review. It looks like this review may have been attached to PR #5677 by mistake. The comments reference commit `a13f0d882`, which is currently the head of PR #5742, and target files that are not part of the #5677 diff. The current head of #5677 is `5cf3bebe9` and contains only the Windows rename-related changes. Could you please check whether this review was intended for #5742? Thanks.

houseme (changes requested):
This remains a high-risk Windows rename/publication change. The PR is not currently merge-ready and its own description says native Windows CI and final high-risk validation are pending. Please rebase, run the native Windows workflow, and complete the required high-risk gate before re-review.

houseme (approved):

overtrue (approved):
The handle-relative publication path, identity guards, retry and cancellation lifetime, replacement compatibility, and regression coverage look sound. I found no blocking issue on this head.

PR #5679, opened 2026-08-03
perf(ecstore): batch local EC shard preads on GET

rdiperri-wasabi:
Performance testing of warm GET on single-node multi-disk setups showed that local EC shard reads were spending a lot of time on per-shard blocking-pool round-trips. That showed up clearly around ~1.25 MB objects under concurrent load. This change batches those positioned reads into one `spawn_blocking` when all online shards are local.

## Related Issues

N/A (performance improvement for single-node multi-disk warm GET).

## Summary of Changes

- Add `batch_shard_pread` to batch local EC shard positioned reads in one `spawn_blocking`.
- Add `get_object_path_if_local` / `adjust_shard_read_params` helpers.
- Fast-path in `create_bitrot_readers_until_quorum_with_preference` when mmap-read is on and all online shards are local; bail to the existing path for inline data, any remote disk, empty batch, or length above `object_mmap_read_max_length()`.
- Unit tests: `test_batch_shard_pread_basic`, `test_batch_shard_pread_partial_errors`.

## Verification

- `make fmt`
- `make pre-commit`
- `make pre-pr` (clippy + script-tests passed; two unrelated uring fd-cache nextest failures on low `RLIMIT_NOFILE`)
- `cargo test -p rustfs-ecstore --lib test_batch_shard_pread` (2/2)
- Lab A/B: same `main` tip (`04c592185`) clean vs patched binaries

## Impact

Single-node 8-disk EC:1, warm GET @ 50 threads / 5 min (main tip A/B):

- **≈1.25 MB (`1000K-1500K`):** ~1,843 → ~4,736 MB/s (**+157%**); avg latency ~32 → ~12 ms
- **5 MB:** ~5,402 → ~5,222 MB/s (−3.3%); latency ~45 → ~47 ms — flat within noise; this does **not** raise the mid-band GET ceiling

## Additional Notes

Bailouts preserve correctness for remote disks, inline objects, and the mmap-copy size cap.

Made with [Cursor](https://cursor.com)


rdiperri-wasabi:

overtrue:
Hey! Just finished reviewing this PR — nice work on the batch pread optimization. The +157% throughput bump on ~1.25MB objects is solid. Here are my thoughts:

## What's looking good

- **Bailout logic is spot on.** Inline data, remote disks, empty batches, size caps — all the edge cases gracefully fall back to the existing path. No surprises there.
- **Path safety is clean.** The `get_object_path` → `check_local_disk_valid_path` chain keeps things secure, no path traversal worries.
- **Good test coverage.** The happy path and partial error cases are both covered.

## A couple things to flag

**Medium — sequential I/O could get dragged by a slow disk**

Right now `batch_shard_pread` reads all shards sequentially in a single `spawn_blocking`. In an 8-disk setup, if one disk has a latency spike (bad sectors, SSD GC, etc.), the whole batch gets held up. The old per-shard approach had natural I/O concurrency.

Might be worth considering `std::thread::scope` or parallel spawns inside the batch function — or at least documenting this as a known tradeoff for now.

**Medium — `as u64` cast**

```rust
file.read_at(&mut buf[total..], (offset + total) as u64)
```

On 64-bit this is fine, but per repo conventions all `as` casts need justification. Could swap to `u64::try_from()` for defensive coding, or just add a comment explaining why it's safe here.

**Minor — test should check error type**

`test_batch_shard_pread_partial_errors` only checks `results[1].is_err()`. Would be better to assert the specific `DiskError` variant:

```rust
assert!(matches!(results[1].as_ref().unwrap_err(), DiskError::Io(_)));
```

**Nit — `use std::io::Cursor` import location**

The `Cursor` import in `io_primitives.rs` is only used in the `#[cfg(unix)]` path. Could move it into the function's local `use` block to stay consistent.

## TL;DR

The core optimization is solid and the bailout design is conservative in a good way. The sequential I/O tradeoff (M2) is the main thing to think about for a follow-up. Everything else is minor.

**Verdict: Approve with minor comments** 👍

PR #5696, opened 2026-08-04
fix(replication): restrict metadata replication targets

overtrue:
### Motivation

- Prevent metadata-only operations (`PutObjectTagging`/`DeleteObjectTagging` etc.) from causing a full-object copy to a target that never previously held the object, which could allow exfiltration when tagging permissions exist but read permissions do not.

### Description

- Add a per-target eligibility check on metadata replication by introducing `MustReplicateOptions::metadata_target_is_eligible` which reads the persisted per-target replication state from internal metadata and returns true only when the target’s state is `COMPLETED` for `ReplicationType::Metadata` operations. (Added constant `SUFFIX_REPLICATION_STATUS` and internal metadata use.)
- Apply the eligibility check when building each target decision in `must_replicate`, changing `replicate = cfg.replicate(&sopts)` to `replicate = cfg.replicate(&sopts) && mopts.metadata_target_is_eligible(&arn)` so metadata updates do not admit targets without an existing completed replica.
- Preserve existing object-data replication paths and semantics (object replication is unchanged) and retain RustFS/MinIO internal metadata compatibility when reading replication state.
- Add a unit test `metadata_replication_requires_completed_target_state` in `crates/replication` exercising `PENDING`, `FAILED`, `COMPLETED`, missing-target, and ordinary-object cases to lock down the new behavior.

Files changed: `crates/replication/src/operation.rs`, `crates/replication/src/http.rs`, `crates/ecstore/src/bucket/replication/replication_object_config.rs`.

### Testing

- Ran the focused unit test `cargo test -p rustfs-replication metadata_replication_requires_completed_target_state` and it passed.
- Invoked the repository pre-commit gate (`make pre-commit`) and the formatting/guard checks covering the changed crates completed successfully; workspace-wide quick-check compilation progressed to the changed crates (limited by the execution environment time window).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a714a443e648329b28187cc9623ee9c)


overtrue (commented):
Adversarial review at head `a53544344904946a648e58a5297fc0dd139dc353`.

Major — transient replication state is being used as proof that the target has ever held the object (`crates/replication/src/operation.rs:89`). After metadata update A is sent successfully, the worker releases its object reader before persisting the final COMPLETED status. A concurrent metadata update B can commit in that window, observe PENDING, and be excluded by `metadata_target_is_eligible`; then worker A writes COMPLETED for its older snapshot. B has no decision, pending marker, queue task, or MRF entry, so the target can remain permanently stale even though it is eligible and already contains the object.

Please separate durable target admission/history from the transient PENDING/FAILED/COMPLETED state, or otherwise serialize the eligibility decision through status persistence. Add a deterministic regression that pauses the first metadata replication after the remote write but before local status write-back, performs a second metadata update, and proves the second update reaches the existing target. The current unit test only checks static status strings and cannot detect this race.

CI note: Quick Checks passed, but the main test/build lanes were cancelled, so there is no full high-risk validation for this head.

PR #5697, opened 2026-08-04
fix(policy): require unscoped KMS bundle grants

overtrue:
### Motivation

- KMS backup/restore admin endpoints were authorized with an empty key scope, letting resource-scoped KMS policies unintentionally match every key and thereby escalate a single-key operator to cluster-wide bundle actions.
- The intent is to require an explicit action-only (unscoped) Allow for global bundle operations while preserving Deny semantics that should still block those operations.

### Description

- Add a guard in `Statement::kms_key_scope_matches` so `kms:Backup` and `kms:Restore` require an action-only (no Resource/NotResource) Allow to succeed, and a resource-scoped Deny still blocks the all-key operation. (change in `crates/policy/src/policy/statement.rs`).
- Add a regression test `test_kms_bundle_actions_require_an_unscoped_statement` covering scoped Allow, NotResource, scoped Deny, and unscoped Allow cases. (change in `crates/policy/src/policy/policy.rs`).
- Keep existing behavior for other KMS actions and preserve warnings for mixed/non-KMS resources.

### Testing

- Ran the focused test: `cargo test -p rustfs-policy test_kms_bundle_actions_require_an_unscoped_statement` and the new regression passed.
- Ran `cargo test -p rustfs-policy` to exercise the package; the relevant policy tests (including the new test) passed in the package run, although an earlier interrupted run encountered an unrelated OPA test that stalled and was aborted during ad-hoc validation.
- `make pre-pr` was attempted but blocked by baseline logging guardrail checks referencing a missing handler file, so the full pre-pr gate was not completed in this environment.
- Note: this checkout had no configured GitHub auth/remote in the environment, so PR creation was not performed from this session.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a71628f221c8329b01bbdbac82171ce)


overtrue (commented):
Adversarial review at head `d88f0748420eebe458d5983d3e9ffe5460cd589b`.

High — `NotAction` statements bypass the new global-bundle restriction. At `crates/policy/src/policy/statement.rs:254`, KMS resource semantics are selected with `self.is_kms()`. A statement such as `Effect: Allow`, `NotAction: s3:*`, `Resource: arn:aws:s3:::*` has an empty Action set, so it passes the requested `kms:Backup`/`kms:Restore` action filter but is evaluated by the generic resource path instead of `kms_key_scope_matches`. With the handlers' empty bucket/object scope, the wildcard matches and authorizes the global bundle operation. Route based on the requested action (`args.action`) after Action/NotAction filtering, and add Allow/Deny `NotAction` regressions.

Medium — the early branch checks raw Resource/NotResource non-emptiness before filtering KMS resources. Existing policies carrying legacy non-KMS resources were explicitly treated as unscoped with a warning; this head silently revokes their Backup/Restore authorization and makes rolling-upgrade nodes disagree. Apply the restriction to filtered KMS resources and add compatibility tests.

The policy unit tests do not cover either case. CI compiled the changed policy crate but the workspace later failed on an unrelated `HealWalkCollector.decode_error` error.

PR #5698, opened 2026-08-04
fix(s3select): enforce SSE-KMS read authorization

overtrue:
### Motivation

- S3 Select (`SelectObjectContent`) could read SSE-KMS encrypted objects without exercising per-key `kms:Decrypt` authorization, enabling an S3-authorized caller to obtain plaintext via the Select API even when missing KMS permissions. 
- The change enforces the intended per-key KMS authorization on the Select code path so the Select API is subject to the same SSE-KMS key policy checks as ordinary object reads.

### Description

- Preserve the request identity and build a request principal with `SseKmsPrincipal::from_request` in `execute_select_object_content` and thread it into the preflight checks. 
- Add a `read_principal` parameter to the select preflight and call `authorize_sse_kms_object_read(read_principal, &info.user_defined).await?` to require `kms:Decrypt` when the object metadata indicates SSE-KMS. 
- Change `SseKmsPrincipal::from_request` to represent anonymous S3-boundary requests as an empty-account principal (distinct from internal `None`) so anonymous callers are not treated as trusted internal callers by the KMS gate. 
- Add a unit test that validates anonymous S3 requests build a KMS principal and added the small test coverage around the new behavior.

### Testing

- Ran `cargo fmt --all --check` and `git diff --check`, both of which passed. 
- Ran staged-diff checks (`git diff --cached --check`) and a local commit verification, which succeeded locally. 
- Attempted targeted builds/tests with `cargo test -p rustfs --lib sse_kms_object_read_authorization --no-default-features` and `cargo check -p rustfs --lib --no-default-features`, but full dependency compilation and the test run did not complete within the execution window so the runtime unit-test gate could not be fully exercised. 
- The change adds a focused unit test for anonymous principal construction that is included in the diff and intended to be run by CI as part of the `sse_kms_object_read_authorization` test set.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a714070599c8329b633d4584fd276c1)


overtrue (commented):
Adversarial review at head `0583e40eec02cc48dc24ab0b5a58f384ed810156`.

High — S3 Select authorizes one revision's KMS key and may decrypt a later revision under the trusted internal principal. `rustfs/src/app/select_object.rs:406` authorizes metadata from preflight, but carries only size into query execution. The Select object store later performs an unpinned latest-object lookup, and the encryption resolver runs with `principal: None`. A caller allowed to decrypt key A can therefore pass preflight on revision A, race a replacement encrypted under denied key B, and receive B's plaintext. Bind authorization and reading to the same immutable object snapshot, or propagate the request principal to the exact reader/decryption metadata. Add a paused preflight-to-reader replacement test for ScanRange and non-ScanRange.

High — the new direct import from `crate::storage::sse` in `rustfs/src/app/select_object.rs:7` violates the app storage boundary. The live Quick Checks job fails `scripts/check_architecture_migration_rules.sh`. Re-export the existing SSE facade through `app/storage_api.rs` and import through that boundary.

No existing substantive review duplicates these findings.

PR #5699, opened 2026-08-04
fix(ecstore): retain namespace read locks for streaming GETs to prevent quota-bypass

overtrue:
### Motivation
- Prevent a quota-bypass availability bug where streaming GETs released namespace read locks while holding renewable snapshot leases, allowing repeated overwrites to leave leased old data directories on disk and exhaust physical storage.
- Keep the optimization that releases locks for fully-materialized reads while ensuring streaming reads continue to block mutating operations until the response completes or is dropped.

### Description
- Stop acquiring snapshot leases on the streaming GET hot path and keep the namespace read lock for streaming readers by wrapping the stream with `SetDiskLockGuardedReader` until EOF or drop. (`crates/ecstore/src/set_disk/mod.rs`, `crates/ecstore/src/set_disk/ops/object.rs`).
- Remove the streaming-path snapshot-lease replacement logic and the producer-side lease handoff so concurrent overwrites/deletes cannot be performed while a streaming reader is still holding the namespace lock. (`crates/ecstore/src/set_disk/ops/object.rs`).
- Disable snapshot-lease acquisition and renewal at the NodeService RPC layer by returning a protocol-compatible `UnsupportedDisk` response for `acquire`/`renew`, while preserving `release` for cleanup; add a helper `snapshot_lease_disabled_response`. (`rustfs/src/storage/rpc/node_service/disk.rs`, `rustfs/src/storage/rpc/node_service.rs`).
- Add focused regression tests that assert a stalled response blocks overwrites and deletes until EOF and that EOF or dropping the reader releases the lock, plus unit tests asserting the disabled-lease RPCs fail closed. (`crates/ecstore/src/set_disk/mod.rs`, `rustfs/src/storage/rpc/node_service/disk.rs`, `rustfs/src/storage/rpc/node_service.rs`).

### Testing
- Ran `cargo fmt --all --check` and `git diff --check`, both passed for the edited files.
- Ran targeted ecstore tests: `cargo test -p rustfs-ecstore streaming_reader_holds_read_lock_until_eof -- --nocapture` (passed) and `cargo test -p rustfs-ecstore streaming_get_blocks_concurrent -- --nocapture` (all 3 added/modified tests passed).
- Added RPC-level and unit tests for the disabled lease responses and exercised them through `cargo test` where the ecstore unit tests executed successfully; the full `rustfs` workspace `snapshot_lease_acquire_and_renew_fail_closed` target was attempted but the full workspace compilation was long-running and was partially interrupted in this environment (no remote push/CI run was performed here).
- Ran `make pre-commit` checks up to the fast gates; workspace-wide pre-commit started but heavy compilation steps were interrupted due to environment limits; no repository guard was weakened to obtain green checks.

Commit: the change was recorded in the local branch with message `fix(ecstore): retain locks for streaming GETs` and includes the modified files listed above.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7153ff6c108329976b4c6b523763fd)


overtrue (commented):
Adversarial review at head `24fa71233e352c5bf978e675796454c40dd7d663`.

Analytical verdict: COMMENT / no code finding established, but CI is not merge-ready.

I attacked EOF/drop lock release, overwrite/delete/delete-objects blocking, reader cancellation, mixed-version snapshot-lease RPC behavior, lock ordering, and the buffered-body fast path. The new `SetDiskLockGuardedReader` retains the namespace read lock for streaming bodies and releases it on EOF or drop; focused tests cover the main mutation interleavings. Disabling acquire/renew while preserving release fails closed for old lease clients.

Residual risk: slow or stalled clients intentionally retain the object read lock for the response lifetime, so this trades snapshot-retention storage growth for write availability. That is the stated design, but it should be monitored with the existing slow-lock diagnostics.

CI currently has a failing `io_uring Integration (real)` check and many cancelled lanes. The observed compile failure concerns a missing `HealWalkCollector.decode_error` initializer and appears outside this diff, but the required high-risk gate is still incomplete.

PR #5700, opened 2026-08-04
fix(iam): disambiguate OIDC virtual parent IDs

overtrue:
### Motivation
- The previous OIDC virtual parent derivation concatenated `openid: + subject + ':' + issuer` without length delimiters or escaping, which allows distinct `(sub, iss)` pairs that contain colons to collide and enables cross-identity STS policy mapping.
- The goal is to prevent such identity collisions while preserving the existing parent-id hashing and encoding surface (`SHA256` then URL-safe base64).

### Description
- Replace the ambiguous colon-delimited encoding in `crates/iam/src/federation/model.rs` with a fixed-width length-prefixed encoding that frames `subject` and `issuer` as 8-byte big-endian lengths followed by their bytes before hashing. 
- Preserve the `openid:` prefix, `HashAlgorithm::SHA256` hashing, and `base64_simd::URL_SAFE_NO_PAD` output so the external format remains a single stable string derived from the unambiguous source bytes. 
- Add a regression unit test that demonstrates the previously-colliding `(subject, issuer)` pair no longer produces the same parent, and add/update a golden-vector assertion to lock the canonical encoding/endianness used by the hash. 
- Keep the behavior that empty or missing `iss`/`sub` returns `None` (fail-closed) and avoid adding any legacy fallback lookup for old ambiguous IDs.

### Testing
- Ran the targeted unit tests with `cargo test -p rustfs-iam federation::model::tests::oidc_virtual_parent --lib`, which executed the new regression and related tests and reported `4 passed; 0 failed`. 
- Verified `git diff --check` (no whitespace errors) and the file-level diff shows only the intended changes to `crates/iam/src/federation/model.rs`. 
- Tried the repository pre-PR gate `make pre-commit`, which is blocked by an unrelated, pre-existing stale doc path referencing `rustfs/src/admin/handlers/table_catalog.rs` and therefore failed outside this change; this is not caused by the IAM fix.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a716ad914b483298ce1f94925a92717)


overtrue (commented):
Adversarial review at head `e170332f9c33cbce50c92cf587d7cb430d848d62`.

High compatibility regression — the change replaces the deterministic OIDC virtual-parent ID for every existing `(subject, issuer)` pair, but provides no migration or dual-lookup path. Policies, STS identities, and service accounts persist and resolve by `parent_user`; after upgrade, the next login derives the new length-prefixed hash and no longer finds mappings stored under the old MinIO-compatible ID. Existing federated users can lose authorization, while previously issued credentials continue to reference the old parent until expiry.

Please version the derivation and migrate policy/identity keys atomically, or retain a collision-safe compatibility lookup that can distinguish the ambiguous legacy pair without orphaning non-colliding identities. Add an upgrade test that seeds a policy and service account under the old ID, performs login with the same verified subject/issuer, and proves authorization plus enumeration remain intact. The current test only proves two crafted inputs no longer collide.

CI is also red (`Test and Lint (rio-v2)` and `io_uring Integration (real)`), with downstream lanes cancelled.

PR #5701, opened 2026-08-04
fix(auth): enforce object-lock actions for POST uploads

overtrue:
### Motivation

- Fix an authorization bypass where an early return for HTTP `POST` skipped `s3:PutObjectRetention` and `s3:PutObjectLegalHold` checks, allowing callers with only `s3:PutObject` (including anonymous public-write) to set object-lock retention or legal-hold via POST form uploads.
- Restore the intended IAM boundary so object-lock fields on uploads always require the corresponding object-lock IAM actions.

### Description

- Remove the unconditional `POST` early-return in `S3Access::put_object` so the object-lock-specific authorizations (`PutObjectLegalHold` / `PutObjectRetention`) are evaluated for POST form uploads when the corresponding fields are present (`rustfs/src/storage/access.rs`).
- Ensure replication-only header authorization is applied in the non-POST path to preserve the expected header behavior while not bypassing object-lock checks for POST uploads (`rustfs/src/storage/access.rs`).
- Convert the two in-tree e2e scenarios that previously accepted anonymous POST uploads with object-lock fields into negative regression tests that assert a `403 AccessDenied` response when only `s3:PutObject` is granted (`crates/e2e_test/src/multipart_auth_test.rs`).
- Apply formatting and commit the change with the message `fix(auth): enforce object-lock actions for POST uploads`.

### Testing

- Ran `cargo fmt --all --check` which passed.
- Ran `git diff --check` and `git diff --cached --check` which passed with no whitespace or indexing issues.
- Attempted a targeted test build with `cargo test -p rustfs storage::access::tests:: --lib`; compilation started successfully but the full test run could not finish within the execution window (build progressed and crates compiled before interruption).
- Attempted to open a GitHub PR with `gh pr create --title 'fix(auth): enforce object-lock actions for POST uploads' --body-file /tmp/rustfs_pr_body.md` but GitHub CLI authentication was not available in the environment, so the PR was not published from this session.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a71898a97c0832995bb34c4b162156c)


overtrue (commented):
Adversarial review at head `6c8ca3d636562f289c9588a1229ff75a00f4bf4c`.

High — `rustfs/src/storage/access.rs:2323` skips `authorize_replication_only_put_headers` for every POST. That helper is already presence-gated. A caller with PutObject and the object-lock action, but without ReplicateObject, can submit a POST carrying `x-amz-replication-status` or source-replication internal headers and bypass the established replication authorization used by PUT/Copy/multipart paths. Call the helper unconditionally; ordinary POSTs pay no authorization cost because it returns immediately when those headers are absent.

Major coverage gap — both changed tests use an anonymous principal lacking both adjacent object-lock actions and assert only generic AccessDenied. Swapping PutObjectRetention and PutObjectLegalHold in production would keep both tests green. Add an authenticated one-permission matrix, a both-fields case requiring both actions, signed POST coverage, and a post-denial HeadObject assertion. The changed E2E tests did not execute in this PR's CI run.

Minor — bucket-generation lookup is unwrapped before the newly enabled lock-action checks, exposing NoSuchBucket/storage errors before AccessDenied. Keep the deferred result until all required actions are authorized, consistent with the other write paths.

overtrue (commented):
Supplemental high-severity finding from the security pass:

POST object-lock authorization conditions are evaluated from raw HTTP headers, while the values persisted by storage come from multipart form fields. A caller can put an allowed retention date in `x-amz-object-lock-retain-until-date` HTTP header, but a disallowed 2099 date in the signed/validated POST form. The IAM condition check reads the benign header (`rustfs/src/auth.rs:712`), while storage applies the form-derived `PutObjectInput` value (`rustfs/src/app/object_usecase.rs:5062`). This can bypass conditional PutObjectRetention policy and create an effectively undeletable version.

Use the multipart-derived values as the canonical authorization condition source for POST and reject conflicting header/form values. Add signed and unsigned regressions with an allowed header plus denied form date.

PR #5705, opened 2026-08-04
fix(ci): block PR-controlled execution in performance A/B workflow

overtrue:
### Motivation

- Close a supply-chain/CI vulnerability where the Performance A/B workflow could run untrusted PR-controlled code on the `sm-standard-2` runner after a maintainer-applied label, and remove a path that granted PR write-capable tokens to that job.
- Ensure the performance gate still runs for trusted contexts (nightly and explicit dispatch) while preventing any PR checkout from being executed on sensitive runners.

### Description

- Remove pull-request triggers and PR-specific conditions from `.github/workflows/performance-ab.yml` so the job only runs on `schedule` and `workflow_dispatch`, and eliminate the workflow's `pull-requests: write` permission and PR-commenting step. 
- Stop invoking PR-controlled workspace scripts from the workflow by removing the PR execution/notification steps and by tightening the `if:` condition to the trusted events only. 
- Add `scripts/security/check_performance_ab_workflow.sh`, a lightweight guard that rejects `pull_request`/`pull_request_target` events and PR-write or broad-write permissions patterns in the performance workflow. 
- Wire the new guard into the security audit (`.github/workflows/audit.yml`) so the repository's audit job runs `./scripts/security/check_performance_ab_workflow.sh` on relevant events and paths. 

### Testing

- Ran the new guard directly with `./scripts/security/check_performance_ab_workflow.sh` and confirmed it rejects injected poison patterns such as `pull_request`, `pull_request_target`, `pull-requests: write`, and `permissions: write-all`. (passed)
- Performed mutation checks by appending representative bad lines to `.github/workflows/performance-ab.yml` and verified the checker fails for each poison case. (passed)
- Validated YAML parsing with `ruby -e 'require "yaml"; YAML.load_file(ARGV.fetch(0), aliases: true)'` for `audit.yml` and `performance-ab.yml`, ran `bash -n` on the checker script, and ran `git diff --check` and `cargo fmt --all --check` to satisfy repository checks. (all passed)
- `actionlint` and `shellcheck` could not be executed in this environment due to network/tooling constraints and are listed as skipped in the verification notes. (skipped)

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a71bb4d88bc8329bda061667343941a)


overtrue (commented):
Adversarial review at head `13067db0d2977a5504a5988e20ba8554bd11571b`.

High — `scripts/security/check_performance_ab_workflow.sh` denies selected raw spellings instead of semantically allowlisting trusted events and permissions. Valid YAML quoting/escaping and alternative events can restore PR-controlled execution while the regex guard passes (for example an `issue_comment` trigger, quoted `pull-requests: "write"`, and checkout of `refs/pull/.../head`). Parse YAML and assert the normalized trigger set and permission tree, with mutation fixtures covering quoted keys/values, flow mappings, job-level permissions, and alternate events.

High — workflow_dispatch can execute the selected repository branch/tag on `sm-standard-2` without a trusted-ref restriction. If dispatch is part of the trusted boundary, enforce `main` (or an explicit immutable ref) before checkout/local actions; otherwise document and separately authorize branch-level code execution on the sensitive runner.

Operationally, the workflow is currently `disabled_manually`, so retained schedule/dispatch triggers do not run until an out-of-band enablement step occurs. Also, the live CI has an unrelated io_uring failure and incomplete downstream lanes.

PR #5706, opened 2026-08-04
fix(api): bound bucket rate limiter keys

overtrue:
### Motivation

- Prevent an unauthenticated remote client from causing unbounded memory retention by supplying oversized path-style or virtual-host prefixes that were stored verbatim as bucket keys by the new per-bucket rate limiter.

### Description

- Introduce `MAX_BUCKET_NAME_LENGTH = 63` and a helper `bounded_bucket_name()` to accept only non-empty bucket strings up to 63 bytes before they are used as limiter keys. 
- Use the bound for both virtual-host-prefix extraction and first-path-segment extraction in `request_bucket()` so oversized values are skipped and never retained by the limiter. 
- Add regression tests that accept a 63-byte bucket and reject a 64-byte bucket for both path-style and virtual-hosted-style extraction. 
- Commit created with message `fix(api): bound bucket rate limiter keys` to apply the minimal fix.

### Testing

- Ran formatting and static checks: `cargo fmt --all --check` and `git diff --check` (passed). 
- Ran repository quality scans for common Rust anti-patterns (`rg` checks) on the changed file (no findings). 
- Added targeted unit tests in `rate_limit.rs`; attempted `cargo test -p rustfs --lib server::rate_limit::tests`, and the build progressed (dependencies and crate compiled) but the full test run could not complete within the environment execution-time limit, so tests were not fully executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a71ddee8a508329acfef6547da8f29a)

### DIFF
--- a/rustfs/src/server/rate_limit.rs
+++ b/rustfs/src/server/rate_limit.rs
@@ -84,6 +84,7 @@ pub(crate) const RATE_LIMIT_SCOPE_CONSOLE: &str = "console";
 pub(crate) const LABEL_RATE_LIMIT_DIMENSION: &str = "dimension";
 pub(crate) const RATE_LIMIT_DIMENSION_CLIENT_IP: &str = "client_ip";
 const RATE_LIMIT_DIMENSION_BUCKET: &str = "bucket";
+const MAX_BUCKET_NAME_LENGTH: usize = 63;
 
 // Header names shared with the Swift error mapping (crates/protocols swift/errors.rs).
 const X_RATE_LIMIT_LIMIT: &str = "x-ratelimit-limit";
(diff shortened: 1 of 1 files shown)


overtrue (commented):
Adversarial review at head `db3473cfb0a038719f300bebb45d2fb984667cf2`.

Analytical verdict: COMMENT / no code finding established, but CI is not merge-ready.

I attacked empty, 63-byte, 64-byte, path-style, virtual-hosted-style, dotted, and invalid oversized bucket candidates. Bounding the borrowed key before inserting it into the per-bucket limiter closes the unbounded-key retention path without adding allocations. The same helper is applied to both extraction paths, and the boundary tests cover max and max+1.

Residual coverage: the targeted unit tests were not completed locally, and this PR's `Test and Lint (rio-v2)` check failed while the remaining heavy lanes were cancelled. Quick Checks passed. The failing lane must be investigated or shown to be an unrelated base-branch failure before merge.

PR #5715, opened 2026-08-04
fix(replication): make bucket replication rules editable from clients

reatang:
## Related Issues

N/A

## Summary of Changes

Two server-side fixes that make bucket replication rules manageable from the console (and from `mc`), plus the dev harness used to verify them:

1. **Accept explicit `Destination.StorageClass=STANDARD` in replication configs** (`crates/replication/src/config.rs`). The replication engine never reads this rule field — replica placement comes from the bucket-target config or the source object — but the validator rejected any config carrying it. The console's add-rule form always sends `StorageClass=STANDARD`, so every rule created through it failed with `InvalidRequest`. Explicit `STANDARD` is now tolerated as a no-op (semantically identical to omitting the field); every other value stays rejected because it would be silently ignored rather than honored. AWS S3 and MinIO both accept `STANDARD` here, so this narrows a parity gap. The field remains deliberately absent from the replication capability contract (documented inline).

2. **MinIO-style partial updates for `set-remote-target?update=true`** (`rustfs/src/admin/handlers/replication.rs`). The update path previously replaced every stored field and required complete credentials, so flipping sync/async from the console forced operators to re-enter the secret key — and real `mc replicate update` bodies (madmin `Clone()` strips the secret) failed to deserialize at all. The handler now implements MinIO's `TargetUpdateType` contract:
   - Query params `creds` / `sync` / `bandwidth` / `path` name the field groups to overlay onto the stored target; unnamed groups keep their persisted values; zero ops is a no-op (matching MinIO).
   - Ops MinIO defines but RustFS cannot honor (`proxy`, `healthcheck`, `edge`, `edgeSyncBeforeExpiry`) fail loudly instead of being silently ignored.
   - Credentials updates are skipped for site-replication peer targets (operator never knows the site replicator's credentials); peer ownership is probed via both scheme derivations of the stored endpoint plus the stored deployment id, and a body-supplied deployment id is ignored on update since it anchors peer identity.
   - madmin JSON aliases (`bandwidthlimit`, `storageclass`, `resetID`, `deploymentID`, `sessionToken`) let genuine mc bodies parse under `deny_unknown_fields`.

3. **Two-site replication lab manager** (`scripts/test/site_replication_smoke.py`): spawns and manages two local rustfs processes, pairs them via the site-replication admin API (idempotent), and verifies bidirectional replication. Used to verify both fixes end to end; kept as a dev harness.

The companion console PR ([rustfs/console#195](https://github.com/rustfs/console/pull/195)) adds the replication rule edit form that consumes the new update contract.

## Verification

- `make pre-commit` — pass
- `cargo test -p rustfs-replication --lib config` — 23 pass (includes new wire-XML acceptance test for the console's exact request shape and a lowercase-rejection guard)
- `cargo test -p rustfs --lib handlers::replication::tests` — 21 pass (includes ops-parser contract, mc wire-shape body, arn requirement, credential requirements per path)
- `cargo test -p e2e_test --lib test_set_remote_target_partial_update_preserves_credentials` — pass (credential-free sync-only update preserves stored connection; zero-ops update with a hostile body changes nothing)
- `cargo test -p e2e_test --lib test_set_remote_target_update` — 2 pass
- `cargo clippy -p rustfs --lib` / `-p rustfs-replication --all-targets` — clean
- Manual end-to-end on a local two-site lab (`scripts/test/site_replication_smoke.py up`): console-shaped `PutBucketReplication` with `StorageClass=STANDARD` accepted (`STANDARD_IA` still 400), sync-mode flip via `update=true&sync=true` with an empty secret succeeded with credentials preserved (`site-replicator-0` untouched), bidirectional replication smoke passed after every step.

## Impact

- API-visible, loosening only: `PutBucketReplication` accepts rules carrying `Destination.StorageClass=STANDARD` (previously rejected); all other values keep the exact same error.
- `set-remote-target?update=true` semantics change from full-replace to per-group overlay. Zero-ops updates become no-ops (MinIO parity). No released client depends on the old semantics: the console gains its first edit UI in the companion PR (co-developed against this contract), internal site-replication flows manipulate `BucketTargetSys` directly and never call this endpoint with `update=true`, and mc could not use the old path at all.
- `mc admin bucket remote edit` / `mc replicate update` against RustFS starts working for sync/bandwidth/path edits.
- Known boundary: madmin has no fields for RustFS's TLS extensions (`skipTlsVerify`, `caCertPem`), so an mc-driven `creds` update resets them to defaults for that target; the console co-sends them and is unaffected.

## Additional Notes

Adversarial validation (high-risk tier, all seven roles, per repo policy) was run for both behavior changes; one-line verdicts:

- Correctness: attacked downstream readers, set_target merge semantics, reconciler interplay, no-op contract — findings (stale e2e assertion, deployment-id trust) fixed; no remaining break.
- Simplicity: minimal in-place shape confirmed; one duplicate test found and removed.
- Security: no authz change, no validation bypass, no secret in logs; body-supplied deployment id no longer trusted on update; SR-peer credential guard hardened (dual-scheme + deployment-id probe).
- Concurrency/durability: suspected in-lock-await ABBA refuted (ordering precedented, ns locks timeout-bounded); RMW fully under the per-bucket targets guard; persistence order unchanged.
- Compatibility: verified against MinIO source and madmin/mc wire bodies; e2e assertion aligned; madmin aliases added so the contract is actually reachable from mc.
- Performance: admin config-mutation path only; no hot-path cost.
- Test coverage: every behavior claim has a named test that fails on revert; merge semantics pinned by the new e2e.

### DIFF
--- a/crates/e2e_test/src/replication_extension_test.rs
+++ b/crates/e2e_test/src/replication_extension_test.rs
@@ -2754,7 +2754,7 @@ async fn test_set_remote_target_update_requires_arn() -> Result<(), Box<dyn Erro
 
     assert_eq!(status, StatusCode::BAD_REQUEST);
     assert!(body.contains("InvalidRequest"), "unexpected response: {body}");
-    assert!(body.to_ascii_lowercase().contains("arn is empty"), "unexpected response: {body}");
+    assert!(body.to_ascii_lowercase().contains("arn is required"), "unexpected response: {body}");
 
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


reatang:
**S3 Implemented Tests failure is pre-existing on `main`, not introduced by this PR.**

Evidence:

- This PR, both runs ([first](https://github.com/rustfs/rustfs/actions/runs/30937111276), [rerun](https://github.com/rustfs/rustfs/actions/runs/30937372088)): a single pytest-xdist worker starts cascading `ERROR` (not `FAILED`) at exactly `test_s3.py::test_object_metadata_replaced_on_put` (~21%); the underlying exception is a teardown `DeleteBucket` → `BucketNotEmpty` after the fixture's object cleanup, which then poisons every subsequent test on that worker.
- `main` @ `15c2bade5` ([run 30923749194](https://github.com/rustfs/rustfs/actions/runs/30923749194)): identical signature — same starting test, same cascade — hours before this branch existed. The next `main` run skipped the suite entirely because its E2E job failed.

This PR touches replication config validation and the `set-remote-target` admin API only; the failing tests are basic object/metadata operations covered by the passing Test and Lint (×3), End-to-End (×2), ILM, and lifecycle jobs. All other 15 checks are green.

Issue #5716, opened 2026-08-04
Production-readiness warning: beta.11 entered a recovery loop and beta.12 failed live compatibility testing


anthonymartin:
# Production-readiness warning: beta.11 entered a recovery loop and beta.12 failed live compatibility testing

**Runtime-tested releases:**

- `1.0.0-beta.11` — sustained workload and migration testing
- `1.0.0-beta.12` — live upgrade qualification, subsequently rolled back

**Observation date:** August 4, 2026  
**Deployment:** Single-node Talos Kubernetes, 8 vCPU, 16 GiB RAM, mirrored NVMe-backed storage  
**Workload:** Approximately 334 GiB across four buckets, including a versioned container registry and a 289,220-object application dataset

## Updated summary

I’m posting this warning so other users do not mistake the current RustFS beta releases for production-ready software.

We did not observe confirmed object loss. Full source-to-destination verification and subsequent S3 read/write tests passed on beta.11. However, beta.11 entered an OOM → unclean restart → full-heal → log-flood cycle, while beta.12 failed live compatibility testing across quotas, multipart uploads, registry metadata operations, and private OIDC discovery.

We attempted to upgrade the shared storage server to beta.12 specifically to determine whether it was a safer basis for production. It was not. The upgrade had to be rolled back to beta.11.

Based on these results, we would treat both releases as alpha-quality for production use.

## Beta.12 live-upgrade results

RustFS `1.0.0-beta.12` was deployed using its immutable OCI digest while retaining the same 5 TiB persistent volume. The pod became Ready and its basic health endpoints passed. Those checks did not reflect actual data-plane health.

### Private OIDC discovery stopped working

Beta.12 applied its outbound SSRF policy to OIDC discovery. Our private Authentik provider stopped working until its exact HTTPS origin was added to `RUSTFS_OUTBOUND_ALLOW_ORIGINS`.

Requiring an allowlist for private discovery can be a defensible security policy, but it was a compatibility-affecting upgrade requirement. A previously working administrative login path stopped functioning immediately after the upgrade.

### Quota-enabled writes returned 503

Our pull-through cache bucket had a configured 50 GiB quota. After the beta.12 upgrade, writes failed with:

```text
Bucket quota check temporarily unavailable, please retry
```

RustFS returned `503 ServiceUnavailable`, which the Distribution registry exposed to clients as HTTP 500. A TaxLint preview subsequently could not pull public base images and therefore could not build any of its four containers.

[...]

PR #5721, opened 2026-08-05
fix(s3): degrade multipart listings per upload instead of failing the bucket

overtrue:
# fix(s3): degrade multipart listings per upload instead of failing the bucket

## Problem

Issue #5716 (beta.12 live-upgrade report): OCI registry pushes through Docker Distribution failed with `InternalError: Corrupted format` at multipart completion. Distribution calls `ListMultipartUploads` on every PATCH/commit to resume its upload, and beta.12's cross-set listing rewrite (#5195/#5362) made that call fragile in two ways:

1. **Cross-bucket poisoning.** The multipart staging namespace is one flat set of `sha256(bucket/object)` directories shared by every bucket, and the bucket-scoped listing reads every upload's metadata. An upload whose stored owner bucket differs from the listed bucket fell into the `_ => Err(DiskError::CorruptedFormat)` arm — so *any concurrent in-flight upload in another bucket* turned every `ListMultipartUploads` of this bucket into an InternalError. This is deterministic and matches the reporter's concurrent multi-bucket registry workload (regression test: `list_multipart_uploads_ignores_other_buckets_uploads`, which fails with `CorruptedFormat` before this fix).
2. **Debris poisoning.** A single upload directory whose `xl.meta` is torn or undecodable (unclean shutdown mid-create — the reporter's beta.11 OOM/restart loop) failed the whole listing permanently (regression test: `list_multipart_uploads_skips_undecodable_upload_dirs`).

## Fix

- `list_multipart_uploads_for_incarnation` now degrades per upload: healthy uploads that belong to another bucket are skipped silently; upload directories whose metadata is affirmatively corrupt (at discovery quorum, or unidentifiable after a quorum read) are skipped with a `debug!` log. Every other decode failure — quorum loss from offline disks, timeouts, transport errors — still fails the listing so clients retry instead of silently losing entries. The degrade-vs-propagate decision is a named classifier (`is_corrupt_upload_metadata_error`, same corrupt family as `classify_metadata_response_error`) with a unit test pinning both sides, because a naive transient-error blacklist turned out to be dead code: `reduce_read_quorum_errs` filters `OBJECT_OP_IGNORED_ERRS`, so transient quorum loss surfaces as `ErasureReadQuorum`, never as the underlying `DiskNotFound`/`FaultyDisk`.
- `FileMeta::check_xl2_v1` now classifies a missing/wrong `XL2 ` magic (including a file too short to hold it) as `Error::FileCorrupt` instead of an anonymous `Error::other(...)` that surfaced as `DiskError::Io`. A torn header is affirmative evidence of damage, and the corrupt classification is what lets the listing (and heal/metrics classification generally) distinguish deterministic damage from transient IO faults.

Skipping (rather than stubbing) corrupt directories follows MinIO's listing behavior; the debris remains reclaimable through the existing stale-upload cleanup (24h expiry sweep), and the pre-existing missing-metadata reclaimable-stub branch for exact-key listings is unchanged.

## Adversarial validation

Seven-role review per AGENTS.md (correctness, simplicity, security, concurrency/durability, compatibility, performance, test coverage) via four independent reviewer passes. Findings acted on: the initial transient-error blacklist was replaced with the corrupt-family allowlist (correctness/security/compatibility all independently proved the blacklist unreachable and the degrade arm over-broad); the per-listing `warn!` was demoted to `debug!` (the issue is partly about log flooding); the corrupt-arm stub was replaced with a skip (the stub was not abortable, and MinIO skips). Null results: no cross-bucket information leak (only the sha-hashed path is logged), no pagination/marker corruption from skipped entries, no lock/durability surface touched, no added IO on the listing hot path.

## Testing

- `cargo test -p rustfs-ecstore --lib -- list_multipart_uploads_ignores_other_buckets_uploads list_multipart_uploads_skips_undecodable_upload_dirs list_multipart_uploads_survives_sub_quorum_corruption corrupt_upload_metadata_classification_excludes_transient_faults` (new regression tests; the first two fail without the fix)
- `cargo test -p rustfs-filemeta` (223 tests, header reclassification)
- Full `set_disk::ops::multipart` module run; `make pre-commit`; `cargo clippy -p rustfs-ecstore -p rustfs-filemeta --all-targets` clean

Refs #5716

### DIFF
--- a/crates/ecstore/src/set_disk/ops/multipart.rs
+++ b/crates/ecstore/src/set_disk/ops/multipart.rs
@@ -357,6 +357,20 @@ fn paginate_upload_page(remaining: &[MultipartInfo], max_uploads: usize) -> (Vec
     (page, is_truncated, next_upload_id_marker)
 }
 
+/// Deterministic per-upload damage during a multipart listing: the metadata
+/// exists but is torn, undecodable, or not identifiable as an upload. The same
+/// corrupt family as `classify_metadata_response_error`'s corrupt group. This
+/// deliberately excludes transient fault shapes (`DiskNotFound`, `Timeout`,
+/// `Io`, `ErasureReadQuorum` from offline-disk quorum loss, ...): degrading on
+/// those would silently drop healthy uploads from a 200 listing while disks
+/// are merely unreachable (issue #5716 review).
+fn is_corrupt_upload_metadata_error(err: &DiskError) -> bool {
+    matches!(
+        err,
+        DiskError::FileCorrupt | DiskError::CorruptedFormat | DiskError::CorruptedBackend | DiskError::OutdatedXLMeta
+    )
+}
+
 async fn multipart_upload_paths_on_disk(disk: DiskStore, bucket: &str) -> disk::error::Result<Vec<String>> {
     if !disk.is_online().await {
         return Err(DiskError::DiskNotFound);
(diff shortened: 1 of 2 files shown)


houseme (approved):

overtrue:
CI note: the S3 Implemented Tests failure here is the known main-level BucketNotEmpty cascade from issue #5703 (inline→non-inline overwrite leaks a synthesized rollback directory that DeleteBucket then trips over), which is fixed by the pending #5724 — the same cascade previously blocked #5717. The failing tests are unrelated setup/teardown DeleteBucket calls, and this PR does not touch object write/delete paths (listing degradation + xl.meta header error classification only). Will rebase and re-run once #5724 lands.

PR #5720, opened 2026-08-05
refactor(ecstore): unify remaining heal logs to structured event style

overtrue:
## Related Issues

Residual cleanup for the log-amplification portion of #5716. The amplification fix itself landed in #5719, which superseded this PR's original demotion diff (see PR history and comments).

## Summary of Changes

#5719 fixed the per-object heal log amplification: per-object statements were demoted, `heal_object` spans are now forced to TRACE by a guardrail with self-tests, and raw object/disk metadata dumps are banned at every level. After rebasing, this PR carries only the remaining log-consistency cleanup on top of it, all in `crates/ecstore/src/set_disk/ops/heal.rs`:

- Convert the remaining bare-field and format-arg log statements to the file's now-dominant structured convention (`event`/`component`/`subsystem` + context fields + short message): the missing-object skip, disk-marked-for-healing, the two cannot-reconstruct `error!`s, the dangling-cleanup `error!`, the missing-`data_dir` `error!`, the xl.meta-regeneration `warn!`, and the orphan-reclaim failure `warn!` (now matching its structured success sibling).
- Demote the last remaining `info!` in the file — the per-set `heal_format` "set disk formats success, NoHealRequired" no-op message — to a structured `debug!` reporting `error_count` instead of a raw `errs` dump, and drop its whitelist exclusion in `scripts/check_logging_guardrails.sh` so the no-INFO check for set-disk heal files is strict.

No control flow, return values, or behavior changes; log levels are unchanged everywhere except the one no-op `info!` demotion.

## Verification

- `cargo fmt --all --check`
- `./scripts/check_logging_guardrails.sh` (including the #5719 self-tests, with the tightened strict no-INFO check)
- `cargo check -p rustfs-ecstore`
- Earlier revisions of this PR passed full adversarial validation per AGENTS.md and `make pre-commit`; the rebased residual diff is a strict subset of that reviewed surface (field additions and one level demotion).

## Impact

- Operational only: heal logs in this file share one structured shape for filtering (`component="ecstore"`, `subsystem="heal"/"set_disk"`); the per-set format-heal no-op message moves from INFO to DEBUG. No S3 API, on-disk format, or configuration impact.

## Additional Notes

Follow-ups tracked separately: #5727 demotes per-object heal task logging in `crates/heal` and takes over the CI wiring for `check_logging_guardrails.sh`; FileInfo Debug redaction landed in #5725.

PR #5727, opened 2026-08-05
fix(heal): demote per-object heal logs and cap erasure-set failure warns

### LINKED ISSUE #5716: Production-readiness warning: beta.11 entered a recovery loop and beta.12 failed live compatibility testing

overtrue:
## Related Issues
Fixes #5716 (remaining heal-side amplification sources; the ecstore per-object heal INFO demotion landed separately).

## Summary of Changes
A recovery loop that queues hundreds of thousands of per-object heal tasks (MRF replay, autoheal, scanner) amplified into per-object `info!`/`warn!`/`error!` lines from `crates/heal`. This PR removes the remaining amplification sources while keeping aggregate task kinds, foreground (admin/internal) requests, and all unique failure diagnostics at operator-visible levels.

- **Shared level-split macro**: new `demote_to_debug_when!(cond, level, target, { fields })` in `task.rs` (plus `HealType::is_per_object()` classifying Object/Metadata/MRF/ECDecode vs Cluster/Bucket/Prefix/ErasureSet). A future `HealType` variant defaults to aggregate, i.e. louder logging (fail-safe direction).
- **`task.rs` lifecycle logs**: `Heal task started`/`completed` (2 `info!` lines per object), `timed_out` (`warn!`), and the generic `failed` `error!` demote to `debug!` for per-object task kinds. The failed arm was a strict-subset duplicate: per-attempt object context still `error!`s inside `heal_object` (`result = "failed"`), and terminal failures still `error!` at the scheduler (`state = "task_failed"`); previously one recoverable per-object failure emitted up to 9 `error!` lines across the three layers.
- **`task.rs` missing-object warn**: `Heal target object is missing` fired per missing object; background sources (scanner/MRF/autoheal/read-repair) routinely race object deletion, so it demotes to `debug!` for them and keeps `warn!` (now with a `source` field) for admin/internal requests.
- **`erasure_healer.rs` sweep warns**: the per-failed/per-skipped-object `transient_skip` and `failed` `warn!`s are now sample-capped per bucket via a shared `take_failure_log_sample` helper (extracted from the existing `heal_bucket_objects` precedent, `MAX_BUCKET_FAILURE_LOG_SAMPLES = 5`); over-cap occurrences log at `debug!`. A flapping rebuild disk can no longer produce ~289k warns per sweep — the end-of-pass `retry_scheduled` warn / `failed_after_retries` error still report full failed/skipped counts.
- **`manager.rs` admission and retry logs**: queue-full/drop/displacement/duplicate-drop `warn!`s and the scheduler `task_retrying` warn level-split by request kind; the per-retry admission decision logs (`retry_enqueued` etc.) demote to `debug!` unconditionally — they fire once per retrying task (per object under MRF retry storms), repeat every backoff cycle while the queue is full, and duplicate both `admit_request_to_queue`'s own logging (context `retry`) and the scheduler `task_retrying` event. `rustfs_heal_admission_total`, the queue-length gauge, and the `heal_queue_state` backlog event keep the aggregate signal. The previously unmetered duplicate-admission short-circuit now records `rustfs_heal_admission_total{context="duplicate"}` so demoted per-object duplicate drops stay observable.
- **Guardrails**: `scripts/check_logging_guardrails.sh` gains four checks pinning the demotions (retry-admission messages must not appear at `info!`/`warn!`, count-based checks on the macro/sampling call sites). The regex window and the count thresholds were verified by injection: flipping any demoted site back fails the script, and it passes on the clean tree. The script now also runs in CI quick-checks — it previously only ran via `make pre-commit`/`pre-pr` — added to both `ci.yml` and its byte-identical `ci-docs-only.yml` mirror per the mandate in those files.

## Verification
- `./scripts/check_logging_guardrails.sh` — passes; regression injection verified both directions (flipping `retry_enqueued`/`retry_cancelled` back to `info!`/`warn!` in a scratch copy is caught; single-site reverts of counted macro/sampling sites drop below the thresholds).
- `cargo clippy -p rustfs-heal --all-targets -- -D warnings` — clean.
- `cargo test -p rustfs-heal --lib` — 233 passed (includes new `per_object_heal_types_are_classified_for_log_demotion` covering all 8 `HealType` variants and `failure_log_sampling_caps_at_max_samples`).
- `cargo test -p rustfs-heal --tests` — all pass except `heal_b920_subquorum_union_test` (2 failures), which fails identically on unmodified `main` (3/3 runs, panic in `ecstore/src/runtime/sources.rs` global disk map setup when the whole binary runs; each test passes in isolation) — pre-existing, tracked for a separate fix.
- `make pre-commit` — passes.
- `actionlint .github/workflows/ci.yml .github/workflows/ci-docs-only.yml` and `./scripts/check_ci_paths_sync.sh` — pass; quick-checks step lists verified mirrored.

Adversarial validation (high-risk tier, all seven roles): correctness — attacked macro expansion/side-effect-once semantics, sampling counter equivalence with the replaced inline code, and control-flow identity at every touched branch; no behavior break. Simplicity — consolidated three hand-written level-split shapes into the shared macro (~90 duplicated lines removed) and de-brittled the guardrail checks per findings. Security — verified every demoted path keeps an aggregate metric or summary signal (no failure mode is invisible at INFO and unmetered after the duplicate-admission metric fix); no content moved to higher levels. Concurrency — sample counters are single-consumer locals (concurrent page futures never touch them); no log moved relative to lock drops/state mutations; no cancellation points changed. Compatibility — no consumer (including `log-analyzer` heal rules) anchors on the demoted events/messages; no metrics or admin-visible types changed; CI mirror-job mandate honored. Performance — no new allocation on demoted paths (lazy field sigils, `matches!` gating); residual per-object missing-object/timeout/failed duplication found by review is fixed in this PR. Test coverage — guardrail dead-regex finding fixed and injection-verified; count thresholds sit exactly at the current site counts so single-site reverts fail CI.

## Impact
Operational logging only: per-object heal noise moves from `info!`/`warn!`/`error!` to `debug!`; operators keep aggregate signals (`rustfs_heal_admission_total`, `rustfs_heal_task_start_total`, queue gauges, `heal_queue_state` backlog event, per-bucket sampled warns, end-of-sweep summaries, terminal `task_failed` errors). No API, storage-format, metric, or configuration changes. Log-based alerting that matched the demoted per-object messages at INFO/WARN (none found in-repo) would need to switch to the aggregate events or metrics.

## Additional Notes
`heal_b920_subquorum_union_test` failures are pre-existing on `main` and unrelated (verified by stashing this diff); flagged separately for a dedicated fix.

### DIFF
--- a/.github/workflows/ci-docs-only.yml
+++ b/.github/workflows/ci-docs-only.yml
@@ -102,6 +102,9 @@ jobs:
       - name: Check architecture migration rules
         run: ./scripts/check_architecture_migration_rules.sh
 
+      - name: Check logging guardrails
+        run: ./scripts/check_logging_guardrails.sh
+
       - name: Check tokio io-uring feature guard
         run: ./scripts/check_no_tokio_io_uring.sh
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,9 @@ jobs:
       - name: Check architecture migration rules
         run: ./scripts/check_architecture_migration_rules.sh
 
+      - name: Check logging guardrails
+        run: ./scripts/check_logging_guardrails.sh
+
       - name: Check tokio io-uring feature guard
         run: ./scripts/check_no_tokio_io_uring.sh
 
--- a/crates/heal/src/heal/erasure_healer.rs
+++ b/crates/heal/src/heal/erasure_healer.rs
@@ -16,7 +16,7 @@ use crate::heal::{
     progress::HealProgress,
     resume::{CheckpointManager, ResumeManager, ResumeUtils, compose_key},
     storage::{HealStorageAPI, next_heal_listing_token},
-    task::is_missing_object_dir_heal_result,
+    task::{demote_to_debug_when, is_missing_object_dir_heal_result, take_failure_log_sample},
 };
 use crate::{Error, Result};
 use futures::{StreamExt, stream::FuturesUnordered};
@@ -612,6 +612,12 @@ impl ErasureSetHealer {
         let page_concurrency_limit =
             Self::effective_heal_page_object_concurrency_for_source(self.source, self.heal_opts.scan_mode);
         let in_flight = Arc::new(AtomicUsize::new(0));
+        // Per-bucket sample caps for per-object warn! lines: a flapping rebuild
+        // disk can fail/skip hundreds of thousands of versions in one sweep, so
+        // only the first few occurrences warn and the rest demote to debug!.
+        // The end-of-pass summary reports the full failed/skipped counts.
+        let mut transient_skip_samples_logged = 0_u64;
+        let mut failure_samples_logged = 0_u64;
 
         // backlog#920: select the per-erasure-set DISK-WALK union enumerator when
         // the scan is Deep OR the request came from AutoHeal — these are the paths
@@ -748,8 +754,7 @@ impl ErasureSetHealer {
                     Err(Error::TransientSkip { message }) => {
                         *skipped_objects += 1;
                         checkpoint_manager.add_skipped_object(key).await?;
-                        warn!(
-                            target: "rustfs::heal::erasure_healer",
+                        demote_to_debug_when!(!take_failure_log_sample(&mut transient_skip_samples_logged), warn, target: "rustfs::heal::erasure_healer", {
                             event = EVENT_HEAL_ERASURE_OBJECT_STATE,
                             component = LOG_COMPONENT_HEAL,
                             subsystem = LOG_SUBSYSTEM_ERASURE_HEALER,
@@ -760,13 +765,12 @@ impl ErasureSetHealer {
                             state = "transient_skip",
                             error = %message,
                             "Erasure set object heal skipped due to transient error"
-                        );
+                        });
                     }
                     Err(err) => {
                         *failed_objects += 1;
                         checkpoint_manager.add_failed_object(key).await?;
-                        warn!(
-                            target: "rustfs::heal::erasure_healer",
+                        demote_to_debug_when!(!take_failure_log_sample(&mut failure_samples_logged), warn, target: "rustfs::heal::erasure_healer", {
                             event = EVENT_HEAL_ERASURE_OBJECT_STATE,
                             component = LOG_COMPONENT_HEAL,
                             subsystem = LOG_SUBSYSTEM_ERASURE_HEALER,
@@ -777,7 +781,7 @@ impl ErasureSetHealer {
                             state = "failed",
                             error = %err,
                             "Erasure set object heal failed"
-                        );
+                        });
                     }
                 }
 
--- a/crates/heal/src/heal/manager.rs
+++ b/crates/heal/src/heal/manager.rs
@@ -15,7 +15,7 @@
 use crate::heal::{
     progress::{HealProgress, HealStatistics},
     storage::HealStorageAPI,
-    task::{HealOptions, HealPriority, HealRequest, HealTask, HealTaskStatus, HealType},
+    task::{HealOptions, HealPriority, HealRequest, HealTask, HealTaskStatus, HealType, demote_to_debug_when},
 };
 use crate::{Error, Result};
 use metrics::{counter, gauge};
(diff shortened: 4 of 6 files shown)


overtrue:
Heads-up on overlap with #5720 (the ecstore side of #5716): both PRs append checks to `scripts/check_logging_guardrails.sh` (textual conflict for whichever lands second) and both add a "Check logging guardrails" step to the `ci.yml` quick-checks job at different positions — a naive merge could end up with the step duplicated. This PR also adds the step to the byte-identical `ci-docs-only.yml` mirror per the mandate in both workflow files; #5720 currently does not. Whichever PR merges second should rebase, keep a single `ci.yml` step, and make sure the `ci-docs-only.yml` mirror carries it too.

overtrue:
Overlap with #5720 is now resolved: this PR's guardrail-script checks moved to a non-overlapping anchor (after the stdout-sink check) and rebased onto latest main; #5720 dropped its duplicate `ci.yml` step (this PR carries the step in both quick-checks jobs per the mirror mandate). Verified with `git apply --check` that each PR's diff applies cleanly on top of the other, in both merge orders.

overtrue:
The "S3 Implemented Tests" failure here is the known main-level BucketNotEmpty cascade from the inline-rollback leak (#5703): this branch is based on a main commit that predates the fixes. #5724 (synthetic rollback dirs) merged after this branch's base, and #5732 (rollback backups) is still open. This PR only changes log levels in crates/heal and cannot produce BucketNotEmpty. Will rebase once #5732 lands to pick up both fixes.

PR #5719, opened 2026-08-05
fix(heal): bound per-object logging

anthonymartin:
## Related Issues

Related to #5716

## Summary of Changes

This bounds the ECStore heal telemetry that amplified an unclean-restart recovery into sustained CPU, disk-write, and log-volume pressure during the beta.11 incident described in #5716.

The change is built around four invariants:

1. Default `INFO` logging does not scale with healed object, wrapper, shard, or part count.
2. Writer creation failures emit at most one structured `WARN` per healed object, including an aggregate count and the first failed part/disk.
3. Heal diagnostics never dump raw `FileInfo`, disk handles, erasure distributions, or storage paths.
4. Detailed per-object/per-part progress remains available at `TRACE` for targeted, temporary diagnosis.

Concretely, this PR:

- moves per-object wrapper spans and object/part progress events from `INFO` to `TRACE`;
- replaces raw metadata and disk `Debug` dumps with bounded structured counts and states;
- demotes expected per-disk metadata divergence diagnostics to structured `DEBUG`;
- retains actionable failures, but aggregates heal-writer failures into one redacted `WARN` per object;
- reports I/O failures by `ErrorKind` so filesystem paths and backend details are not exposed;
- adds hermetic partial- and all-target writer-failure tests; and
- extends the logging guardrail with negative fixtures for multiline spans, long events, old sentence-style logs, and raw structured/positional metadata dumps.

The heal decisions and data-path behavior are unchanged. Invalid-distribution error text is intentionally made bounded, while the same conditions still fail closed.

## Verification

- `make pre-pr`
  - all formatting, architecture, security, logging, schema, and repository policy guards passed;
  - workspace Clippy passed with warnings denied;
  - `cargo nextest run --all --exclude e2e_test`: 12,492 passed, 0 failed, 163 skipped;
  - `cargo test --all --doc`: passed.
- `cargo nextest run -p rustfs-ecstore should_heal_object_on_disk`
- `cargo nextest run -p rustfs-ecstore heal_writer_`
- `cargo nextest run -p rustfs-ecstore 'erasure::coding::heal'`
- `cargo clippy -p rustfs-ecstore --all-targets -- -D warnings`
- Manual mutation check: changing a protected heal event back to `INFO` is rejected by `scripts/check_logging_guardrails.sh`.
- Adversarial review covered correctness, durability/concurrency, security, compatibility, performance, simplicity, and test coverage; all roles completed without remaining findings.

## Impact

[...]

PR #5742, opened 2026-08-05
fix(scanner): publish bounded observational usage

anthonymartin:
## Related Issues

Related to #5716 and follows #5719.

## Summary of Changes

An object scan can finish structurally complete but become superseded when the bucket namespace changes during the scan. Today that useful result is discarded, while the scanner can immediately repeat the same expensive full traversal. Under sustained namespace activity this leaves the console stale and can turn scanner recovery into continuous CPU and disk-I/O pressure.

This change separates quota authority from operational observation and makes retry behavior bounded:

1. `.usage.v2.json` remains the sole authoritative snapshot for quotas and write admission.
2. A complete-but-superseded scan is published to `.usage.observed.json` for admin and metrics visibility only.
3. An observed snapshot is selectable only when it is newer and names the exact authoritative baseline against which it was scanned.
4. Only a durably persisted authoritative scan clears dirty generations or acknowledges remote changes.
5. Repeated superseded scans retry with exponential backoff, capped at 30 minutes; dirty and ordinary cluster-change wakes cannot bypass that backoff.

The admin account-info and cluster-usage surfaces now expose an additive optional `usage_snapshot_converged` field, and metrics expose `rustfs_cluster_usage_objects_snapshot_converged`. Startup may run promptly when a valid newer observation proves convergence is still pending.

## Verification

- `make pre-commit`
- `make clippy-check`
- Full `make pre-pr` passed before rebasing from `f0c4fbd28` onto the current `upstream/main`.
- On the exact rebased scanner commit (`d0beeb4b`), every pre-PR target except `s3s-footprint-check` passed: all guardrails, Clippy, 12,546/12,546 nextest tests, and all workspace doc tests.
- `s3s-footprint-check` self-failed identically on the then-current `upstream/main` and this branch (1,687 invocations versus its 1,686 baseline); this branch adds zero `s3_error!` invocations. Upstream fixed that baseline independently in #5747.
- On the CI-isolation follow-up (`033b051fa`), `cargo nextest show-config test-groups --profile ci --groups embedded-test-ports -p rustfs --no-pager` resolved all six active port-probing embedded-server tests into one max-concurrency-1 group.
- `cargo nextest run --profile ci -p rustfs -E 'binary(/^embedded.*_test$/) | binary(admin_diagnostic_capability_e2e)'` passed 6/6 in 11.489 seconds. The worst-case serialization penalty versus perfect overlap is about 8.9 seconds; compilation and the rest of the suite remain parallel.
- All other `make pre-commit` targets, including `quick-check`, passed on `033b051fa`; the only local aggregate failure was the already-fixed #5747 baseline mismatch inherited by the branch.
- The CI isolation uses no retries or sleeps: first-attempt server failures remain visible.
- On replacement CI run `31035982211`, the sftp leg ran all 3,300 tests successfully, including every embedded integration test. GitHub subsequently marked the job failed only because matrix fail-fast cancelled sibling jobs after an unrelated swift-leg failure.
- The swift failure exposed an existing load-sensitive assumption in `execute_get_object_generation_change_bypasses_old_cold_fill_plan`: its intentionally blocked GET inherited a five-second dynamic production deadline while the test performed a real 1.3 MiB replacement write. A test-only policy seam now disables dynamic sizing for that scenario without changing production timeouts, adding retries, or removing its 2-second join and 10-second completion assertions.
- On the exact follow-up (`a13f0d882`), `cargo nextest run --profile ci -p rustfs --features swift -E 'test(execute_get_object_generation_change_bypasses_old_cold_fill_plan)'` passed in 2.760 seconds. Production-mode `quick-check`, formatting, diff, and every pre-commit target except the branch-inherited #5747 baseline mismatch passed.
- Focused data-usage tests cover additive wire compatibility, exact-baseline selection, clock skew, namespace mutation, structural completeness, and convergence ordering.
- Focused scanner tests cover real-channel publication, authoritative/observed persistence separation, startup convergence, durable dirty-state retention, exponential backoff, and deferred wake sources.
- Focused ECStore, madmin, and observability tests cover the selector boundary, additive API compatibility, and the convergence metric.

## Impact

- Quota enforcement and S3 write admission continue to consume only `.usage.v2.json`; observational state cannot authorize writes or raise a quota baseline.
- Admin and metrics consumers can receive a fresher, explicitly non-converged complete snapshot while a clean authoritative scan is pending.
- A new internal metadata object, `.usage.observed.json`, is best-effort and is never required for bucket operations.
- API and persisted-structure changes are additive and optional; legacy snapshots decode with convergence unknown.
- No new configuration is required. Superseded retry backoff starts at the smaller of the configured interval and 60 seconds, doubles per consecutive superseded cycle, and caps at 30 minutes.
- Five integration-test binaries that use the best-effort embedded-port probe are serialized as a small nextest group to close their cross-process TOCTOU race.

## Additional Notes

- Specification review: PASS — the change addresses stale operational usage and unbounded superseded retries without changing quota authority.
- Correctness review: PASS — exact authoritative-baseline identity fences stale observations, and only durable authoritative completion clears change state. The CI follow-up covers every integration binary that calls `find_available_port`, in both default and CI profiles.
- Security review: PASS — observations are excluded from quota decisions and cannot inflate the degraded authoritative fallback.
- Compatibility review: PASS — wire fields are optional/additive, legacy data defaults safely, and the existing authoritative object remains unchanged in role.
- Performance review: PASS — admin reads use a separate 30-second cache, scans do not add live object listing, retries are exponentially bounded, and embedded-test serialization adds at most about 8.9 seconds in the measured subset.
- Simplicity review: PASS — the two-object model makes authority explicit and keeps retry policy in scanner scheduling state. A dedicated nextest group is smaller than changing embedded-server listener ownership and avoids hiding failures with retries. A test-only timeout-policy seam is narrower and more deterministic than process-wide environment mutation or a production timeout change.
- Test review: PASS — revert-detecting tests exercise publication, selection, persistence, startup, acknowledgement, backoff, API/metric boundaries, and the complete port-probing integration-test inventory.


crates/scanner/src/scanner.rs
```diff
@@ -354,6 +365,36 @@ struct ScannerCleanIdleBackoff {
...
+        }
+    }
+
+    fn retry_interval(self, configured_interval: Duration) -> Option<Duration> {
+        let exponent = self.consecutive_cycles.checked_sub(1)?.min(31);
+        let multiplier = 1u32.checked_shl(exponent).unwrap_or(u32::MAX);
+        let base_interval = configured_interval
+            .max(Duration::from_secs(1))
+            .min(SUPERSEDED_RETRY_BASE_INTERVAL);
+        Some(base_interval.saturating_mul(multiplier).min(SUPERSEDED_RETRY_MAX_INTERVAL))
```
overtrue: This cap makes the superseded retry cadence independent of the operator's configured cycle interval. With `cycle_interval = 24h` under sustained write load, every cycle classifies as Superseded (any object write during the scan regrows the dirty set), so the scanner converges to a full namespace traversal every 30 minutes — ~48 full scans/day where the operator configured 1. For the supersession case with no dirty wake pending, this is also a frequency increase over current behavior, which waits out the full configured interval. Operators pick long intervals precisely to bound scan I/O on large clusters. Please either make the retry interval respect the configured interval as an upper bound (e.g. `.min(SUPERSEDED_RETRY_MAX_INTERVAL).min(configured_interval.max(base_interval))`), make the cap configurable, or document in the PR why a fixed 30-minute ceiling should override operator configuration.

crates/ecstore/src/data_usage/mod.rs
```diff
@@ -551,6 +588,23 @@ where
...
+        bucket,
+        DATA_USAGE_REMOVE_CAS_RETRIES,
+        None,
+        guard,
+    )
+    .await
+    {
+        // The authoritative timestamp was already advanced above, so admin
+        // selection rejects this observation even if optional cleanup fails.
+        // Never make an admin-only freshness artifact block DeleteBucket.
```
overtrue, line 664: Bucket delete is currently the only path that ever removes or rewrites `.usage.observed.json`. After the next authoritative scan completes, the observation is permanently fenced by the baseline identity check but the object itself lingers indefinitely, and both scanner startup (`persisted_usage_cache_is_cold_for_startup`) and every admin cache refresh keep paying a read for a file that can never be selected again. Please add a best-effort delete (same never-block, log-on-failure posture as here) of the observed object after a successful authoritative save, so the steady state of a converged cluster has no stale observation on disk.

crates/obs/src/metrics/stats_collector.rs
```diff
@@ -102,6 +103,7 @@ async fn load_obs_data_usage_from_backend(store: Arc<ObsStore>) -> ObsEcstoreRes
     Ok(ObsDataUsageInfo {
         last_update: data_usage.last_update,
         usage_snapshot_complete,
+        usage_snapshot_converged: data_usage.usage_snapshot_converged != Some(false),
```
overtrue: `!= Some(false)` maps `None` (unknown) to converged=1. The admin selector upgrades complete authoritative snapshots from `None` to `Some(true)` before this point, so the remaining `None` cases reaching this line are precisely the degraded/non-authoritative fallbacks — an incomplete snapshot whose bucket data was discarded will report `snapshot_converged 1`. That inverts the metric's stated meaning for exactly the situations operators most need to notice. Please map `None` to non-converged (or gate on `usage_snapshot_complete`), or document why unknown should read as converged.

overtrue (changes requested):
Thanks for the very thorough PR. I did an adversarial review across six axes and found no blocking correctness defects in the code itself:

- **Quota authority**: verified that quota admission (`get_real_time_usage`, memory overlay, `lookup_degraded_bucket_usage_baseline`) never consumes `.usage.observed.json`; observational saves skip `replace_bucket_usage_memory_from_info`, and `store_data_usage_in_backend` hard-rejects `converged == Some(false)`. The "observations cannot inflate the degraded fallback" claim holds.
- **Baseline fencing**: the exact `(last_update, scanner_cycle, scanner_epoch)` identity match is clock-independent, and the load-bearing premise that every bucket delete rewrites the authoritative identity holds (`advance_data_usage_last_update` is monotonic), including the old-binary rollback delete/recreate case.
- **Persistence/wire compat**: both new fields are `Option` + `skip_serializing_if`, `DataUsageInfo` is JSON-persisted without `deny_unknown_fields`, so a rolled-back binary reads the new authoritative object safely; the msgpack array-encoding invariant only constrains `DataUsageCache`/`DataUsageEntry`, untouched here. madmin `AccountInfo` field is additive.
- **Normalization**: `is_complete_bucket_usage_snapshot` is internal-consistency only, so a selected observation cannot be gutted by `discard_incomplete_bucket_usage`; the memory overlay is an absolute replace gated on `usage_updated_at`, so no double counting on top of an observation.
- **Concurrency**: the split generation counters are wired correctly, the lock order is globally quota-cache -> admin-cache with no reverse acquisition, and dirty clear/ACK is correctly gated on `Complete` + durable persist (locked by `finalizing_a_durable_superseded_snapshot_keeps_dirty_work_pending`).
- **Performance**: no hot-path cost; new I/O is one small JSON write per superseded cycle plus one extra metadata read per 30s admin refresh.

I'm requesting changes for three points that need an author response or adjustment before this merges:

1. **Superseded retry cadence can far exceed the operator-configured cycle interval** (inline at `crates/scanner/src/scanner.rs`). With a 24h configured interval and sustained write load (any object write during a scan regrows the dirty set, so every cycle classifies as Superseded), full namespace traversals now run roughly every 30 minutes (~48/day) instead of once a day. For the supersession-without-dirty-wake case this is a frequency *increase* over today's behavior (which waited out the full configured interval). Please either justify the fixed 30-minute cap or make the policy respect the configured interval (e.g. cap the retry interval at the configured cycle interval, or make the cap configurable).
2. **Stale `.usage.observed.json` is never cleaned up after convergence** (inline at `crates/ecstore/src/data_usage/mod.rs`). Once a newer authoritative scan lands, the observation is permanently fenced but the object lingers forever, costing one startup read and one admin-refresh read indefinitely. Please add a best-effort delete of the observed object after a successful authoritative save.
3. **`rustfs_cluster_usage_objects_snapshot_converged` reports 1 for unknown/degraded snapshots** (inline at `crates/obs/src/metrics/stats_collector.rs`). Mapping `None` to converged=1 means a degraded, incomplete fallback snapshot reports as converged. Please fix the mapping or explain why unknown should read as converged.

Non-blocking notes, no action strictly required:

- `invalidate_data_usage_snapshot_cache` in `crates/ecstore/src/data_usage/mod.rs` holds both cache write locks simultaneously, while `prepare_bucket_usage_for_namespace_change` and the guarded invalidation drop the first before taking the second. There is no reverse-order path today so no deadlock, but please unify on drop-then-acquire to keep the lock order trivially auditable.
- The nextest `embedded-test-ports` group and the `#[cfg(test)]` GET-timeout policy seam in `object_usecase.rs` are unrelated CI-flake fixes bundled into a scanner PR. Both are small, well-documented, and verifiable (the binary filter matches the five `find_available_port` users under `rustfs/tests/`), so I won't insist on splitting them, but please keep such fixes in separate PRs going forward.
- Baseline identity via the `(last_update, cycle, epoch)` triple works, but the CAS machinery already has revision etags that would make an exact-baseline check a single opaque string comparison; worth considering as a follow-up simplification.

CI is fully green on the current head (`a13f0d882`).

houseme (approved):

PR #5745, opened 2026-08-05
feat(table-catalog): complete namespace REST contracts

marshawcoco:
## Related Issues

N/A

## Summary of Changes

- Aligns namespace paths with the Iceberg REST multi-level namespace separator while retaining the documented dotted-path compatibility fallback.
- Adds parent-scoped direct-child namespace listing with scoped pagination tokens and store-level pagination for object-backed and durable-strong catalogs.
- Adds the standard namespace property update endpoint with bounded request parsing, operation-specific authorization, request validation, and atomic durable-strong persistence.
- Treats active tables and views as namespace existence evidence so implicit parent namespaces remain discoverable and usable without synthetic marker writes.
- Keeps migration inventory, warehouse-index recovery, and data-plane resource resolution complete when catalog resources exist without explicit namespace entries.
- Adds route, handler, persistence, pagination-boundary, migration, restart, failure, and backing-contract coverage for the new behavior.

## Verification

- `cargo clippy -p rustfs --tests --no-deps -- -D warnings`
- `git diff --check`

## Impact

Namespace listing now follows Iceberg REST direct-child semantics and supports the standard `parent` query parameter. Namespace paths advertise and accept the `%1F` separator, while existing dotted paths remain accepted for compatibility. Durable-strong catalogs advertise and support namespace property updates; object-backed catalogs return `406 Not Acceptable` for that operation.

## Additional Notes

The dotted namespace compatibility fallback is registered in the architecture cleanup register so it can be removed after the minimum supported client release consistently uses the Iceberg REST separator.


houseme:
@codex review

PR #5726, opened 2026-08-05
fix(auth): restore filtered ListBuckets fallback

overtrue:
## Related Issues

Fixes rustfs/backlog#1673

Follow-up to #5688

## Summary of Changes

- Return `AccessDenied` when IAM does not allow the service-level `s3:ListAllMyBuckets` action instead of attempting bucket-policy evaluation with an empty bucket name.
- Let the existing MinIO-compatible `ListBuckets` handler filter results by `s3:ListBucket` or `s3:GetBucketLocation`.
- Add an E2E regression that proves the scoped user sees only the authorized bucket while direct bucket access still succeeds.
- Add the regression to the `e2e-smoke` PR profile and reuse the shared S3 client configuration for explicit credentials.

## Verification

- `make pre-pr`
- `cargo clippy -p rustfs --lib -- -D warnings`
- `cargo build --bin rustfs`
- `CARGO_BIN_EXE_rustfs=target/debug/rustfs cargo nextest run --profile e2e-smoke -p e2e_test -E 'test(bucket_scoped_policy_returns_only_authorized_bucket)'`

## Impact

Bucket-scoped IAM users without `s3:ListAllMyBuckets` now receive the intended filtered bucket list instead of HTTP 500 `errBucketMetadataNotInitialized`. Explicit `s3:ListAllMyBuckets` grants still return the full bucket list. There are no configuration, API shape, storage format, or migration changes.

## Additional Notes

This preserves MinIO's `ListBucketsHandler` contract: any denied global list check enters the per-bucket `ListBucket` / `GetBucketLocation` filter, and buckets without either permission remain hidden.

Adversarial validation verdicts:

- Correctness: exercised explicit grant, implicit deny, empty credentials, owner, and hidden-bucket cases; no break found.
- Simplicity: reduced the fix to one in-place guard and reused shared E2E client setup; no remaining diff-scoped finding.
- Security: unauthorized bucket names remain filtered; explicit-deny fallback matches the pinned MinIO contract; no diff-scoped leak found.
- Concurrency/durability: the runtime guard adds no lock, await, mutation, or persistence path; no break found.
- Compatibility: explicit grants retain full listing and denied grants retain MinIO-compatible filtering; no wire or storage change.
- Performance: the new guard avoids the erroneous empty-bucket metadata lookup and adds no allocations or I/O; existing per-bucket filtering cost is unchanged.
[...]

PR #5746, opened 2026-08-05
fix(auth): align ListBuckets discovery with IAM policies

GatewayJ:
## Related Issues

Follow-up to #5726.

## Summary of Changes

- Rebase the IAM-only `ListBuckets` discovery change onto current `main` and retain the existing quiet structured implicit-deny audit event.
- Keep bucket discovery restricted to IAM policies while preserving bucket-policy enforcement for direct bucket operations.
- Cover OPA initialization and actual HTTP 503 validation failures with fail-closed STS and `ListBuckets` regression tests.

## Verification

- `CARGO_INCREMENTAL=0 make pre-pr`
- 12,610 workspace tests passed under nextest; all doc-tests passed.
- `cargo test -p rustfs-iam test_policy_plugin_state_fails_after_opa_validation_returns_503 --lib`
- `cargo test -p e2e_test --lib list_buckets_filters_with_iam_bucket_resources`
- `cargo test -p e2e_test --lib test_list_buckets_opa_contract`
- `cargo test -p e2e_test --lib test_sts_and_list_buckets_fail_closed_after_opa_validation_failure`

## Impact

`ListBuckets` uses IAM-only discovery semantics: a bucket-policy-only Allow does not make a bucket visible. An IAM Allow keeps a bucket visible even when its bucket policy explicitly Denies access; direct bucket operations remain subject to that Deny. No configuration, migration, wire-format, or protocol changes are required.

## Additional Notes

- The existing quiet structured `iam_implicit_deny` event is preserved when global IAM discovery falls back to bucket-level IAM visibility checks.
- High-risk adversarial validation completed: correctness, simplicity, security, concurrency/durability, compatibility, performance, and test coverage found no unresolved issue.

---


rustfs/src/storage/access.rs
```diff
@@ -859,7 +948,7 @@ pub async fn authorize_request<T>(req: &mut S3Request<T>, action: Action) -> S3R
         }
 
         if action == Action::S3Action(S3Action::ListAllMyBucketsAction) {
-            return Err(s3_error!(AccessDenied, "Access Denied"));
+            return Err(S3Error::with_message(S3ErrorCode::AccessDenied, "Access Denied".to_string()));
```
overtrue: This exact line conflicts with main: #5755 (ea8dbf49a) already replaced the `s3_error!` call here with `ApiError::access_denied().into()` to hold the s3s ratchet at baseline. When rebasing, keep main's version and drop this edit. Note that after this PR no production caller invokes `authorize_request` with `ListAllMyBucketsAction` anymore (the handler goes through `prepare_list_buckets_iam_authorization` instead), so this branch becomes defensive-only; worth a brief comment or a mention in the PR body.

rustfs/src/storage/access.rs
```diff
@@ -580,6 +584,104 @@ fn auth_fs() -> &'static FS {
...
+            .await
+    }
+}
+
+/// Prepare the IAM-only authorization used to decide which buckets are visible in ListBuckets.
+/// Bucket policies are intentionally excluded from this discovery decision, matching MinIO.
+pub(crate) async fn prepare_list_buckets_iam_authorization<T>(req: &S3Request<T>) -> S3Result<ListBucketsIamAuthorization> {
+    let req_info = req_info_ref(req)?;
+    let Some(cred) = req_info.cred.as_ref() else {
+        return Err(S3Error::with_message(S3ErrorCode::AccessDenied, "Access Denied".to_string()));
```
overtrue: After rebasing onto main, prefer `ApiError::access_denied().into()` here (introduced by #5755) over constructing `S3Error::with_message(S3ErrorCode::AccessDenied, ...)` directly. It keeps the AccessDenied code/message pair defined in one place and follows the gateway-side error convention from rustfs/backlog#1677 F1. Same applies to the other new AccessDenied constructions added by this PR.

crates/e2e_test/src/sts_query_compat_test.rs
```diff
@@ -523,10 +527,96 @@ async fn test_sts_assume_role_opa_contract() -> TestResult {
...
+        "ListBuckets should not make redundant OPA evaluations"
+    );
+
+    env.stop_server();
+    Ok(())
+}
+
+#[tokio::test]
+#[serial]
+async fn test_sts_and_list_buckets_fail_closed_while_opa_is_initializing() -> TestResult {
```
overtrue, line 649: Coverage regression: the old `start_delayed_unavailable` variant exercised the 503-during-validation path, i.e. the transition into `PolicyPluginState::Failed`, end to end. `start_blocked` only exercises the Initializing state. The Failed state is now covered solely by the new unit test in crates/iam/src/sys.rs, which constructs the enum variant directly and never drives the actual 503 -> Failed transition. I understand the old test's `sleep(200ms)` was race-prone, but consider keeping a deterministic Failed-state e2e (e.g. a mock that returns 503 immediately on validation) or stating the trade-off in the PR body.

overtrue (changes requested):
Adversarial review of the IAM-only ListBuckets discovery change. Direction is approved: the new semantics match MinIO's ListBucketsHandler point by point (IAM-only filtering with bucket policy excluded from discovery, ListBucket -> GetBucketLocation short-circuit order, injected prefix=""/delimiter="/" condition values, AccessDenied when zero buckets are visible, owner passthrough), and preparing the identity policy context once per request instead of re-merging per bucket per action is a real improvement.

Security verdict: no leak or privilege escalation found. Users without any bucket-scoped IAM grant get AccessDenied and cannot enumerate bucket names (the PutObject-only e2e case pins this), OPA Initializing/Failed states fail closed via PreparedIamMode::Deny, and IAM store read failures also resolve to Deny. The prepared context in crates/iam/src/sys.rs is identity-scoped rather than action-scoped, so reusing it across ListAllMyBuckets/ListBucket/GetBucketLocation evaluations is sound. All principal paths (regular user, group, inherited and session-bound service accounts, STS inherited and session policy, OPA) have e2e coverage.

Blocking: this branch must be rebased onto latest main. The base (6f10ca18a) predates #5747 and #5755, and there is a real textual conflict at rustfs/src/storage/access.rs line 951: this PR rewrites the same line that #5755 (ea8dbf49a) already rewrote on main. Resolution: keep main's `ApiError::access_denied().into()` and drop this PR's `S3Error::with_message` edit on that line. Post-rebase the s3s footprint ratchet is unaffected (this PR adds no `s3_error!` lines; the count stays at 1686 under the 1687 baseline).

Also blocking (documentation, not code): this PR makes two deliberate behavior changes that reviewers and operators need to see explicitly in the PR body, please spell them out there. (1) Users whose only grant is a bucket policy no longer see those buckets in ListBuckets (the #5726 fallback consulted bucket policy via authorize_request; this PR removes that), which affects console users relying on bucket-policy shares. (2) Conversely, an explicit bucket-policy Deny no longer hides a bucket from discovery when IAM allows it; the name is disclosed, though direct operations remain blocked by the explicit-deny gate at rustfs/src/storage/access.rs:880-897. Both are MinIO-consistent and intentional, but they must be visible in the PR description, not only in test assertions.

Non-blocking: the OPA Failed-state e2e coverage regression noted inline, and reusing `ApiError::access_denied()` for the new AccessDenied constructions after rebase.

CI on the current (stale) base is fully green, including the new list_buckets_iam_filter e2e; `cargo check -p rustfs-iam` passes locally on this branch.

PR #5751, opened 2026-08-05
fix(site-replication): translate policy mapping userType at MinIO wire boundary

reatang:
## Related Issues
Related to rustfs/backlog#1675 (MinIO replication compatibility review, finding P0-4).

## Summary of Changes
Policy-mapping `userType` wire values diverged from MinIO: RustFS internal encoding is None=0/Svc=1/Sts=2/Reg=3 while MinIO uses unknown=-1/reg=0/sts=1/svc=2. Consequences: MinIO group mappings (sent as -1) failed u64 deserialization and the whole IAM item was rejected, so group policy mappings never replicated from MinIO; MinIO STS mappings (1) were misfiled as service-account mappings, silently dropping federated users' permissions; a low-probability privilege-escalation existed via name collision with the misfiled cache. This PR changes `SRPolicyMapping.user_type` (and `SRCredInfo.iam_user_type`) to i64 and introduces a dedicated wire codec at the site-replication boundary: outbound Reg->0/Sts->1/Svc->2 with groups always 0; inbound -1->None/0->Reg/1->Sts/2->Svc plus the permanent legacy alias 3->Reg; unknown values fail closed. The internal `UserType::to_u64/from_u64` encoding used by intra-cluster node RPC is deliberately untouched to keep rolling restarts safe.

## Verification
- `cargo test -p rustfs-madmin --lib` (152), `cargo test -p rustfs-iam --lib` (235), `cargo test -p rustfs --lib -- site_replication` (186), `-- policies policy` (92) — all passed
- `cargo clippy -p rustfs-madmin -p rustfs-iam --all-targets`, `cargo clippy -p rustfs --lib`
- `cargo fmt --all --check`

## Impact
Compatibility matrix verified in all four directions: MinIO<->new RustFS (-1/0/1/2), old RustFS->new RustFS (3 alias, group 0), new RustFS->old RustFS (0 lands in the same users path as legacy Reg=3). Literal-value contract tests pin the MinIO table and forbid "unifying" back to the internal encoding.

## Additional Notes
`SRCredInfo.iam_user_type` currently has no consumer; the codec must be reused when credential-change replication is implemented.

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -63,7 +63,7 @@ use rustfs_config::{
 };
 use rustfs_iam::error::is_err_no_such_service_account;
 use rustfs_iam::federation::OIDC_VIRTUAL_PARENT_CLAIM;
-use rustfs_iam::store::{MappedPolicy, UserType};
+use rustfs_iam::store::{MappedPolicy, UserType, sr_wire_user_type, user_type_from_sr_wire};
 use rustfs_iam::sys::{
     NewServiceAccountOpts, SITE_REPLICATOR_SERVICE_ACCOUNT, UpdateServiceAccountOpts, get_claims_from_token_with_secret,
 };
@@ -4303,7 +4303,7 @@ fn local_idp_settings() -> IDPSettings {
 fn mapped_policy_to_sr_mapping(name: String, is_group: bool, user_type: UserType, mapping: MappedPolicy) -> SRPolicyMapping {
     SRPolicyMapping {
         user_or_group: name,
-        user_type: user_type.to_u64(),
+        user_type: sr_wire_user_type(user_type, is_group),
         is_group,
         policy: mapping.policies,
         updated_at: Some(mapping.update_at),
@@ -7824,7 +7824,8 @@ async fn apply_iam_item(item: SRIAMItem) -> S3Result<()> {
             let Some(mapping) = item.policy_mapping else {
                 return Err(s3_error!(InvalidRequest, "policyMapping is required"));
             };
-            let user_type = UserType::from_u64(mapping.user_type).ok_or_else(|| s3_error!(InvalidRequest, "invalid userType"))?;
+            let user_type =
+                user_type_from_sr_wire(mapping.user_type).ok_or_else(|| s3_error!(InvalidRequest, "invalid userType"))?;
             iam_sys
                 .policy_db_set(&mapping.user_or_group, user_type, mapping.is_group, &mapping.policy)
                 .await
@@ -11910,7 +11911,7 @@ mod tests {
             "alice".to_string(),
             SRPolicyMapping {
                 user_or_group: "alice".to_string(),
-                user_type: UserType::Reg.to_u64(),
+                user_type: sr_wire_user_type(UserType::Reg, false),
                 policy: "readwrite".to_string(),
                 updated_at: Some(OffsetDateTime::UNIX_EPOCH),
                 ..Default::default()
--- a/crates/iam/src/store.rs
+++ b/crates/iam/src/store.rs
@@ -139,6 +139,59 @@ impl UserType {
     }
 }
 
+/// Encode a [`UserType`] as the site-replication wire value for
+/// `SRPolicyMapping.userType` / `SRCredInfo.iamUserType`.
+///
+/// The wire uses MinIO's `IAMUserType` table (cmd/iam.go):
+///
+/// | wire | MinIO meaning |
+/// |------|---------------|
+/// | -1   | unknown       |
+/// | 0    | regUser       |
+/// | 1    | stsUser       |
+/// | 2    | svcUser       |
+///
+/// This is deliberately distinct from the internal encoding
+/// [`UserType::to_u64`]/[`UserType::from_u64`] (None=0, Svc=1, Sts=2, Reg=3),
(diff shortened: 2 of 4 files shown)


rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -4303,7 +4303,7 @@ fn local_idp_settings() -> IDPSettings {
 fn mapped_policy_to_sr_mapping(name: String, is_group: bool, user_type: UserType, mapping: MappedPolicy) -> SRPolicyMapping {
     SRPolicyMapping {
         user_or_group: name,
-        user_type: user_type.to_u64(),
+        user_type: sr_wire_user_type(user_type, is_group),
```
overtrue, line 4306: Non-blocking, follow-up suggestion: the callers of this function in `get_sr_info` only export Reg user mappings (`load_mapped_policies(UserType::Reg, false, ...)`) plus group mappings, while MinIO's `SiteReplicationMetaInfo` also exports stsUser and svcUser mapped policies (`addPolicy(regUser/stsUser/svcUser, ...)` in cmd/site-replication.go). With this PR, MinIO->RustFS STS mappings replicate and are stored correctly, but they remain invisible to RustFS's own SRInfo, so cross-site status/heal cannot see or reconcile them. Worth a dedicated follow-up issue rather than this PR: fixing it also has to deal with pre-fix RustFS peers misdecoding 1/2 during rolling upgrades, which deserves its own compatibility analysis.

overtrue (changes requested):
The substance of this PR passes adversarial verification on every axis I checked; the only blocker is mechanical: the branch conflicts with current main and needs a rebase.

**What I verified (all good):**

- **MinIO wire table**: checked against upstream sources. `cmd/iam.go` defines `unknownIAMUserType = iota - 1`, so unknown=-1/reg=0/sts=1/svc=2 — exactly the table pinned by `sr_wire_user_type`/`user_type_from_sr_wire`. MinIO's group policy mappings are sent with `UserType: int(unknownIAMUserType)` (-1) in both the initial-sync and heal paths (`cmd/site-replication.go`), which confirms the root cause: serde rejects -1 for a `u64` field and `#[serde(default)]` does not rescue type errors, so the whole IAM item was dropped.
- **Group encoding choice (0, not -1)**: correct and deliberate. MinIO's `PeerPolicyMappingHandler` routes group mappings by `isGroup` and ignores userType for the storage path, so it accepts 0; and pre-fix RustFS peers have a `u64` wire field that cannot parse -1 at all. 0 is the one value every peer generation accepts.
- **Mixed-version matrix**: old RustFS emits only 3 (users and groups via the change-hook path) or 0 (groups via SRInfo), both decoded correctly by the new `3 -> Reg` permanent alias and `0 -> Reg`. New RustFS emits only 0 today (the only `mapped_policy_to_sr_mapping`/policies.rs callers pass Reg or None), which old peers decode to `None` and file under the same users/groups prefixes (`get_mapped_policy_path` routes `None` with the `_` arm). The theoretically dangerous new->old Sts=1 path is unreachable because RustFS has no Sts/Svc outbound emitter yet.
- **Boundary discipline**: the internal `to_u64`/`from_u64` node-RPC encoding is untouched; the only remaining `from_u64` call in the binary is `node_service.rs`, which is exactly where it belongs. No status/heal comparison includes userType on either side (RustFS compares `contains_key` presence; MinIO's `isPolicyMappingReplicated`/`isPolicyMappingEqual` compare IsGroup/Policy/UserOrGroup only), so the outbound 3->0 change cannot cause mismatch storms.
- **Fail-closed**: unknown wire values decode to `None` and `apply_iam_item` rejects with InvalidRequest. Good.
- **Tests**: the literal-value contract tests genuinely pin the MinIO table and the negative-value serde acceptance; I re-ran `cargo test -p rustfs-iam --lib store::tests` and `cargo test -p rustfs-madmin --lib site_replication` locally, all green. CI is fully green on the branch.

**Blocking (the reason for Request Changes):**

The branch is based on a pre-#5749 main and GitHub reports it as CONFLICTING; #5750 has just been squash-merged into main as well, so the conflict surface now includes both. The conflict is confined to `crates/madmin/src/site_replication.rs` (#5749 rewrote the SRInfo field attributes/tests area, #5750 added the `SR_IAM_ITEM_STS_ACC` constants at the top of the same file) and resolution is mechanical — your hunks (SRPolicyMapping/SRCredInfo i64 fields plus tests) do not semantically overlap with either. Please rebase onto latest main, resolve, and let CI re-run; after that this is ready to merge from my side.

reatang:
Rebased onto latest main (now includes #5749 and #5750). The conflict was confined to the test-module import line in `crates/madmin/src/site_replication.rs` as you predicted; resolution is the union of both imports, no semantic overlap touched.

Re-verified on the rebased branch: `cargo test -p rustfs-madmin --lib` 155 passed, `cargo test -p rustfs-iam --lib` 235 passed, `cargo test -p rustfs --lib -- site_replication` 189 passed (counts increased over the pre-rebase run precisely because the branch now carries #5749's PascalCase/nil-map fixtures and #5750's inbound alias tests — both suites stay green alongside this PR's i64/codec changes), `cargo fmt --all --check` clean.

Your non-blocking finding on `get_sr_info` only exporting Reg mappings is now tracked as rustfs/backlog#1773, including the rolling-upgrade caveat from the codec doc comment (enabling Sts/Svc outbound will mis-decode on pre-fix peers, so the export fix needs its own compatibility gate). Thanks for the thorough matrix verification — ready for re-review.

PR #5753, opened 2026-08-05
fix(replication): schedule replication for CopyObject and snowball extracted objects

reatang:
## Related Issues
Related to rustfs/backlog#1675 (MinIO replication compatibility review, finding P0-6).

## Summary of Changes
`execute_copy_object` wrote through `store.copy_object` with no replication decision, no PENDING stamp, and no scheduling, so objects created via CopyObject (cross-bucket copies, rename workflows, metadata REPLACE self-copies) never replicated synchronously; snowball auto-extract members had the same gap. Worse, replication-status metadata inherited from the source could fake a COMPLETED state on the destination and suppress even scanner-based backfill. This PR mirrors the PutObject contract (single decision drives both stamp and schedule): strip inherited replication-status metadata first (equivalent to MinIO `filterReplicationStatusMetadata`, only for non-inbound requests), compute `must_replicate_object` once, stamp PENDING before the write, schedule after success; the same wiring is applied per extracted snowball member. A secondary defect found and fixed along the way: authorized inbound replication PUTs are no longer re-dispatched as snowball extract on the target (members inherited the auto-extract header and replication went permanently FAILED).

## Verification
- White-box: `must_replicate_object` call-count tests for copy and extract (red-light: 0 calls before fix); 415-test nextest sweep over copy/replication/extract/snowball filters
- e2e (dual instance): `test_copy_object_replicates_to_target`, `test_snowball_extract_replicates_members_to_target`
- `cargo fmt --all --check`, `cargo clippy -p rustfs --lib --tests`, `clippy -p e2e_test`
- `make pre-commit` guards pass except a pre-existing s3s-footprint baseline drift on main (tracked separately; this diff adds zero `s3_error!` call sites)

## Impact
CopyObject and snowball-extracted objects now participate in replication with correct PENDING/COMPLETED progression; inbound replica writes and the existing authorization path are unchanged; REPLACE self-copies schedule replication, matching MinIO. Landing note: land after fix/p0-5. Review note: scanner backfill via existing-object resync can eventually cover empty-status objects when ExistingObjectReplication is enabled, but synchronous replication semantics were still broken and inherited-COMPLETED / ExistingObjectReplication=Disabled cases leaked permanently; e2e for the scanner-compensation boundary is tracked as follow-up.

## Additional Notes
N/A

### DIFF
--- a/crates/e2e_test/src/replication_extension_test.rs
+++ b/crates/e2e_test/src/replication_extension_test.rs
@@ -2435,6 +2435,107 @@ async fn build_replication_pair(
     Ok((source_env, target_env, source_bucket.to_string()))
 }
 
+/// P0-6: CopyObject creates a new object on the destination key, so it must be
+/// scheduled for bucket replication exactly like PutObject (MinIO
+/// CopyObjectHandler parity). Before the fix the copy path never consulted the
+/// replication config: the destination object stayed local forever (its status
+/// metadata was inherited wholesale from the source, so the scanner heal pass
+/// skipped it too — no PENDING/FAILED marker meant nothing to re-drive).
+#[tokio::test]
+#[serial]
+async fn test_copy_object_replicates_to_target() -> TestResult {
+    init_logging();
+
+    let (source_env, target_env, source_bucket) = build_replication_pair(true).await?;
+    let source_client = source_env.create_s3_client();
+    let target_client = target_env.create_s3_client();
+    let target_bucket = "replication-check-dst";
+
+    let src_key = "copy-repl-source.txt";
+    let dst_key = "copy-repl-destination.txt";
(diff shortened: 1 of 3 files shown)


overtrue:
Merged after re-running the sftp lane (the earlier 55m failure was a job-level hang, not a test assertion; the rerun passed in normal time). Two non-blocking follow-ups noted during review: (1) extracted snowball member objects persist x-amz-meta-snowball-auto-extract in user metadata, so replication replay toward an older RustFS target without this PR's receiver guard can still misroute member PUTs into the extract path — consider stripping that key when writing members (and checking what MinIO does); (2) the extract path does not strip replication-status metadata inherited from archive PUT headers the way the copy path does — impact is marginal but worth aligning.

PR #5758, opened 2026-08-05
feat(table-catalog): validate Iceberg metadata graphs

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Adds bounded, specification-aware validation for Iceberg table metadata before catalog state changes.

- Validates required Iceberg v1/v2 table metadata fields, identifier references, snapshot refs, sequence numbers, and supported update transitions.
- Reads GZIP-compressed metadata JSON using both Iceberg filename conventions with encoded and decoded size limits.
- Parses manifest lists and manifests with bounded Avro container, decompression, record-count, and aggregate commit budgets.
- Restricts online physical validation to new or changed snapshots and active registration heads, while charging shared manifest and file-reference budgets once per unique object.
- Validates manifest partition specs, content types, sequence inheritance, required v2 counts, warehouse ownership, and live data/delete object reachability while preserving upgraded v1 manifest compatibility and delete files stored under data paths.
- Applies the same bounded metadata and manifest readers to registration, import, rollback, external synchronization, maintenance planning, recovery, and both catalog backings.
- Adds focused coverage for malformed metadata, dangling references, compressed metadata, manifest limits, shared objects, relocated warehouses, deleted-file semantics, and v1-to-v2 compatibility.

## Verification

- `/root/.local/bin/rustfs-cargo-safe fmt --all -- --check`
- `CARGO_TARGET_DIR=/data/rustfs-target-5758 /root/.local/bin/rustfs-cargo-safe check -p rustfs --tests`
- `git diff --check`

## Impact

Malformed, oversized, cross-warehouse, or incomplete Iceberg metadata changes are rejected before catalog pointer updates without synchronously traversing unchanged table history. Valid GZIP metadata, shared manifests, delete files under data paths, embedded v2 manifests, and v1 manifests retained by upgraded v2 tables remain supported. No configuration migration is required.

## Additional Notes

The branch is rebased onto the current `main` and preserves the merged namespace conflict guards. Focused runtime tests were attempted twice, but the local 16 GiB host reached its enforced 45-minute safety timeout while compiling/linking the complete RustFS library test binary before test execution; runtime coverage is therefore left to CI.


rustfs/src/table_catalog/iceberg/validation.rs
```diff
@@ -191,29 +238,1119 @@ where
...
+    budget.charge_manifests(manifests.len())?;
+    manifests
+        .iter()
+        .map(|manifest| {
+            manifest
+                .as_str()
+                .filter(|manifest| !manifest.is_empty())
+                .map(|manifest| SnapshotGraphManifestLocation {
+                    manifest_path: manifest.to_string(),
+                    format_version: 1,
```
overtrue: **Blocking.** Embedded-`manifests` snapshots are hardcoded to `format_version: 1` with `sequence_number: None` here, and the caller in `snapshot_graph_manifest_references` then passes `manifest_sequence_number = 0` for the `format_version == 1` case. But the decoded manifest's format version is inferred from its writer schema, so a v2-schema manifest referenced via the embedded `manifests` field decodes as v2 with real sequence numbers, and `validate_snapshot_graph_data_file_reference` rejects it with "Iceberg v2 manifest entry sequence number exceeds its manifest sequence" because every entry sequence number exceeds the hardcoded bound of 0. This is exactly what CI shows in `row_level_conflict_allows_v1_manifest_snapshot` — a case this PR's description promises to keep supported. Fix: for embedded manifests, bound entry sequence numbers by `snapshot_sequence_number` instead of 0, and do not force v1 at the location level.

rustfs/src/table_catalog/iceberg/validation.rs
```diff
@@ -191,29 +238,1119 @@ where
...
+                .ok_or_else(|| TableCatalogStoreError::Invalid("snapshot manifest-list object is missing".to_string()))?;
+            budget.charge_avro_bytes(manifest_list_object.data.len())?;
+            let decoded_manifest_list = decode_manifest_list_avro_async(manifest_list_object.data).await?;
+            budget.charge_decoded_avro_bytes(decoded_manifest_list.decoded_size)?;
+            budget
+                .manifest_lists
+                .insert(manifest_list_key, decoded_manifest_list.references.clone());
+            decoded_manifest_list.references
+        };
+        budget.charge_manifests(references.len())?;
```
overtrue: **Blocking.** `charge_manifests` (and `charge_file_references` in the caller) is charged on every traversal, including when the manifest list was served from the `budget.manifest_lists` cache. Since the budget is shared across the whole `validate_table_snapshot_graph` call and manifest lists are heavily shared between snapshots, the effective limit is `snapshots x manifests-per-snapshot` against `TABLE_COMMIT_MAX_MANIFESTS = 10_000` and `snapshots x manifests x entries` against `TABLE_COMMIT_MAX_FILE_REFERENCES = 1_000_000`. A moderate table with 100 retained snapshots of ~200 manifests each is rejected with a 400 even though each individual snapshot is far below the nominal limits, and the constants are not configurable. Please charge budgets against deduplicated keys (the `budget.manifests` / `validated_live_objects` maps already exist for this) or make the budget per-snapshot — and note this interacts with the broader ask in the review body to only validate new/changed snapshots at commit time.

rustfs/src/table_catalog/iceberg/validation.rs
```diff
@@ -191,29 +238,1119 @@ where
...
+    if context.entry.table_bucket != context.table_bucket {
+        return Err(TableCatalogStoreError::Invalid("snapshot object is outside the table bucket".to_string()));
+    }
+    let object_key = table_catalog_object_key_from_location(context.table_bucket, location)
+        .ok_or_else(|| TableCatalogStoreError::Invalid("snapshot object location is invalid".to_string()))?;
+    let warehouse_object_prefix = table_warehouse_object_prefix(context.entry)?;
+    let object_kind =
+        table_maintenance_object_kind(context.namespace, context.table, Some(&warehouse_object_prefix), &object_key)
+            .ok_or_else(|| TableCatalogStoreError::Invalid("snapshot object is outside the table warehouse".to_string()))?;
+    if object_kind != expected_kind {
```
overtrue: **Blocking.** This kind check, combined with the classifier in `table_maintenance_object_kind` (recovery.rs), requires `DeleteFile` objects to live under `<warehouse>/delete/`. The Iceberg spec has no delete directory: position and equality delete files are written by the same LocationProvider as data files, i.e. under the data path (`<table-location>/data/` by default). So any real v2 table with row-level deletes written by Spark, Flink, or PyIceberg classifies its delete files as `DataFile`, hits `object_kind != expected_kind`, and every commit/register fails with "snapshot object kind does not match manifest metadata". Similarly, tables using a custom `write.data.path`/`write.metadata.path` outside the table location are rejected as "outside the table warehouse". Fix: accept delete files under the data prefix (the manifest `content` field is the authority on data-vs-delete, not the directory), and consider making the warehouse-containment check configurable or advisory.

overtrue (changes requested):
Thanks — the untrusted-bytes handling in this draft is solid and worth keeping: size prechecks plus `take(max+1)` streaming reads, the hand-rolled Avro container pre-scan that bounds decompressed size before `apache_avro::Reader` inflates anything, checked arithmetic throughout the varint/zigzag decoding, `spawn_blocking` for all parsing, gzip metadata support under both filename conventions, fail-safe retention of Invalid metadata on the GC path, and dropping the over-strict "manifest-list must be non-empty" check (empty snapshots are legal). Direction is partially endorsed on that basis, but the validation scope and several compatibility assumptions need adjustment before this can land.

## Direction: full-graph physical validation on every commit is over-reach

`validate_table_snapshot_graph` walks every snapshot in the metadata, reads every manifest list and manifest, and stats every live data file — on every commit, register, import, and external sync. An Iceberg REST catalog's job at commit time is to validate requirements and metadata shape, not full physical reachability of the historical graph. As written, a single commit request can read up to 512MB of Avro and issue up to 1M existence checks synchronously in the request path, and the cost repeats on every commit. Please scope commit-time validation to the snapshots that are new or changed relative to the current metadata (diff against the current pointer), and leave full-graph reachability to the maintenance/recovery paths where it already conceptually belongs.

## Blocking

1. **Embedded-manifests snapshots are hardcoded to v1/seq=0 and reject a case this PR explicitly promises to support** — see inline comment at validation.rs:1111. This is confirmed by CI: `row_level_conflict_allows_v1_manifest_snapshot` fails with "Iceberg v2 manifest entry sequence number exceeds its manifest sequence".
2. **Budgets are charged per traversal instead of per unique object, so legitimate tables get rejected** — see inline comment at validation.rs:1076.
3. **Delete files are required to live under a `delete/` directory that the Iceberg spec does not define** — see inline comment at validation.rs:1350. Any real v2 table with row-level deletes written by Spark/Flink/PyIceberg will fail commit/register.
4. **Stacked base is missing the final commit of #5745 and a careless rebase would silently revert a merged guard.** The merged #5745 squash (7211f2949) includes a fifth commit, "fix(table-catalog): reject implicit namespace creates", which added the `has_active_namespace_object`/`has_active_namespace_descendant` Conflict guard in `store/object.rs` and the matching guard in `store/strong.rs`, plus Conflict assertions in `tests.rs`. This branch's base predates that commit, so the branch content deletes that guard and its tests relative to main. When you rebase onto main, the rebase must keep main's version of those guards — please double-check the create_namespace paths and the removed `assert_matches!(..., Err(TableCatalogStoreError::Conflict(_)))` assertions after rebasing.
5. **CI is red with 8 failing unit tests in the touched area** (all three Test and Lint variants), including one of this PR's own new tests:
   - `admin::handlers::table_catalog::tests::row_level_conflict_allows_v1_manifest_snapshot`
   - `admin::handlers::table_catalog::tests::metadata_location_api_validates_relocated_snapshot_graph_under_target_warehouse` (new in this PR; the entry's warehouse_location is not adopted from the relocated metadata location)
   - `admin::handlers::table_catalog::tests::table_location_updates_must_stay_inside_bucket`
   - `admin::handlers::table_catalog::tests::row_level_conflict_rejects_delete_of_non_current_file`
   - `admin::handlers::table_catalog::tests::row_level_conflict_rejects_changed_inherited_manifest_identity`
   - `admin::handlers::table_catalog::tests::row_level_conflict_rejects_manifest_outside_table_warehouse`
   - `admin::handlers::table_catalog::tests::row_level_conflict_rejects_missing_manifest_before_pointer_update`
   - `admin::handlers::table_catalog::tests::table_metadata_maintenance_helper_commits_snapshot_expiration`

## Non-blocking suggestions

- **Avro codec whitelist (null/deflate only)**: this covers the Java default (gzip maps to deflate) and PyIceberg, but tables configured with `write.avro.compression-codec=snappy` or `zstd` will be rejected. At minimum surface this as `Unsupported` rather than `Invalid`; ideally add snappy.
- **v2 snapshot `summary.operation` restricted to append/replace/overwrite/delete**: Java's `SnapshotParser` only requires the field to be present and does not validate the value set, so this is stricter than the reference implementation.
- **`synchronize_table_metadata_version_fields` strips dual-write fields from v1 metadata** (removes `schemas`, `current-schema-id`, `partition-specs`, `sort-orders`, etc.). The spec recommends v1 writers keep both forms for forward compatibility; actively discarding them degrades interop for readers that prefer the v2-style fields.
- **Manifest-list detection via the `snap-` filename prefix** is a convention, not spec; worth a comment noting nonconforming names get classified as manifest files and rejected.
- **Deflate blocks are decompressed twice** (once in the pre-scan to a sink, once by the real reader) — up to 2x 512MB of CPU per commit with no cross-request concurrency cap; consider a global semaphore.
- **`create_table` reads `format-version` from properties but leaves it in the stored properties**; most catalog implementations strip it.
- **Graph validation is wired only at the handler layer**; `store.commit_table` itself does not enforce it. REST is the only entry today, but the invariant deserves a doc comment on the trait.

Also note the current diff-vs-main appears to touch `site_replication`, `error.rs`, and `check_s3s_footprint.sh` — that is purely the stale base (#5748/#5749/#5755 landed after the merge-base) and will disappear on rebase, but it underlines that the rebase should happen before further review rounds.

PR #5763, opened 2026-08-06
fix(logging): bound hot-path span amplification

anthonymartin:
## Related Issues

Follow-up to #5716, #5719, and #5727.

## Why

A production five-minute sample emitted 2,983 lines, of which 2,508 (84%) were tracing span-close records. Normal S3 reads fan out through multiple thin wrappers, and the pinned `s3s` authentication path emits several INFO spans per request. Healthy readiness polling was also still logged at INFO because request events used the module target `rustfs::server::layer`, while the production filter targeted `rustfs::server::http`.

## Invariants

- Normal successful object reads and authentication do not multiply into INFO records at every internal wrapper.
- WARN/ERROR authentication and HTTP failures remain visible.
- Explicit `RUST_LOG` target directives restore diagnostics when requested.
- A strict or unrelated `RUST_LOG` never gets broadened by automatic defaults.
- Write, heal, rebalance, and admin spans are unchanged.
- No storage, locking, request, or response behavior changes.

## What changed

- Demote the measured namespace-lock, ListObjectsV2, and object-info wrapper spans to TRACE while preserving their fields for opt-in diagnostics.
- Cap `s3s` at WARN for normal non-verbose production configuration.
- Preserve explicit `s3s` target directives and verbose configurations.
- Avoid adding `s3s=warn` when the effective configuration is ERROR/OFF or names only unrelated targets.
- Give HTTP completion/failure events the stable `rustfs::server::http` target that the existing production filter actually controls.
- Add source guardrails for the narrowly enumerated hot spans and HTTP target.
- Add event-level filter tests, not just filter-string assertions.

## Verification

- `scripts/check_logging_guardrails.sh` — passed.
- `cargo test -p rustfs-obs telemetry::filter::tests --lib` — 16/16 passed.
- `cargo check --locked -p rustfs` — passed.

The event-level tests prove that:

- default INFO suppresses `s3s` INFO while retaining WARN;
- explicit `s3s=info` restores INFO diagnostics;
- HTTP success events are suppressed at the stable target while errors remain;
- ERROR and unrelated target-only filters do not accidentally enable `s3s`.

## Impact and rollback

This removes formatting/export work and log writes for redundant hot-path spans. It does not alter scanner cadence or legitimate data-disk I/O. Rollback is a single-commit revert.

### DIFF
--- a/rustfs/src/server/layer.rs
+++ b/rustfs/src/server/layer.rs
@@ -391,6 +391,7 @@ impl RequestLogContext {
 
         if status.is_server_error() {
             error!(
+                target: "rustfs::server::http",
                 event = HTTP_REQUEST_COMPLETED_EVENT,
                 component = LOG_COMPONENT_SERVER,
                 subsystem = LOG_SUBSYSTEM_HTTP,
@@ -407,6 +408,7 @@ impl RequestLogContext {
             );
         } else {
             info!(
+                target: "rustfs::server::http",
                 event = HTTP_REQUEST_COMPLETED_EVENT,
                 component = LOG_COMPONENT_SERVER,
                 subsystem = LOG_SUBSYSTEM_HTTP,
@@ -429,6 +431,7 @@ impl RequestLogContext {
         E: std::fmt::Display,
     {
         error!(
+            target: "rustfs::server::http",
             event = HTTP_REQUEST_FAILED_EVENT,
             component = LOG_COMPONENT_SERVER,
             subsystem = LOG_SUBSYSTEM_HTTP,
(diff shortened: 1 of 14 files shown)


rustfs/src/server/layer.rs
```diff
@@ -391,6 +391,7 @@ impl RequestLogContext {
 
         if status.is_server_error() {
             error!(
+                target: "rustfs::server::http",
```
houseme: It is recommended to unify the format and define constants.

line: 411 , 434

houseme (approved):

PR #5770, opened 2026-08-06
fix(capacity): back off timed-out scans

anthonymartin:
## Related Issues

N/A

## Summary of Changes

- apply exponential backoff to scheduled capacity refreshes that time out, fail, or return degraded results; the delay starts at the configured interval, doubles to a 30-minute ceiling, and resets after a clean scan
- preserve single-flight refresh behavior by joining an in-flight refresh before selecting the next delay
- distinguish timeout fallback estimates from ordinary successful sampling, so large complete scans retain the normal cadence
- choose overflow samples before calling `metadata()`, avoiding stats for unsampled entries while preserving the exact prefix, file count, sampling positions, and estimate calculation
- add table-driven backoff/reset coverage, a paused-time scheduler test, and scan tests that pin both estimate equivalence and the reduced metadata-read count

## Verification

- `cargo test -p rustfs-object-capacity` (79 passed)
- `cargo clippy --all-targets -p rustfs-object-capacity -- -D warnings`
- `make pre-commit`
- `cargo fmt --all --check`
- `git diff --check`

## Impact

Repeated capacity scans no longer retry at the base cadence after they have already exhausted their time budget. At the default configuration, consecutive unclean refreshes wait 2, 4, 8, 16, then 30 minutes; a clean refresh resets the delay to 2 minutes. Configured intervals above 30 minutes are not shortened.

For a one-million-file scan with the default 200,000-file exact prefix and 1-in-200 overflow sampling, metadata reads drop from 1,000,000 to 204,000. Capacity values and sampling positions remain unchanged for stable filesystems.

`CapacityUpdate` gains a `timed_out` field so successful fallback estimates remain distinguishable from complete sampled scans. There are no configuration, persistence, or wire-format changes.

## Additional Notes

The existing `RUSTFS_CAPACITY_SCHEDULED_INTERVAL` remains the base cadence. This PR does not add an idle/write gate and does not change the timeout or sampling configuration defaults.


houseme:
Thanks for the capacity backoff work. The general direction makes sense, but I do not think this is ready to merge yet.

- In `crates/object-capacity/src/scan.rs`, the new overflow path decides whether an entry is sampled before calling `metadata()`. That means metadata failures on unsampled overflow entries are now completely skipped and no longer set `had_partial_errors`. A scan with inaccessible or racing files in the overflow region can therefore look clean, update the per-disk cache, clear dirty disks, and reset the scheduled backoff even though part of the tree was not actually readable. Please preserve the previous partial-error semantics, or add a test that demonstrates this behavior is intentionally safe.
- The PR checks are not green: `ILM Integration (serial)` failed, and the regular `End-to-End Tests` job was cancelled after the 30-minute limit. Please rerun or fix those before merge.

Also, this branch is still based on `efd5481b...` while `main` is now `e26b869...`, so please rebase/merge current `main` before the next review pass.

PR #5779, opened 2026-08-06
fix(table-catalog): harden commit publication

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Hardens Iceberg table commit publication across REST handlers and both catalog backings.

- Makes standard REST commit retries deterministic by reusing client commit IDs or idempotency keys and rebuilding generated metadata with the persisted commit timestamp.
- Authorizes exact metadata, manifest, and data object keys with the corresponding S3 actions before storage access, including table-ref PUT and DELETE commit routes.
- Collects referenced object keys before publication, sorts and deduplicates them under one global `(bucket, key)` order, then reacquires read guards and revalidates object state before catalog pointer publication.
- Verifies a canonical metadata digest again inside the catalog store and releases retained guards immediately after the catalog commit completes.
- Recovers historically committed staged records without moving a newer table pointer, while rejecting failed records and conflicting idempotency indexes.
- Keeps existing persisted commit records replay-compatible and applies the same publication checks to object-backed and durable-strong catalog stores.
- Adds regression coverage for exact object authorization failures, table-ref route guarding, reverse-order lock contention, post-lock object revalidation, historical recovery, metadata replacement, and cross-backing behavior.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- Focused regression tests added to the RustFS test suite:
  - `commit_publication_denies_generated_metadata_write_before_pointer_advance`
  - `commit_publication_denies_referenced_data_read_before_pointer_advance`
  - `commit_publication_authorizes_referenced_objects`
  - `commit_publication_holds_referenced_object_locks_until_pointer_publish`
  - `commit_publication_acquires_discovered_object_locks_in_key_order`
  - `commit_publication_revalidates_objects_after_ordered_lock_acquisition`
  - `table_ref_write_handlers_install_commit_publication_guard`
  - `table_ref_write_responses_use_commit_guard_and_protect_deletes`
  - `commit_publication_replays_historical_standard_commit_across_backings`
  - `catalog_backings_reject_metadata_changed_after_validation`
  - `object_table_catalog_store_recovers_historical_staged_commit_after_later_commit`
  - `table_commit_recovery_does_not_overwrite_conflicting_idempotency_index`

Runtime execution of the focused tests is covered by the pull request CI test matrix.

## Impact

Table commits now fail before pointer advancement when the caller lacks access to a generated or referenced object, when a validated object changes before publication, or when persisted idempotency state conflicts. Table-ref write routes use the same guarded publication path as generic commits. Existing REST response shapes and catalog configuration remain unchanged.

## Additional Notes

High-risk verification coverage:

- Authorization: exact-key GetObject and PutObject checks are exercised on generic and table-ref commit surfaces, including a denial case that confirms metadata location, version token, and generation remain unchanged.
- Concurrency and durability: referenced keys are acquired in deterministic order, revalidated while guards are held, and retained through catalog publication; reverse discovery order and object replacement are covered by regression tests.
- Recovery: staged, historically committed, failed, and conflicting-idempotency states are covered without allowing recovery to move a newer table pointer.
- Compatibility: object-backed and durable-strong stores share the same request matching, digest, replay, and conflict expectations, with no REST wire-shape change.


houseme (changes requested):
This change spans authorization, read-lock lifetime, durable commit records, idempotency recovery, and two catalog backings. cargo check is not sufficient validation for those behaviors. Please add or run focused commit, recovery, and authorization tests plus the required high-risk verification before re-review.

rustfs/src/admin/handlers/table_catalog/mod.rs
```diff
@@ -1098,34 +1100,310 @@ async fn authorize_table_catalog_resource_request(
...
+        Ok(())
+    }
+
+    async fn retain_read_lock(&self, bucket: &str, object: &str) -> crate::table_catalog::TableCatalogStoreResult<()> {
+        let key = (bucket.to_string(), object.to_string());
+        let mut publication = self.publication.lock().await;
+        if publication.locked_objects.contains(&key) {
+            return Ok(());
+        }
+        let guard = self.backend.acquire_read_lock(bucket, object).await?;
```
overtrue: These read locks are retained in discovery order, but DeleteObjects acquires the same object write locks in sorted order. A commit that reads z then a can hold z while waiting for a, while a concurrent delete holds a and waits for z, so both time out. Please collect, sort, and deduplicate the keys under one global order, then revalidate after acquiring the guards, with a reverse-order contention test.

rustfs/src/admin/handlers/table_catalog/table.rs
```diff
@@ -106,17 +106,20 @@ pub struct RestCommitTableHandler {}
...
-        authorize_table_catalog_resource_request(&req, &resource, AdminAction::CommitTableAction).await?;
+        let principal = authorize_table_catalog_resource_request(&req, &resource, AdminAction::CommitTableAction).await?;
+        install_table_catalog_s3_request_info(&mut req, &principal)?;
         ensure_table_bucket_enabled(&warehouse).await?;
-        let request = read_json_body::<RestCommitTableRequest>(req.input).await?;
+        let request = read_json_body::<RestCommitTableRequest>(std::mem::take(&mut req.input)).await?;
         let metadata_backend = table_catalog_backend()?;
         let store = table_catalog_store_from_backend(metadata_backend.clone())?;
-        let response = commit_table_response(&store, &metadata_backend, &warehouse, &namespace, &table, request).await?;
+        let commit_backend = TableCommitObjectBackend::for_request(metadata_backend, req);
```
overtrue, line 124: This wrapper is only installed for the generic commit and metadata-location handlers. PutTableRefHandler and DeleteTableRefHandler also reach standard_commit_table_response with the raw backend, so they can read and write generated metadata and publish the same table pointer without the exact-key S3 checks or retained locks added here. Please route every commit surface through the guarded backend and add an alternate-route authorization test.

overtrue (changes requested):
The new publication guard is bypassed by table-ref commits and has a cross-operation lock-order deadlock.

rustfs/src/admin/handlers/table_catalog/table.rs
```diff
@@ -159,18 +162,22 @@ pub struct UpdateTableMetadataLocationHandler {}
...
+            authorize_table_catalog_resource_request(&req, &resource, AdminAction::SetTableMetadataLocationAction).await?;
+        install_table_catalog_s3_request_info(&mut req, &principal)?;
         ensure_table_bucket_enabled(&warehouse).await?;
-        let request = read_json_body::<UpdateTableMetadataLocationRequest>(req.input).await?;
+        let request = read_json_body::<UpdateTableMetadataLocationRequest>(std::mem::take(&mut req.input)).await?;
         let metadata_backend = table_catalog_backend()?;
         let store = table_catalog_store_from_backend(metadata_backend.clone())?;
-        let response =
-            update_table_metadata_location_response(&store, &metadata_backend, &warehouse, &namespace, &table, request).await?;
+        let commit_backend = TableCommitObjectBackend::for_request(metadata_backend, req);
```
overtrue, line 181: The guarded backend is still absent from the other pointer-publishing paths: rollback passes the raw backend to rollback_table_response, and external sync validates with the raw backend before calling store.commit_table. A caller allowed the catalog action but denied s3:GetObject can therefore publish referenced objects through either route without the exact-key checks or publication locks added here. Please route both through the guarded request backend and cover denied reads on those handlers.

rustfs/src/admin/handlers/table_catalog/mod.rs
```diff
@@ -1098,34 +1100,509 @@ async fn authorize_table_catalog_resource_request(
...
+                        "table commit publication has already completed".to_string(),
+                    ));
+                }
+            }
+        };
+
+        let prepared = async {
+            let mut guards = Vec::with_capacity(expected.len());
+            for (bucket, object) in expected.keys() {
+                guards.push(self.backend.acquire_read_lock(bucket, object).await?);
```
overtrue: Accepted limits allow up to 1,000,000 live file references and 512 MiB of Avro. This loop retains one distributed read guard per observed key, and observed_objects.clone above deep-copies all retained bodies before re-reading them. A single valid request can create around a million lock RPCs and guards plus over 1 GiB of object buffers. Please bound the publication state and locks, and cover the limit behavior.

rustfs/src/admin/handlers/table_catalog/mod.rs
```diff
@@ -1098,34 +1100,509 @@ async fn authorize_table_catalog_resource_request(
...
+                        max_size,
+                    } => {
+                        let actual = match max_size {
+                            Some(max_size) => self.backend.read_object_unlocked_limited(bucket, object, *max_size).await?,
+                            None => self.backend.read_object_unlocked(bucket, object).await?,
+                        };
+                        actual == *expected
+                    }
+                    TableCommitObservedObject::Existence(expected) => {
+                        self.backend.object_exists_unlocked(bucket, object).await? == *expected
```
overtrue: object_exists observations record only true for live data files. A concurrent PUT can replace or delete-and-recreate a referenced key after discovery but before this lock; this recheck is still true, so the catalog can publish a manifest describing different bytes. Please compare a stable fingerprint or etag under the lock and add an overwrite/recreate test.

rustfs/src/admin/handlers/table_catalog/mod.rs
```diff
@@ -4243,6 +4744,23 @@ where
...
+    request: crate::table_catalog::TableCommitRequest,
+) -> S3Result<crate::table_catalog::TableCommitResult>
+where
+    S: crate::table_catalog::TableCatalogStore + ?Sized,
+{
+    metadata_backend
+        .prepare_table_commit_publication()
+        .await
+        .map_err(catalog_store_error)?;
+    let result = store.commit_table(request).await;
```
overtrue: prepare_table_commit_publication returns with warehouse object read guards held, and this call then waits for the catalog table-entry write lock. Maintenance deletion takes that table lock first and then waits for object write locks. If cleanup targets an object referenced by the pending commit, the two operations form a lock-order cycle and time out. Please establish one lock hierarchy and add this cross-operation test.

rustfs/src/admin/handlers/table_catalog/mod.rs
```diff
@@ -4342,7 +4935,8 @@ where
         &target_metadata,
     )
     .await?;
-    let result = store.commit_table(request).await.map_err(catalog_store_error)?;
+    request.requirements.push(metadata_digest_requirement(&target_metadata)?);
```
overtrue: requirements is persisted in the commit log, and older nodes compare it exactly in commit_log_matches_request. A legacy metadata-location commit accepted here stores this private digest; if its response is lost and the retry lands on an older node during a rolling upgrade, that node reconstructs only the client requirements and returns a conflict even though the pointer already advanced. Please keep the internal fingerprint outside the client requirement vector or add mixed-version replay coverage.

rustfs/src/admin/handlers/table_catalog/mod.rs
```diff
@@ -4427,13 +5063,198 @@ where
...
+    if operation != commit.operation {
+        return Err(iceberg_rest_error(
+            ICEBERG_ERROR_COMMIT_FAILED,
+            StatusCode::CONFLICT,
+            "commit retry operation does not match the original commit",
+        ));
+    }
+    let requirements = replay_commit_requirements(&commit, &request.requirements, &target_metadata)?;
+    let committed_metadata_location = commit.new_metadata_location.clone();
+    let result = publish_table_commit(
```
overtrue, line 5451: A retry of a staged pre-CAS commit reaches this call without rerunning snapshot-graph validation. If the first attempt writes the staged log and then the table CAS fails, a manifest or data file can be deleted or access can be revoked before retry; only the metadata JSON objects are observed here, so this call can advance the pointer to a dangling snapshot and skip exact-key reauthorization. Please rerun graph validation when the current table still matches the staged base, and cover deletion or revocation between attempts.

overtrue (changes requested):
The ordered relock and ref-route coverage fix the previous findings, but six blocking gaps remain in authorization coverage, retry validation and compatibility, object identity, lock ordering, and resource bounds.

rustfs/src/table_catalog/store/object.rs
```diff
@@ -3773,6 +3795,33 @@ where
             None => None,
         };
 
+        if let Some(existing) = existing_idempotency_commit.as_ref()
+            && !commit_log_matches_request(existing, &request, &current.table_id)
```
overtrue, line 3960: commit_log_matches_request does not compare new_version_token, so an idempotency index that differs only in that generated token passes this check. The committed or historical retry path then overwrites it via finalize_commit_log, even though recovery classifies the same payload mismatch as conflicting. Please compare the full recovery payload and test a token-only conflict in both backings.

rustfs/src/table_catalog/iceberg/commit.rs
```diff
@@ -39,6 +41,44 @@ pub(crate) fn table_matches_staged_base(table: &TableEntry, commit_log: &CommitL
...
+    if table.table_id != target.table_id || matches!(target.status, CommitLogStatus::Failed) {
+        return false;
+    }
+    let target_state = (target.new_metadata_location.as_str(), target.new_version_token.as_str());
+    let mut state = (table.metadata_location.as_str(), table.version_token.as_str());
+    if state == target_state {
+        return true;
+    }
+
+    let mut by_new_state = BTreeMap::<(&str, &str), Option<&CommitLogEntry>>::new();
```
overtrue: This rebuilds the full BTreeMap for one target, but recovery, reporting, and migration call the helper once for every commit. A table with N commit logs now does O(N² log N) map work and allocations, while the strong backing holds its state lock. Please build the history index once per table and add a scale test.

overtrue (changes requested):
There are still blocking publication, compatibility, recovery, and scaling gaps in this head.

rustfs/src/admin/handlers/table_catalog/mod.rs
overtrue: The new 10,000-object cap avoids unbounded guards by rejecting workloads the validator already accepts. A manifest with 10,001 distinct live files is within the 1,000,000-reference and Avro budgets but now fails in record_observation before publication. Please keep the established boundary coherent or use a publication scheme that scales to it, with an end-to-end boundary test.

rustfs/src/admin/handlers/table_catalog/table.rs
overtrue: Rollback and external sync are guarded now, but REST register and catalog import still use the raw backend at table.rs:66 and :223, validate object graphs without exact-key GetObject checks, and call register_table after releasing those reads. The same caller can still publish a replaced or inaccessible graph through those two routes. Please apply the guard there and cover denied reads plus replacement.

rustfs/src/admin/handlers/table_catalog/mod.rs
overtrue: The fingerprint still is not bound to the bytes returned by the production backend: read_object_with_options stats A, then opens a separate reader, but observed_content records A's ETag and mod_time. On a versioned bucket, PUT B between those calls and delete B's version after the read to restore A; validation parses B, prepare sees A and passes, then A is published. Please bind identity to the returned bytes and cover this ABA sequence.

rustfs/src/admin/handlers/table_catalog/mod.rs
overtrue: The fence fixes the hierarchy only when both nodes run this head. During a rolling upgrade, an old maintenance node still holds the catalog table lock before taking warehouse object write locks, while a new commit holds the publication and object locks and waits for that catalog lock at 4785. The cross-version pair can still deadlock until timeout. Please make the order compatible with old nodes or gate this path, and add a mixed-version lock test.

overtrue (changes requested):
This head only merges main, so the seven existing blockers remain: unguarded register/import paths, the 10k compatibility cap, the stat/read ABA, unfenced internal commits, mixed-version lock ordering, token-only idempotency conflicts, and quadratic history recovery.

rustfs/src/table_catalog/store/strong.rs
```diff
@@ -1328,12 +1410,26 @@ where
...
+        let publication = TableCommitLockPublication::new(&self.object_backend);
+        self.commit_table_with_publication(request, &publication).await
+    }
+
+    async fn commit_table_with_publication(
+        &self,
+        request: TableCommitRequest,
+        publication: &(dyn TableCommitPublication + Sync),
+    ) -> TableCatalogStoreResult<TableCommitResult> {
         let _write_guard = self.write_lock.lock().await;
```
overtrue: This global write lock remains held while publication.prepare serially acquires and revalidates up to 10,000 warehouse-object locks. A blocked late object therefore stalls unrelated strong-catalog writes for the lock timeout. Please keep cross-table progress independent and add a blocked-object concurrency test for the strong backing.

overtrue (changes requested):
The current head fixes several earlier publication and recovery gaps, but the 10,000-object compatibility cap and maintenance paths without guarded requirements remain, and the strong backing now holds its global write lock while waiting on warehouse-object locks.

overtrue (changes requested):
The strong-store lock wait is fixed, but the 10,000-object compatibility cap and unfenced maintenance publications still block this head.

rustfs/src/admin/handlers/table_catalog/mod.rs
overtrue: Fresh commits now have no store-side fingerprint. Snapshot expiration and compaction still validate through raw backends and then call commit_table with empty requirements, so objects can change before pointer publication. Please route both maintenance paths through the guarded publication flow.

rustfs/src/storage/access.rs
```diff
@@ -1410,6 +1526,9 @@ async fn authorize_table_data_plane_if_needed(
         })
         .await;
     if allowed {
+        if table_data_plane_content_mutation(action) {
+            retain_table_data_plane_publication_guard(req, &resource).await?;
```
overtrue: A second write permission never reaches this retained-key check until after it resolves the table again. The first Put/Delete authorization can hold this publication read guard while a commit takes the catalog table-entry write lock and waits for publication write; legal-hold, retention, or bypass authorization then waits for that catalog entry, completing the cycle. Please cache the resolved table before reading catalog again and cover this interleaving.

rustfs/src/admin/handlers/table_catalog/mod.rs
```diff
@@ -1098,34 +1101,542 @@ async fn authorize_table_catalog_resource_request(
...
+
+        let prepared = async {
+            let content_object_count = expected
+                .values()
+                .filter(|object| matches!(&object.identity, TableCommitObjectIdentity::ContentSha256(_)))
+                .count();
+            let mut guards = Vec::with_capacity(content_object_count.saturating_add(1));
+            guards.push(self.backend.acquire_write_lock(table_bucket, &publication_lock).await?);
+            for ((bucket, object), expected_object) in &expected {
+                if matches!(&expected_object.identity, TableCommitObjectIdentity::ContentSha256(_)) {
```
overtrue: Only content reads retain object guards here, so live data files now rely entirely on the new table fence. That fence is absent on older nodes, and register/import skips it because the table entry and index are not visible yet. Either path can replace or delete a live file after the metadata recheck below and before pointer publication. Please keep a compatible guard for these writers and cover rolling-upgrade plus first-publication interleavings.

overtrue (changes requested):
This head fixes the previous object-count and maintenance-publication gaps, but the new table fence has a lock-order deadlock and does not cover first publication or mixed-version writers.

overtrue (approved):
The publication fences, request-scoped resource reuse, and rolling-upgrade object guards close the outstanding correctness gaps.

PR #5648, opened 2026-08-02
fix: address rc.1 release blockers

overtrue:
## Related Issues

N/A

## Summary of Changes

- Harden bucket delete/recreate fencing, namespace locks, multipart cleanup, and Object Lock fail-closed behavior.
- Make replication target refresh, delete-marker replay, MRF persistence, resync, and target admission safer under retries, restarts, and config changes.
- Tighten storage RPC authorization, disk mutation guards, durability reporting, lifecycle recovery, and regression coverage.
- Scale adversarial validation roles by risk and impact instead of agent count.

## Verification

- cargo fmt --all --check
- git diff --check
- ./scripts/check_layer_dependencies.sh
- CARGO_BUILD_JOBS=2 cargo check -p rustfs-ecstore --tests -j 2
- CARGO_BUILD_JOBS=2 cargo check -p rustfs --tests -j 2
- CARGO_BUILD_JOBS=2 cargo nextest run -p rustfs-ecstore -E test(/mrf|replication_put_object_options/) -j 2 --no-fail-fast: 34 passed
- CARGO_BUILD_JOBS=2 cargo nextest run -p rustfs-ecstore --lib -j 2 --no-fail-fast: 3653 passed, 1 skipped
- CARGO_BUILD_JOBS=2 cargo nextest run -p rustfs -E test(/bucket_generation|extract_entry_pax|complete_multipart|restore/) -j 2 --no-fail-fast: 33 passed

## Impact

Core storage, replication, lifecycle, Object Lock, and admin durability paths are affected. Rolling mixed-version deployment is not supported for these fencing changes; upgrade all nodes together and drain old in-flight multipart work first.

## Additional Notes

This stays draft while broader release validation and deployment-level checks continue.

PR #5782, opened 2026-08-06
perf(ecstore): memoize bucket-incarnation fence validation under lifecycle read-lock coverage

overtrue:
## Why

The PUT commit fence introduced in #5648 validates the bucket incarnation with an uncached read on every PUT commit: a distributed metadata-transaction read lock plus an EC quorum read of the bucket metadata (`get_bucket_incarnation_id_from_disk`), on top of the per-PUT bucket lifecycle read lock. Under 64-concurrency 4KiB PUT load on a 4-node cluster this adds two quorum round-trips per PUT; the resulting lock-manager pressure produced a sustained storm of ~1,000 client-visible `Lock acquisition timeout` failures per 5-minute benchmark window, and an isolation experiment that swapped only this validation for a cached read eliminated the error storm entirely (full details and raw data in rustfs/backlog#1776).

A plain cached read is not a safe replacement on its own: peer bucket-metadata cache invalidation is best-effort and happens outside the lifecycle write lock, so a stale cache hit could admit a PUT into a recreated bucket — which is exactly why #5648 read from disk.

## What

Memoize the validation per node while lifecycle read-lock coverage is continuous. Bucket deletion/recreation requires the bucket lifecycle WRITE lock, so while at least one lifecycle read guard on this node has been held continuously, no write lock can have been granted anywhere in the cluster and the incarnation cannot have changed. The first fenced PUT in a coverage window performs exactly the pre-existing authoritative disk validation; PUTs whose guards overlap that window reuse its result. The memo is cleared when the node's last guard drops or any guard observes `is_lock_lost` (lost coverage means a write lock may have been granted), so the next PUT revalidates from disk. The memo's validity is derived purely from locks this process itself holds, so best-effort peer cache invalidation is irrelevant to its correctness.

Fence semantics are unchanged — only redundant re-validations under continuous coverage are elided. The multipart fence call sites keep the per-call disk validation for now (follow-up candidate).

Also right-sizes the s3s footprint ratchet baselines: -1 `s3_error!` line from this change, and +1 s3s-importing file inherited from #5763 (`crates/obs/src/telemetry/filter.rs`) which landed on main without the baseline bump.

## Measurements

4-node × 4-NVMe cluster, `warp put --concurrent=64 --duration=5m --obj.size=4KiB`, fresh bucket per run, uniform config:

| build | obj/s | client-visible errors |
|---|---|---|
| main @ efd5481b, fence as-is | 440.97 | 942 × lock-acquisition timeout |
| main @ efd5481b, validation via cached read (isolation experiment) | 428.23 | 0 |
| main @ 77f2b948c (base, same night) | 639.73 | 1000 (503s, unrelated mode — see backlog#1776 §3) |
| main @ 77f2b948c + this PR | 684.07 | 1000 (same unrelated 503 mode); **0 lock-timeout errors** |

The throughput delta at HEAD (+7%) is within the environment's noise band; the demonstrated win is the elimination of the fence's per-PUT quorum round-trips and its lock-timeout error class. A separate, larger PUT-throughput regression relative to 1.0.0-beta.12 remains under investigation in rustfs/backlog#1776 (bisected to two independent steps; not addressed here).

## Verification

- `cargo test -p rustfs-ecstore --lib store::` — 416 passed (includes 4 new registry unit tests: memo valid only while guards overlap, lost-lock clears memo, bucket isolation, no-guard memoize ignored)
- `cargo test -p rustfs-ecstore --lib set_disk::` — 553 passed
- `cargo clippy -p rustfs-ecstore --lib` — clean; `make pre-commit` — all checks passed
- Cluster deployment verified: health 200, 5-minute 64-concurrency PUT run with zero lock-timeout errors


overtrue:
**Measurement addendum (supersedes the PR body's throughput table)**: subsequent ABBA-paired benchmarking on an isolated cluster with per-run disk reformatting showed that the PUT throughput deltas quoted in this PR (and the lock-timeout error storms) were environmental measurement artifacts — the test environment swings 3.7x on the same binary within an hour, and the error storms only reproduce atop polluted on-disk state (details: rustfs/backlog#1776, final-verdict comment). Clean-state paired runs show beta.12 vs main within ±6%, and zero lock-timeout errors on either side.

Reframing this PR accordingly: it is a **cost/robustness optimization, not a regression fix** — it removes two distributed quorum round-trips (metadata-transaction read lock + EC quorum metadata read) per PUT commit under continuous lifecycle read-lock coverage, with fence semantics provably unchanged. That reduction is architecturally real regardless of benchmark weather, and matters most on high-latency internode links, but I no longer claim a measured throughput win. Happy to close if maintainers prefer to keep the simpler always-from-disk validation.

PR #5791, opened 2026-08-07
fix(s3): resume in-flight GET streams after rebalance relocation

deadspacewii:
## Related Issues

N/A

## Summary of Changes

Large-object GETs pin the pool and `FileInfo` at request start, but open multipart files lazily. If rebalance moves the object to another pool and deletes a not-yet-opened source part, the in-flight download used to abort. This PR resumes the remaining range from the relocated object while preserving the original response identity.

- Adds a bounded streaming resume state machine for typed relocation failures and premature EOF. Reopened metadata must match the original version, modification time, size, ETag, and transform state, so replacement or changed objects fail closed instead of splicing versions.
- Drops the failed reader before backoff, admission, and reopen. This releases its disk-read permit and namespace read guard, preventing self-starvation when the read pool is saturated or has a single permit.
- Constructs resume context and retry state lazily only for the disk-streaming path. Buffered, cached, and materialized GETs do not allocate an `Arc`, boxed reopen closure, or retry state.
- Replays only the SSE-C header trio and preserves header sensitivity so tracing cannot expose key material.
- Emits successful per-object resume diagnostics at `debug`, with a logging guard that prevents this event from returning to `info`.
- Adds deterministic production-path coverage for a multipart object relocated from pool 0 to pool 1 while `execute_get_object` is streaming. The test verifies the complete 72 MiB body byte-for-byte, distinct multipart boundaries, one reopen, and source removal after the target metadata is published.

No API, configuration, or on-disk/on-wire format changes are introduced.

## Verification

- `cargo nextest run -p rustfs -E 'test(get_object_resume) | test(get_object_streaming_reader)'` - 26/26 passed.
- Focused sole-permit regression - passed; proves byte-exact resume, exactly one reopen, and balanced permit release.
- Production two-pool relocation regression - passed repeatedly; exercises `execute_get_object` through the real storage path.
- `cargo fmt --all --check` - passed.
- `git diff --check` - passed.
- `./scripts/check_logging_guardrails.sh` - passed, including an explicit mutation check from `debug!` back to `info!`.
- `make pre-pr` - stopped at the requester's direction after repository guards, full-workspace Clippy, and script self-tests passed; the test phase was compiling targets and had reported no attributable failure when interrupted.
- High-risk adversarial validation - all seven roles completed with no remaining findings: correctness, simplicity, security, concurrency/durability, compatibility, performance, and test coverage.

## Impact

In-flight S3 GETs can transparently continue when rebalance relocates the object between pools. Deleted, overwritten, identity-mismatched, or unrecoverable objects retain fail-closed behavior and cannot produce a mixed-version response. The normal GET path avoids eager resume allocations; recovery remains bounded to three jittered attempts and the existing stall timeout.

## Additional Notes

The latest review findings were valid and are addressed by commit `51a0e5d08`:

- The old reader is now taken and dropped before reopen, removing disk-permit and read-guard overlap.
- A sole-primary-permit regression pins the admission deadlock scenario. `RUSTFS_OBJECT_DISK_DEGRADED_READ_CAP=0` mirrors the primary cap rather than creating a literal zero-capacity degraded pool, but the reported self-starvation still occurred whenever primary and degraded admission were saturated.
- A deterministic two-pool, four-disks-per-pool production test now covers successful relocation and body integrity.
- Resume context construction is lazy for the disk-streaming branch.
- Successful resume logging is demoted to `debug` and protected by the logging guard.

---


deadspacewii:

houseme (approved):

cxymds:
I completed a high-risk review of the final `0fdab84a8` diff. The resume direction is sound and the identity/SSE-C checks fail closed, but I found one blocking concurrency issue and three follow-ups.

## Blocking: the failed reader self-starves resume admission

`GetObjectStreamingReader::poll_resume` keeps `self.inner` alive until the replacement reader has fully reopened (`rustfs/src/app/object_usecase.rs:1557-1575`). The old stream therefore still owns its `DiskReadPermitReader` permit after a typed relocation error, while `GetObjectResumeContext::reopen` requests a second permit before opening the replacement (`rustfs/src/app/object_usecase.rs:2031-2042`).

This is not limited to an extreme theoretical storm. `RUSTFS_OBJECT_DISK_DEGRADED_READ_CAP` defaults to `0`, and the primary wait defaults to 5 seconds. With `primary_cap=1, degraded_cap=0`, one GET holding the only primary permit can hit a typed relocation error, wait five seconds for the same permit it still owns, receive `SlowDown`, repeat until the resume budget is exhausted, and fail the body. With a saturated normal primary pool, all affected streams have the same failure mode. This defeats transparent recovery exactly under the rebalance pressure this PR targets.

Please drop/release the already-failed reader at `begin` before admission/reopen (for example, make the inner reader takeable), then add a regression test with primary cap 1 and degraded cap 0 that proves byte-exact resume and balanced permit release. The same change also removes the supported `RUSTFS_OBJECT_LOCK_OPTIMIZATION_ENABLE=false` read-guard overlap described in the swap comment.

## Missing production success-path regression

`get_object_streaming_reader_resumes_after_relocation_error` exercises a synthetic reader/reopen closure, while `execute_get_object_resume_exhausts_budget_when_object_data_vanishes` exercises only the real-store failure path. `get_object_resume_reopen_rejects_a_replaced_object_version` validates a direct reopen, not an in-flight `execute_get_object` relocation. The PR body also records the deterministic multi-pool success test as a residual gap.

For this high-risk S3/rebalance behavior, please add a deterministic two-pool test that starts a GET, copies the same version to the target pool, removes the source before a later part opens, and asserts the complete body byte-for-byte through the production handler. Include the default/saturated admission case above; otherwise the advertised success behavior can regress while the current suite remains green.

## Non-blocking follow-ups

- `execute_get_object` unconditionally builds `GetObjectResumeContext`, clones `ObjectOptions` and object/header strings, allocates an `Arc` plus boxed closure, and initializes randomized retry state at `rustfs/src/app/object_usecase.rs:6398-6411` / `2076-2092`, including GETs that are ultimately served from a buffered or cache body. Please construct this lazily only for the disk streaming branch, or provide measurements justifying the per-GET overhead.
- A successful resume emits `info!` per object at `rustfs/src/app/object_usecase.rs:1653-1667`. RustFS logging governance reserves `info` for low-frequency lifecycle/mode transitions and puts repetitive success-path events at `debug`/`trace`; a rebalance storm can make this noisy. Please demote it and retain metrics for aggregate visibility.

I did not find an authz, on-wire/on-disk compatibility, mixed-version splice, or SSE-C secret exposure in this diff. The main blocker is the old permit/reader lifetime across reopen.

cxymds:
I reproduced and addressed all four findings in `51a0e5d08`.

- **Reader/permit self-starvation:** `GetObjectStreamingReader.inner` is now takeable, and the failed reader is dropped before backoff, admission, or reopen. This releases both `DiskReadPermitReader` admission and the namespace read guard. A sole-primary-permit regression proves byte-exact recovery, exactly one reopen, and balanced permit release. One nuance: `RUSTFS_OBJECT_DISK_DEGRADED_READ_CAP=0` mirrors the primary cap rather than constructing a literal zero-capacity degraded pool; the reported failure still exists when both admission classes are saturated, so the lifetime fix is required.
- **Production relocation coverage:** added an isolated two-pool/four-disks-per-pool test through `execute_get_object`. It starts a 72 MiB multipart GET from pool 0, publishes the same target metadata in pool 1, removes the later source parts and metadata, then verifies the complete response byte-for-byte with distinct per-part fill bytes and exactly one reopen.
- **Normal-path allocation:** resume context/control construction now uses stack-allocated `FnOnce` factories and occurs only when the disk streaming body is selected. Buffered, cache, and materialized paths do not allocate the resume `Arc`, boxed reopen closure, or retry state.
- **Logging:** the per-object successful resume event is now `debug!`; `scripts/check_logging_guardrails.sh` prevents it from returning to `info!`.

Focused verification is green: 26/26 resume/streaming tests, the sole-permit regression, the production two-pool test, formatting, diff checks, and the logging guard. All seven high-risk adversarial roles completed with no remaining findings. The full `make pre-pr` gate is still running on this head and I will report its final result separately.

cxymds (approved):

cxymds:
Local `make pre-pr` was stopped at the requester's direction after the repository guards, full-workspace Clippy, and script self-tests passed. The test phase was still compiling targets and had reported no attributable failure when interrupted. The focused 26-test resume suite and both new regressions remain green; task-owned build artifacts were cleaned afterward.

PR #5793, opened 2026-08-07
fix(admin): allow owner service account updates

### LINKED ISSUE #5778: from admin, if edit created keys on other users return 403 error
RUSTFS v1.0.0-beta.12

I have created user in users menu with added keys for access to storage, but if i first step to create key and setting up custom access policy - all right, it works, but if i create keys with default policy (inherit from base policy) and next I try to edit this keys and setup custom policy - rustfs return 403 error (access denied) or i just click on "edit" and next click "submit" - same error.

iam using admin right account.

Image

ccccpj:
## Related Issues

Fixes #5778.

## Summary of Changes

The update-service-account handler already required `admin:UpdateServiceAccount`, but it then unconditionally required the caller's effective parent to match the target service account's parent. As a result, the authenticated owner/root passed the admin-action check but still received `403 AccessDenied` when editing a service account owned by another user.

This change keeps the explicit admin-action check and target lookup in their existing order, then lets only the authentication layer's existing `owner` result bypass the parent-scope comparison. Possessing `consoleAdmin` or `admin:UpdateServiceAccount` alone does not grant cross-parent update access. That restriction matters because this endpoint can replace the secret key and session policy.

The real Admin API regression test covers this permission matrix:

| Caller | Target parent | Expected |
| --- | --- | --- |
| owner/root | another ordinary user | allow no-op and implied-to-custom policy updates |
| parent user with explicit update action | self | allow |
| ordinary non-owner | another user | deny |
| non-owner with explicit update action (including the relevant consoleAdmin capability) | another user | deny |
| service-account credential | another user | deny |
| STS temporary credential | another user | deny |
| owner/root | missing target | fail closed with `NoSuchResource` |

Denied takeover attempts are also verified not to mutate the target. A narrow unit regression verifies malformed IAM credential data maps to an internal error instead of being accepted.

## Verification

- Red-light regression proof on unpatched `origin/main`:  
  `cargo test -p e2e_test test_update_service_account_enforces_owner_and_parent_scope -- --nocapture` failed at the first owner/root cross-parent no-op update with expected `204`, actual `403 AccessDenied`.
- Patched real-handler E2E (final run):  
  `cargo test -p e2e_test test_update_service_account_enforces_owner_and_parent_scope -- --nocapture` — passed (1 passed).
- Targeted lookup-error unit tests:  
  `cargo test -p rustfs map_service_account_lookup_error_ -- --nocapture` — passed (2 passed).
- Formatting:  
  `cargo fmt --all --check` — passed.
- High-risk gate:  
  `make pre-pr` was executed. Formatting, all repository guard scripts, workspace Clippy with `-D warnings`, and script tests passed. The full nextest run completed with 12,676/12,677 passing; the sole failure was an unrelated host-DNS behavior where the reserved nonexistent hostname was synthesized to `198.18.0.19`/`fdfe:dcba:9876::1d` instead of returning a resolver error.
- Full no-fail-fast confirmation:  
  `cargo nextest run --all --exclude e2e_test --no-fail-fast` — 12,676 passed, only the same unrelated DNS-environment test failed; no other failures.
- Isolated DNS-environment confirmation in the network-isolated sandbox:  
  `cargo nextest run --all --exclude e2e_test -E 'test(resolve_domain_preserves_system_resolver_error_provenance)'` — passed (1 passed).
- Documentation tests:  
  `cargo test --all --doc` — passed.

Local builds emitted the existing macOS linker warning that the `__eh_frame` section is too large; it is unrelated to these IAM changes.

## Impact

Administrators authenticated as the existing owner/root identity can edit service accounts belonging to other users from the Console, including submitting an unchanged form or replacing an inherited policy with a custom session policy.

Non-owner behavior remains scoped to the caller's effective parent. This does not change Add, Delete, Info, request/response schemas, routes, persistence order, or site-replication hooks.

## Additional Notes

- **Correctness:** pass — action authorization, lookup, parent scope, mutation, and replication-hook ordering are preserved; owner/root cross-parent updates now reach the existing mutation path.
- **Simplicity:** pass — the production change is one short-circuit condition plus a security-boundary comment; helper additions exist only to drive the required API matrix.
- **Security:** pass — only the existing authenticated `owner` result bypasses parent matching; consoleAdmin/action-only users, ordinary users, service accounts, and temporary credentials remain denied across parents, and lookup/decode errors fail closed.
- **Concurrency/durability:** pass — no locks, awaits, persistence operations, or replication ordering changed.
- **Compatibility:** pass — no wire, schema, route, status, or non-owner behavior changes; the fix restores the existing owner cross-user service-account management contract.
- **Performance:** pass — one boolean short-circuit was added with no allocation, I/O, synchronization, or extra IAM lookup.
- **Test-coverage skeptic:** pass — reverting the production condition reproduces the issue through the real handler; the final E2E covers success, denial, mutation resistance, temporary/derived credentials, no-op, policy transition, and missing-target behavior.

### DIFF
--- a/crates/e2e_test/src/admin_iam_crud_test.rs
+++ b/crates/e2e_test/src/admin_iam_crud_test.rs
@@ -26,7 +26,9 @@
 //! Later batches tracked on backlog#1154: config get/set, info, pools status,
 //! group lifecycle, import/export IAM.
 
-use crate::common::{RustFSTestEnvironment, admin_ok, admin_request, init_logging};
+use crate::common::{
+    RustFSTestEnvironment, admin_ok, admin_request, admin_request_with_session_token, build_test_sts_client, init_logging,
+};
 use aws_sdk_s3::config::{Credentials, Region};
 use aws_sdk_s3::primitives::ByteStream;
 use aws_sdk_s3::{Client, Config};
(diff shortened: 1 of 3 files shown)


houseme (commented):
The owner and parent permission matrix is well covered. This PR is still draft and not merge-ready; please complete final checks and update the review state when it is ready.

houseme:
The reviewed implementation from this PR has been carried forward and consolidated in #5889: https://github.com/rustfs/rustfs/pull/5889

To keep a single active review and merge path, this PR is being closed in favor of #5889. Thank you for the contribution.

Issue #5837, opened 2026-08-08
Bitrot in `xl.meta` is never healed: CRC failure surfaces as a generic Io error, so the heal skips the drive and reports success


terem42:
## Summary

When the `xl.meta` file of an object is corrupted on disk (bitrot), no heal path repairs it — not the background scanner's deep scan, not an explicit admin deep heal — while **every heal reports success** (`result: ok`). The corrupted copies stay corrupted indefinitely, the object's effective metadata redundancy silently drops below the erasure guarantee, and the scanner re-submits the same no-op heal forever.

Payload bitrot (`part.1`) in the identical setup **is** detected and repaired (with #5822 applied), which isolates the failure to the metadata path and rules out setup problems: same cluster, same object, same heal request — the part case repairs in seconds, the `xl.meta` case never.

## Root cause

The `xl.meta` reader detects the corruption correctly — its CRC check fails — but raises the failure with the wrong `io::ErrorKind`, and the heal's disk-classification logic only recognises corruption by error type:

1. `FileMeta` decoding fails the CRC comparison and returns `Error::other("xl file crc check failed")` — i.e. `ErrorKind::Other` (`crates/filemeta/src/filemeta/codec.rs`, both CRC sites — lines 100 and 167 on current `main`).

2. The disk error conversion maps **only** `ErrorKind::InvalidData` to `DiskError::FileCorrupt` (`crates/ecstore/src/disk/error_conv.rs`). `ErrorKind::Other` falls through to a generic `DiskError::Io`. This is what `heal_object` actually receives per drive — captured from a live cluster with `RUST_LOG=rustfs_ecstore::set_disk=info` during a deep heal of an object with two dd-corrupted `xl.meta` copies:

   ```
   errs: [Some(Io(Custom { kind: Other, error: "xl file crc check failed" })),
          Some(Io(Custom { kind: Other, error: "io error xl file crc check failed" })),
          None, None, None, None, None, None, None, None, None, None]
   ```

   (first entry local, second via peer RPC — both drives holding corrupt `xl.meta`, all other ten clean.)

3. `should_heal_object_on_disk` (`crates/ecstore/src/set_disk/mod.rs`) marks a drive for metadata healing only for `FileNotFound`, `FileVersionNotFound` or `FileCorrupt`. Any **other** error returns `(false, false, err)` — the drive is skipped entirely:

   ```rust
   if let Some(err) = err
[...]

PR #5844, opened 2026-08-08
fix(multipart): recover part transactions by write quorum

### LINKED ISSUE #5839: UploadPart returns non-retryable InternalError "Faulty disk" while a node is down: the part-commit transaction scan ignores write quorum
## Summary

With one node of a 3-node / 12-drive EC:4 cluster hard-powered off — 4 of 12 drives lost, exactly the documented tolerance — every `UploadPart` request fails with `500 InternalError` and the message `"Faulty disk"`, while ordinary `PutObject` against the same endpoint in the same seconds succeeds without a single error. `InternalError` is not in any AWS SDK's default retryable set, so multipart clients see a permanent server fault and give up — for a degraded-but-fully-functional cluster state that every other operation routes around. Large uploads are precisely the ones that use multipart, i.e. the ones where a forced restart costs the most.

Measured twice, 25 hours apart, on two builds (stock `1.0.0-beta.12`, and beta.12 + #5822 + the #5837 fix), identical result. In the second run, during the 38-second failure window: 9 of 9 `UploadPart` requests failed (7 × `InternalError "Faulty disk"`, 1 × `ServiceUnavailable`, 1 × connection dropped at the instant of the power cut) while the background load completed 65 `PutObject`s with 0 errors — 8,170 requests, 0 errors, over the whole run. `CompleteMultipartUpload` succeeded against the same degraded cluster, assembling the parts that had landed.

## Root cause

Three pieces, all on `main` @ `778f1dfa2`:

**1. `UploadPart`'s commit runs the only non-quorum, fail-fast disk fanout in ecstore.** `put_object_part` survives the dead node at every quorum-reduced step (upload-id check, bitrot-writer setup, erasure encode with `write_quorum = 8`), then dies in `rename_part` → `recover_part_transaction` (`crates/ecstore/src/set_disk/core/io_primitives.rs:3216`), which reads part-transaction markers from **all 12 disks** and aborts on the first error:

```rust
let transaction_meta = match disk.read_all(RUSTFS_META_MULTIPART_BUCKET, &transaction_meta_path).await {
    Ok(meta) => Some(meta),
    Err(DiskError::FileNotFound) => None,
    Err(err) => return Err(err),          // <- FaultyDisk from a dead peer's disk escapes here
};
...
let observations = join_all(reads).await.into_iter().collect::<disk::error::Result<Vec<_>>>()?;
```

The `collect::<Result<Vec<_>>>()?` returns the first `Err`, so one unreachable disk out of twelve kills the request regardless of quorum. Every other per-disk fanout in the file folds errors into `Vec<Option<DiskError>>` and passes them through `reduce_*_quorum_errs`; this one is the exception. (The third read in the same closure — `current_meta` — already swallows every error, so the tolerant intent is visibly there; the first two arms are not.) A dead peer's disks produce `DiskError::FaultyDisk` via the remote-disk health gate (`crates/ecstore/src/cluster/rpc/remote_disk.rs`, `execute_with_timeout_for_op_and_health_action`) or `UnexpectedEof → FaultyDisk` in `crates/ecstore/src/disk/error_conv.rs` while the connection is first dying.

`rename_part` calls this scan both **before** the renames (`io_primitives.rs:3378`) and **after** they have succeeded (`io_primitives.rs:3491`) — so even a part durably written to 8/12 drives is answered with a 500.

**2. `PutObject` does not hit this because its commit primitive reduces by quorum.** Same disk set, same `write_quorum = 8`: `rename_data` folds per-disk errors and calls `reduce_write_quorum_errs(&errs, OBJECT_OP_IGNORED_ERRS, write_quorum)` (`io_primitives.rs:2753`), and `FaultyDisk` is in `OBJECT_OP_IGNORED_ERRS` (`crates/ecstore/src/disk/error_reduce.rs`) — 4 × FaultyDisk + 8 × Ok ⇒ success. The asymmetry between the two halves of the write API is exactly this one primitive. `CompleteMultipartUpload` survives for the same reason: its listing fanout in `recover_part_transactions` (plural) **is** quorum-reduced (`io_primitives.rs:3354`), and it only descends into the fail-fast singular scan when a transaction marker actually exists.

**3. The S3 error mapping has no arm for degraded-cluster disk errors.** `impl From<StorageError> for ApiError` (`rustfs/src/error.rs:263`) maps quorum and lock failures to retryable codes (`SlowDown`, `ServiceUnavailable`), but `StorageError::FaultyDisk` (and `FaultyRemoteDisk`, `DiskNotFound`, `TooManyOpenFiles`) fall through `_ => S3ErrorCode::InternalError`, and the `InternalError` branch then uses `err.to_string()` as the client-facing message — which is where the literal `"Faulty disk"` leaks out. The storage layer itself classifies these errors as ignorable/degraded (`OBJECT_OP_IGNORED_ERRS`; the retryable group in `crates/ecstore/src/config/com.rs`), so the S3 layer's mapping disagrees with ecstore's own taxonomy.

The one `ServiceUnavailable` we observed came from the namespace-lock path (`StorageError::NamespaceLockQuorumUnavailable → ServiceUnavailable`, `rustfs/src/error.rs:278`) — that request lost the lock quorum race before reaching `rename_part`. The correct mapping exists one match-arm away from the missing one.

The fail-fast scan arrived with the part-transaction mechanism in `b432f31c2` ("fix(multipart): preserve retried parts on quorum failure", #5363).

## Reproduction

3-node / 12-drive cluster, EC parity 4 (any geometry where a node loss stays within parity):

```sh
# 1. start a multipart upload with parts spaced a couple of seconds apart,
#    pointed at node1
# 2. mid-upload, hard-kill a DIFFERENT node (virsh destroy / power pull)
# 3. every subsequent UploadPart fails: 500 InternalError "Faulty disk"
#    (occasionally 503 ServiceUnavailable when the lock quorum race is lost first)
# 4. in parallel, ordinary PutObject through the same endpoint succeeds throughout
# 5. CompleteMultipartUpload with the parts that landed succeeds
```

Deterministic in shape; only the number of failed parts depends on where in the upload the cut lands.

## Expected behaviour

A single-node outage inside the erasure tolerance should be handled by `UploadPart` the way `PutObject` already handles it — served from the surviving quorum. Failing that, the error surfaced to the client must be retryable (`ServiceUnavailable`/`SlowDown`, 503), not `InternalError` with an internal disk-state string: SDKs retry the former by default and treat the latter as fatal.

## Suggested fix

1. **Primary:** in `recover_part_transaction` (`io_primitives.rs:3216`), fold the per-disk observations into `Vec<Option<DiskError>>` and apply `reduce_read_quorum_errs(&errs, OBJECT_OP_IGNORED_ERRS, read_quorum)` like every other fanout in the file — at minimum treat `OBJECT_OP_IGNORED_ERRS` members as absent markers, which is what the `current_meta` read in the same closure already does. The post-commit call at `io_primitives.rs:3491` needs the same treatment, or a part that was durably committed still 500s.
2. **Defence in depth:** add `StorageError::FaultyDisk | FaultyRemoteDisk | DiskNotFound | TooManyOpenFiles => S3ErrorCode::SlowDown` (or `ServiceUnavailable`) to the mapping in `rustfs/src/error.rs`, so no degraded-cluster condition can surface as a non-retryable 500 carrying an internal message.
3. Cosmetic: `put_object_part`'s writer setup (`crates/ecstore/src/set_disk/ops/multipart.rs:540`) could mirror `put_object`'s `disk.is_online().await` pre-filter (`ops/object.rs:956`).

Happy to submit a PR for (2) immediately; can attempt (1) as well if the quorum-reduction approach looks right to you.

## Impact

- Any node restart or crash fails every in-flight multipart upload cluster-wide with an error class clients will not retry — application-visible failures for a fault the cluster tolerates by design.
- The failure is worst for exactly the workloads multipart exists for: large uploads that are expensive to restart.
- `"Faulty disk"` as a client-facing message sends operators hunting for a broken drive when no drive is broken.

## Environment

- RustFS `1.0.0-beta.12`, 3 nodes / 12 drives, one erasure set, parity 4, musl static build; identical results with the #5822 and #5837 fixes applied.
- Client: boto3 with `retries=0` to make the error classes visible; the codes themselves are what any SDK receives.
- Code references verified on `main` at `778f1dfa2` (the `1.0.0-rc.1` version bump).
- Related: #5363 / `b432f31c2` introduced the part-transaction scan; #5821/#5822, #5837/#5838 are unrelated heal-path issues from the same test campaign.

houseme:
## Related Issues
Fixes #5839

## Summary of Changes
- Recover multipart part transaction marker reads by collecting per-disk errors and reducing them with write quorum semantics instead of failing on the first degraded disk.
- Keep transaction settle results index-aligned, return committed only when commit reaches write quorum, return rolled back only when rollback reaches write quorum, and fail with write quorum when neither side can settle enough disks.
- Map degraded disk and local resource pressure errors (`FaultyDisk`, `FaultyRemoteDisk`, `DiskNotFound`, `TooManyOpenFiles`) to retryable `SlowDown` API errors instead of client-visible `InternalError`.
- Add a degraded `UploadPart` regression test for the data==parity case where read quorum is lower than write quorum.

## Verification
- `cargo test -p rustfs-ecstore put_object_part_recovers_transaction_with_one_faulty_disk_at_write_quorum`
- `cargo test -p rustfs-ecstore complete_multipart_upload_recovers_interrupted_part_retry`
- `cargo test -p rustfs-ecstore upload_part_retry_quorum_failure_preserves_old_part_across_ec_geometries`
- `cargo test -p rustfs-ecstore rename_part_quorum_failure_without_old_part_removes_new_shards`
- `cargo test -p rustfs test_api_error_from_storage_error_mappings`
- `cargo fmt --all --check`
- `git diff --check`
- `make pre-commit`

## Impact
Multipart `UploadPart` can now commit/recover successfully when ignored degraded disk errors remain within write quorum tolerance. Clients now receive retryable S3 `SlowDown` responses for transient degraded-disk/resource-pressure storage errors. No configuration, migration, or on-disk format changes.

## Additional Notes
- Rebased onto latest `origin/main` before opening the PR.
- Adversarial validation: correctness attacked quorum-1, exact write quorum, data==parity read/write quorum split, and split commit/rollback outcomes; no break found after targeted regression tests.
- Adversarial validation: simplicity attacked helper extraction and constant reuse; the local observation/outcome structs are scoped to one recovery flow and no existing helper matched the required transaction-settle semantics.
- Adversarial validation: security attacked client-visible error leakage; retryable mappings now avoid exposing degraded disk errors as `InternalError`.
- Adversarial validation: concurrency/durability attacked partial settle, crash-after-commit, and retry interleavings; commit/rollback still uses the existing settle primitive and only changes quorum classification after fan-out.
- Adversarial validation: compatibility attacked S3 API behavior and on-disk/on-wire format drift; API behavior becomes retryable and no persisted format changes are introduced.
- Adversarial validation: performance attacked added hot-path allocation and disk I/O; the diff adds O(disks) `Bytes` refcount clones and no additional disk fan-out beyond existing transaction recovery reads.
- Adversarial validation: test coverage attacked revert detection; the new degraded `UploadPart` test fails on the old fail-fast marker read, API mapping cases fail on the old `InternalError` fallback, and existing multipart recovery tests continue to cover rollback and quorum failure behavior.

---

### DIFF
--- a/crates/ecstore/src/set_disk/core/io_primitives.rs
+++ b/crates/ecstore/src/set_disk/core/io_primitives.rs
@@ -3412,6 +3412,18 @@ impl SetDisks {
     }
 
     async fn recover_part_transaction(&self, dst_object: &str, write_quorum: usize) -> disk::error::Result<bool> {
+        struct PartTransactionObservation {
+            transaction_meta: Option<Bytes>,
+            current_meta: Option<Bytes>,
+            rollback: bool,
+            err: Option<DiskError>,
+        }
+
+        enum PartTransactionOutcome {
+            Commit,
+            Rollback,
+        }
+
         let disks = self.get_disks_internal().await;
         let transaction_path = part_transaction_path(dst_object);
         let transaction_meta_path = format!("{transaction_path}/{PART_TRANSACTION_NEW_META}");
@@ -3425,35 +3437,75 @@ impl SetDisks {
             let current_meta_path = current_meta_path.clone();
             async move {
                 let Some(disk) = disk else {
-                    return Ok((None, None, false));
+                    return PartTransactionObservation {
+                        transaction_meta: None,
+                        current_meta: None,
+                        rollback: false,
+                        err: Some(DiskError::DiskNotFound),
+                    };
                 };
                 let transaction_meta = match disk.read_all(RUSTFS_META_MULTIPART_BUCKET, &transaction_meta_path).await {
                     Ok(meta) => Some(meta),
                     Err(DiskError::FileNotFound) => None,
-                    Err(err) => return Err(err),
+                    Err(err) => {
+                        return PartTransactionObservation {
+                            transaction_meta: None,
+                            current_meta: None,
+                            rollback: false,
+                            err: Some(err),
+                        };
+                    }
                 };
                 let rollback = match disk.read_all(RUSTFS_META_MULTIPART_BUCKET, &rollback_path).await {
                     Ok(_) => true,
                     Err(DiskError::FileNotFound) => false,
-                    Err(err) => return Err(err),
+                    Err(err) => {
+                        return PartTransactionObservation {
+                            transaction_meta,
+                            current_meta: None,
+                            rollback: false,
+                            err: Some(err),
+                        };
+                    }
                 };
                 let current_meta = match disk.read_all(RUSTFS_META_MULTIPART_BUCKET, &current_meta_path).await {
                     Ok(meta) => Some(meta),
                     Err(DiskError::FileNotFound | DiskError::DiskNotFound) => None,
                     Err(_) => None,
                 };
-                Ok((transaction_meta, current_meta, rollback))
+                PartTransactionObservation {
+                    transaction_meta,
+                    current_meta,
+                    rollback,
+                    err: None,
+                }
             }
         });
-        let observations = join_all(reads).await.into_iter().collect::<disk::error::Result<Vec<_>>>()?;
-        if observations.iter().all(|(transaction, _, _)| transaction.is_none()) {
+        let observations = join_all(reads).await;
+        let read_errs = observations
+            .iter()
+            .map(|observation| observation.err.clone())
+            .collect::<Vec<_>>();
+        if let Some(err) = reduce_write_quorum_errs(&read_errs, OBJECT_OP_IGNORED_ERRS, write_quorum) {
+            return Err(err);
+        }
+
+        if observations
+            .iter()
+            .filter(|observation| observation.err.is_none())
+            .all(|observation| observation.transaction_meta.is_none())
+        {
             return Ok(false);
         }
 
         let mut current_counts: HashMap<Bytes, usize> = HashMap::new();
-        for (_, current, _) in &observations {
-            if let Some(current) = current {
+        let mut transaction_meta_values = HashSet::new();
+        for observation in observations.iter().filter(|observation| observation.err.is_none()) {
+            if let Some(current) = &observation.current_meta {
                 *current_counts.entry(current.clone()).or_default() += 1;
+            }
+            if let Some(transaction_meta) = &observation.transaction_meta {
+                transaction_meta_values.insert(transaction_meta.clone());
             }
         }
         let current_quorum = current_counts
@@ -3462,11 +3514,29 @@ impl SetDisks {
 
         let old_meta_path = format!("{transaction_path}/{PART_TRANSACTION_OLD_META}");
         let old_meta_absent_path = format!("{transaction_path}/old.meta.absent");
+        let mut outcomes = Vec::with_capacity(observations.len());
+        for observation in &observations {
+            let outcome = if observation.err.is_none() && observation.transaction_meta.is_none() {
+                match &observation.current_meta {
+                    Some(current_meta) if transaction_meta_values.contains(current_meta) => Some(PartTransactionOutcome::Commit),
+                    _ => Some(PartTransactionOutcome::Rollback),
+                }
+            } else {
+                None
+            };
+            outcomes.push(outcome);
+        }
         let decisions = observations
             .iter()
             .enumerate()
-            .filter_map(|(index, (transaction_meta, _, rollback))| {
-                transaction_meta.as_ref().map(|meta| (index, meta.clone(), *rollback))
+            .filter_map(|(index, observation)| {
+                if observation.err.is_some() {
+                    return None;
+                }
+                observation
+                    .transaction_meta
+                    .as_ref()
+                    .map(|meta| (index, meta.clone(), observation.rollback))
             })
             .map(|(index, transaction_meta, rollback)| {
                 let disk = disks[index].clone();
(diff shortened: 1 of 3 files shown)


terem42:
Verified this PR end-to-end on the same physical test rig that produced #5839, under the original fault. Result: the failure mode is gone.

**Setup:** 3-node / 12-drive cluster, one erasure set, EC parity 4, musl static builds. This PR's commit (`a0cd74ef`) cherry-picked onto the same base the issue was reproduced on (1.0.0-beta.12 + #5822 + the #5837 fix), so the delta measured is exactly this change. Same scenario as the issue: a 200 MiB multipart upload (13 parts, 2 s apart) against node1, `virsh destroy` on node3 (hard power-off, no shutdown) 12 s into the upload, with 20 rps of mixed GET/PUT/LIST background load against the same endpoint throughout.

**Before** (twice, 2026-08-07 stock and 2026-08-08 patched-base): every part after the cut failed — 7× `InternalError "Faulty disk"`, 1× `ServiceUnavailable`, 1× connection dropped mid-flight. 9/9 failed, both runs, identical distribution.

**With this PR**, same cut point: parts 1–4 completed before the cut as always. Part 5, in flight at the instant of the power-off, lost its connection — expected and unavoidable. Part 6 landed in the window where the dead peer's connections were still timing out and got `ServiceUnavailable` — the correct, retryable class; a default SDK retry policy would have recovered it. Parts 7 through 13 **all succeeded** against the degraded cluster. No `InternalError`, no `"Faulty disk"` anywhere in the run. `CompleteMultipartUpload` assembled the successful parts. The background load recorded 7,637 requests with 0 errors across the fault, and a full byte-level verification of the 336-object fixture passed after recovery (0 missing, 0 corrupt), so the quorum-relaxed marker scan did not admit any inconsistency we could detect.

Also ran the PR's five new tests locally on the PR branch — all pass — and the cherry-pick onto the beta.12-era base applied clean, for what that's worth to backporting.

One observation, not an objection: part 6 shows there is still a short transitional window (the seconds while the dead peer's connections are timing out) where `UploadPart` returns 503 rather than being served from the surviving quorum the way parts 7–13 were. That is correct behaviour from the client's point of view — retryable and honest — just noting that the window exists and is small (one part at 2 s spacing).

Thanks for the quick turnaround on this — the settle-quorum accounting in `recover_part_transaction` is a more complete treatment than the minimal fix we had prepared on our side.

PR #5853, opened 2026-08-08
fix(build): support non-Linux Unix targets (illumos/Solaris/*BSD)

### LINKED ISSUE #3195: # Feature Request: Support building on Illumos/OmniOS and other non-Linux Unix systems
## Summary

RustFS currently fails to build on **OmniOS (Illumos)** and likely on **FreeBSD**, **Solaris**, and other non-Linux Unix platforms due to two independent issues. Both are fixable without changing RustFS core logic — only the build infrastructure needs updating.

---

## Environment

- **OS:** OmniOS r151054 (Solaris fork/Illumos kernel)
- **Rust:** 1.96.0 (stable, target: `x86_64-unknown-illumos`)
- **Cargo:** 1.96.0
- **RustFS:** 1.0.0-beta.7

---

### Root cause

`pulsar v6.8.0` uses `prost-build` in its `build.rs` to generate Rust structs from `.proto` files. The generated file is included via:

```rust
include!(concat!(env!("OUT_DIR"), "/pulsar.rs"));
```

On OmniOS, `protoc` (the Protocol Buffers compiler) is not available in the system package manager (`pkg`). The build fails with 50–150 errors because `pulsar.rs` is never generated:

```
error[E0425]: cannot find function `client_version` in module `proto`
error[E0252]: the name `BaseCommand` is defined multiple times
error[E0277]: the trait bound `CompressionType: TryFrom<i32>` is not satisfied
error[E0308]: mismatched types — expected `Option<i32>`, found `Option<String>`
... (dozens more)
```

These errors are all symptoms of `pulsar.rs` not being generated — the `proto` module is empty or missing.

### Proposed fix

Add `protobuf-src` as a build-dependency in `pulsar`'s `Cargo.toml`. This compiles `protoc` from bundled C++ sources during the Rust build, requiring only a C compiler (`gcc` or `clang`) — which is always present on Unix systems:

```toml
# In pulsar's Cargo.toml or your Cargo.toml patch
[build-dependencies]
prost-build = "0.12"
protobuf-src = "1.1.0"
```

```rust
// In pulsar's build.rs
fn main() {
    std::env::set_var("PROTOC", protobuf_src::protoc());
    prost_build::compile_protos(&["proto/PulsarApi.proto"], &["proto/"]).unwrap();
}
```

Alternatively, use `prost-build` with the `vendored` feature which bundles a pre-compiled `protoc` binary for common platforms, or vendor the generated `pulsar.rs` directly into the repository as a fallback for platforms where `protoc` is unavailable.

---

### Root cause

The `clocksource` crate (used by RustFS for timing) uses `libc::CLOCK_MONOTONIC_COARSE`, which is a Linux-specific extension. On Illumos/OmniOS:

```
error[E0425]: cannot find value `CLOCK_MONOTONIC_COARSE` in crate `libc`
```

### Note on target_os

Illumos reports `target_os = "illumos"` (not `"solaris"`) since Rust 1.37. Any fix must use `cfg(target_os = "illumos")`, not `cfg(target_os = "solaris")`.

### Proposed fix

In `clocksource` (or wherever `CLOCK_MONOTONIC_COARSE` is used), add a platform guard:

```rust
#[cfg(any(target_os = "linux", target_os = "android"))]
const CLOCK_ID: libc::clockid_t = libc::CLOCK_MONOTONIC_COARSE;

#[cfg(not(any(target_os = "linux", target_os = "android")))]
const CLOCK_ID: libc::clockid_t = libc::CLOCK_MONOTONIC;
```

`CLOCK_MONOTONIC` is POSIX-standard and available on all Unix platforms including Illumos, FreeBSD, and macOS.

---

## Current workaround

We maintain a build script that patches both issues at build time:
- Replaces `CLOCK_MONOTONIC_COARSE` via `sed`
- Injects a hand-written `proto` module to replace the missing `pulsar.rs`

This works but is fragile and breaks with every upstream update. A proper fix in the upstream crates would make RustFS buildable on all Unix platforms without any patching.

---

## Impact

Fixing these two issues would enable RustFS to build on:
- **Illumos / OmniOS** (current blocker)
- **FreeBSD** (`CLOCK_MONOTONIC_COARSE` also missing; `protoc` availability varies)
- **Solaris** (`target_os = "solaris"`, same clock issue)
- **macOS** (likely already works but would benefit from vendored protoc)

The changes are isolated to build infrastructure — no RustFS runtime logic needs to change.

yyyCode:
## Related Issues

Closes #3195

## Summary of Changes

Two independent build-infrastructure blockers kept RustFS from building on non-Linux Unix platforms (illumos/OmniOS, Solaris, the BSDs). Neither change touches runtime logic.

**1. pulsar protobuf codegen needs `protoc`.** `pulsar` regenerates its protobuf bindings from `PulsarApi.proto` in `build.rs` on every build, which requires a `protoc` binary. Platforms without a packaged protoc now enable pulsar's `protobuf-src` feature through a `cfg`-gated dependency in `crates/targets/Cargo.toml`, which builds a vendored `protoc` from C++ sources using the platform toolchain. Mainstream targets (Linux/macOS/Windows) keep the lean dependency and their existing system/CI protoc — their feature set is unchanged. This follows @houseme's suggestion on the issue.

**2. clocksource used the Linux-only `CLOCK_MONOTONIC_COARSE`.** `clocksource 0.8.3` was pulled in transitively by `ratelimit 0.10`. Rather than patching clocksource, this upgrades `ratelimit` to `2.0`, which **dropped the clocksource dependency entirely** — removing the portability problem at the root. The bandwidth throttle's best-effort bulk `consume()` is rewritten onto ratelimit 2.0's `try_wait_n`, preserving the existing partial-consumption semantics.

## Verification

Local native builds on this Windows box use a GNU-host toolchain (no MSVC build tools present), but the checks below all pass:

- `cargo check -p rustfs-ecstore` — clean
- `cargo test -p rustfs-ecstore --lib bucket::bandwidth::monitor` — 14/14 pass (consume/deficit/concurrent/burst)
- `cargo tree --target x86_64-unknown-illumos -e features` — confirms `protobuf-src` is enabled **only** on illumos/Solaris/*BSD
- `cargo tree --target x86_64-pc-windows-gnu` — confirms `protobuf-src` is **not** enabled on mainstream targets
- `cargo tree -i clocksource` — no longer in the dependency graph

**The final on-platform illumos build has not been run here** (no illumos target / cmake locally). @guenther-alka has been maintaining OmniOS build scripts on the issue and could confirm end to end.

## Impact

- No runtime behavior change; changes are confined to build configuration and a dependency upgrade.
- `protobuf-src` adds a C++ build step **only** for illumos/Solaris/*BSD targets; mainstream builds are unaffected and do not pay the vendored-protoc cost.
- `ratelimit 0.10 -> 2.0` and removal of `clocksource` are reflected in `Cargo.lock`.

## Additional Notes

`protobuf-src` requires a C/C++ toolchain (and cmake) on the target platform to compile the vendored protoc — always available on the Unix targets in scope.


guenther-alka:
Confirmed end-to-end on OmniOS r151058 (illumos), as requested.

git fetch origin pull/5853/head:pr-5853, checked out clean, applied only the existing illumos TLS-provider/pprof-disable patches from my local build script (unrelated to this PR — aws-lc-rs→ring, disable pyroscope/jemalloc_pprof/mimalloc), then ran cargo build -p pulsar as a scoped test of the actual fix.

Result: pulsar v6.8.0 compiles cleanly, exit 0, ~7m36s.

protobuf-src builds its vendored protoc via cmake exactly as described — but two platform prerequisites had to be added on my end that aren't mentioned in the PR notes and are worth documenting for other illumos/Solaris users:

cmake isn't in the base OmniOS image — needs pkg install ooce/developer/cmake.
/usr/bin/make on illumos is Sun's dmake, not GNU Make. protobuf-src's generated CMake build invokes make with a GNU-style bare -j (jobserver) flag, which dmake rejects outright (No dmake max jobs argument after -j flag). Fix: pkg install developer/build/gnu-make, then make sure gmake resolves as make ahead of /usr/bin in PATH (I used a make -> gmake symlink in a directory prepended to PATH, rather than touching the system /usr/bin/make).

cargo tree -i clocksource confirms the dependency is fully gone with the ratelimit 2.0 bump — no leftover patch needed there either.

houseme:
@yyyCode 
Thanks for the follow-up work here. Before we keep the target-specific `protobuf-src` feature enabled for `pulsar`, could you please verify one alternative?

If a system `protoc` is installed first, for example with `sudo apt install protobuf-compiler` on Debian/Ubuntu-style environments, does the build pass with the lean dependency instead?

```toml
pulsar = { workspace = true, default-features = false, features = ["tokio-rustls-runtime", "telemetry"] }
```

The current target-specific dependency enables:

```toml
pulsar = { workspace = true, default-features = false, features = ["tokio-rustls-runtime", "telemetry", "protobuf-src"] }
```

That pulls in a vendored `protoc` build via CMake/C++, which is heavier and adds extra toolchain requirements. If installing the system `protobuf-compiler` is enough for Pulsar's protobuf codegen, I would prefer documenting `protoc` as a platform prerequisite and avoiding `protobuf-src` unless there is no reliable system package path for the target.

guenther-alka:
Verified — yes, the lean dependency works fine on illumos too, and it's noticeably faster.

Tested by installing the packaged protoc (pkg install ooce/developer/protobuf, gives libprotoc 34.1 at /opt/ooce/bin/protoc), then temporarily forcing the lean feature set in crates/targets/Cargo.toml for the illumos target (features = ["tokio-rustls-runtime", "telemetry"], i.e. dropping protobuf-src) and building rustfs-targets from clean:

Compiling pulsar v6.8.0
Compiling rustfs-targets v1.0.0-rc.1
Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 34s
LEAN_EXIT=0

That's ~4 minutes faster than the protobuf-src path I tested earlier (7m36s, mostly spent compiling libprotobuf from C++ source).

ooce/developer/protobuf is a maintained OOCE package (currently at 34.1, also available at 33.2), so I'd agree: documenting protoc as a platform prerequisite is preferable to pulling in the vendored CMake/C++ build on illumos. Happy to have illumos drop protobuf-src in favor of the lean path if that's the direction you want to take.

houseme (approved):

houseme:
Thanks again @yyyCode and @guenther-alka — especially for the OmniOS validation and for checking the system-protoc alternative. I pushed one more commit to this PR branch: `d6ec8be58c8f4f88b8fbdccc1ec40a29e25516c3`, which removes the target-specific `protobuf-src` feature and keeps `pulsar` on the lean feature set.

The intended rule is now: install a system `protoc` on platforms where it is not already available, instead of building a vendored `protoc` through CMake/C++.

Could you both please try the latest PR branch again in your normal environments and confirm the end-to-end experience?

- `protoc --version` resolves from `PATH` before the build.
- `cargo build -p rustfs-targets` or your usual RustFS build path still passes.
- OmniOS/illumos no longer needs the extra `protobuf-src` / CMake / GNU make workaround for `pulsar`.
- The `ratelimit` 2.0 behavior remains unchanged for bandwidth throttling.

Really appreciate the careful platform testing here; it made the lighter dependency path much clearer.

yyyCode:
Thanks @houseme for the lean-path commit and @guenther-alka for the on-platform illumos testing. I re-verified the latest branch (`d6ec8be58`) locally on Windows and both parts check out:

**1. Lean pulsar feature with a system `protoc`**
- Installed the official prebuilt `protoc` (v35.1, `libprotoc 35.1`) and put it on `PATH` — no CMake/C++ toolchain involved.
- `protoc --version` resolves from `PATH` before the build.
- `cargo check -p rustfs-targets` passes (exit 0, ~3m19s); the log shows `Compiling pulsar v6.8.0` succeeding with the lean feature set (`tokio-rustls-runtime`, `telemetry`). Since this box has no CMake, a vendored `protobuf-src` path would have failed here — confirming the build takes the system-`protoc` route.

**2. ratelimit 2.0 bandwidth throttle**
- `cargo test -p rustfs-ecstore --lib bucket::bandwidth::monitor` → 16 passed, 0 failed (consume/deficit/concurrent/burst/continuous-refill). Behavior is unchanged from the 0.10 path.

So the direction is confirmed on Windows too: documenting a system `protoc` as the prerequisite works cleanly, and no `protobuf-src`/CMake workaround is needed. LGTM.

guenther-alka:
ok, build tested with smaller optimized build script
https://illumos.topicbox.com/groups/omnios-discuss/T3f99b32a3a89273b-M8786b2d4979535aca271280f

PR #5856, opened 2026-08-08
fix(ecstore): honor scoped object heal placement

cxymds:
## Related Issues
N/A

## Summary of Changes
- Honor `HealOpts.pool` when dispatching object heal across ECStore pools.
- Honor `HealOpts.set` when selecting the erasure set inside a pool instead of always falling back to the object hash.
- Fail closed with `InvalidArgument` for scoped object-heal requests that do not match any pool/set, preventing a scoped heal from silently expanding to the whole deployment.
- Add regression tests for pool filtering, requested-set selection, and invalid scoped coordinates.

## Verification
- `git diff --check`
- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore heal_object_scope --lib`
- `cargo test -p rustfs-ecstore handle_heal_object_rejects_unmatched_scoped_pool --lib`
- `cargo test -p rustfs-ecstore heal_object_uses_requested_set_instead_of_object_hash --lib`
- `cargo test -p rustfs-ecstore heal_object_rejects_requested_set_outside_pool --lib`
- `make pre-commit`
- `CARGO_BUILD_JOBS=1 cargo clippy --all-targets -- -D warnings`

## Impact
Scoped object heal now targets only the requested placement. Unscoped object heal keeps the existing all-pool fan-out behavior. Invalid scoped heal coordinates now surface as `InvalidArgument` instead of falling through to unrelated pools. No S3 API, RPC wire format, or on-disk format changes.

## Additional Notes
The default `make pre-pr` wrapper was attempted, but the local machine repeatedly terminated the default workspace clippy process with SIGTERM/SIGINT under normal parallelism. I reran the full workspace clippy command successfully with `CARGO_BUILD_JOBS=1`, and kept the focused ecstore tests listed above for the changed behavior.

### DIFF
--- a/crates/ecstore/src/core/sets.rs
+++ b/crates/ecstore/src/core/sets.rs
@@ -286,6 +286,23 @@ impl Sets {
         self.get_disks(self.get_hashed_set_index(key))
     }
 
+    fn get_heal_object_disks(&self, object: &str, opts: &HealOpts) -> Result<Arc<SetDisks>> {
+        if let Some(set_idx) = opts.set {
+            return self.disk_set.get(set_idx).cloned().ok_or_else(|| {
+                Error::InvalidArgument(
+                    "heal".to_string(),
+                    "scope".to_string(),
+                    format!(
+                        "invalid object heal set index {set_idx} for pool {} with {} sets",
+                        self.pool_idx, self.set_count
+                    ),
+                )
+            });
+        }
+
+        Ok(self.get_disks_by_key(object))
+    }
+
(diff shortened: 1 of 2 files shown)


cxymds:
Closing as superseded by #5855, which is already merged into main and covers the same ECStore pool/set-scoped object-heal routing, plus the corresponding heal-manager scoped repair limits. Resolving this conflict by retaining both implementations would duplicate the behavior and risk diverging error semantics.

PR #5857, opened 2026-08-08
fix(heal): count scoped object heals in set bulkhead

cxymds:
## Related Issues
N/A

## Summary of Changes
- Include set-scoped object heal tasks in the scheduler's per-set running counts.
- Reuse the existing erasure-set key format for object heal requests that carry both `pool_index` and `set_index`.
- Keep unscoped object heal behavior unchanged: only requests with explicit pool and set coordinates participate in the set bulkhead.
- Add regression tests for scoped object admission and active running-count aggregation.

## Verification
- `cargo fmt --all --check`
- `cargo test -p rustfs-heal can_schedule --lib`
- `cargo test -p rustfs-heal running_heal_set_counts --lib`
- `git diff --check`
- `CARGO_BUILD_JOBS=1 make pre-commit`

## Impact
Set-scoped object heal work now respects the same per-set scheduler bulkhead as erasure-set heal work. This prevents multiple scoped object heal tasks from bypassing `max_concurrent_per_set` on the same erasure set. There are no RPC, on-disk, S3 API, or configuration format changes.

## Additional Notes
The default parallel `make pre-commit` quick-check was terminated locally by SIGTERM during dependency compilation. Re-running the same pre-commit gate with `CARGO_BUILD_JOBS=1` completed successfully.

### DIFF
--- a/crates/heal/src/heal/manager.rs
+++ b/crates/heal/src/heal/manager.rs
@@ -2475,7 +2475,7 @@ impl HealManager {
             return;
         }
 
-        let mut running_per_set = running_erasure_set_counts(&active_heals_guard);
+        let mut running_per_set = running_heal_set_counts(&active_heals_guard);
         let mut tasks_started = 0usize;
         let mut delayed_by_mainline_throttle = false;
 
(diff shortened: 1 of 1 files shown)


cxymds:
Closing as superseded by #5855, which is already merged and includes the same scoped-object per-set bulkhead accounting and regression coverage. The remaining diff only factors the shared set-key match into a helper and does not provide independent runtime behavior worth resolving the conflict for.

PR #5864, opened 2026-08-08
fix(replication): retry, persist and replay failed delete-marker purges

reatang:
## Summary

P1-21 from the replication compatibility review (rustfs/backlog#1675, batch B2 — data consistency). The delayed delete-marker purge was fire-and-forget: the target DELETE discarded its result (`let _ =`), a missing target client was silently skipped, and nothing recorded the intent — one transient target error stranded the replicated marker on the target forever, with no log, no metric, no retry channel (the scanner cannot see a source version that is gone). Separately, `replicate_delete_with_outcome` included `!requires_delayed_purge` in its outcome, pinning every delete-marker MRF entry to Missed so the durable backlog retained them permanently (the companion finding folded into this fix).

## Changes

- **Per-target purge results** (`replicate_delete_marker_purge_to_targets`): failures now warn with structured events + `rustfs_replication_delete_marker_purge_total` counters, including the previously silent `target_client_missing` skip; the function returns the failed ARNs and supports retrying only those. A target-side `NoSuchKey`/`NoSuchVersion` counts as purge success (strict-404 targets must not retain the intent forever; RustFS/MinIO answer 204 anyway).
- **Watch-window retries** (`watch_and_purge_source_delete_marker`): the 5x1s watcher retries failed targets across its remaining rounds; on exhaustion it persists the purge intent to the durable MRF journal via the new `ReplicationPoolTrait::persist_mrf_entry`. Journal-only on purpose: live re-dispatch would loop unboundedly against a down target. Intent entries are shaped as marker-creation deletes so replay funnels into the stale-marker branch.
- **Stale-marker branch purges** (source marker gone while replication was pending): previously returned success without touching the targets — a latent leak that also would have acked replayed purge intents. It now purges and reports the result as the replay outcome. Heal callers retry for the full window (the startup MRF processor runs before bucket metadata/target clients initialize — observed in the e2e as `target_client_missing` on the first pass); live callers attempt once and fall back to a fresh durable intent so a down target cannot pin a replication worker.
- **Outcome formula** extracted as `replicate_delete_outcome` and pinned by a unit test: successfully replayed delete-marker entries are acknowledged instead of retained forever.

- Red commit (80c2c2e) pins both failure modes with journal evidence: 3 faulted DELETE attempts and a permanently stranded target marker; zero post-restart replay.
- `test_delayed_delete_marker_purge_retries_after_transient_target_failure`: 4 scripted 503s outlast every pre-fix channel; a watch-round retry converges (~15s).
- `test_delayed_delete_marker_purge_exhaustion_persists_to_mrf_and_replays_on_restart`: exhaustion -> durable MRF -> restart replay purges -> second restart issues zero replays (acknowledgement pinned).
- Unit tests: MRF intent shaping; outcome formula (the exhaustive e2e staging for a marker-still-present durable entry would require saturating the worker queues, hence the formula-level guard).
- `make pre-commit`, logging guardrails, `cargo clippy --all-targets` (ecstore + e2e_test) clean; full ecstore lib suite 3729 passed (3 pre-existing local-DNS kubernetes endpoint test failures reproduce on the base commit without this change).

## Adversarial validation

Seven roles (correctness, simplicity, security, concurrency/durability, compatibility, performance, test coverage) ran as independent reviewer agents; the single blocking finding (no revert detection for the outcome hunk) was resolved with the formula guard test. Non-blocking findings applied: live/Heal retry split, NoSuchVersion-as-success mapping, log/loop dedup, `admitted_target_arns` hoist.

## Known residuals (recorded, not regressions)

- **Watcher shutdown window**: the watcher is a detached task; a shutdown during the watch window loses an unpersisted intent. Present before this fix too (which additionally had no replay half at all); closing it needs a write-ahead intent, tracked for follow-up.
- **Rolling downgrade**: an old binary winning the MRF recovery lock replays new purge-intent entries through its old stale branch (`return true`) — acked without purging, equal to pre-fix behavior; no corruption.
- **Targets that mint their own version ids**: replay reconstructs a synthetic state without the recorded per-target marker id and falls back to the source id, so the purge no-ops there. Same family as the P1-19 version-identity work (next in B2), which will reject such targets explicitly; `MrfReplicateEntry` gaining a recorded-id field is the eventual fix.
- The nightly workflow's "27 tests" comment predates this PR and needs a workflow-scoped credential to update (split out separately). `.config/nextest.toml` counts updated here (20 fast + 38 nightly = 58).

## Merge note

Two commits (red test evidence -> fix); squash-merge if per-commit CI green is required — the red commit fails by design.

### DIFF
--- a/crates/ecstore/src/bucket/replication/replication_resyncer.rs
+++ b/crates/ecstore/src/bucket/replication/replication_resyncer.rs
@@ -1495,15 +1495,6 @@ pub(crate) async fn replicate_delete_with_outcome<S: ReplicationStorage>(
     let is_version_purge = is_version_delete_replication(&dobj.delete_object);
 
     let requires_delayed_purge = should_retry_delete_marker_purge(&dobj.delete_object);
-    if requires_delayed_purge {
-        let bucket_clone = bucket.clone();
-        let dobj_clone = dobj.clone();
-        let dsc_clone = dsc.clone();
-        let storage_clone = storage.clone();
-        tokio::spawn(async move {
-            watch_and_purge_source_delete_marker(bucket_clone, dobj_clone, dsc_clone, storage_clone).await;
-        });
-    }
 
     let (replication_status, prev_status) = if !is_version_purge {
         (
@@ -1544,6 +1535,24 @@ pub(crate) async fn replicate_delete_with_outcome<S: ReplicationStorage>(
     );
     if replication_status != prev_status {
         drs.replication_timestamp = Some(OffsetDateTime::now_utc());
+    }
+
+    if requires_delayed_purge {
+        // Hand the watcher the MERGED replication state: `drs` folds this
+        // round's per-target results into the previous state, including the
+        // version ids the targets assigned to the markers they just created.
+        // Spawning with the pre-merge `dobj` made the purge fall back to a
+        // source-derived id, which a target that mints its own ids answers
+        // with an idempotent 204 — the intent was then dropped while the
+        // real marker stayed behind.
+        let bucket_clone = bucket.clone();
+        let mut dobj_clone = dobj.clone();
+        dobj_clone.delete_object.replication_state = Some(drs.clone());
+        let dsc_clone = dsc.clone();
+        let storage_clone = storage.clone();
+        tokio::spawn(async move {
+            watch_and_purge_source_delete_marker(bucket_clone, dobj_clone, dsc_clone, storage_clone).await;
+        });
     }
 
     let event_name = if replication_status == ReplicationStatusType::Completed {
@@ -1712,22 +1721,12 @@ async fn replicate_delete_marker_purge_to_targets(
         {
             continue;
         }
-        let Some(tgt_client) = ReplicationTargetStore::remote_target_client(bucket, &tgt_entry.arn).await else {
-            warn!(
-                event = EVENT_DELETE_MARKER_PURGE_FAILED,
-                component = LOG_COMPONENT_ECSTORE,
-                subsystem = LOG_SUBSYSTEM_REPLICATION_RESYNC,
-                bucket,
-                object = dobj.delete_object.object_name,
-                arn = tgt_entry.arn,
-                reason = "target_client_missing",
-                "Delete-marker purge attempt failed"
-            );
-            counter!(METRIC_DELETE_MARKER_PURGE_TOTAL, "state" => "failed").increment(1);
-            failed_arns.push(tgt_entry.arn.clone());
-            continue;
-        };
-
+        // Decide the version first: refusing to guess is a per-target
+        // FAILURE, not a silent skip. Reporting it as success would let the
+        // watcher and the MRF replay drop the purge intent while the marker
+        // is still on the target — the leak stays visible instead (the
+        // entry is retained and keeps warning) until an operator repairs
+        // the metadata.
         let Some(purge_version_id) = delete_marker_purge_version_id(
             dobj.delete_object.replication_state.as_ref(),
             &tgt_entry.arn,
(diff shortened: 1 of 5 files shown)


houseme (commented):
The retry and MRF replay path improves transient failures. The remaining shutdown interval before the detached watcher persists its own intent can still lose a purge. Please either persist the intent before acknowledging the parent delete or clearly scope this as a known non-durable window with follow-up tracking before merge.

crates/ecstore/src/bucket/replication/replication_resyncer.rs
```diff
@@ -1485,21 +1495,7 @@ pub(crate) async fn replicate_delete_with_outcome<S: ReplicationStorage>(
...
-                        delete_marker_version_id,
-                    )
-                    .await
-                {
-                    replicate_delete_marker_purge_to_targets(&bucket_clone, &dobj_clone, &dsc_clone).await;
-                    break;
-                }
-                tokio::time::sleep(TokioDuration::from_secs(1)).await;
-            }
+            watch_and_purge_source_delete_marker(bucket_clone, dobj_clone, dsc_clone, storage_clone).await;
```
overtrue: The watcher receives dobj before rinfos is merged into the persisted replication state. If a target minted its own delete-marker version ID, the purge falls back to the source UUID; a no-op 204 then clears the retry set and leaves the real marker behind. Please pass the merged target-version mapping into this path and cover a target that ignores the source-version header.

overtrue (changes requested):
The live purge can acknowledge the wrong target version and strand a delete marker.

crates/ecstore/src/bucket/replication/replication_resyncer.rs
```diff
@@ -1686,25 +1734,217 @@ async fn replicate_delete_marker_purge_to_targets(bucket: &str, dobj: &DeletedOb
...
+                    reason = "target_delete_failed",
+                    "Delete-marker purge attempt failed"
+                );
+                counter!(METRIC_DELETE_MARKER_PURGE_TOTAL, "state" => "failed").increment(1);
+                mark_replication_target_offline_if_needed(&tgt_client, &error).await;
+                failed_arns.push(tgt_entry.arn.clone());
+            }
+        }
     }
+    failed_arns
```
overtrue, line 1802: An ARN skipped for inconsistent version metadata is not added to failed_arns. Returning an empty list then makes both the watcher and MRF replay treat a purge with no DELETE as successful and drop the intent. Keep that target in a non-success outcome and cover corrupt persisted state.

overtrue (changes requested):
The target-version race remains, and corrupt version metadata can now be skipped and reported as a successful purge.

reatang:
Both findings were real and are fixed on the new head; the shutdown-window question is answered below.

**1. Watcher ran on pre-merge state → wrong purge version** (`replication_resyncer.rs`)

Confirmed exactly as described. The watcher was spawned before `drs = get_replication_state(&rinfos, ...)` folded this round's per-target results into the state, so `target_delete_marker_version_ids` — the ids the targets had just assigned to the markers they created — were invisible to it. `delete_marker_purge_version_id` then fell back to the source-derived id, a minting target answered the versioned DELETE with an idempotent 204, and that "success" cleared the retry set while the real marker stayed behind.

The spawn now happens after the merge and carries `drs` as the watcher's `replication_state`.

Coverage: `test_delayed_delete_marker_purge_uses_target_assigned_version` drives a fake target with a new `assign_own_version_ids` switch — it validates but does **not** mirror the forwarded `x-*-source-version-id` header, for objects and delete markers alike — and asserts the replicated marker is really gone from the target (the assertion is version-agnostic now, since the target's id differs from the source's). It fails on the old code.

**2. Refused purge counted as success** (`replication_resyncer.rs`)

Also confirmed. An ARN skipped for inconsistent recorded metadata never entered `failed_arns`, so an empty result made both the watcher and the MRF replay treat a purge that issued no DELETE as complete and drop the intent.

The refusal is now a per-target **failure** with its own metric label (`state="refused"`), so the intent is retained and keeps warning rather than being acknowledged — an unresolvable purge stays visible for an operator instead of disappearing. I also moved the version decision ahead of the target-client lookup: refusing to guess is decided from the recorded metadata alone, before any remote call.

Coverage: `test_delete_marker_purge_reports_corrupt_recorded_version_as_failure` pins it with corrupt persisted state and **no** registered target client, so it can only pass if the refusal itself produces the failed outcome.

**3. @houseme — the detached-watcher shutdown window**

Scoping it as a known non-durable window, now written at the watcher itself rather than only in the PR body. Rationale: closing it needs a write-ahead intent persisted before the parent delete is acknowledged, which would put a journal write on *every* delete-marker replication for a purge that almost never runs. The window is also strictly better than before this PR, which had neither a durable channel nor a replay half. Recorded as follow-up with that exact shape (write-ahead intent) rather than left implicit.

Verification on the new head: 7 e2e green (the 4 fake-target self-tests, the new minting case, retry, and exhaustion→MRF→replay), `replication_resyncer` unit suite 41 green, fmt/clippy/logging guardrails clean.

crates/ecstore/src/bucket/replication/replication_resyncer.rs
```diff
@@ -1663,48 +1692,286 @@ fn delete_marker_purge_version_id(
...
+        DELETE_MARKER_PURGE_WATCH_ROUNDS
+    } else {
+        1
+    };
+    let mut failed_arns = replicate_delete_marker_purge_to_targets(bucket, dobj, &dsc, None).await;
+    for _ in 1..rounds {
+        if failed_arns.is_empty() {
+            break;
+        }
+        tokio::time::sleep(DELETE_MARKER_PURGE_WATCH_INTERVAL).await;
```
overtrue, line 1965: Every failed Heal replay can sleep four seconds here, and start_mrf_processor handles entries serially while holding the recovery lock. A down target with 1,000 purge intents delays every later MRF entry for at least 66 minutes; the journal cap permits 200,000. Please use a bounded global readiness window or defer failed entries without a per-entry sleep.

overtrue (changes requested):
The previous correctness findings are fixed, but per-entry retry delays can stall the entire durable recovery backlog for hours.

PR #5865, opened 2026-08-08
fix(storage): pin resumed get object version

cxymds:
## Related Issues
N/A

## Summary of Changes
- Pin mid-stream GetObject resume reopens to the object version resolved by the initial committed response.
- Keep explicit `versionId` requests unchanged and avoid fabricating a version pin for objects without a resolved version identity.
- Add regression coverage for latest versioned GETs where a newer PUT advances the key before resume: the resumed reader now reopens the original version instead of resolving latest and failing/splicing.

## Verification
- `cargo fmt --all && cargo test -p rustfs get_object_resume_opts_pin_latest_request_to_resolved_version --lib && cargo test -p rustfs get_object_resume_reopen_uses_initial_version_after_newer_put --lib && cargo test -p rustfs get_object_resume_reopen_rejects_a_replaced_object_version --lib`
- `cargo fmt --all --check && cargo clippy -p rustfs --lib --tests -- -D warnings && CARGO_BUILD_JOBS=1 make pre-commit`

## Impact
Versioned latest GetObject responses that resume after relocation now keep serving the exact committed object version even if a newer version is written before the resume attempt. There are no API, on-disk, wire-format, or configuration changes.

## Additional Notes
High-risk adversarial review covered correctness, simplicity, security, concurrency/durability, compatibility, performance, and test coverage. No remaining findings.

### DIFF
--- a/rustfs/src/app/object_usecase.rs
+++ b/rustfs/src/app/object_usecase.rs
@@ -1943,6 +1943,15 @@ impl GetObjectResumeIdentity {
     }
 }
 
+fn pin_get_object_resume_opts_to_resolved_version(mut opts: ObjectOptions, info: &ObjectInfo) -> ObjectOptions {
+    if opts.version_id.is_none()
+        && let Some(version_id) = info.version_id
+    {
+        opts.version_id = Some(version_id.to_string());
+    }
+    opts
+}
+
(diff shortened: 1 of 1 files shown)


cxymds:
Closing as superseded by #5859, which implements the same runtime fix and is already merged.

PR #5866, opened 2026-08-08
docs(ci): make e2e-replication-nightly test-count comments drift-resistant

reatang:
## Summary

The header comment in `.github/workflows/e2e-replication-nightly.yml` claimed the lane runs **27** nightly replication e2e tests (lines 18 and 35) with a per-category breakdown. The lane has since grown to **36** on `main` (and will become 38 once the in-flight `fix/p1-21-delayed-purge-failure-handling` branch lands), so both the headline counts and the itemized list had drifted.

Rather than bumping the numbers (which would go stale again on the very next filterset change, and whose correct value depends on merge order with in-flight branches), this drops the exact counts and per-category tallies from the workflow comment entirely and points to the single authority instead:

- `cargo nextest list -p e2e_test --profile e2e-repl-nightly` for authoritative membership and count
- the count invariant maintained next to the filtersets in `.config/nextest.toml` (deliberately not duplicated in the workflow)

The qualitative description of what the lane covers (remote-target TLS validation, data-plane/helper convergence tests, `_real_*_node` site-replication tests) is retained.

## Verification

- Comment-only change; no workflow behavior affected (`on:`, jobs, and steps untouched).
- Counts cross-checked on current `main`: `cargo nextest list -p e2e_test --profile e2e-repl-nightly` → 36 tests; smoke replication subset → 20; matches the `20 here + 36 nightly = 56 total` invariant at `.config/nextest.toml`.

### DIFF
--- a/.github/workflows/e2e-replication-nightly.yml
+++ b/.github/workflows/e2e-replication-nightly.yml
@@ -14,25 +14,24 @@
 
 # Nightly full replication e2e lane (backlog#1147 repl-1, deps: ci-4).
 #
-# The per-PR gate (ci.yml `e2e-tests` job, `--profile e2e-smoke`) runs the 20
-# FAST replication tests. This scheduled lane runs the remaining 27
-# heavier replication e2e tests that are unfit for a per-PR gate:
-#
-#   * 2 remote-target TLS validation tests.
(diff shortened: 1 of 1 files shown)


houseme (approved):

PR #5873, opened 2026-08-08
test(replication): cover rule id byte limit

overtrue:
## Related Issues
N/A

## Summary of Changes
Adds a focused regression test for replication rule ID validation so the 255 limit stays byte-based, matching the MinIO/Go `len` contract. The test accepts a 254-byte multibyte ID and rejects a 256-byte ID whose character count is still below the limit, which catches an accidental switch to character counting.

## Verification
- `RUSTC_WRAPPER= cargo test -p rustfs-replication structure_validation_counts_rule_id_limit_in_bytes -- --nocapture`
- `cargo fmt --all --check`
- `git diff --check`
- `RUSTC_WRAPPER= make pre-commit`

## Impact
Test-only. No runtime behavior change.

## Additional Notes
N/A

### DIFF
--- a/crates/replication/src/config.rs
+++ b/crates/replication/src/config.rs
@@ -795,6 +795,33 @@ mod tests {
     }
 
     #[test]
+    fn structure_validation_counts_rule_id_limit_in_bytes() {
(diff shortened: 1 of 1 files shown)


houseme (approved):

PR #5874, opened 2026-08-09
fix(policy): accept object lock mode condition

overtrue:
## Related Issues
N/A

## Summary of Changes
Accept `s3:object-lock-mode` as a valid policy condition key. The latest main e2e-full run failed while parsing a policy that uses this Object Lock condition key in the signed snowball/PAX auth path.

## Verification
- `cargo test -p rustfs-policy object_lock_mode --lib`
- `cargo fmt --all --check`
- `cargo test -p rustfs-policy --lib`
- `make pre-commit`

## Impact
Policies can now parse and evaluate Object Lock mode conditions alongside the existing retain-until-date condition key.

## Additional Notes
The targeted e2e test was attempted locally but did not complete within 180s, so it was interrupted and is not counted as passing.

### DIFF
--- a/crates/policy/src/policy/function/key_name.rs
+++ b/crates/policy/src/policy/function/key_name.rs
@@ -198,6 +198,9 @@ pub enum S3KeyName {
     #[strum(serialize = "s3:object-lock-retain-until-date")]
     S3ObjectLockRetainUntilDate,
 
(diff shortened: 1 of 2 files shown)


houseme (approved):

PR #5877, opened 2026-08-09
test(replication): pin the scanner existing-object compensation matrix

reatang:
## Summary

P1-20 from the replication compatibility review (rustfs/backlog#1675, batch B2 — data consistency). **Test-only.** No prior test in the repo wrote objects *before* the replication rule arrived: every existing e2e configures replication first, so the scanner's existing-object resync pass — the only channel that can move pre-existing objects — had zero end-to-end coverage, and the enqueue truth table was partially unpinned at unit level.

## What is pinned

**e2e** (negative cells are contracts asserted over multiple fast-scanner cycles, always next to a replicated control key proving the scanner and live path are running — an absent key on a dead scanner proves nothing):

- `test_scanner_compensates_existing_objects_across_write_paths` — objects that predate an `ExistingObjectReplication=Enabled` rule converge via scanner compensation regardless of write path (plain PUT, CopyObject, Snowball auto-extract), while a null-version object (PUT before the bucket became versioned) is pinned as **never** compensated (`scanner_folder.rs` heal gate skips nil-version objects).
- `test_scanner_never_compensates_when_existing_object_replication_disabled` — Disabled is a contract, not a delay: pre-rule keys stay absent while post-rule writes replicate normally.

**Unit truth-table pins** (`crates/replication`, the two cells the review found uncovered plus one cheap sibling):

- `queue.rs`: an empty replicate decision (Disabled existing-object / inbound REPLICA) skips heal queueing for every status; Completed without a resync decision skips.
- `operation.rs`: existing-object resync without a reset replicates exactly the never-replicated (Empty) objects — no recorded status is re-queued.

Helper: `put_bucket_replication_with_statuses` parameterizes the previously hardcoded `ExistingObjectReplication` status (existing callers unchanged). nextest count comments 36→38 nightly / 56→58 total (standalone off main; resolves additively against #5864's count bump).

## Product defect surfaced (fix ships separately)

Writing the third planned variant — inbound-replica anti-cascade — uncovered a real defect: **inbound bucket-replication PUTs never persist the REPLICA status** (`options.rs` stores it only on `opts.delete_replication`, which the PUT persistence path does not consume; the Snowball inbound path does this correctly via metadata insertion). Consequences: HEAD/GET return no `x-amz-replication-status` for replicas, and the scanner's existing-object pass **cascades inbound replicas onward** (A→B→C topologies; live-path anti-loop only blocks immediate scheduling). The anti-cascade e2e reproduces this deterministically and ships as the red light of the follow-up fix PR rather than being `#[ignore]`d here.

Not covered here, recorded: a positive "replication-reset refills the target" e2e remains a gap (unit-level reset cells are pinned in `operation.rs`); the resync helpers exist for a follow-up.

## Verification

Both e2e green twice consecutively (~11s each, nightly lane by name filter); `cargo test -p rustfs-replication` 157 passed; fmt / clippy (e2e_test + rustfs-replication) / logging guardrails clean. Test-only diff — validated at the light adversarial tier (coverage-skeptic requirements were built into the design: every negative assertion has a liveness control, every truth-table cell names its code anchor).

## Merge note

Independent of #5864; whichever merges second resolves the trivial nextest count-comment conflict additively (20 fast + 41 nightly = 61 once all three replication test PRs land).

### DIFF
--- a/crates/e2e_test/src/replication_extension_test.rs
+++ b/crates/e2e_test/src/replication_extension_test.rs
@@ -8336,7 +8336,7 @@ async fn test_scanner_never_compensates_when_existing_object_replication_disable
     init_logging();
     let source_bucket = "scanner-disabled-src";
     let target_bucket = "scanner-disabled-dst";
-    let (source_env, target_env) = build_scanner_compensation_pair(source_bucket, target_bucket).await?;
+    let (source_env, mut target_env) = build_scanner_compensation_pair(source_bucket, target_bucket).await?;
     let source_client = source_env.create_s3_client();
     let target_client = target_env.create_s3_client();
 
@@ -8367,8 +8367,31 @@ async fn test_scanner_never_compensates_when_existing_object_replication_disable
         .await?;
     wait_for_replicated_object(&target_client, target_bucket, control_key, control_payload).await?;
 
-    // With the live path proven, the pre-existing key must stay absent across
-    // multiple fast-scanner cycles.
+    // Scanner-only witness. A live-path control key alone would let this test
+    // pass while the existing-object scanner is disabled or wedged, so make
+    // the scanner itself observable: an object whose replication FAILED while
+    // the target was down can only be re-driven by the data scanner's
+    // replication heal pass (see FAST_SCANNER_ENV), and that pass is NOT
+    // gated by ExistingObjectReplication. The witness lives in the same
+    // bucket and prefix as the pre-existing key, so a heal pass that reached
+    // it necessarily walked the pre-existing key in the same scan.
+    let witness_key = "scanner-witness.txt";
+    let witness_payload = "scanner witness payload";
+    target_env.stop_server();
+    source_client
+        .put_object()
+        .bucket(source_bucket)
+        .key(witness_key)
+        .body(ByteStream::from_static(witness_payload.as_bytes()))
+        .send()
+        .await?;
+    wait_for_source_replication_pending_or_failed(&source_client, source_bucket, witness_key).await?;
+    target_env.restart_server_preserving_data(vec![], &[]).await?;
+    let target_client = target_env.create_s3_client();
+    wait_for_replicated_object(&target_client, target_bucket, witness_key, witness_payload).await?;
+
+    // The scanner demonstrably swept this bucket; the pre-existing key must
+    // still be absent, and stay absent over further cycles.
     assert_replication_key_absent(&target_client, target_bucket, existing_key, Duration::from_secs(6)).await?;
 
     Ok(())
--- a/crates/e2e_test/src/replication_extension_test.rs
+++ b/crates/e2e_test/src/replication_extension_test.rs
@@ -8385,7 +8385,7 @@ async fn test_scanner_never_compensates_when_existing_object_replication_disable
         .body(ByteStream::from_static(witness_payload.as_bytes()))
         .send()
         .await?;
-    wait_for_source_replication_pending_or_failed(&source_client, source_bucket, witness_key).await?;
+    wait_for_source_replication_status(&source_client, source_bucket, witness_key, "FAILED", false).await?;
     target_env.restart_server_preserving_data(vec![], &[]).await?;
     let target_client = target_env.create_s3_client();
     wait_for_replicated_object(&target_client, target_bucket, witness_key, witness_payload).await?;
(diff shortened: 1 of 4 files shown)


houseme (approved):

overtrue:
@reatang plz fix conflicts

crates/replication/src/operation.rs
```diff
@@ -635,6 +635,35 @@ mod tests {
...
+    /// not.
+    #[test]
+    fn resync_target_without_reset_replicates_only_empty_status() {
+        let user_defined = HashMap::new();
+        let object = ReplicationResyncTargetObject {
+            mod_time: Some(OffsetDateTime::UNIX_EPOCH + Duration::seconds(10)),
+            user_defined: &user_defined,
+        };
+
+        for (status, expected) in [
```
overtrue, line 651: The claim says any recorded status must stay out of resync, but COMPLETE is parsed as the reachable CompletedLegacy variant and is missing here. A regression for legacy metadata would pass this table. Please add CompletedLegacy with an expected false result.

crates/e2e_test/src/replication_extension_test.rs
```diff
@@ -6975,3 +6986,213 @@ async fn test_replication_put_and_create_multipart_carry_source_version_id_query
...
+        .bucket(source_bucket)
+        .key(control_key)
+        .body(ByteStream::from_static(control_payload.as_bytes()))
+        .send()
+        .await?;
+    wait_for_replicated_object(&target_client, target_bucket, control_key, control_payload).await?;
+
+    // With the live path proven, the pre-existing key must stay absent across
+    // multiple fast-scanner cycles.
+    assert_replication_key_absent(&target_client, target_bucket, existing_key, Duration::from_secs(6)).await?;
```
overtrue, line 8395: The control object only proves the live replication path is running. If the existing-object scanner is disabled, wedged, or simply has not covered this key during the fixed six-second window, this absence still passes. Please wait for an observable scanner-only witness or completed scan covering this object before asserting the Disabled contract.

overtrue (changes requested):
The new matrix still has two false-green gaps: the Disabled case does not prove a scanner pass occurred, and the no-reset status table omits the reachable legacy completed state.

overtrue (changes requested):
Both test gaps remain on this head: the Disabled case still has no scanner-only witness, and the no-reset matrix still omits CompletedLegacy.

reatang:
Both gaps are fixed on the new head — the false-green analysis was right on target.

**1. `CompletedLegacy` missing from the no-reset table** (`crates/replication/src/operation.rs`)

Added with an expected `false`. Confirmed the variant is reachable exactly as described: `"COMPLETE"` on disk parses to `ReplicationStatusType::CompletedLegacy` (`crates/replication/src/filemeta.rs`), so legacy metadata really had no row. The production decision already handles it (it only replicates `Empty`), so this is pure regression coverage — but the table was claiming more than it proved.

**2. The Disabled case had no scanner-only witness** (`crates/e2e_test/src/replication_extension_test.rs`)

Agreed: the live-path control key proves the *live* path is alive and nothing more, so a disabled/wedged existing-object scanner — or one that simply had not reached the key inside the fixed six-second window — would still have passed.

The test now waits for an observable scanner-only witness before asserting the contract:

- with the target stopped, write a witness object into the **same bucket and prefix** as the pre-existing key, and wait for its source status to reach PENDING/FAILED;
- restart the target and wait for the witness to appear remotely. Re-driving a persisted FAILED object is only done by the data scanner's replication heal pass (the mechanism documented on `FAST_SCANNER_ENV`), and that pass is **not** gated by `ExistingObjectReplication` — so its arrival proves a scan swept this bucket;
- only then assert the pre-existing key is still absent, and stays absent over further cycles.

Because the witness and the pre-existing key share the bucket and prefix, a heal pass that reached the witness necessarily walked the pre-existing key in the same scan — the "completed scan covering this object" you asked for.

Verification: both matrix e2e tests green locally (~11s each, including the target stop/restart and the heal cycle); `cargo test -p rustfs-replication` green.

crates/e2e_test/src/replication_extension_test.rs
overtrue: The witness is still not scanner-only because this helper accepts PENDING, which is stamped before the original live worker finishes. Restarting the target then can let that live attempt deliver the object. Please require persisted FAILED or direct failed-attempt evidence before restart, then use the later delivery as the scanner witness.

overtrue (changes requested):
The legacy status case is fixed, but the witness can still complete through the original live replication attempt, so the Disabled scanner contract remains unproven.

overtrue (approved):
Verified the scanner witness now reaches FAILED before the target restart.

PR #5878, opened 2026-08-09
fix(replication): persist REPLICA status on inbound replication writes

reatang:
## Summary

Surfaced by the P1-20 scanner compensation matrix work (rustfs/backlog#1675, batch B2; companion to #5877). Inbound bucket-replication PUTs and multipart uploads parsed `X-Amz-Replication-Status: REPLICA` into `opts.delete_replication` only — a field the object PUT persistence path never consumes — so the replica status was silently dropped on the target:

- **HEAD/GET on a replicated object returned no `x-amz-replication-status`** (AWS/MinIO parity break for anything observing replica state);
- **the scanner's existing-object resync pass cascaded inbound replicas onward**: with the persisted status missing, a replica looks like a never-replicated existing object, so any outbound rule on the target bucket pushes it downstream (A→B→C topologies). The live-path anti-loop only blocks immediate scheduling, not scanner compensation — each scan cycle re-drives it.

The Snowball auto-extract inbound path already did this correctly (metadata insertion in `object_usecase.rs`); plain PUT/multipart never got the same treatment. The read side needed no change — `ObjectInfo` already derives REPLICA from this metadata key (`object_api/types.rs`) and `must_replicate`'s anti-cascade guard consumes it. The write half was simply missing.

## Fix

`apply_replica_status_from_headers` (shared by the PUT, multipart, and copy option builders; gated on the signature-verified `ReplicateObjectAction` authorization — replication-only headers force that authz in `access.rs`) now also persists the status into `opts.user_defined`, normalizing key casing, mirroring the Snowball path. ~15 lines of production code, one function.

For multipart, the on-disk stamp comes from the same header at initiate time (create-multipart builds options through the same authorized path and persists them into the upload metadata); the completion-time insert feeds the completion-side anti-cascade check.

- Red commit (0f9dd56): `test_scanner_never_cascades_inbound_replicas` — dual-instance A→B with an existing-object rule from B to a third bucket; on main the HEAD precondition shows `replication_status = None` and the replica cascades to the third bucket (journal evidence in the P1-20 investigation).
- Green: HEAD returns REPLICA; the replica never cascades across 5+ fast-scanner cycles while a native control object written to B replicates onward normally (liveness proof for the negative assertion).
- Unit tests pin the persisted key on both builders and that unauthorized clients cannot forge it (forged header → request rejected by the authz gate, not silently ignored).
- fmt / clippy / logging guardrails clean; `rustfs` lib option tests 66 passed.

## Adversarial review

All seven roles ran on the diff: non-blocking. Verified in the process: the PUT persistence chain never strips the key; the key is excluded from the cross-disk quorum-metadata hash (drift cannot split quorum); MinIO stores the same canonical key/value in xl.meta so drop-in replica objects derive correctly; the authorization gate covers all four builder call sites.

## Merge note

Independent of #5864/#5877 in code, but shares the test file tail with #5877 — merge #5877 first for a clean apply (this branch's e2e was deliberately split out of that test-only PR per review agreement: one defect, one PR).

### DIFF
--- a/crates/e2e_test/src/replication_extension_test.rs
+++ b/crates/e2e_test/src/replication_extension_test.rs
@@ -6975,3 +6975,87 @@ async fn test_replication_put_and_create_multipart_carry_source_version_id_query
     target.shutdown().await;
     Ok(())
 }
+
+/// P1-20: inbound replicas never cascade. An object replicated A->B carries
+/// x-amz-replication-status=REPLICA on B; `must_replicate` returns an empty
+/// decision for replicas, so even an ExistingObjectReplication=Enabled rule
+/// configured on B AFTER the replica landed (making it an "existing object"
+/// for that rule) must never push it onward — while B's own native objects
+/// flow to the onward bucket, proving B's outbound replication and scanner
+/// are live.
+#[tokio::test]
+#[serial]
+async fn test_scanner_never_cascades_inbound_replicas() -> TestResult {
+    init_logging();
+
+    let mut env_a = RustFSTestEnvironment::new().await?;
+    let mut env_a_vars = replication_fast_env();
+    env_a_vars.extend_from_slice(LOOPBACK_REPLICATION_TARGET_ENV);
+    env_a.start_rustfs_server_with_env(vec![], &env_a_vars).await?;
+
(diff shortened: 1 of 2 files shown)


houseme (approved):

PR #5880, opened 2026-08-09
fix(rpc): negotiate authenticated file writes

cxymds:
## Related Issues

N/A

## Summary of Changes

- Probe and authenticate peer support before enabling the internode PUT body trailer protocol.
- Keep legacy peers on the original PUT route and send authenticated V1 writes through a route old servers cannot consume.
- Bind V1 writes to the authenticated server epoch and fail closed on peer restart, replacement, malformed responses, or attempted downgrade.
- Share success, legacy, and typed failure probe results across each per-endpoint singleflight wave, including cancellation-safe initializer handoff and stale-generation protection.
- Keep cache hits allocation-free and behind a short shared map lookup; unrelated endpoints probe concurrently.
- Classify capability negotiation under its own low-cardinality operation instead of PUT transfer metrics.

## Verification

- `cargo test -p rustfs-ecstore put_file_capability --lib` (12 passed)
- `cargo test -p rustfs-rio cloned_internode_http_error --lib` (2 passed)
- `cargo test -p rustfs-rio internode_rpc_operation_maps_known_routes --lib` (1 passed)
- `cargo test -p rustfs put_file_ --lib` (8 passed)
- `cargo test -p rustfs authenticated_put_route_checks_server_epoch_before_disk_lookup --lib` (1 passed)
- `cargo test -p rustfs internode_http_operation_maps_only_known_routes --lib` (1 passed)
- `cargo test -p rustfs internode_rpc_subsystem_matches_known_operations --lib` (1 passed)
- `cargo test -p rustfs-io-metrics operation_metric_names_and_low_cardinality_values_are_stable --lib` (1 passed)
- `cargo fmt --all --check`
- `make pre-pr` (12,918 workspace tests passed; Clippy, doctests, and repository guards passed)

## Impact

Mixed-version clusters no longer append the V1 authentication trailer as legacy file data. Rollout remains compatible with old peers through the legacy route; V1 failures never fall back after a write attempt. Concurrent writes now share one failed probe window instead of serializing repeated timeouts, while preserving typed retry classification for every waiter.

## Additional Notes

The isolated resolver A/B used a 4 KiB request context, debug build, seven-round medians, and the same synthetic probe for main and the candidate. It measures capability-resolution behavior rather than end-to-end distributed S3 throughput.

| Concurrency | Scenario | Main median | Candidate median | Main probes | Candidate probes |
| --- | --- | ---: | ---: | ---: | ---: |
| 64 | Cache hit | 14,822 us | 13,536 us | 0 | 0 |
| 64 | TTL refresh | 13,137 us | 13,157 us | 1 | 1 |
| 64 | Unreachable peer | 240,770 us | 4,241 us | 64 | 1 |
| 128 | Cache hit | 29,483 us | 27,360 us | 0 | 0 |
| 128 | TTL refresh | 13,180 us | 13,197 us | 1 | 1 |
| 128 | Unreachable peer | 482,268 us | 4,706 us | 128 | 1 |

Deploy server-first and avoid binary downgrade during the transition.

High-risk adversarial review verdicts:

- Correctness: attacked empty/expired V1 and legacy state, shared failures, epoch replacement, cancellation, retry, and stale completion; no unresolved break found.
- Simplicity: rejected DashMap/Moka, negative error TTL, spawned background tasks, and manual notification in favor of existing locks plus `OnceCell`; no smaller equivalent safe diff found.
- Security: attacked unsigned or malformed probes, proof mutation, downgrade statuses, route confusion, and stale epochs; the exact signed route, V1 pin, and fail-closed behavior held.
- Concurrency/durability: attacked initializer and all-caller cancellation, ABA completion, endpoint contention, and lock-across-await; tests show takeover, recovery, generation fencing, and per-endpoint parallelism, with no storage durability change.
- Compatibility: checked old-to-new, new-to-old, new-to-new, restart, and endpoint replacement paths; legacy and V1 routes remain separated and epoch-bound.
- Performance: checked hot cache allocation/locking, TTL refresh, and 64/128-way failed probes; cache hits remain local and each concurrent failure wave emits one probe.
- Test coverage: each claimed behavior has a reverting test, including the real HTTP/1 handler route and typed failure classification; no uncovered changed branch remained.

---


houseme (changes requested):
The mixed-version routing, authenticated capability proof, epoch binding, and no-downgrade behavior look sound. I am requesting changes for a failure-path performance and availability problem in the capability singleflight.

`resolve_put_file_auth_capability` holds the per-endpoint `RwLock` write guard while awaiting `probe()`. When a probe times out or returns another error, `?` returns without installing a result. Each queued writer then acquires the write lock, sees the same expired/empty state, and starts a fresh probe.

For N concurrent writes to an unreachable endpoint, this turns one 5-second probe timeout into approximately N sequential 5-second waits. This is on the RemoteDisk create/append path, so a short peer outage can unnecessarily serialize and stall all writes targeting that peer.

Please preserve singleflight for failure results too: model an in-flight probe with a shared completion result, so all waiters receive the same success, legacy, or explicit failure outcome. Do not turn a failure into legacy fallback. Add a concurrent timeout/error regression test that proves only one probe runs and that all waiters complete within one probe window.

Please also address the related performance/observability follow-ups in the same update:
- The cached `open_write` fast path currently takes a global `Mutex<HashMap<...>>` and eagerly allocates `endpoint.to_string()` on every call. Avoid the allocation on hits and avoid a process-wide contention point for unrelated endpoints.
- Capability GETs are classified as `put_file_stream`; an expected legacy 404 therefore contributes to PUT error/latency metrics. Give capability negotiation its own operation/metrics, or exclude the expected fallback from transfer metrics.
- Add a 4 KiB/high-concurrency comparison against main covering cache hits, TTL refresh, and an unreachable peer.

houseme (approved):
Reviewed the current head after #5879 landed. The capability negotiation preserves legacy routing, authenticates the V1 response, pins the server epoch, and uses per-endpoint singleflight caching. No blocking functional or performance concern found.

houseme (changes requested):
Updating my review for 8569d6972ecc2f7dd1a4263900d9afc2a3787240: REQUEST_CHANGES.

The prior approval is superseded by this review. The current head still holds the per-endpoint RwLock write guard across probe().await in resolve_put_file_auth_capability. If the cache is empty or expired and the probe times out or errors, the error returns without installing a shared result. Queued writers then acquire the lock one at a time and each launch another probe.

For N concurrent RemoteDisk create/append operations to an unreachable peer, a 5-second probe timeout can therefore become approximately N sequential 5-second waits. This is a tail-latency and recovery-throughput regression on the write path.

Please model an in-flight probe with a shared completion result. All waiters in a probe wave must receive the same V1, legacy, or explicit failure outcome; a transient failure must not downgrade to legacy. Add a deterministic concurrent timeout/error test proving one probe, shared failure, and completion within one probe window.

Please also address the performance follow-ups before approval:
- avoid endpoint.to_string() allocation on global capability-cache hits and remove or narrow the process-wide Mutex contention point;
- prevent expected legacy capability GET 404 responses and probe latency from being classified as PUT transfer errors/latency;
- provide a 4 KiB high-concurrency comparison against main for cache hits, TTL refresh, and an unreachable peer.

CI is not currently merge-ready. The only completed failing job reports E0425 for two_set_test_sets in crates/ecstore/src/core/sets.rs, a file outside this PR; I have not attributed that failure to this patch, while the remaining jobs were cancelled.

cxymds:
Addressed the requested capability singleflight, hot-path, metrics, and benchmark changes in the current head.

- Failure outcomes now participate in the per-endpoint singleflight wave through a cancellation-safe `OnceCell`; 64 concurrent failing callers run one probe and all retain the same retryable `ConnectionRefused` classification.
- Initializer cancellation is taken over by a waiter, all-caller cancellation leaves the flight reusable, a later wave can retry immediately, and generation plus flight identity prevents stale completion from overwriting newer state.
- The global cache uses a short shared lookup and allocates the endpoint key only on miss; unrelated endpoints probe in parallel.
- Capability negotiation has a dedicated signed route and low-cardinality `put_file_capability` operation, so expected legacy 404s no longer pollute PUT transfer metrics.
- A real HTTP/1 handler regression proves the signed capability route is distinct from the legacy PUT route and that epoch validation happens before disk lookup.

The requested isolated 4 KiB resolver A/B (debug build, seven-round medians) produced:

| Concurrency | Scenario | Main | Candidate | Probe count main/candidate |
| --- | --- | ---: | ---: | ---: |
| 64 | Cache hit | 14,822 us | 13,536 us | 0 / 0 |
| 64 | TTL refresh | 13,137 us | 13,157 us | 1 / 1 |
| 64 | Unreachable peer | 240,770 us | 4,241 us | 64 / 1 |
| 128 | Cache hit | 29,483 us | 27,360 us | 0 / 0 |
| 128 | TTL refresh | 13,180 us | 13,197 us | 1 / 1 |
| 128 | Unreachable peer | 482,268 us | 4,706 us | 128 / 1 |

This A/B isolates capability-resolution behavior rather than full distributed S3 throughput. `make pre-pr` is green on the updated branch: Clippy, repository guards, doctests, and 12,918/12,918 workspace tests passed. The branch is also merged with current `origin/main` (`73e4ef4dd`). Please re-review the current head.

houseme (approved):
Re-reviewed `4d019d7ca` after the capability-probe timeout retry fix.

The timeout now retains `InternodeHttpErrorKind::ConnectTimeout`, so `RemoteDisk::open_write_with_retry` classifies it as retryable. The new classification and RemoteDisk retry regressions cover the previous error-path gap.

The prior singleflight, fast-path allocation/contention, capability-metrics, and 4 KiB high-concurrency A/B requirements are addressed in the current head and documented benchmark results. CI is fully green, including the main test/lint, E2E, ILM, io_uring, and S3 suites. No blocking correctness or performance concern remains.

PR #5881, opened 2026-08-09
fix(replication): probe the version-identity contract in replication-check

reatang:
## Summary

P1-19 from the replication compatibility review (rustfs/backlog#1675, batch B2 — data consistency; plan B as recorded in the review: capability probing + explicit signal, not full source→target version mapping). Replication only converges on targets that **adopt the source version id**: version-addressed deletes and heal re-drives address the source id, so a target that mints its own ids (a generic S3 service) silently diverges — deletes no-op, heal re-uploads forever, and nothing surfaces it. `?replication-check` already captured the probe PUT's response version id but never compared it.

Scope note (agreed during batch alignment): the probe lives in the explicit `?replication-check` diagnostic plus a runtime spot-check — `set-remote-target`/`validate_target` are deliberately untouched, so registering a target gains no new write side effects.

## Changes

- **Probe PUT carries `?versionId=`** — the exact shape live replication uses since P0-5 (#5752), and the only shape MinIO consumes. Without it the probe would pass against targets the real data path drifts on (MinIO ignores the internal source-version-id header). Reuses ecstore's `append_version_id_query` through the api facade.
- **New `VersionFidelity` phase** in the check report: the probe PUT's response version id must equal the sent source id. On mismatch: phase FAILED with the machine-readable extension key `"Code": "BucketRemoteTargetVersionMismatch"` (new optional `Code` field on phase statuses; Go decoders ignore unknown keys), target FAILED, the later version-addressed mutation phases SKIPPED, and **cleanup still removes the probe** via the id the target actually assigned (list-based sweep as backstop when the target returns no version id).
- **Runtime spot-check**: `TargetClient::put_object` now returns the assigned version id (mirroring `remove_object`); the replication PUT path audits it — every drifting PUT increments `rustfs_replication_version_identity_drift_total`, and the first drift per target ARN logs a structured warning pointing at `?replication-check`. Drift judgment is a pure function with an exemption-matrix test (empty / literal `"null"` / nil-uuid sources carry no contract — the P0-5 null-version mapping stays exempt).
- `docs/operations/replication-check.md` documents the phase and the code.
- Test infra: FakeS3Target gains `assign_own_version_ids` (models a generic S3 service) and a prefix+max-keys `ListObjectVersions` (probe key allocation requires it).

- Red commit (4a40f53): against a version-minting fake target, `?replication-check` reported `Status "OK"` — the drift was invisible.
- Green: the check fails with the full expected shape (phase, code, skips, cleanup) and the probe leaves no residue; the probe's `?versionId=` query is pinned by a journal assertion (reverting the query hunk alone fails it); probe-level unit tests cover the mismatch/mirror matrix including cleanup addressing the minted id; the existing success e2e asserts `VersionFidelity: OK` against a RustFS target.
- fmt / clippy (rustfs, ecstore, e2e_test) / logging + doc-path guardrails clean; probe unit suite 9 passed.

## Adversarial review

Seven roles on the diff: **non-blocking**. Verified: cleanup addressing chain incl. the no-version-id backstop; MinIO targets pass (they adopt the query id) while pre-P0-5 targets that reject the query were already failing live replication — the check reports reality, not a false positive; the warned-ARN set is bounded by configured targets; `TargetClient.endpoint` carries no credentials. Noted follow-ups (advisory): the multipart runtime path is not audited (the probe phase pins the contract itself), and a reconfigured target only re-warns after a restart.

## Merge note

Independent of #5864/#5877/#5878 in logic; shares the FakeS3Target `stored_versions` accessor with #5864 (identical code — whichever merges second resolves trivially) and the e2e file tail with #5877/#5878.

### DIFF
--- a/crates/e2e_test/src/fake_s3_target/mod.rs
+++ b/crates/e2e_test/src/fake_s3_target/mod.rs
@@ -30,10 +30,10 @@ use s3s::access::{S3Access, S3AccessContext};
 use s3s::auth::SimpleAuth;
 use s3s::dto::{
     AbortMultipartUploadInput, AbortMultipartUploadOutput, CompleteMultipartUploadInput, CompleteMultipartUploadOutput,
-    CreateMultipartUploadInput, CreateMultipartUploadOutput, DeleteObjectInput, DeleteObjectOutput, ETag,
+    CreateMultipartUploadInput, CreateMultipartUploadOutput, DeleteMarkerEntry, DeleteObjectInput, DeleteObjectOutput, ETag,
     GetBucketVersioningInput, GetBucketVersioningOutput, GetObjectInput, GetObjectOutput, HeadBucketInput, HeadBucketOutput,
-    HeadObjectInput, HeadObjectOutput, PutObjectInput, PutObjectOutput, StreamingBlob, Timestamp, TimestampFormat,
-    UploadPartInput, UploadPartOutput,
+    HeadObjectInput, HeadObjectOutput, ListObjectVersionsInput, ListObjectVersionsOutput, ObjectVersionId, PutObjectInput,
+    PutObjectOutput, StreamingBlob, Timestamp, TimestampFormat, UploadPartInput, UploadPartOutput,
 };
 use s3s::service::{S3Service, S3ServiceBuilder};
 use s3s::validation::{AwsNameValidation, NameValidation};
@@ -91,6 +91,7 @@ pub enum Operation {
     GetObject,
     HeadObject,
     DeleteObject,
+    ListObjectVersions,
     CreateMultipartUpload,
     UploadPart,
     CompleteMultipartUpload,
@@ -659,6 +660,7 @@ fn parse_request(method: &Method, uri: &Uri) -> ParsedRequest {
     let operation = match (method, key.is_some()) {
         (&Method::HEAD, false) => Operation::HeadBucket,
         (&Method::GET, false) if query.contains_key("versioning") => Operation::GetBucketVersioning,
+        (&Method::GET, false) if query.contains_key("versions") => Operation::ListObjectVersions,
         (&Method::PUT, true) if upload_id.is_some() && part_number.is_some() => Operation::UploadPart,
         (&Method::PUT, true) if upload_id.is_some() || query.contains_key("partNumber") => Operation::Unknown,
         (&Method::POST, true) if query.contains_key("uploads") => Operation::CreateMultipartUpload,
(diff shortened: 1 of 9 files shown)


houseme (commented):
The probe correctly surfaces a version-identity mismatch, but the replication data path still continues after drift is observed. Please state whether that is an explicit diagnostic-only contract and ensure the high-risk replication gate runs on the final head.

crates/e2e_test/src/fake_s3_target/mod.rs
```diff
@@ -1148,7 +1242,8 @@ impl S3 for FakeBackend {
             ));
         }
 
-        let version_id = new_version_id(&headers)?;
+        let assign_own = lock(&self.store).assign_own_version_ids;
```
overtrue: This function already holds the store mutex in state, so locking self.store again deadlocks every delete that creates a marker. The same nested lock appears in create_multipart_upload. Read assign_own_version_ids from the existing state guard in both places.

crates/ecstore/src/bucket/replication/replication_resyncer.rs
```diff
@@ -2603,6 +2656,13 @@ impl ReplicateObjectInfoExt for ReplicateObjectInfo {
             let result = tgt_client
                 .put_object(&tgt_client.bucket, &object, size, byte_stream, &put_opts)
                 .await
+                .map(|assigned_version_id| {
```
overtrue, line 2927: The audit is only in the non-multipart branch, and replication-check also probes only PutObject. A target can mirror PutObject version IDs but mint one at CreateMultipartUpload, so the check reports OK while multipart deletes and heals silently miss. Please probe and audit the multipart identity path as well, with a target that behaves differently for PUT and multipart.

overtrue (changes requested):
Two paths need changes: the fake target now deadlocks on delete-marker and multipart initiation, and the version-fidelity contract still leaves multipart drift undetected.

overtrue (changes requested):
The fake-target deadlock is fixed, but multipart version-identity drift is still neither probed nor audited.

reatang:
Multipart is now probed and audited — the gap was real, and the fake-target deadlock you caught first is fixed on this head too.

**Multipart identity was unenforced** (`rustfs/src/admin/router.rs`, `crates/ecstore/.../replication_resyncer.rs`)

Exactly as described: the probe only exercised PutObject and the runtime audit sat in the non-multipart branch, so a target that adopts PutObject version ids while minting its own at CreateMultipartUpload passed the check while multipart deletes and heals addressed a version that never existed.

- The check gains a **multipart leg** on the same probe key: initiate carrying the source version as `?versionId=` (the point where the target fixes the version), one 8-byte part, complete, and compare the version the completion reports — completion is the only place the multipart path reveals it. `VersionFidelity` now passes only when **both** paths adopted the source id, and the error names the path that drifted (`... on CreateMultipartUpload`). The multipart version joins the cleanup set; the existing prefix sweep already covers it, and the e2e asserts the target is left empty.
- The **runtime audit** now also runs after `complete_multipart_upload`, so a drifting multipart target is counted and warned about in production, not just in the diagnostic.

Coverage uses a target that behaves differently per path, as you asked: the fake target grew an `assign_own_multipart_version_ids` switch (PutObject mirrors, multipart mints), and `test_replication_check_flags_multipart_only_version_minting_target` asserts FAILED + `BucketRemoteTargetVersionMismatch` + the CreateMultipartUpload mention, with `Put` still OK so the multipart leg is provably what failed. The probe unit tests cover the mirror/mismatch matrix through the new leg.

**@houseme — is drift diagnostic-only?**

Yes, and deliberately so, now stated explicitly rather than implied. The data path continues after drift is observed: this PR is plan B from the review (capability probing plus an explicit signal), not a fail-closed gate. Stopping replication on the first drifting response would turn a target-compatibility problem into an outage for objects that are still being copied correctly, and the drift is not per-object — it is a property of the target. So the contract is: `?replication-check` refuses the target explicitly (machine-readable code), and the runtime path records `rustfs_replication_version_identity_drift_total` and warns once per target ARN pointing at the check. Making it fail-closed is a separate decision that needs an operator-visible switch; I would rather not fold it into this PR.

The high-risk replication gate ran on this head: 11 e2e green (every replication-check case plus the fake-target self-tests), probe unit suite 9 green, `rustfs-ecstore` and `rustfs` clippy/fmt clean, logging and doc-path guardrails clean. The nightly replication lane will exercise the rest on CI.

rustfs/src/admin/router.rs
```diff
@@ -2245,8 +2330,109 @@ async fn put_replication_probe_object(
...
+        .client
+        .upload_part()
+        .bucket(target_bucket)
+        .key(probe_key)
+        .upload_id(&upload_id)
+        .part_number(1)
+        .content_length(8)
+        .body(AwsByteStream::from_static(b"aaaaaaaa"))
+        .send()
+        .await
```
overtrue, line 2409: Once initiation succeeds, either failure here or at completion leaves an incomplete upload. The outer cleanup only lists object versions and delete markers, so it cannot see or abort MPU state; repeated failed checks accumulate parts while cleanup may still report OK. Abort the upload on every post-init failure and cover upload-part and completion failures.

overtrue (changes requested):
Multipart identity coverage is now present, but failed probes can leak incomplete uploads.

overtrue (approved):
The multipart probe now aborts failed uploads, reports cleanup failures separately, and handles ambiguous completion safely.

PR #5882, opened 2026-08-09
fix(site-replication): route state RMW through one locked transaction

reatang:
## Summary

P1-15 PR1 from the replication compatibility review (rustfs/backlog#1675, batch B2 — data consistency; PR2 will migrate the remaining call sites and remove the transitional process mutex). The site-replication state object (`config/site-replication/state.json`, which also carries the retry-event queue) was mutated through read-modify-write sequences with inconsistent locking:

- the **retry-event writers** (enqueue/dequeue), which hang off *every* hook broadcast path (bucket ops, IAM changes), took **no lock at all** — a single process could lose a concurrent locked writer's update (the red commit pins the deterministic interleaving: a stale retry-event persist wipes a committed rotation ack);
- the **RPC-driven service reload** (fanned out cluster-wide) rewrote the whole state file with no lock;
- no writer held a **distributed** lock across the whole RMW, so cross-node writers lost updates even where the process mutex was held.

## Changes

- **New `admin/site_replication_state` module** — the state transaction boundary `with_site_replication_state_lock[_on]`: process mutex + the distributed config-object write lock (the pattern proven by the repair state's `update_site_replication_repair_state`), plus the shared path constant (previously defined twice). The process mutex is transitional until PR2.
- **Typed `update_site_replication_state`** in the handlers: no-lock load / persist-or-clear inside the boundary, normalizing the peer map exactly once (retires the double-clone/double-normalize persist path — P2-22). Migrated call sites: retry-event enqueue, dequeue (lock-free probe first, transaction only on hit — it sits on every successful broadcast), and both pending-rotation/remove ack markers.
- **Service reload** now runs its tolerant byte-level read→normalize→save inside the same boundary via no-lock IO. Normalization semantics deliberately untouched (the tolerant JSON path is the startup/reload repair channel; all six service-side normalization tests unchanged and green).
- **Add/PeerJoin/Edit handlers release the state guard before their peer fan-out** — the adversarial review caught that the transport helpers' retry-event bookkeeping now re-enters the state transaction, and three handlers crossed it while holding the guard (re-entrancy deadlock on any peer send failure). The fix mirrors the Remove/Rotate handlers' existing narrow-guard pattern. The Edit non-refresh branch now commits before fanning out: the old fanout-first order recorded retry events pointing at a state the local site had not saved.
- **ecstore**: `delete_config_no_lock` (+ facade/bridge exports) so the clear half of persist-or-clear works under the held object lock.
- The lock object is the state path itself — the future retry-queue drain (P1-3, batch B3) plugs into the same transaction with no extra lock.

- Red commit (313a70d): deterministic lost-update — replaying enqueue's exact three steps around a completed `mark_pending_rotation_peer_acked` wipes the ack.
- Green: the same test now drives the **real** functions concurrently (8 rounds of enqueue × ack) and asserts every retry event and every ack survives.
- Unit suites: handlers site-replication 171 passed, service 6 passed (normalization semantics pinned). Dual-node e2e green: `test_site_replication_state_edit_fresh_and_stale_real_dual_node`, `test_site_replication_replicates_object_with_bucket_versioning_real_dual_node`. fmt / clippy (rustfs + ecstore) / logging guardrails clean. On-disk format unchanged.

## Adversarial review

Seven roles; **one blocking finding** — the re-entrancy deadlock above — fixed in this PR and re-verified by a second full pass over all 30 lock sites and the complete Add/Join/Edit call graphs (lock-order matrix: process mutex → state object lock, globally consistent; no ABBA against the repair-exec or bucket-targets locks). Non-blocking follow-ups recorded for PR2: a miss-skip return so the ack markers stop persisting on miss (persist-or-clear semantics), the Add handler's pre-existing guard-across-peer-probe availability debt, and a timeout-guarded unreachable-peer regression test for the fan-out paths.

## Merge note

Independent of the other B2 PRs (#5864 / #5877 / #5878 / #5881); two commits (red evidence → fix), squash if per-commit CI green is required.

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -46,7 +46,7 @@ use crate::config::get_config_snapshot;
 use crate::error::ApiError;
 use crate::server::{ADMIN_PREFIX, RemoteAddr};
 use crate::storage::storage_api::{
-    lock_bucket_targets_metadata, read_config_no_lock, save_config_no_lock, with_config_object_read_lock,
+    delete_config_no_lock, lock_bucket_targets_metadata, read_config_no_lock, save_config_no_lock, with_config_object_read_lock,
     with_config_object_write_lock,
 };
 use base64::Engine;
@@ -111,7 +111,9 @@ const LOG_COMPONENT_ADMIN: &str = "admin";
 const LOG_SUBSYSTEM_SITE_REPLICATION: &str = "site_replication";
 const EVENT_ADMIN_SITE_REPLICATION_STATE: &str = "admin_site_replication_state";
 const SERVICE_ACCOUNT_ENVELOPE_VERSION: u64 = 2;
-const SITE_REPLICATION_STATE_PATH: &str = "config/site-replication/state.json";
+use crate::admin::site_replication_state::{
+    SITE_REPLICATION_STATE_LOCK, SITE_REPLICATION_STATE_PATH, with_site_replication_state_lock,
+};
 const SITE_REPLICATION_REPAIR_STATE_PATH: &str = "config/site-replication/repair-state.json";
 const SITE_REPLICATION_REPAIR_EXECUTION_LOCK_PATH: &str = "config/site-replication/repair-execution.lock";
 const SITE_REPL_ADD_SUCCESS: &str = "Requested sites were configured for replication successfully.";
@@ -344,7 +346,7 @@ impl TryFrom<&PeerSite> for PeerConnection {
 
 static SITE_REPLICATION_PEER_CLIENT: LazyLock<Mutex<Option<SiteReplicationPeerClientCache>>> = LazyLock::new(|| Mutex::new(None));
 // Lock order: lifecycle -> bucket operation -> repair admission -> state -> per-bucket metadata.
-static SITE_REPLICATION_STATE_LOCK: LazyLock<Mutex<()>> = LazyLock::new(|| Mutex::new(()));
+// The state mutex now lives in crate::admin::site_replication_state (P1-15).
 static SITE_REPLICATION_LIFECYCLE_LOCK: LazyLock<Mutex<()>> = LazyLock::new(|| Mutex::new(()));
 static SITE_REPLICATION_BUCKET_OP_LOCK: LazyLock<RwLock<()>> = LazyLock::new(|| RwLock::new(()));
 static SITE_REPLICATION_ADD_BOOTSTRAP: LazyLock<StdMutex<Option<SiteReplicationAddBootstrap>>> =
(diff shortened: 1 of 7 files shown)


rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -8937,6 +9016,15 @@ impl Operation for SiteReplicationEditHandler {
...
             } else {
+                // Commit before the peer fan-out (mirrors the add/join
+                // handlers): a failed notification is recorded as a retry
+                // event and converges from the committed local state —
+                // fanning out first meant the retry event pointed at a state
+                // the local site had not saved. Releasing the guard is also
+                // required: the transport helpers' retry-event bookkeeping
+                // re-enters the state transaction (P1-15).
+                save_site_replication_state(&state).await?;
+                drop(state_guard.take());
```
overtrue: This release breaks the ordering previously provided by the state guard. Edit A can save and stall here, edit B can save and reach a remote first, then A arrives last and overwrites B. The peer edit handler accepts the payload without a generation or updated-at check, and a successful stale delivery is not repaired by the retry queue. Please keep delivery ordered or add a version fence that rejects stale edits, with a concurrent regression test.

overtrue (changes requested):
Releasing the edit lock before fan-out allows successful peer edits to be delivered out of order and leaves sites with different state. Please preserve ordering or add stale-update fencing.

rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -14752,4 +14840,66 @@ mod tests {
...
+    /// P1-15 (rustfs/backlog#1675 B2): every state RMW — including the
+    /// retry-event writers that hang off the hook broadcast paths — now runs
+    /// through `update_site_replication_state`, which holds the process
+    /// mutex plus the distributed state-object write lock for the whole
+    /// load -> mutate -> persist. Before the fix the retry writers took no
+    /// lock at all: this concurrent mix deterministically lost one side
+    /// (the red-light commit pinned the exact interleaving).
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    #[serial]
+    async fn test_retry_event_persist_must_not_wipe_concurrent_locked_rmw() {
```
overtrue, line 15374: Both tasks use the same process mutex and the same object lock, so this still passes if either guard is removed and does not cover the cross-node lost-update claim. Please add separate regression coverage for a process-only legacy writer and for two nodes.

overtrue (changes requested):
The edit-ordering race is still present on this head, and the new concurrency test does not isolate either side of the two-lock contract.

reatang:
Both points were right — the ordering one was a regression I introduced, and the test really did not isolate either guard. Fixed on the new head.

**1. Edit ordering** (`rustfs/src/admin/handlers/site_replication.rs`)

Confirmed: releasing the guard before the fan-out traded away the ordering the guard used to provide. A can commit and stall, B can commit and reach a peer first, and A arrives last and wins — the peer edit handler applies whatever arrives (no generation or updated-at fence), and a *successful* stale delivery is never repaired by the retry queue. That is a silent divergence, and it was my deadlock fix that opened it.

The fan-out is back **under** the guard. What genuinely cannot run under it is the retry-event bookkeeping, since that re-enters the state transaction — so the edit branch now delivers with the plain transport and settles the retry queue after the guard is released: every delivered target is dequeued, the first failure is enqueued and returned. Ordering and bookkeeping both preserved, no re-entrancy.

The add handler gets the same treatment: its peer-edit finalize fan-out stays under the guard (it already used the plain transport), and the guard is released only before bootstrap/back-fill, which send bucket-ops rather than peer edits through retry-event transports — those are not ordering-sensitive against a concurrent edit.

**2. The concurrency test proved neither guard**

Also correct: both tasks took the process mutex *and* the object lock, so the test survived removing either one and said nothing about the cross-node claim. It is replaced by two tests that isolate one guard each, and I verified both by mutation rather than by argument:

- `test_transaction_serializes_against_a_process_only_legacy_writer` — one writer uses the legacy shape the ~26 unmigrated call sites still use (process mutex, then plain load/save with their own per-IO locks), the other runs the transaction. Deleting the process mutex from the transaction makes it fail (verified).
- `test_state_object_lock_serializes_writers_from_separate_nodes` — both writers bypass the process mutex, which is what separate nodes do, and drive the **production** object-lock path. I factored `with_site_replication_state_object_lock` out of the boundary precisely so this test exercises production code; replacing that lock with a bare `operation().await` makes it fail (verified).

**On availability**: holding the guard across the peer fan-out is the pre-PR behavior, so this restores rather than adds the "state writes queue behind a slow peer" window. Narrowing it needs the stale-update fence you mention (a generation on the peer-edit payload plus a per-source last-applied record on the receiving side), which I would rather do as its own change than fold into the lock consolidation.

Verification: handlers 173 + service 6 unit tests green; dual-node and three-node site-replication edit e2e green; arch/layer/logging guardrails, fmt, clippy clean.

rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -8937,21 +9047,52 @@ impl Operation for SiteReplicationEditHandler {
...
             } else {
-                for target in remote_targets {
+                // Commit before the peer fan-out (mirrors the add/join
+                // handlers): a failed notification is recorded as a retry
+                // event and converges from the committed local state —
+                // fanning out first meant the retry event pointed at a state
+                // the local site had not saved.
+                save_site_replication_state(&state).await?;
+
+                // The fan-out stays UNDER the state guard: releasing it here
```
overtrue: SITE_REPLICATION_STATE_LOCK is process-local, so this ordering holds only when both admin edits hit the same node. Edits accepted concurrently by two nodes can still save and fan out in opposite order, while the peer handler has no generation fence. Please use distributed ordering or a fencing token and cover two-node delivery.

rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -8937,21 +9047,52 @@ impl Operation for SiteReplicationEditHandler {
...
+                            failure = Some((target.clone(), err));
+                            break 'fanout;
+                        }
                     }
+                    delivered.push(target.clone());
+                }
+                drop(state_guard.take());
+
+                for target in &delivered {
+                    dequeue_site_replication_retry_event(target, SITE_REPLICATION_PEER_EDIT_PATH).await;
```
overtrue: After dropping the guard, this dequeue can erase a newer failure. If edit A succeeds and pauses here, edit B can commit, fail delivery, and enqueue the same peer/path; when A resumes, it deletes the event for B. Local state is B, the peer is A, and no retry remains. Please fence settlement by edit generation or id and cover this interleaving.

overtrue (changes requested):
The new ordering is still process-local, and retry settlement can discard a newer failed edit.

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in fc0786d5c — you are right that holding the process mutex across the fan-out orders nothing beyond the node that took it.

Edits now carry a fencing token. `SiteReplicationState::edit_generation` is incremented inside the same commit as the edit (`next_peer_edit_generation` runs in the state transaction, i.e. under the distributed state-object lock), so two nodes of this site can never take the same generation. It travels on the peer-edit request as `?editOrigin=<deployment id>&editGeneration=<n>`, and `SRPeerEditHandler` drops a delivery whose generation is at or below `applied_edit_generations[origin]` — acking it rather than erroring, since the newer edit already landed and there is nothing for the sender to retry. The mark is written in the same commit as the edit it fences, so a crash between the two cannot let the superseded delivery apply on a later attempt.

Compatibility: a peer that predates the fence sends no query parameters, `peer_edit_fence` returns `None`, and the edit is applied exactly as before — mixed-version meshes keep the current behaviour rather than failing closed.

One thing the fence needed on top: a site that leaves the mesh drops below two peers, which clears its state object and restarts its counter at zero. A leftover mark would then reject every edit it sends after rejoining, so `parse_site_replication_state` drops marks for deployment ids that are no longer peers (`peer_edit_marks_do_not_outlive_the_peer_that_earned_them`).

Coverage:
* `test_peer_edit_generations_are_unique_across_nodes` — two writers that bypass the process mutex (what two nodes actually are) allocate 16 generations; they are unique only because the allocation is inside the transaction. Mutation-checked: dropping the object lock from `update_site_replication_state_as_separate_node` produces `[1, 1, 2, 2, ...]` and the test fails.
* `peer_edit_fence_rejects_a_delivery_the_newer_edit_already_passed` — the path round-trip plus the staleness predicate at the boundaries (older / equal / newer generation, a second origin site, and both unstamped-sender forms).
* `test_sr_peer_edit_handler_uses_site_replication_operation_action` now also pins the handler wiring — rejecting and recording — plus the sender allocating inside the committed state.

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in fc0786d5c — the interleaving you describe is real: a success only proves the peer reached the state *that* delivery carried, so removing every matching event also removed a newer edit's retry, leaving local state on B, the peer on A, and nothing queued to converge them.

Settlement is now fenced by the same edit generation. `SiteReplicationRetryEvent` records the generation whose delivery failed, and `settle_site_replication_retry_events` keeps a matching event when its recorded generation is newer than the one being settled. `upsert_site_replication_retry_event` takes `max` of the recorded and incoming generation, so an older delivery failing afterwards cannot lower the fence and let its own success clear B's event.

The broadcast paths keep settling unconditionally (`generation = None`): their retry events live under their own paths and never collide with a peer-edit delivery, so fencing them would need a pre-read on every successful hook broadcast for no behavioural gain.

`retry_settlement_must_not_erase_a_newer_generation_failure` pins exactly your interleaving: enqueue for generation 6, settle for 5 (must remove nothing), then settle for 6 (must remove it). Mutation-checked: make `settle_site_replication_retry_events` unconditional again and the test fails on the first assertion.

PR #5891, opened 2026-08-09
fix(docker): support TZ environment variable

### LINKED ISSUE #5843: Allow setting timezone on Docker container via environment variable
It is very useful to be able to set the timezone of a Docker container when creating it, however this is not possible with RustFS containers. This makes reading logs and status messages difficult as they are all displayed in UTC as opposed to the local timezone.

If an image contains the `tzdata` package then it is possible to simply define the `TZ` environment variable and the [timezone inside the container will be updated](https://stackoverflow.com/questions/57607381/how-do-i-change-timezone-in-a-docker-container). So I would like to request that this package be added to the list of installed packages in the project's Dockerfile(s).

**Additional context**
Considering that the server status endpoint at `/health/ready` returns a timestamp which includes the timezone, it seems strange that this cannot be easily set when creating a new instance with Docker.

```json
{
  "status": "ok",
  "ready": true,
  "service": "rustfs-endpoint",
  "timestamp": "2026-08-08T10:51:29.186032284Z[Etc/Unknown]",
  "version": "1.0.0-beta.12",
  "details": {
    "storage": {
      "status": "connected",
      "ready": true
    },
    "iam": {
      "status": "connected",
      "ready": true
    },
    "lock": {
      "status": "connected",
      "ready": true
    }
  },
  "degradedReasons": []
}
```

RamakrishnaChilaka:
## Related Issues

Fixes #5843.

## Summary of Changes

- Install `tzdata` in the published Alpine and Ubuntu runtime images so `TZ` can select an IANA timezone.
- Verify IANA timezone resolution while building each runtime image.
- Add a regression guard covering both final Dockerfile stages.

## Verification

- `./scripts/test_docker_runtime_timezone.sh`
- `make pre-pr`

## Impact

Docker deployments can set `TZ` to an IANA timezone such as `Asia/Kolkata`. Existing deployments that do not set `TZ` retain their current default behavior. Runtime images gain the `tzdata` package.

## Additional Notes

`Dockerfile.source` already installs `tzdata`; this change aligns both published binary image variants.

---

### DIFF
--- a/.config/make/tests.mak
+++ b/.config/make/tests.mak
@@ -25,6 +25,7 @@ TEST_THREADS ?= 1
 script-tests: ## Run shell script tests
 	@echo "Running script tests..."
 	./scripts/test_build_rustfs_options.sh
+	./scripts/test_docker_runtime_timezone.sh
 	./scripts/test_entrypoint_credentials.sh
 	./scripts/test_internode_grpc_ab_bench.sh
 	./scripts/test_object_batch_bench_enhanced.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,7 +91,12 @@ LABEL name="RustFS" \
 # Upgrade base-image packages so published images pick up security fixes
 # (e.g. openssl/libssl3 CVEs) without waiting for a new Alpine point release.
 RUN apk upgrade --no-cache && \
-    apk add --no-cache ca-certificates coreutils curl
+    apk add --no-cache \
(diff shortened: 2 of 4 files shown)


RamakrishnaChilaka:

overtrue (approved):
Looks good. Both published runtime images now install tzdata, and the regression guard checks the final stages and IANA timezone resolution.

PR #5897, opened 2026-08-09
perf(get): reduce response body chunk overhead

houseme:
## Related Issues

## Summary of Changes
- Replace the generic `ReaderStream` wrapper used by streamed GetObject responses with a bounded GetObject-specific stream that reads at most the remaining response length for each body chunk.
- Avoid creating a per-poll `Instant` when GetObject output handoff attribution is disabled.
- Surface an unexpected short EOF from the response stream as an explicit stream error instead of a clean end of body.
- Add cached, low-cardinality HTTP response body chunk size, chunk latency, and stream duration metrics for server-side tail attribution.

## Verification
- `cargo fmt --all --check`
- `cargo test -p rustfs get_object_reader_stream --lib`

## Impact
- Streaming GetObject responses preserve the same `Content-Length` contract and now bound each read by the remaining response bytes, reducing unnecessary final-chunk buffer pressure and making short reads fail loudly.
- New metrics are low-cardinality and use cached metric handles on the response body hot path. The existing response body byte counter remains unchanged.
- No configuration or API changes. Four-node performance validation has not been run for this PR yet.

## Additional Notes
Adversarial validation summary:
- Correctness: attacked exact-length, final-partial-chunk, zero-remaining, and short-EOF cases; added tests for bounded read capacity and short EOF.
- Simplicity: kept the change local to the existing GetObject body handoff and HTTP trace hooks; no new cross-crate abstraction.
- Concurrency/performance: no new locks, awaits, disk I/O, or high-cardinality labels on the hot path; metric handles are cached and gated by the existing metrics switch for the new histograms.
- Compatibility: no S3 API/header format changes; short EOF now surfaces as a stream error instead of a clean body end.
- Test coverage: the new tests fail if the stream asks for the full capacity when fewer bytes remain, or if a short reader ends cleanly.

### DIFF
--- a/rustfs/src/app/object_usecase.rs
+++ b/rustfs/src/app/object_usecase.rs
@@ -115,7 +115,7 @@ use crate::delete_tail_activity::{DeleteTailActivityGuard, DeleteTailStage};
 use crate::error::ApiError;
 use crate::server::convert_ecstore_object_info;
 use crate::table_catalog;
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use futures::{Stream, StreamExt, TryStreamExt};
 use http::{HeaderMap, HeaderValue, StatusCode};
 use md5::{Digest as Md5Digest, Md5};
@@ -198,7 +198,9 @@ use time::{OffsetDateTime, format_description::well_known::Rfc3339};
 use tokio::io::{AsyncRead, ReadBuf};
 use tokio::sync::{OwnedSemaphorePermit, RwLock};
 use tokio_tar::Archive;
-use tokio_util::io::{ReaderStream, StreamReader};
+#[cfg(test)]
+use tokio_util::io::ReaderStream;
+use tokio_util::io::{StreamReader, poll_read_buf};
 use tracing::{debug, error, instrument, warn};
 use uuid::Uuid;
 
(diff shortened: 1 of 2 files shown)


rustfs/src/app/object_usecase.rs
```diff
@@ -1331,12 +1337,20 @@ where
...
+        let reader = match this.reader.as_mut().as_pin_mut() {
+            Some(reader) => reader,
+            None => return Poll::Ready(None),
+        };
+        let read_capacity = (*this.capacity).min(*this.remaining);
+        if this.buf.capacity() < read_capacity {
+            this.buf.reserve(read_capacity - this.buf.capacity());
+        }
+
+        let result: Poll<Option<Self::Item>> = match poll_read_buf(reader, cx, &mut *this.buf) {
```
overtrue: BytesMut::reserve guarantees at least the requested additional capacity; it does not cap the buffer at read_capacity. After split(), a later reserve can restore the original allocation. For example, capacity 64, expected 65, and a 66-byte reader can yield 64 then 2 bytes, so the remaining subtraction underflows and the stream may emit past Content-Length. Please present an actually bounded writable slice here and cover this multi-chunk case.

overtrue (changes requested):
The new stream can read past the declared final chunk and underflow its remaining-length accounting. Please cap the actual read buffer and add a multi-chunk overlong-reader regression test.

overtrue (approved):

PR #5914, opened 2026-08-10
fix(s3): return InvalidArgument for mismatched ListMultipartUploads key-marker

ylw510:
fix(s3): return InvalidArgument for mismatched ListMultipartUploads key-marker

A key-marker that does not start with the request prefix is invalid input, not an unimplemented feature.

## Related Issues
N/A

## Summary of Changes
When `ListMultipartUploads` is called with a `key-marker` that does not start with the request `prefix`, RustFS previously returned `NotImplemented`. That is incorrect: this is invalid client input, not a missing feature.

This change returns `InvalidArgument` instead, consistent with other parameter validation in the same path (e.g. `max-uploads`), and updates the related unit tests.

## Verification
- Updated assertions in:
  - `rustfs/src/storage/s3_api/multipart.rs`
  - `rustfs/src/app/multipart_usecase.rs`
- Intended local check:
  - `cargo test -p rustfs --lib rejects_invalid_key_marker`
- Local run was blocked by dependency fetch TLS (`datafusion`); relying on CI for full verification.

## Impact
S3 clients that pass a `key-marker` outside the given `prefix` now receive `InvalidArgument` instead of `NotImplemented`. No change to successful list/pagination behavior.

## Additional Notes
N/A


houseme (approved):

houseme:
> 
> Missing signatures:
> 
> * @ylw510
> 
> 
> 
> CLA document: https://github.com/rustfs/cla/blob/main/cla/v2.md

@ylw510

ylw510:

PR #5915, opened 2026-08-10
fix(iam): merge OIDC extra root CAs

### LINKED ISSUE #5908: SSL_CERT_FILE replaces the entire TLS trust store instead of augmenting it, silently breaking outbound HTTPS to unrelated hosts
### Summary

Setting `SSL_CERT_FILE` to trust one internal/self-signed CA (e.g. a Kubernetes cluster's own apiserver CA, for the built-in Kubernetes OIDC workload-identity provider) causes RustFS to reject TLS connections to *every other* HTTPS destination — including completely unrelated OIDC providers, S3 targets, or webhook endpoints — because the trust store is replaced rather than extended.

### Environment

- RustFS version: `1.0.0-beta.12-preview.1` (Helm chart `0.12.0`)
- Deployment: single-node Kubernetes, standalone mode

### Steps to reproduce

1. Configure the built-in (`default`) OIDC provider against an internally-signed issuer, e.g.:
   ```
   RUSTFS_IDENTITY_OPENID_CONFIG_URL=https://kubernetes.default.svc.cluster.local/.well-known/openid-configuration
   SSL_CERT_FILE=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
   ```
2. Configure a second, named OIDC provider against a public, well-known-CA-signed issuer (e.g. Authentik behind Let's Encrypt/Cloudflare):
   ```
   RUSTFS_IDENTITY_OPENID_CONFIG_URL_myidp=https://auth.example.com/application/o/rustfs/.well-known/openid-configuration
   ```
3. Restart RustFS and observe OIDC discovery for `myidp`.

### Expected

Discovery for `myidp` succeeds — its certificate chains to a publicly trusted root, unrelated to the internal CA configured for the `default` provider.

### Actual

Discovery for `myidp` fails with a generic transport error (`error sending request for url ...`). Confirmed via `wget` against the exact same discovery URL *from inside the same pod* succeeding immediately (a different TLS stack, unaffected by `SSL_CERT_FILE`), while RustFS's own process fails consistently. Removing `SSL_CERT_FILE` entirely fixes the second provider immediately, with no other change.

### Why this matters

`SSL_CERT_FILE` is presented (implicitly, by working at all for the internal CA case) as the way to add trust for a custom/internal CA. In practice it *replaces* the default trust bundle for the whole process, so it's only safe to use if literally every outbound HTTPS destination RustFS talks to shares that one CA — which won't be true in any multi-provider OIDC setup, and likely isn't true for other outbound integrations (webhooks, S3 replication targets, etc.) either.

### Suggested fix

Either:
- Merge the specified cert file into the default trust store rather than replacing it (i.e. treat it as "trust this too," not "trust only this"), or
- Document the full-replacement behavior prominently and provide a supported way to layer an internal CA onto the default bundle (e.g. reading `SSL_CERT_FILE` *in addition to* the compiled-in default set, or supporting a separate `RUSTFS_EXTRA_CA_CERT` env var).

houseme:
## Related Issues
Fixes #5908.

## Summary of Changes
- Scope `RUSTFS_EXTRA_CA_CERT` to OIDC outbound TLS instead of publishing it as process-wide outbound root material, preserving native/public roots for unrelated TLS clients such as tier targets.
- Add a reload-aware OIDC extra root CA provider that merges the current published RustFS outbound TLS roots with the OIDC-only extra CA bundle and refreshes parsed reqwest certificates when the material generation changes.
- Validate configured OIDC extra CA bundles with `rustls::RootCertStore::add`, so malformed DER inside valid PEM framing fails closed during startup/admin validation instead of disabling OIDC later.
- Reuse the same OIDC CA material path for startup discovery and admin provider validation, including TLS reload generations and `RUSTFS_EXTRA_CA_CERT`.
- Add HTTPS OIDC coverage for static extra roots, generation rotation from CA A to CA B, malformed DER rejection, invalid PEM rejection, and isolation from global outbound TLS roots.

## Verification
- `cargo test -p rustfs oidc_extra_ca_cert_loads_configured_bundle --offline`
- `cargo test -p rustfs oidc_extra_ca_cert_rejects_invalid_pem --offline`
- `cargo test -p rustfs oidc_extra_ca_cert_rejects_malformed_der_certificate --offline`
- `cargo test -p rustfs load_tls_material_does_not_append_oidc_extra_ca_cert --offline`
- `cargo test -p rustfs-iam oidc_discovery_accepts_extra_root_ca_for_https_provider --offline`
- `cargo test -p rustfs-iam oidc_discovery_refreshes_extra_root_ca_when_generation_changes --offline`
- `cargo test -p rustfs-iam test_validate_oidc_provider_config --offline`
- `cargo fmt --all --check`
- `git diff --check`
- `cargo clippy -p rustfs-iam --all-targets --offline -- -D warnings`
- `cargo clippy -p rustfs -p rustfs-protocols --all-targets --features swift --offline -- -D warnings`
- `cargo clippy -p rustfs -p rustfs-protocols --all-targets --features sftp --offline -- -D warnings`
- `cargo clippy -p rustfs -p rustfs-protocols --all-targets --features rio-v2 --offline -- -D warnings`
- `make pre-commit`

## Impact
Operators should use `RUSTFS_EXTRA_CA_CERT=/path/to/ca-bundle.pem` for private or self-signed OIDC issuer roots. Existing `SSL_CERT_FILE` behavior is unchanged, but it should no longer be needed as a workaround for OIDC private CAs. Invalid OIDC extra CA bundles now fail closed during TLS material initialization or admin validation. The OIDC-only bundle no longer changes the trust roots used by unrelated outbound TLS clients.

## Additional Notes
High-risk adversarial validation summary:
- Correctness: attacked no-TLS-path extra CA, TLS-path coexistence, admin validation, and malformed PEM/DER paths; no break found after OIDC scoping and DER root-store validation.
- Simplicity: attacked duplicate cache setup and patch layering; reduced repeated cache initialization and kept one OIDC CA material path for startup/admin validation.
- Security: attacked trust expansion, SSL_CERT_FILE replacement semantics, OIDC private CA trust, and secret/log exposure; extra trust is scoped to OIDC and logs only path metadata.
- Concurrency/durability: attacked TLS reload generation, token/JWKS refresh after CA rotation, and lock holds; OIDC reloads certs outside cache locks and rotation coverage passes.
- Compatibility: attacked public-root preservation and unrelated TLS consumers; global outbound roots no longer receive OIDC-only extra CA material.
- Performance: attacked per-request file reads and certificate parsing; parsing is generation-cached and no storage/tier hot path consumes the OIDC-only bundle.
- Test coverage: revert tests exist for static HTTPS discovery, CA generation rotation, malformed DER, invalid PEM, and global outbound-root isolation.

Review feedback addressed:
- Scoped `RUSTFS_EXTRA_CA_CERT` to OIDC to avoid breaking unrelated public TLS clients.
- Refreshed OIDC transport certificates when outbound/OIDC CA material generation changes.
- Validated X.509 DER through the root store and added malformed-DER coverage.

---


rustfs/src/startup_tls_material.rs
```diff
@@ -56,11 +45,34 @@ pub(crate) async fn init_outbound_tls_material(config: &Config) -> Result<()> {
         if startup_runtime_sources::observability_metric_enabled() {
             startup_runtime_sources::init_tls_metrics();
         }
+    } else if let Some(outbound) = crate::server::tls_material::load_standalone_outbound_tls_material()
+        .await
+        .map_err(|err| Error::other(err.to_string()))?
+    {
+        publish_outbound_tls_material(&outbound, None).await;
```
overtrue: Publishing this extra-only bundle as the process-wide root set breaks the additive contract for existing consumers. Without `RUSTFS_TLS_PATH`, TransitionClient takes its custom-root branch and replaces native roots, so configuring a private OIDC CA also breaks public AWS/R2/etc tier TLS. Please either scope this CA to OIDC or preserve native roots in every consumer, with coexistence coverage.

crates/iam/src/oidc.rs
```diff
@@ -678,7 +714,12 @@ fn trusted_aud(other_audiences: &[String], audience: &Audience) -> bool {
 impl OidcSys {
     /// Parse environment variables and discover all configured OIDC providers.
     pub async fn new() -> Result<Self, String> {
-        let http_client = ReqwestHttpClient::new()?;
+        Self::new_with_extra_root_ca(None).await
+    }
+
+    /// Parse environment variables and discover providers with an additional outbound root CA bundle.
+    pub(crate) async fn new_with_extra_root_ca(root_ca_pem: Option<&[u8]>) -> Result<Self, String> {
+        let http_client = ReqwestHttpClient::new_with_extra_root_certs(oidc_extra_root_certs(root_ca_pem)?)?;
```
overtrue, line 823: This snapshots roots for the lifetime of `OidcSys`, while TLS reload publishes newer generations. Token exchange and JWKS refresh keep using the original client, so rotating CA A to B rejects B and still trusts A until restart even though admin validation sees B. Please refresh the OIDC transport when the outbound generation changes and add a rotation test.

rustfs/src/server/tls_material.rs
```diff
@@ -267,6 +285,47 @@ fn map_runtime_tls_error(err: rustfs_tls_runtime::TlsRuntimeError) -> TlsMateria
...
+        path = ?path,
+        "TLS root CA set enriched"
+    );
+    Ok(true)
+}
+
+fn validate_cert_bundle(data: &[u8], path: &Path) -> Result<(), TlsMaterialError> {
+    let mut reader = Cursor::new(data);
+    let mut found = false;
+    for cert in CertificateDer::pem_reader_iter(&mut reader) {
```
overtrue, line 313: `pem_reader_iter` validates PEM framing, not the X.509 DER inside it. A `CERTIFICATE` block containing base64 for `not a cert` passes here; reqwest later rejects it while building OIDC requests, and startup continues with no providers. Validate the certificate in the actual root store/client here and cover malformed DER inside valid PEM framing.

overtrue (changes requested):
The extra CA path currently breaks unrelated public TLS clients, does not follow hot reloads in OIDC, and accepts malformed DER that later disables OIDC. These need to fail closed and preserve the additive trust contract.

Issue #5869, opened 2026-08-08
Automatic heal after drive replacement reports completion while objects are left a shard short — every health signal green, redundancy silently below EC guarantee


terem42:
## Summary

The official documentation ([Hard Drive Failures](https://docs.rustfs.com/en/troubleshooting/driver)) makes three promises about drive replacement:

> "RustFS ensures read/write access can still be provided when some disks fail through mechanisms similar to erasure coding, and **automatically heals data after replacing disks**."

> "Once RustFS detects a freshly formatted disk at a known mount point, its background scanner heals the missing data onto it automatically — **no manual command is required**."

> "The healing process will complete in the background [...] **After healing is complete, the tool will report success or list failed objects.**"

In our testing the third promise is the one that breaks, and it breaks in the worst direction: the automatic rebuild starts correctly, runs for a few minutes, then **stops with objects still missing a shard on the replacement drive — while reporting success and listing no failed objects**. Every signal an operator has confirms completion: `healing.bin` is removed from the drive, the heal queue reports `idle` with 0 queued and 0 active tasks, `onlineDisks` shows the full count, the console shows the disk healthy, and every object still reads back byte-perfect (via parity reconstruction). Nothing anywhere indicates that the affected objects are now one drive failure closer to loss than the EC configuration promises — and the documentation explicitly tells the operator no further action is needed.

## Observed

3-node / 12-drive cluster, one erasure set, EC parity 4. Realistic replacement path: a data disk detached from a running node, a blank disk attached in its place, node rebooted, disk formatted and mounted at the original mount point — exactly the documented procedure.

Reproduced 4 times in 6 attempts across three builds, including the current release:

- `1.0.0-beta.12`, twice out of two attempts: the replacement refills at full speed for ~3 minutes, then stops. Final census: 11 drives × 86 `part.1` shards, the replacement 76 — ten objects at 11/12 shards, parity margin 3 of 4. The gap then stays byte-constant for 12+ minutes while the drive tracks new writes perfectly — it keeps up with fresh data and never resumes the backlog.
- A main-derived build, once out of three attempts (intermittent): replacement at 76 shards vs 88, and on this occurrence **an explicit recursive deep heal of every bucket did not close the gap either** — post-heal census still uneven. (On the other occurrences the explicit deep heal did repair everything in ~40 s, which makes it a workaround that usually works — but demonstrably not always.)
[...]

PR #5918, opened 2026-08-10
test(heal): add privileged replacement rebuild e2e

houseme:
## Related Issues
Related to rustfs/backlog#1791 and rustfs/rustfs#5869.

## Summary of Changes
Adds a privileged Linux-only E2E that exercises the automatic replacement path in a 3 node x 4 drive cluster without issuing an Admin deep-heal request. The test launches itself inside a private mount namespace through the system `unshare` command, mounts every drive as an independent tmpfs, detaches a target drive at runtime, waits for replacement recovery status, mounts a blank replacement at the same endpoint, restarts the affected node, and proves recovery with a per-version physical census of `xl.meta` and `part.N` on the replacement target.

The test is ignored by default and requires `RUSTFS_PRIVILEGED_REPLACEMENT_E2E=1` plus root or CAP_SYS_ADMIN. It deliberately removes `RUSTFS_UNSAFE_BYPASS_DISK_CHECK` from the harness environment so production mount readiness and durable replacement admission are exercised. The implementation avoids Rust `unsafe` blocks; privileged namespace and mount operations are delegated to standard Linux tooling.

## Verification
- `cargo fmt --all --check`
- `git diff --check`
- `RUSTFLAGS="-D unsafe-code" cargo check -p e2e_test --tests`
- `cargo clippy -p e2e_test --tests -- -D warnings`

## Impact
No runtime impact. This adds ignored Linux privileged E2E coverage only. Normal CI compiles the test; execution requires an explicit privileged Linux runner with `unshare`, `mount`, and `umount` available.

## Additional Notes
The intended privileged runner command is:

```bash
sudo env RUSTFS_PRIVILEGED_REPLACEMENT_E2E=1 cargo nextest run -p e2e_test --run-ignored ignored-only -E 'test(/^replacement_privileged_e2e_test::/)'
```

This PR intentionally keeps #1791 test infrastructure separate from #1790 status protocol work.


crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -0,0 +1,500 @@
...
+            tick.tick().await;
+        }
+    }
+
+    async fn run_replacement_e2e(parity: usize, test_name: &str) -> Result<(), Box<dyn Error + Send + Sync>> {
+        init_logging();
+        if !privileged_run_enabled()? {
+            return Ok(());
+        }
+        if std::env::var_os(NAMESPACE_ENV).is_none() {
```
overtrue, line 742: This environment variable is not proof that the test is in a private mount namespace. If it is inherited or set manually, the privileged test skips unshare and then runs mount --make-rprivate / plus all tmpfs mounts in the host namespace. Please verify that the current mount namespace is isolated before any mount operation instead of trusting the sentinel.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -0,0 +1,500 @@
...
+        verify_bodies(&clients[0], &versions).await?;
+
+        mount_ns.detach(&target_disk)?;
+        let missing_after_detach = incomplete_versions(&target_disk, &versions)?;
+        assert_eq!(
+            missing_after_detach.len(),
+            versions.len(),
+            "detached target path must not still expose baseline shards"
+        );
+        wait_for_replacement_state(&cluster, &target_disk, &["waiting_for_replacement", "running"], 180).await?;
```
overtrue: Accepting running lets the test move on at exactly the state the regression produces. It immediately stops the node and installs the blank disk, so a broken scanner can enter running while the target is absent and the normal post-restart heal still makes the final census pass. Please keep the target absent across a checkpoint opportunity and assert that it never runs, completes, or reports false idle during that window.

overtrue (changes requested):
The privileged scenario can still mutate the host mount namespace when its internal sentinel is set, and it does not keep the drive absent through a checkpoint opportunity to prove the replacement admission gate.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
overtrue: The 20-second window still proves only that the initial waiting record stayed unchanged. A delayed scanner can leave it untouched until after the window, so the admission/checkpoint regression still passes. Please wait for an observable scanner or checkpoint generation while the disk is absent, then assert the target never leaves waiting.

overtrue (changes requested):
The mount-namespace guard is fixed, but the absent phase still has no observable scanner or checkpoint witness, so the required regression can remain untested.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -0,0 +1,694 @@
...
+            .to_str()
+            .ok_or_else(|| format!("tmpfs target path is not UTF-8: {target:?}"))?;
+        run_command("mount", ["-t", "tmpfs", "-o", MOUNT_SIZE, label, target])
+    }
+
+    fn detach_mount(target: &Path) -> Result<(), Box<dyn Error + Send + Sync>> {
+        let target = target
+            .to_str()
+            .ok_or_else(|| format!("umount target path is not UTF-8: {target:?}"))?;
+        run_command("umount", ["-l", target])
```
overtrue: umount -l does not make this disk absent to RustFS. LocalDisk keeps the mount alive through its open lease fd and all disk I/O uses /proc/self/fd/<fd>/., so the node can still read the old tmpfs after this returns; only the pathname census sees it gone. Please use a fault that makes the held mount or device inaccessible and prove the scanner observes that state.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -0,0 +1,694 @@
...
+        loop {
+            let status = replacement_status(cluster).await?;
+            let states = target_record_states(&status, target_disk);
+            if states.len() != 1 || !states.contains("waiting_for_replacement") {
+                return Err(format!(
+                    "replacement target must stay waiting while absent; observed states {states:?} in status {status}"
+                )
+                .into());
+            }
+            let witness = replacement_intent_witness(cluster, target_disk)?;
```
overtrue: On a genuinely absent path, this witness cannot exist: HealTask checks replacement_targets_ready before new_replacement_intent, and readiness rejects an absent or non-mount endpoint. The intent is admission state, not evidence of an absent scanner pass. Please wait on an explicit scanner generation or observation instead of requiring this intent and a fixed 20-second window.

overtrue (changes requested):
The privileged scenario still does not produce a real absent disk, and its new witness cannot exist on the guarded path.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -0,0 +1,687 @@
...
+        let deadline = Instant::now() + Duration::from_secs(timeout_secs);
+        let mut tick = interval(Duration::from_secs(1));
+        let target = target_disk.to_string_lossy();
+        loop {
+            let log = match fs::read_to_string(log_path) {
+                Ok(log) => log,
+                Err(error) if error.kind() == std::io::ErrorKind::NotFound => String::new(),
+                Err(error) => return Err(format!("failed to read target node log {log_path:?}: {error}").into()),
+            };
+            if log.contains("replacement_path_unavailable") && log.contains(target.as_ref()) {
```
overtrue: This log is emitted mid-scan, before durable-recovery discovery and queue admission finish. Returning here lets the test check empty state and kill the node while that same scan is still running, so removing the production `continue` after this event can still pass if the test wins the race. Please wait for a post-scan generation or barrier before asserting zero side effects.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -0,0 +1,687 @@
...
+        cluster: &RustFSTestClusterEnvironment,
+        target_disk: &Path,
+        states: &[&str],
+        timeout_secs: u64,
+    ) -> Result<serde_json::Value, Box<dyn Error + Send + Sync>> {
+        let deadline = Instant::now() + Duration::from_secs(timeout_secs);
+        let mut tick = interval(Duration::from_secs(1));
+        loop {
+            let status = replacement_status(cluster).await?;
+            if target_record_has_state(&status, target_disk, states) {
```
overtrue: Please require `cluster.definitive == true` before trusting these records. The status API deliberately returns partial local records with `definitive: false` when a peer is unavailable or unsupported, so one reachable node's `completed` record can currently satisfy this predicate without an authoritative cluster result. The absent-state check needs the same guard.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -0,0 +1,687 @@
...
+
+    fn incomplete_versions(
+        target_disk: &Path,
+        versions: &[BaselineVersion],
+    ) -> Result<BTreeSet<String>, Box<dyn Error + Send + Sync>> {
+        let mut missing = BTreeSet::new();
+        for version in versions {
+            let actual =
+                census_object_version_on_disk(target_disk, &version.bucket, &version.key, version.version_id.as_deref())?;
+            if !actual.matches_manifest(&version.expected) {
```
overtrue, line 690: `matches_manifest` only compares `xl.meta` presence, the data-dir UUID, and `part.N` filenames. Empty shards or shards for the wrong erasure slot still pass, while the final GET can reconstruct from the other eleven disks. Please include the target erasure index and per-part size or hash in the baseline proof.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -0,0 +1,687 @@
...
+        cluster.set_env("RUSTFS_STORAGE_CLASS_STANDARD", format!("EC:{parity}"));
+        cluster.set_node_env(TARGET_NODE, "RUST_LOG", "rustfs=info,rustfs::heal::manager=debug,rustfs_notify=debug")?;
+        cluster.start().await?;
+
+        let clients = cluster.create_all_clients()?;
+        let target_disk = PathBuf::from(&cluster.nodes[TARGET_NODE].data_dirs[TARGET_DRIVE]);
+        let versions = seed_baseline(&clients[0], &target_disk).await?;
+        verify_bodies(&clients[0], &versions).await?;
+
+        cluster.stop_node(TARGET_NODE)?;
```
overtrue: Stopping the node before unmounting drops the live `LocalDisk` mount lease and changes the reported runtime detach into a clean restart with an unmounted directory. A regression in the live lease/device-loss path can still pass this test. Please make the running process lose access to the target, observe that scanner phase, and reboot only after the replacement is installed.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -0,0 +1,687 @@
...
+        let missing_before_restart = incomplete_versions(&target_disk, &versions)?;
+        assert_eq!(
+            missing_before_restart.len(),
+            versions.len(),
+            "blank replacement must start without any baseline version shards"
+        );
+        cluster.start_node(TARGET_NODE).await?;
+
+        wait_for_replacement_state(&cluster, &target_disk, &["completed"], 420).await?;
+        wait_for_physical_census(&target_disk, &versions, 420).await?;
```
overtrue: Waiting for `completed` and then allowing another 420 seconds for the census accepts premature completion: operator state can clear first while shard writes finish later, and the test still passes. Poll status and census together and fail if `completed` appears before the full target manifest matches.

overtrue (changes requested):
The scenario still has false-green gaps in live-detach coverage, scanner synchronization, cluster status, shard proof, and completion ordering.

overtrue (changes requested):
This head only merges current main; the five unresolved test gaps are unchanged.

PR #5928, opened 2026-08-10
fix(rpc): negotiate replay-safe mutation authentication

### LINKED ISSUE #5922: Rolling upgrade from beta.11 to rc.1 broken: lock RPCs rejected as Unauthenticated
**Describe the bug**

A 1.0.0-rc.1 node rolled into a 1.0.0-beta.11 cluster **starts successfully** (the #5689 FileInfo compatibility fix works — format quorum and metadata reads against beta.11 peers are fine) and serves correctly for a few minutes, then degrades to returning 503 `ServiceUnavailable` for essentially all object requests. Its log shows that every distributed-lock RPC against every beta.11 peer is rejected as **`Unauthenticated` / "No valid auth token"**, so lock quorum is unreachable and every lock-gated request errors out:

```json
{"timestamp":"2026-08-10T13:42:21.572250816+08:00","level":"WARN","message":"Remote lock RPC returned tonic error","addr":"https://node1.internal:9000/mnt/rustfs11","op":"lock","timeout_ms":"3000","resource_summary":"bucket-a/bucket-a@latest","tonic_code":"Unauthenticated","tonic_message":"No valid auth token","transport_failure":false,"target":"rustfs_ecstore::cluster::rpc::remote_locker","filename":"crates/ecstore/src/cluster/rpc/remote_locker.rs","line_number":194,"threadName":"rustfs-worker","threadId":"ThreadId(144)"}
{"timestamp":"2026-08-10T13:42:21.572278635+08:00","level":"WARN","message":"Failed to acquire lock on client from response: 0, error: Remote lock RPC failed: Internal error: lock RPC failed: code: 'The request does not have valid authentication credentials', message: \"No valid auth token\"","resource":"bucket-a/bucket-a@latest","owner":"node4.internal:9000","target":"rustfs_lock::distributed_lock","filename":"crates/lock/src/distributed_lock.rs","line_number":814,"threadName":"rustfs-worker","threadId":"ThreadId(96)"}
{"timestamp":"2026-08-10T13:42:25.442204491+08:00","level":"DEBUG","message":"acquire_lock_quorum final failure: Unrecoverable quorum failure: 1/2 required; 3 clients failed; last failure: Remote lock RPC failed: Internal error: lock RPC failed: code: 'The request does not have valid authentication credentials', message: \"No valid auth token\"","resource":"bucket-a/bucket-a@latest","owner":"node4.internal:9000","target":"rustfs_lock::distributed_lock","filename":"crates/lock/src/distributed_lock.rs","line_number":606,"threadName":"rustfs-worker","threadId":"ThreadId(96)"}
```

(The same `Remote lock RPC returned tonic error` warning repeats for every peer and every lock operation, ~230 times within minutes; hostnames and bucket names redacted.)

Root cause appears to be that the node-service lock RPCs became `auth-policy: body-bound` in beta.12 (the #5425/#5455 line of work) and remain so in rc.1:

```diff
-  rpc Lock(GenerallyLockRequest) returns (GenerallyLockResponse) {};
+  rpc Lock(GenerallyLockRequest) returns (GenerallyLockResponse) {}; // auth-policy: body-bound
   (same for UnLock, ForceUnLock, Refresh, LockBatch, UnLockBatch)
```

beta.11 does not recognize the auth that rc.1 attaches to these methods — its lock service rejects the requests with `Unauthenticated`/"No valid auth token" (see fragments above) — so mixed-version lock traffic fails closed. This was invisible in beta.12 only because the FileInfo positional break (#5509) killed beta.11-origin canaries at startup before any lock RPC was issued; #5689 fixed that first layer, and the lock-auth layer is what a canary now reaches. Net effect:

| upgrade path | result |
|---|---|
| beta.11 → beta.12 rolling | broken at startup (FileInfo, #5509) |
| beta.11 → rc.1 rolling | starts fine (#5689 works), then loses lock quorum (this issue) |
| beta.12 → rc.1 rolling | expected compatible (both body-bound) |

So **no rolling path currently exists from beta.11 to any newer release**, which is presumably not the intent given that #5689 was landed specifically to keep beta.11 payloads readable during upgrades.

**To Reproduce**

1. Run a 4-node 1.0.0-beta.11 cluster (erasure-coded, TLS internode, `RUSTFS_RPC_SECRET` configured identically on all nodes).
2. Upgrade exactly one node to 1.0.0-rc.1 and restart it (rolling-style canary).
3. The node starts, reaches format quorum, and serves normally at first.
4. Within a few minutes, lock acquisitions start failing as above and object requests return 503 `ServiceUnavailable` (each after ~4 s).
5. Rolling the node back to beta.11 restores it completely — no persistent state is involved, the incompatibility is purely on the wire.

**Expected behavior**

Either of:

1. Rolling upgrade works: when a peer does not advertise body-bound support, rc.1 attaches lock-RPC auth that the older peer accepts (the legacy token scheme) — and accepts the same from that peer — for at least one release. The capability-advertisement machinery added in #5809 for cross-pool fencing looks like exactly the right gate.
2. Failing that, the joining node fails **fast and loud** with an explicit version/auth-capability mismatch error at startup, instead of starting healthy and then degrading into a 503-serving node whose health endpoints stay green (#5921) — the current failure mode can outlive an operator's canary observation window.

**Environment**

- Cluster version: 1.0.0-beta.11; canary: 1.0.0-rc.1
- `RUSTFS_RPC_SECRET` configured identically on all nodes

**Additional context**

- This is the third consecutive release boundary that breaks mixed-version internode operation for us (beta.10→beta.11: RPC secret policy, #5153; beta.11→beta.12: FileInfo layout, #5509; beta.11→rc.1: this issue). A compatibility gate or CI check for internode RPC/DTO changes against the previous release would catch this class before it ships.
- Happy to provide full canary-window logs, and to re-test a patched rc build on the same cluster.

cxymds:
## Related Issues

Fixes #5922.

## Summary of Changes

RC1 body-bound internode mutations generate replay nonces, while beta.11 retains those nonces in a fixed-size cache. Sustained mutation traffic can exhaust that cache and turn otherwise valid lock and disk RPCs into authentication failures.

This change adds an additive, HMAC-authenticated `dynamic-replay-cache-v1` response capability and keeps authenticated boot-epoch discovery independent from that capability. Explicitly marked high-frequency lock and disk mutations use the existing cache-free `UNSIGNED-PAYLOAD` v2 lane only when a peer has no authenticated boot epoch, as on beta.11. Peers with an authenticated boot epoch, including beta.12 and unpatched RC1, retain canonical body binding and replay-scoped v3 authentication, so replay-scope-strict peers accept non-Ping mutations. IAM, service-control, heal-control, tier-mutation, and scanner-activity contracts are never downgraded.

The client state is generation-aware and fail-closed after a previously capable endpoint loses or returns an invalid dynamic-cache proof. Request-local snapshots keep body admission, replay-scope selection, and response state transitions consistent across concurrent upgrades, restarts, rollbacks, and reordered responses. The remote-lock and remote-disk production helpers mark all supported high-frequency mutation paths for the rolling-compatibility lane.

## Verification

- `cargo test -p rustfs-ecstore cluster::rpc::client::tests::` — 21 passed
- `cargo test -p rustfs-ecstore cluster::rpc::http_auth::tests::` — 61 passed
- `cargo test -p rustfs-ecstore cluster::rpc::remote_locker::tests::lock_mutation_helper_marks_single_and_batch_requests_for_rolling_auth` — 1 passed
- `cargo test -p rustfs-ecstore cluster::rpc::remote_disk::tests::disk_mutation_digest_marks_rolling_compatibility` — 1 passed
- `make pre-pr` — passed, including workspace Clippy, nextest, doctests, architecture, logging, and documentation guards
- High-risk adversarial validation: correctness, simplicity, security, concurrency/durability, compatibility, performance, and test-coverage reviews found no remaining break

## Impact

Patched RC1 clients use cache-free unsigned v2 for beta.11 lock and disk mutations when body-digest strict mode is disabled. Beta.12 and unpatched RC1 peers that provide an authenticated boot proof receive body-bound v3 requests, including when replay-scope strict mode is enabled. Patched peers additionally pin the authenticated dynamic-cache capability; losing that proof later fails closed instead of silently downgrading.

There is no new negotiation RPC or network round trip. The per-request capability lookup occurs once in the interceptor and is reused by the channel snapshot. Patched servers add one HMAC computation and two small authenticated response headers. No configuration or protobuf change is required. `RUSTFS_INTERNODE_RPC_BODY_DIGEST_STRICT=true` remains intentionally incompatible with beta.11 during a mixed-version rollout.

## Additional Notes

The compatibility cleanup register documents the beta.11 unsigned-v2 lane, the authenticated boot-only body-bound-v3 lane, and the pinned dynamic-cache capability lane.

Rollback from a patched peer to an older peer at the same endpoint requires restarting the client process to clear the in-memory capability pin; automatic downgrade after a pinned capability is intentionally rejected to prevent response-header stripping from weakening authentication.

---


crates/ecstore/src/cluster/rpc/client.rs
```diff
@@ -303,16 +357,21 @@ where
...
-                match verify_tonic_boot_epoch_response(&audience, challenge, response.headers()) {
-                    Ok(epoch) => remember_peer_boot_epoch(audience, epoch),
-                    Err(error)
-                        if response.headers().contains_key(RPC_BOOT_EPOCH_HEADER)
-                            || response.headers().contains_key(RPC_BOOT_EPOCH_PROOF_HEADER) =>
-                    {
-                        debug!(error = %error, "peer boot epoch response proof was rejected")
-                    }
-                    Err(_) => {}
+                let response_capability = verify_tonic_replay_cache_capability_response(&audience, challenge, response.headers());
```
overtrue: Legacy beta.12 and unpatched rc.1 peers return an authenticated boot-epoch proof but not this new capability proof. Ignoring the old proof means we never attach v3 for them, so a peer with `RUSTFS_INTERNODE_RPC_REPLAY_SCOPE_STRICT=true` accepts bootstrap Ping and then rejects every Lock or disk call. Please track authenticated boot epoch separately from dynamic-cache capability, and cover an old-proof strict peer with a non-Ping request.

crates/ecstore/src/cluster/rpc/http_auth.rs
```diff
@@ -1100,9 +1193,18 @@ pub fn set_tonic_mutation_body_digest<T: rustfs_protos::CanonicalMutationBody>(
         .get_ref()
         .canonical_body()
         .map_err(|_| std::io::Error::other("RPC mutation body length cannot be represented"))?;
-    set_tonic_canonical_body_digest(request, &canonical_body)
+    set_tonic_rolling_mutation_body_digest(request, &canonical_body)
```
overtrue: This also marks low-frequency IAM, SignalService, and decommission mutations as rolling-compatible. For an unknown or beta.11 peer, the interceptor replaces their signed digest with `UNSIGNED-PAYLOAD`, so an on-path attacker can alter fields such as `dry_run`, `subsystem`, or an IAM access key without invalidating v2 auth. Please keep the shared mutation helper body-bound and opt only the high-volume lock and disk methods into this fallback.

overtrue (changes requested):
Two blockers remain in the downgrade path: it breaks strict legacy peers and removes body integrity from first-use destructive control and IAM calls.

crates/ecstore/src/cluster/rpc/client.rs
cxymds: Fixed in 01e6c044b. Authenticated boot epoch is now tracked independently from the dynamic replay-cache capability. A beta.12 or unpatched RC1 boot proof enables body-bound v3 for subsequent non-Ping mutations, while beta.11 remains on unsigned v2. The new legacy_boot_proof_keeps_mutations_body_bound_and_enables_non_ping_v3 regression test exercises the real interceptor-to-channel path and asserts the Lock request carries v3.

crates/ecstore/src/cluster/rpc/http_auth.rs
cxymds: Fixed in 01e6c044b. The shared set_tonic_mutation_body_digest helper is body-bound again. Only the remote-lock and remote-disk production helpers opt into rolling fallback. Added regression coverage proving an unknown-peer DeleteUser request retains its canonical digest, plus production helper tests for single/batch lock and disk mutations.

cxymds:
@overtrue The two requested downgrade-path changes are addressed in 01e6c044b, and both review threads have been resolved. The focused RPC tests and the full local `make pre-pr` gate pass.

The current `Test and Lint` and `Test and Lint (rio-v2)` failures come from `prepared_reader_reuses_metadata_fanout_exactly_once` (`left: 4`, `right: 3`). The same failure is present on `main` in run 31397633107, and its fix is tracked separately in #5932. The S3 job reached 99% with no test failure and was canceled after the workflow failure. Ready for re-review.

overtrue (approved):
Both requested fixes are covered on the current head: legacy boot proofs enable strict non-Ping v3, and only lock/disk mutations can use the rolling fallback. RPC tests pass locally.

overtrue (approved):
The replay compatibility fix now preserves authenticated boot proof independently from dynamic capability state, keeps shared mutations body-bound, and fails closed after capability pinning. I found no remaining blocker on this head.

PR #5929, opened 2026-08-10
perf(ecstore): keep bounded GET fanout opt-in

houseme:
## Related Issues

## Summary of Changes
- Keep latest-version GET data-read metadata early-stop and bounded metadata fanout behind explicit environment switches instead of enabling them by default.
- Preserve the existing safe default: metadata-only latest reads may early-stop, while GET body/data reads keep full metadata fanout unless an operator explicitly opts in.
- Keep the focused opt-in A/B coverage showing 2+2 data-read fanout can drop from 4 `ReadVersion` calls to 3 when both data-read early-stop and bounded fanout are enabled.
- Keep invalid or overly conservative default parity layouts on full fanout so heterogeneous test sets do not produce a zero-sized bounded fanout.

## Verification
- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore bounded_metadata_early_stop -- --nocapture`
- `cargo test -p rustfs-ecstore metadata_early_stop -- --nocapture`
- `cargo test -p rustfs-ecstore prepared_reader -- --nocapture`
- `cargo test -p rustfs-ecstore get_object_reader -- --nocapture`
- `cargo test -p rustfs-ecstore metadata_quorum_accumulator -- --nocapture`
- `cargo test -p rustfs-ecstore second_pool_multipart_uses_its_own_layout_and_round_trips -- --nocapture`
- `cargo test -p rustfs-ecstore --features rio-v2 second_pool_multipart_uses_its_own_layout_and_round_trips -- --nocapture`
- `cargo test -p rustfs-ecstore select_snapshot_reuses_initial_metadata_fanout_for_first_reader -- --nocapture`
- `cargo clippy -p rustfs-ecstore --features rio-v2 --lib --tests -- -D warnings`
- `RUSTFLAGS="--cfg tokio_unstable" cargo clippy --fix --all-targets --all-features --allow-dirty -- -D warnings`
- `git diff --check`

## Impact
The PR now keeps the availability-preserving default path for GET body reads: full metadata fanout remains the default so a slow scheduled disk or a selected shard failure after metadata preparation can still use other healthy disks.

Operators can still run controlled A/B tests with:
- `RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE=true`
- `RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT=true`

The opt-in path remains useful for measuring `ReadVersion` and replay-cache nonce pressure, but it is intentionally not a production default until bounded hedging and post-prepare read-failure tolerance are proven.

## Additional Notes
High-risk adversarial validation summary:
- Correctness: review identified two concrete availability regressions for default bounded GET data-read fanout: slow scheduled disk without hedge and selected-shard failure after metadata preparation. The latest diff fixes this by keeping data-read early-stop and bounded fanout opt-in by default.
- Simplicity: the fix avoids a partial hedging design in this PR and keeps the production diff limited to env defaults, quorum boundary handling, and focused regression updates.
- Security: no auth, secret, RPC signature, parsing, or trust-boundary behavior changes.
- Concurrency/durability: no namespace lock lifetime, disk write, fsync, on-disk format, or commit ordering changes.
- Compatibility: no S3 API, wire format, on-disk metadata, or mixed-version protocol changes; explicit env rollback and opt-in controls remain available.
- Performance: default GET body behavior no longer reduces fanout, by design; the opt-in A/B test still quantifies the possible `ReadVersion` reduction for the next hedged implementation.
- Test coverage: default data-read full-fanout, opt-in bounded fanout, data-read error fallback, invalid parity full-fanout, second-pool multipart, and Select snapshot metadata reuse are covered by focused tests.

### DIFF
--- a/crates/ecstore/src/set_disk/mod.rs
+++ b/crates/ecstore/src/set_disk/mod.rs
@@ -672,10 +672,10 @@ const DEFAULT_RUSTFS_GET_SMALL_OBJECT_DIRECT_MEMORY_THRESHOLD: usize = 128 * 102
 
 const ENV_RUSTFS_GET_METADATA_EARLY_STOP_ENABLE: &str = "RUSTFS_GET_METADATA_EARLY_STOP_ENABLE";
 // Enabled by default (backlog#872): the early-stop path only engages for
-// requests `should_allow_metadata_early_stop` classifies as safe (metadata-only
-// reads by default, without version_id / healing / free-version needs) and
-// still requires a full read-quorum agreement before stopping. Set the env var
-// to `false` to fall back to full-wait metadata fanout.
+// requests `should_allow_metadata_early_stop` classifies as safe (latest-version
+// metadata-only reads by default, without version_id / healing / free-version
+// needs) and still requires a full read-quorum agreement before stopping. Set
+// the env var to `false` to fall back to full-wait metadata fanout.
 const DEFAULT_RUSTFS_GET_METADATA_EARLY_STOP_ENABLE: bool = true;
 
 const ENV_RUSTFS_GET_METADATA_EARLY_STOP_ROLLOUT_PCT: &str = "RUSTFS_GET_METADATA_EARLY_STOP_ROLLOUT_PCT";
@@ -836,10 +836,10 @@ mod prepared_get_object_metadata_tests {
                     .prepare_get_object_metadata(bucket, object, &opts)
                     .await
                     .expect("prepared metadata should resolve");
+                let prepared_calls = calls.total(disk_call_counters::KIND_READ_VERSION);
                 assert_eq!(
-                    calls.total(disk_call_counters::KIND_READ_VERSION),
-                    4,
-                    "preparation should fan out to each online disk exactly once"
+                    prepared_calls, 3,
+                    "default prepared GET metadata should stop after the 2+2 read/write quorum"
                 );
 
                 let mut reader = set_disks
@@ -864,7 +864,7 @@ mod prepared_get_object_metadata_tests {
         );
         assert_eq!(
             calls.total(disk_call_counters::KIND_READ_VERSION),
-            4,
+            3,
             "reader construction must consume prepared metadata instead of repeating the fanout"
         );
     }
--- a/crates/ecstore/src/set_disk/core/io_primitives.rs
+++ b/crates/ecstore/src/set_disk/core/io_primitives.rs
@@ -538,7 +538,7 @@ impl MetadataQuorumAccumulator {
     }
 
     pub(in crate::set_disk) fn default_write_quorum(&self) -> usize {
-        if self.default_parity_count == 0 {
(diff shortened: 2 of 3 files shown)


crates/ecstore/src/set_disk/mod.rs
```diff
@@ -685,10 +685,10 @@ const ENV_RUSTFS_GET_METADATA_VERSION_EARLY_STOP_ENABLE: &str = "RUSTFS_GET_META
 const DEFAULT_RUSTFS_GET_METADATA_VERSION_EARLY_STOP_ENABLE: bool = false;
 
 const ENV_RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE: &str = "RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE";
-const DEFAULT_RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE: bool = false;
+const DEFAULT_RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE: bool = true;
 
 const ENV_RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT: &str = "RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT";
-const DEFAULT_RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT: bool = false;
+const DEFAULT_RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT: bool = true;
```
overtrue: Enabling bounded fanout by default makes a healthy unscheduled disk unable to hedge a slow scheduled disk. With 2+2, disks 0/1/2 are started; if 0 and 1 finish, 2 hangs, and disk 3 is healthy, pending=1 is still considered enough for quorum so disk 3 is never started. The request waits for disk 2, while the previous full fanout completed from 0/1/3. Please keep this default off or add bounded hedging, with a regression covering that schedule.

overtrue (changes requested):
One blocking availability regression remains.

crates/ecstore/src/set_disk/mod.rs
overtrue: The same default also removes read-failure tolerance for non-inline objects. On 16-drive EC:4, early stop returns only 12 selected metadata entries; if one selected part fails after prepare, the body reader has 11 data shards and cannot use the four healthy unqueried disks. Please cover a selected-shard failure after metadata preparation as well.

houseme: Fixed by keeping both GET data-read metadata early-stop and bounded fanout default-off. The default GET body path now preserves full metadata fanout for slow-disk hedging and post-prepare read-failure tolerance; the bounded 2+2 reduction remains available only through explicit env opt-in for controlled A/B testing. Updated the focused default regression back to full fanout and kept the opt-in A/B/error fallback coverage.

Issue #5870, opened 2026-08-08
[Console] Single-node shows "Cluster status is incomplete" + extra Unknown "local" server after upgrade to 1.0.0-rc.1


swiffer:
**Describe the bug**  
After upgrading a single-node Docker deployment from `1.0.0-beta.12` to `1.0.0-rc.1`, the Console shows:

- “Cluster status is incomplete”
- Message: *“The reported health rows do not cover the full cluster topology.”*
- Servers: 1 Online · 0 Offline · **1 Unknown**
- Reported: **0 · Expected: 1**
- An extra server entry **Unknown `local`** (Disks 0/0, Network 0/0, Uptime Unknown)
- Next to the healthy entry **Online `:::9000`** (Disks 1/1, Network 1/1)

Workload Admission is also shown as **Unknown** (`workload admission has non-open classes: scanner`).

Everything else works normally (storage, scanner, buckets, objects).

**To Reproduce**  
1. Run a single-node Docker container with `rustfs/rustfs:1.0.0-beta.12` and a single data volume.
2. Upgrade the image to `rustfs/rustfs:1.0.0-rc.1` (same volume and same run command).
3. Open the Console → Running Status / Server List.
4. Observe the incomplete cluster status and the extra Unknown `local` entry.

**Expected behavior**  
On a single-node deployment only one server entry should appear and the cluster status should be complete/healthy (Reported = Expected).

**Screenshots**  
Screenshots of the Running Status and Server List pages are attached.

**Desktop (please complete the following information):**
 - OS: Linux (Docker host)
 - Browser: [your browser, e.g. Chrome / Firefox]
 - Version: Console of RustFS 1.0.0-rc.1

**Smartphone (please complete the following information):**
 - Device: N/A
 - OS: N/A
 - Browser: N/A
 - Version: N/A

**Additional context**  
- Deployment: Single Node Single Disk (SNSD) via Docker  
- Image: `rustfs/rustfs:1.0.0-rc.1` (upgraded from beta.12)  
- Bind address: default (`:9000` → appears as `:::9000`)  
- Backend Type: Erasure, Parity 0/0 (expected for SNSD)  
- No functional impact observed – read/write, health endpoint and scanner work correctly.  
- The healthy node (`:::9000`) reports correct disk/network/uptime values.

Is the extra `local` / Unknown entry + incomplete cluster status intentional for single-node in rc.1, or a regression in the new topology/health reporting?

PR #5930, opened 2026-08-10
fix: correct SNSD cluster diagnostics

cxymds:
## Related Issues

Related to #5870

## Summary of Changes

- Add an optional `server_info_endpoint` alias to the v4 admin membership response for the local path-based `local` node, using the same runtime identity exposed by v3 ServerInfo.
- Track scanner runtime lifetime independently from active work units so an initialized but idle scanner reports admission as open instead of unknown.
- Preserve unknown scanner admission before runtime initialization and preserve saturation reporting at the configured limit.
- Add regression coverage for handler wiring, optional response compatibility, idle/uninitialized scanner states, and scanner runtime lifetime.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-scanner scanner_runtime_guard_tracks_runtime_lifetime`
- `cargo test -p rustfs --lib scanner_snapshot_`
- `cargo test -p rustfs --lib cluster_snapshot_handler_requires_server_info_admin_permission`
- `cargo test -p rustfs --lib cluster_snapshot_view_serializes_machine_readable_sections`
- `cargo clippy -p rustfs-ecstore -p rustfs-scanner --lib -- -D warnings`

## Impact

The v4 cluster snapshot adds an optional field only for the local path membership; existing consumers remain compatible and the internal `local` node ID is unchanged. Idle initialized scanners now report `open`, removing the misleading scanner warning. Storage, erasure coding, membership, and on-disk formats are unchanged.

## Additional Notes

Console companion: https://github.com/rustfs/console/pull/202

Adversarial validation covered correctness, simplicity, security, concurrency, compatibility, performance, and test reversibility. The initial ecstore public-structure approach was removed in favor of response-layer enrichment to avoid a Rust source-compatibility break.

---

### DIFF
--- a/crates/scanner/src/lib.rs
+++ b/crates/scanner/src/lib.rs
@@ -82,11 +82,16 @@ pub use storage_api::ScannerReplicationConfig as ReplicationConfig;
 pub use storage_api::scan::SCANNER_ACTIVITY_PROTOCOL_VERSION;
 
 static SCANNER_ACTIVE_WORK_UNITS: AtomicU64 = AtomicU64::new(0);
+static SCANNER_RUNTIME_INSTANCES: AtomicU64 = AtomicU64::new(0);
 static SCANNER_FOREGROUND_READ_ACTIVITY: AtomicU64 = AtomicU64::new(0);
 static SCANNER_FOREGROUND_STREAM_READS: AtomicU64 = AtomicU64::new(0);
 
 pub fn current_scanner_activity() -> u64 {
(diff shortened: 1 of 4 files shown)


overtrue (approved):
The optional local alias preserves the existing v4 shape, and the scanner lifetime signal cleanly distinguishes idle from uninitialized. The focused compatibility and state tests cover the changed behavior.

PR #5931, opened 2026-08-10
perf(ecstore): bound strict inline commit syncs

cxymds:
## Related Issues

N/A

## Summary of Changes

- Apply the existing per-disk and process-wide file-sync admission limits to strict inline-object commits.
- Reuse one per-disk admission across adjacent durability barriers while retaining the existing fsync count and crash-consistency order.
- Preserve cancellation isolation: detached blocking work keeps its per-disk permit and namespace lease, while the async waiter returns global capacity for healthy disks.
- Keep relaxed durability on its existing path without consuming strict file-sync capacity.
- Skip limiter acquisition and blocking-task scheduling for non-Unix directory fsyncs, where the platform operation is already a no-op.
- Add production-path contention, cancellation-isolation, overwrite, ancestor-directory, and Windows-lane regression coverage.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore cancelled_file_sync_waiter_keeps_disk_admission_until_blocking_work_finishes -- --nocapture`
- `cargo test -p rustfs-ecstore strict_inline_rename_retains_admission_until_commit_fsync -- --nocapture`
- `cargo test -p rustfs-ecstore windows_and_unix_rename_data_new_inline_object_fsyncs_new_ancestor_dirs -- --nocapture`
- `cargo test -p rustfs-ecstore test_rename_data_writes_old_metadata_backup_for_inline_overwrite -- --nocapture`
- `cargo test -p rustfs-ecstore windows_and_unix_cancelled_inline_preparation_serializes_newer_commit -- --nocapture`
- `cargo test -p rustfs-ecstore relaxed_inline_preparation_does_not_use_file_sync_limit -- --nocapture`
- `make pre-pr` completed all static guards and workspace Clippy; its full nextest run stopped after 8,027 passes on six unrelated `rustfs-heal` tests while another full workspace build was saturating the host. The same six tests passed together in 1.8 seconds when rerun directly.

## Impact

Strict inline-object commits now share bounded file-sync capacity with other sync-heavy paths, preventing unbounded blocking work under concurrent small PUT load. Durability mode semantics, fsync ordering, S3 behavior, and on-disk/wire formats are unchanged. No configuration or migration is required.

## Additional Notes

- Correctness: attacked strict/relaxed branches, inline create/overwrite, rollback, cancellation, quorum, error propagation, and crash ordering; no break found.
- Simplicity: removed redundant test synchronization and retained only the seams needed to deterministically exercise production admission ordering; no remaining finding.
- Security: attacked path confinement, test-hook production reachability, cancellation storms, resource exhaustion, and error disclosure; no break found.
- Concurrency/durability: attacked namespace/disk/global lock ordering, detached blocking work, permit lifecycle, partial failure, slow-disk isolation, and power-loss ordering; no break found.
- Compatibility: no public API, S3, wire, or on-disk behavior changes; mixed-version behavior is unaffected, and non-Unix only skips scheduling around an existing no-op.
- Performance: the review fixes remove non-Unix no-op scheduling and add no production allocations or hooks; the broader PR intentionally bounds concurrency and does not claim a single-request throughput gain.
- Test coverage: reverting production admission retention makes the real `rename_data` contention test fail; reverting the non-Unix bypass fails a test selected by the Windows CI lane; cancellation assertions use a test-local global semaphore.
- The production-path contention test was also mutation-checked by temporarily releasing and reacquiring admission after preparation; the test failed on the intended overtaking state and passed after the mutation was removed.

---


crates/ecstore/src/disk/os.rs
```diff
@@ -1120,6 +1135,61 @@ pub(crate) async fn run_blocking_namespace_operation<T: Send + 'static>(
...
+    result.map_err(|err| io::Error::other(format!("blocking namespace file sync operation failed: {err}")))?
+}
+
+pub(crate) async fn fsync_dir_with_namespace_file_sync_limit(
+    dir: impl AsRef<Path>,
+    lease: Arc<NamespaceMutationLease>,
+    admission: &FileSyncAdmission,
+) -> io::Result<()> {
+    let dir = dir.as_ref().to_path_buf();
+    run_blocking_namespace_file_sync_operation(lease, admission, move || {
```
overtrue: On Windows `fsync_dir_std` is a no-op, but this helper still acquires the global limiter and enters `spawn_blocking` once for the object directory and every new ancestor. Strict inline PUTs previously returned synchronously for these directory flushes. Please keep this Unix-only so the limit does not add several no-op blocking round trips per request.

crates/ecstore/src/disk/local.rs
```diff
@@ -9098,8 +9108,12 @@ impl DiskAPI for LocalDisk {
                     has_dst_buf.is_none(),
                     prepared_metadata_source,
                 ))
-            })
-            .await
+            };
+            let inline_preparation = if let Some(admission) = file_sync_admission.as_ref() {
+                os::run_blocking_namespace_file_sync_operation(mutation_lease.clone(), admission, prepare_inline_metadata).await
```
overtrue, line 9142: This call site is the behavior the PR claims to protect, but the new reuse test only holds a `FileSyncAdmission` manually. Releasing and reacquiring admission after preparation still leaves every added test green and lets a queued commit overtake the already-renamed commit before its durability fsync. Please add a production-path contention test around `rename_data` that detects that mutation.

crates/ecstore/src/disk/os.rs
```diff
@@ -4553,6 +4624,83 @@ mod tests {
...
+            .await
+        });
+
+        tokio::task::spawn_blocking(move || entered_rx.recv_timeout(Duration::from_secs(30)))
+            .await
+            .expect("blocking work waiter should run")
+            .expect("blocking work should start");
+        waiter.abort();
+        assert!(waiter.await.expect_err("waiter should be cancelled").is_cancelled());
+        let returned_global_permit = FILE_SYNC_PERMITS
```
overtrue: `FILE_SYNC_PERMITS` is process-global, while this test only serializes peers tagged `file_sync_probe`. The new untagged admission test and strict-inline tests can queue for the last permit; after abort, the returned permit is handed to that waiter and this `try_acquire()` fails. Please use an injectable test-local global limiter or otherwise isolate all users of this assertion.

overtrue (changes requested):
The strict-inline admission change still adds avoidable Windows blocking work, lacks contention coverage on the production call path, and introduces a race-prone global-permit assertion.

crates/ecstore/src/disk/os.rs
cxymds: Fixed in 9d382b420. fsync_dir_with_namespace_file_sync_limit now uses the limiter and blocking task only on Unix; non-Unix directly executes the existing no-op fsync_dir_std. The real inline ancestor test is also named for the Windows CI filter and asserts that non-Unix directory fsyncs are not limiter-recorded.

crates/ecstore/src/disk/local.rs
cxymds: Fixed in 9d382b420. Added strict_inline_rename_retains_admission_until_commit_fsync using two real LocalDisk::rename_data calls and a one-slot per-disk limiter. It holds the first request in production preparation, queues the second at production admission, and proves the first reaches its commit directory barrier before the second preparation starts. A temporary drop/reacquire mutation made this test fail on the intended overtaking state.

crates/ecstore/src/disk/os.rs
cxymds: Fixed in 9d382b420. The blocking helper now has a private injectable implementation while the production wrapper still supplies FILE_SYNC_PERMITS. The cancellation test uses its own one-slot global semaphore and independently asserts that cancellation returns global capacity while the in-flight blocking operation retains per-disk capacity.

overtrue (approved):
The prior blockers are resolved; no blocker remains.

PR #5933, opened 2026-08-10
fix(scanner): skip disk inventory in scan spans

### LINKED ISSUE #5927: Scanner logs ~340 KB per span: nsscanner_disk records set_disks as a tracing field (11 GB/hour at info)
**Describe the bug**

At `RUSTFS_OBS_LOGGER_LEVEL=info`, a single-node RustFS writes roughly **11 GB of logs per hour** while essentially idle. All of it comes from one tracing span.

`nsscanner_disk` in `crates/scanner/src/scanner_io.rs` takes `set_disks: Vec<Arc<Disk>>`, and that argument is not in the `skip(...)` list on its `#[tracing::instrument]` attribute:

https://github.com/rustfs/rustfs/blob/1.0.0-rc.1/crates/scanner/src/scanner_io.rs#L3884-L3892

```rust
#[tracing::instrument(skip(self, budget, updates, cache))]
async fn nsscanner_disk(
    self: Arc<Self>,
    ctx: CancellationToken,
    budget: Arc<ScannerCycleBudget>,
    set_disks: Vec<Arc<Disk>>,   // <-- recorded as a span field via Debug
    cache: DataUsageCache,
    updates: Option<mpsc::Sender<DataUsageEntry>>,
    scan_mode: HealScanMode,
) -> Result<ScannerDiskScanOutcome> {
```

So `tracing` records the whole `Vec<Arc<Disk>>` as a span field using its `Debug` impl. On a **single-disk standalone deployment** that renders to ~170 KB of text: the `LocalDisk` struct, the raw `format.json` bytes, and the disk's metrics table — 2,280 `TimedActionSlot` entries in one line.

The two sibling functions in the same file skip their heavy arguments correctly (`nsscanner_with_status` at L2297, `nsscanner_cache` at L2641). This one looks like an oversight rather than a decision.

**Measurements**

10 seconds of `kubectl logs -f` from an idle instance: 1,104 lines, **268 MB**. Median line 170 KB, max 341 KB. Grouped by `(target, message)`:

```
178.4 MB  n=529   (rustfs_scanner::scanner_io, "enter")
 89.3 MB  n=529   (rustfs_scanner::scanner_io, "exit")
  0.2 MB  n=3     everything else combined
```

Field sizes within one such line:

```
170463 B  spans        <- ancestor stack, carries the same payload again
170461 B  span
    34 B  filename
    32 B  timestamp
     7 B  message      <- "enter"
```

and within `span`:

```
{'ctx': 43, 'scan_mode': 6, 'set_disks': 170347, 'name': 16}
```

The payload is emitted three times per span — on `enter`, on `exit`, and again in the `spans` ancestor array — so one span costs roughly 340 KB. The scanner emits about a thousand of them per cycle.

**Impact**

- On Kubernetes the kubelet rotates the container log file every second or two, so `kubectl logs` returns almost nothing and most of the output is unrecoverable. A log shipper reading the file cannot keep up: the file grew at 7.6 MB/s while the collector read it at 4.0 MB/s.
- In my cluster this was 99.5% of all container log bytes, roughly 70x every other workload combined.
- Anyone hitting this with `RUSTFS_OBS_LOG_DIRECTORY` set will also hit #2130, since the active file never rotates.

RustFS itself is fine throughout — 0.09 cores, 422 MiB — so nothing else signals a problem. The only symptom is the log volume.

**To Reproduce**

1. Deploy RustFS `1.0.0-rc.1` standalone, one disk, `RUSTFS_OBS_LOGGER_LEVEL=info`, `RUSTFS_OBS_LOG_DIRECTORY=""` (stdout).
2. Put some objects in a bucket so the data scanner has work.
3. `docker logs -f` / `kubectl logs -f` and pipe to `wc -c`.

**Expected behavior**

Scanner span events should be a few hundred bytes, like every other span in the codebase.

**Suggested fix**

Add `set_disks` to the skip list:

```rust
#[tracing::instrument(skip(self, budget, updates, cache, set_disks))]
```

If the disk set is useful context, `fields(disk_count = set_disks.len())` would give it at constant size.

**Versions**

- RustFS `1.0.0-rc.1` (released 2026-08-08), official image `rustfs/rustfs:1.0.0-rc.1`
- Standalone mode, single disk, Linux x86_64, Kubernetes

Still present on `main` as of 2026-08-10. Not fixed by #2086, which cleaned up instrumentation elsewhere.

**Workaround**

`RUSTFS_OBS_LOGGER_LEVEL` accepts `RUST_LOG` syntax (`crates/obs/src/telemetry/filter.rs` splits on commas and builds one directive per entry), so the module can be muted without lowering the global level:

```
RUSTFS_OBS_LOGGER_LEVEL=info,rustfs_scanner::scanner_io=warn
```

That took the same instance from 11.3 GB/hour to about 47 MB/hour, average line length 243 KB -> 804 bytes, with scanner warnings still coming through.

marshawcoco:
## Related Issues

Fixes #5927.

## Summary of Changes

- Exclude `set_disks` from the `nsscanner_disk` tracing span so INFO lifecycle records no longer serialize every disk's format bytes and metrics ring.
- Extend the logging guardrail with positive and negative fixtures that require this parameter to remain skipped.

## Verification

- `/root/.local/bin/rustfs-cargo-safe fmt --all -- --check`
- `/root/.local/bin/rustfs-cargo-safe check -p rustfs-scanner --lib`
- `git diff --check`

## Impact

Substantially reduces scanner span log volume in development/FULL and production/CLOSE logging modes. Scanner behavior, disk selection, APIs, and on-disk data are unchanged.

## Additional Notes

N/A.

### DIFF
--- a/crates/scanner/src/scanner_io.rs
+++ b/crates/scanner/src/scanner_io.rs
@@ -3881,7 +3881,7 @@ impl ScannerIODisk for Disk {
         Ok(size_summary)
     }
 
-    #[tracing::instrument(skip(self, budget, updates, cache))]
+    #[tracing::instrument(skip(self, budget, updates, cache, set_disks))]
     async fn nsscanner_disk(
         self: Arc<Self>,
         ctx: CancellationToken,
--- a/scripts/check_logging_guardrails.sh
+++ b/scripts/check_logging_guardrails.sh
@@ -842,6 +842,34 @@ for file in "${disk_logging_files[@]}"; do
   fi
 done
 
+# `set_disks` expands every Disk through Debug, including raw format bytes and
+# the full per-operation metrics ring. Keep it out of the INFO scanner span.
+scanner_disk_skip_pattern='#\[(tracing::)?instrument\([^]]*skip\([^)]*\bset_disks\b[^)]*\)[^]]*\)\][[:space:]]*async fn nsscanner_disk\b'
+if ! rg -U "$scanner_disk_skip_pattern" crates/scanner/src/scanner_io.rs >/dev/null; then
+  echo "❌ logging guardrail violation: nsscanner_disk must skip set_disks in its tracing instrumentation" >&2
+  exit 1
+fi
+
+for fixture in \
+  $'#[tracing::instrument(skip(self, budget, updates, cache, set_disks))]\nasync fn nsscanner_disk('; do
+  if ! printf '%s\n' "$fixture" | rg -U "$scanner_disk_skip_pattern" >/dev/null; then
+    echo "❌ logging guardrail self-test failed: safe nsscanner_disk span was rejected" >&2
+    echo "$fixture" >&2
+    exit 1
+  fi
+done
+
+for fixture in \
+  $'#[tracing::instrument(skip(self, budget, updates, cache))]\nasync fn nsscanner_disk(' \
+  $'#[tracing::instrument(skip(self, budget, updates, cache), fields(set_disks = set_disks.len()))]\nasync fn nsscanner_disk(' \
+  $'#[tracing::instrument(skip(self, budget, updates, cache, set_disks_count))]\nasync fn nsscanner_disk('; do
+  if printf '%s\n' "$fixture" | rg -U "$scanner_disk_skip_pattern" >/dev/null; then
+    echo "❌ logging guardrail self-test failed: unsafe nsscanner_disk span was accepted" >&2
+    echo "$fixture" >&2
+    exit 1
+  fi
+done
+
 # `forbidden_patterns` above only retires log lines that already shipped, so a
 # newly written sentence-style log passes every check in this script — which is
 # how one reaches review in the first place (PR #5822 added


terem42:
Independently reproduced, and the fix verified against its own parent commit.

**Setup.** Standalone single node, real disks, `RUSTFS_OBS_LOGGER_LEVEL=info`, `RUSTFS_OBS_LOG_DIRECTORY` set, default `RUSTFS_OBS_ENVIRONMENT` (i.e. `production`, so `FmtSpan::CLOSE`). Buckets seeded so the data scanner has work; log growth sampled over a fixed 120 s window after a settle pause. Binaries one commit apart, same tree and toolchain: `1148e7627` (the PR's parent) and `cc44c37cd` (the PR head).

**Reproduced on `1148e7627`**, and the record is almost entirely that one field:

| `nsscanner_disk` record, `1148e7627` | 1 disk | 4 disks |
|---|---|---|
| longest record | 167,244 B | 668,002 B |
| `span.set_disks` within it | 166,835 B | 667,592 B |

The remaining span fields are `ctx` 43 B, `name` 16 B, `scan_mode` 6 B. The payload is the per-disk `metrics: DiskHealthMetricEpoch` — 38 operation entries each carrying a 60-slot `last_minute` ring, so 2,280 `TimedActionSlot` in a single record — plus the raw `format.json` bytes and the parsed `Url`. It is linear in the number of disks in the set, since `set_disks` is the whole set inventory.

**Fixed on `cc44c37cd`.** Same window, same seed, same topology:

| | `1148e7627` | `cc44c37cd` |
|---|---|---|
| `rustfs_scanner::scanner_io` records, 4 disks / 1 bucket | 6 | 6 |
| bytes in them | ≈1.34 MB | ≈3 KB |
| longest record | 668,002 B | **469 B** |
| share of all log bytes | 95.6% | 4.0% |
| same, 4 disks / 10 buckets | ≈13.4 MB — 98.9% | ≈10 KB — 5.9% |

The record count is identical in both arms at every topology (6, 6, 24), so the spans still occur and scanner behaviour is unchanged — only the record stops carrying the inventory. 1,424x smaller per record.

**Two notes on scope.**

The default environment is affected, not only development ones. The `enter`/`exit` records in the report imply `FmtSpan::FULL`, which `crates/obs/src/telemetry/local.rs` selects only when the environment is not `production`; `ENVIRONMENT` defaults to `production` (`crates/config/src/constants/app.rs:50`), giving `FmtSpan::CLOSE`. That is one record per span rather than three, but it still carries the full `set_disks` — the 668,002 B figure above was measured in `CLOSE` mode.

The cost scales with the erasure set rather than with the node, so a single-disk instance is the floor rather than the typical case: 166,835 B of payload at one disk, 667,592 B at four, in the same shape of run.

For anyone hitting this before the fix lands, `RUSTFS_OBS_LOGGER_LEVEL=info,rustfs_scanner::scanner_io=warn` removes it completely — zero records in a 120 s window, against 24 without it — and costs nothing else: that target emits no other records at info, and its warnings and errors still come through.

cxymds:
LGTM.

The change is appropriately scoped: skipping `set_disks` only removes its `Debug` representation from the tracing span and does not affect scanner behavior or argument passing.

The guardrail also provides useful regression coverage: it rejects the previous instrumentation and avoids false matches such as `set_disks_count`. The independent A/B measurements confirm that the span count remains unchanged while the record size drops substantially.

I found no blocking correctness, compatibility, security, or performance concerns.

cxymds (approved):

PR #5936, opened 2026-08-10
fix: report stalled object traffic as unready

### LINKED ISSUE #5921: /health/ready returns 200 on a node that cannot serve object traffic
**Describe the bug**

`/health/ready` (and the console health endpoint) answer HTTP 200 in milliseconds on a node whose **object requests hang**. Any load balancer using these endpoints keeps routing client traffic into a node that serves nothing, indefinitely — we have observed such states persist for over an hour (and, unnoticed, much longer) across 1.0.0-beta.10 and 1.0.0-beta.11, in several distinct "wedge" flavors:

1. **Full wedge** (beta.10): every GET/PUT including 84 KiB inline objects hung ≥60 s; `/health/ready` answered 200 in ~6 ms on the same node at the same moment; process `active`, ~255% CPU. Other nodes served the same objects in 10–320 ms.
2. **Write-only wedge** (beta.10): reads fast, every PUT hung to client timeout (`PutObject store write remains in flight`).
3. **Metadata wedge** (beta.11, multiple occurrences): inline-object GETs (metadata path) time out while erasure GET/PUT still work — observed after a peer outage/rejoin, correlated with stale namespace/bucket-metadata lock entries.

**To Reproduce**

We have no deterministic trigger — the wedges arise from lock/handle state under restart churn (peer outage, rejoin, or rolling restarts) and have recurred six times across three weeks of ordinary operation. They are easy to *verify* when present:

1. Against each node directly, run authenticated timed requests with a deadline: a GET of a small inline object, a GET of a ~4 MiB erasure object, and a small PUT.
2. On a wedged node, one or more request classes hangs past any reasonable deadline (60 s+) while the same requests answer in milliseconds on the other nodes.
3. On the same wedged node, `HEAD /health/ready` answers 200 in single-digit milliseconds throughout, so load-balancer health checks never fail.

**Expected behavior**

Readiness should reflect the node's actual ability to serve object traffic. Concretely (any of these would have prevented the production impact):

1. `/health/ready` (or a new deep-health endpoint, e.g. `/health/object`) performs a periodic internal self-GET of a small canary object with a deadline — an inline object is enough to catch flavors 1 and 3 — and reflects the result in readiness.
2. A foreground-request liveness watchdog: if all in-flight object requests exceed N seconds with zero completions, log at ERROR with lock-table / queue-depth context. Today the node logs nothing about its own wedge.
3. Requests that cannot acquire their namespace lock within the acquire timeout should error out rather than hang past 60 s (flavors 1 and 3).

As a stopgap we are deploying a custom agent script that probes each node's object path and feeds the result to our load balancer's health checks. It works, but every RustFS deployment behind a load balancer would need to rebuild the same thing — the check belongs in the server.

**Environment**

- Versions observed: 1.0.0-beta.10, 1.0.0-beta.11 (current production)
- Load balancer health check: `HEAD /health/ready`

**Additional context**

- In flavor 1 we verified the contrast directly: one node timed out 8/8 GETs (4 MiB erasure + 84 KiB inline) at 60 s while answering `/health/ready` 200 in 6 ms; the other three nodes served the identical objects in 10–320 ms.
- The green-health property also inverts the operator's mental model during incident response: the health surface says "up", monitoring says "up", and only client-side timeouts reveal the wedge.
- Happy to provide full logs/timelines for any of the occurrences, and to test a patched readiness implementation.

cxymds:
## Related Issues

Fixes #5921.

## Summary of Changes

- Track completion progress for bounded foreground object stages: GET metadata/request-context setup, GET object-reader setup, and fully received empty, small-eager, and zero-copy-eager PUT storage commit.
- Keep the readiness hot path O(1): health requests read per-`AppContext` atomic snapshots and never issue self-GET/PUT operations, create canary objects, start per-request tasks, or perform storage I/O.
- Return `503 Service Unavailable` from `/health/ready`, `/minio/health/ready`, and console readiness when a tracked read or write class has active work but no completion within the configured service-progress SLO. Liveness and cluster-probe semantics remain unchanged.
- Add the diagnostic reasons `object_read_stalled` and `object_write_stalled`, and keep the existing dependency fields intact so a stalled data path is not mislabeled as a disconnected object-store handle.
- Add `RUSTFS_HEALTH_OBJECT_PROGRESS_ENABLE` (default `true`) and `RUSTFS_HEALTH_OBJECT_PROGRESS_TIMEOUT_MS` (default `30000`). The effective threshold is never earlier than the canonical namespace-lock acquisition deadline plus a 5-second margin.
- Add deterministic regression coverage for real metadata and namespace-lock stalls, stage boundaries, cancellation, atomic timestamp monotonicity, endpoint aliases, request-scoped instance isolation, metrics, configuration fallback, exact zero-byte EOF validation, empty/zero-copy eager PUT progress, and recovery.

## Verification

- `cargo test -p rustfs object_traffic --lib` — 10 passed.
- `cargo test -p rustfs object_progress --lib -- --test-threads=4` — 4 passed.
- `cargo test -p rustfs object_progress_tracks_zero_byte_and_zero_copy_put_lock_waits --lib -- --test-threads=1` — 1 passed.
- `cargo test -p rustfs health --lib` — 74 passed.
- `make pre-pr` through Nextest — formatting, architecture/unsafe/logging guards, workspace Clippy, script self-tests, and all 13,084 Nextest cases passed. The remaining workspace doctest phase was intentionally stopped at the requester's direction.
- Formal isolated release-build 4 KiB, concurrency-8, five-round A-B-B-A hot-path gate passed with zero errors. GET candidate legs were -0.64% and -1.08% req/s versus baseline, with unchanged p90/p99; PUT candidate legs were +1.43% and -3.94%, with no repeatable latency or tail regression. Closing baseline drift legs were +0.83% GET and +1.24% PUT.
- Benchmark binary provenance: baseline SHA-256 `43e9962969762ec4c4e256c4e8ec9a808bb35ea7ae4d836f8d9adfe72a6fb5db`; candidate SHA-256 `f81377eef79d51731e16ac08ed4e314cd47bada714b4716eeeda3d3fb8e7d176`.

## Impact

- Existing readiness URLs and HEAD/GET response contracts are preserved. The behavior is intentionally stricter: a node can now become unready when bounded foreground object work stops completing for the configured SLO.
- This does not change S3 responses, object formats, RPC formats, startup readiness, liveness, or the existing MinIO cluster-read/cluster-write probe semantics.
- The detector is a foreground class-wide no-completion watchdog, not a deep canary. It intentionally does not claim idle disk writability, fast-error capability, arbitrary per-key partial-wedge detection, response-body streaming progress, or positive-length SSE/compressed/streaming PUT coverage. Empty PUTs and fully buffered small/zero-copy eager PUTs are covered only after request EOF has been validated. Operators can tune the SLO or set `RUSTFS_HEALTH_OBJECT_PROGRESS_ENABLE=false` for rollback.

## Additional Notes

- Correctness — attacked false-green states, GET/PUT stage boundaries, legal lock budgets, timestamp ordering, decoded-zero pending EOF, extra-body poison, recovery, and metric consistency; no break found.
- Simplicity — attacked self-canary alternatives, duplicated helpers and predicates, one-caller abstractions, readiness I/O, and test-fixture shape; the foreground snapshot remained the smallest semantics-correct design, no break found.
- Security — attacked unauthenticated readiness drain amplification, client-controlled pending ingress, cross-instance context confusion, auth bypass, and diagnostic leakage; no break found.
- Concurrency/durability — attacked atomic happens-before, timestamp regression, oneshot cancellation, guard cleanup, initialization singleflight, lock ordering, and persistent side effects; no break found.
- Compatibility — attacked MinIO aliases, console parity, HEAD behavior, exact empty-object S3 semantics, liveness/startup isolation, mixed-version rollout, and on-wire/on-disk formats; no break found.
- Performance — attacked per-GET/tiny-PUT atomic contention, empty/zero-copy eager overhead, cache-line sharing, LB probe overhead, A-B-B-A drift/tails/errors; the formal 4 KiB@8 A-B-B-A gate passed (GET -0.64%/-1.08%, PUT +1.43%/-3.94%, zero errors), no break found.
- Test coverage — mapped each behavior to revert-detecting unit, real-ECStore, endpoint, metric, environment, cancellation, pending-EOF, extra-byte, and eager-PUT tests; no remaining gap found.

---


overtrue (approved):
Looks good.

rustfs/src/app/object_usecase.rs
```diff
@@ -5866,6 +5931,14 @@ impl DefaultObjectUsecase {
             }
         });
 
+        let object_traffic_health = if use_small_eager_put_path {
```
overtrue: This guard is never created for a valid zero-byte PUT because should_use_small_eager_put_path rejects size <= 0; the request still enters put_object_with_old_current_size, so a held namespace lock leaves readiness at 200 indefinitely. The fully buffered zero-copy eager path has the same gap. Please track every PUT whose ingress is already complete before the store call and add a held-lock zero-byte regression.

overtrue (changes requested):
A valid fully received PUT path remains outside the watchdog.

rustfs/src/app/object_usecase.rs
cxymds: Fixed in 3d5ae9d00. Valid empty PUTs now exact-drain and validate request EOF before the storage-progress guard is created, so client-controlled pending ingress cannot withdraw readiness. Fully received empty, small-eager, and zero-copy-eager PUTs are then tracked across the store call. The regression test covers a decoded-zero body with pending EOF, an extra-byte poison case returning UnexpectedContent without storing an object, and real held namespace-lock stalls for empty and zero-copy eager PUTs. The focused test passes, and the full pre-PR gate passed through all 13,084 Nextest cases; the remaining doctest phase was intentionally skipped at requester direction.

overtrue (approved):

PR #5937, opened 2026-08-10
fix(ecstore): publish multipart parts on Windows

### LINKED ISSUE #5419: Multipart UploadPart fails on Windows during rename_part with PATH NOT FOUND reported as “File access denied”

Ivanbeethoven:
## Related Issues

Fixes #5419.

## Summary of Changes

- Close the prepared `new.meta` handle before publishing its parent transaction directory on Windows.
- Reuse the existing Windows-aware `sync_file` helper so strict-durability shard flushes use a write-capable handle.
- Add a Windows regression test with a realistic multipart path longer than 260 characters, covering initial publication and replacement of the same part.
- Pin the regression test to Strict durability so external or concurrent durability overrides cannot skip the shard-sync branch.

The failures had two Windows-specific causes: the transaction prepare path renamed a directory while one of its child files was still open, and the shard durability path attempted `FlushFileBuffers` through a read-only handle.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `$env:RUSTFS_DURABILITY_MODE='none'; cargo test -p rustfs-ecstore --lib disk::local::test::test_rename_part_commits_realistic_windows_multipart_path --target x86_64-pc-windows-msvc -- --exact --nocapture --test-threads=1`
- `cargo test -p rustfs-ecstore --lib windows_ -- --nocapture --test-threads=1` (65 passed)
- `cargo test -p rustfs-ecstore --lib rename_all_ -- --nocapture --test-threads=1` (26 passed)
- `cargo test -p rustfs-ecstore --lib set_disk::ops::multipart::tests::upload_part_retry_quorum_failure_preserves_old_part_across_ec_geometries -- --exact --nocapture --test-threads=1`
- `cargo build --release --target x86_64-pc-windows-msvc -p rustfs --bins`
- Native Windows/NTFS end-to-end: ordinary PUT/GET passed; 20 serial 20 MiB multipart iterations passed; all 100 `UploadPart` calls returned HTTP 200 with valid ETags; restart/resume completed and matched SHA-256; abort disappeared from listing and the old upload ID returned HTTP 404 `NoSuchUpload`.
- Full RustFS logs contained zero `File access denied`, sharing-violation/code-32, `write_quorum_unavailable`, or `reliable_rename` errors.

## Impact

Windows multipart `UploadPart` can complete under strict durability instead of failing during transaction prepare or shard flush. There are no S3 API, wire-format, on-disk-format, configuration, or non-Windows behavior changes.

## Additional Notes

- No test-specific part or transaction files remained after end-to-end validation. The strict evidence checker separately counted the expected `.trash` root plus two empty housekeeping directories created before the five-minute background cleanup interval elapsed.
- Correctness: attacked Windows handle lifetime, error propagation, initial/replacement publication, rollback, recovery, and cleanup idempotency; no break found.
- Simplicity: attacked helper reuse, explicit handle drop versus inner scoping, comment necessity, and test duplication; no smaller design preserving the platform semantics was found.
- Security: attacked traversal-shaped object keys, root-boundary enforcement, and secret/log exposure; this diff adds no untrusted-input or logging surface, and no break was found.
- Concurrency/durability: attacked Strict/Relaxed/None gating, crash ordering, cancellation, same-part retry, and the test override guard; no break found.
- Compatibility: attacked S3, MinIO, on-disk, wire, rolling-upgrade, and non-Windows behavior; the change affects only Windows handle access/lifetime and no break was found.
- Performance: attacked allocations, clones, blocking hops, fsync counts, durability gates, and the 4 KiB path; the production syscall count is unchanged and no regression was found.
- Test coverage: exact reverts of either production fix now fail deterministically under the pinned Strict mode; the test uses a unique temporary directory and has no timing-based assertions.


Ivanbeethoven:

overtrue (approved):

PR #5956, opened 2026-08-11
test(heal): strengthen replacement e2e evidence

overtrue:
## Related Issues

Follow-up to #5918.

## Summary of Changes

- include inline `xl.meta` payload size and SHA-256 in the per-version physical shard census, while deriving present part numbers from the fingerprint map
- require the live-loss barrier to consume only post-fault heal-scanner events in causal order
- sample the physical census before authoritative completion status and confirm a stale incomplete sample before reporting premature completion
- add real inline, erasure-index, shard-size, shard-hash, status, ordering, and log-epoch regression coverage

## Verification

- `cargo fmt --all --check`
- `CARGO_INCREMENTAL=0 cargo test -p e2e_test chaos::tests:: -- --nocapture`
- `CARGO_INCREMENTAL=0 cargo test -p e2e_test replacement_privileged_e2e_test::tests:: -- --nocapture`
- `CARGO_INCREMENTAL=0 cargo clippy -p e2e_test --tests -- -D warnings`
- `CARGO_INCREMENTAL=0 cargo test -p e2e_test reliability_disk_fault_test::tests::test_versioned_shard_census_selects_each_version_data_dir -- --nocapture`

## Impact

Test-only changes. No production API, configuration, storage-format, or deployment impact.

## Additional Notes

The privileged device-mapper replacement tests remain opt-in and were not executed on macOS because they require Linux root or `CAP_SYS_ADMIN`. The Linux-gated module was explicitly compiled, unit-tested, and checked with Clippy; the real non-privileged four-disk census test was executed locally.

---


crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -597,6 +607,32 @@ mod tests {
...
+        let mut saw_live_loss = false;
+        for line in log.lines() {
+            if line.contains("Heal auto-scan disk inspection failed")
+                && line.contains("check_failed")
+                && line.contains(target.as_ref())
+            {
+                saw_live_loss = true;
+                continue;
+            }
+            if saw_live_loss && line.contains("Heal auto-scan cycle completed") {
```
overtrue: This barrier cannot complete in the fault state under test. The failed disk is logged with `check_failed` but is not added to `endpoints`; with every other disk healthy, the scanner emits `Heal auto disk scanner idle` and continues before `Heal auto-scan cycle completed`. Both privileged cases therefore time out after 180 seconds. Please accept the real failure-to-idle sequence as the completed post-fault scan and pin that sequence in the unit test.

crates/e2e_test/src/replacement_privileged_e2e_test.rs
```diff
@@ -694,6 +722,55 @@ mod tests {
...
+        mut status: S,
+    ) -> Result<CompletionSample, Box<dyn Error + Send + Sync>>
+    where
+        C: FnMut() -> Result<BTreeSet<String>, Box<dyn Error + Send + Sync>>,
+        S: FnMut() -> F,
+        F: std::future::Future<Output = Result<serde_json::Value, Box<dyn Error + Send + Sync>>>,
+    {
+        let first = sample_replacement_completion(target_disk, &mut census, &mut status).await?;
+        if matches!(first, CompletionSample::CompletedWithIncomplete(_)) {
+            return sample_replacement_completion(target_disk, census, status).await;
```
overtrue: The second incomplete census is already taken after a definitive `completed` status, so it proves premature completion. Fetching status again can erase that proof: a transient non-definitive response returns `Pending`, and the test can later pass once healing catches up. Please keep the first completed status as the barrier and resample only the census.

overtrue (commented):
Two blockers remain on this head.

Issue #5950, opened 2026-08-11
Heal task stuck in infinite loop for 20+ days, cannot be stopped or cleared, scanner never completes


1399781652:
I have upgraded from version 1.0.0-beta5 all the way to r1; could changes in the metadata structure be causing parsing failures? Additionally, there is an issue where tasks keep getting added without the scan actually starting, and the tasks cannot be stopped.

Environment
RustFS Version: r1

Deployment: 4 nodes, 48 disks (erasure coding)

Total Data: ~2.87 TB

Configuration:

      - RUSTFS_SCANNER_ENABLED=true
      - RUSTFS_SCANNER_SPEED=default
      - RUSTFS_SCANNER_IDLE_MODE=true
      - RUSTFS_HEAL_INTERVAL_SECS=120
      - RUSTFS_HEAL_MAX_CONCURRENT_HEALS=4
      - RUSTFS_HEAL_MAX_CONCURRENT_PER_SET=1
      - RUSTFS_HEAL_TASK_TIMEOUT_SECS=604800
      - RUSTFS_SCANNER_START_DELAY_SECS=600

Logs:

```
{"timestamp":"2026-08-11T13:36:05.442773076+08:00","level":"ERROR","message":"failed to check XL2 v1 format: File corrupt","target":"rustfs_filemeta::filemeta::codec","filename":"crates/filemeta/src/filemeta/codec.rs","line_number":126,"threadName":"rustfs-worker","threadId":"ThreadId(32)"}
{"timestamp":"2026-08-11T13:36:05.442788397+08:00","level":"ERROR","message":"failed to check XL2 v1 format: File corrupt","target":"rustfs_filemeta::filemeta::codec","filename":"crates/filemeta/src/filemeta/codec.rs","line_number":126,"threadName":"rustfs-worker","threadId":"ThreadId(32)"}
{"timestamp":"2026-08-11T13:36:05.442891868+08:00","level":"ERROR","message":"failed to check XL2 v1 format: File corrupt","target":"rustfs_filemeta::filemeta::codec","filename":"crates/filemeta/src/filemeta/codec.rs","line_number":126,"threadName":"rustfs-worker","threadId":"ThreadId(32)"}
```

## Current Heal Status：
Heal Status: In Progress

  Scan Mode:     deep
  Scan Cycle:    7855
  Tasks:         1402 queued, 8 active
  Items:         0 scanned, 89582 healed, 0 failed
  Data:          0 B scanned, 0 B healed
[...]

PR #5970, opened 2026-08-12
fix(scanner): defer cycles during data movement

### LINKED ISSUE #5953: Scanner error occur again
**Describe the bug**
after I click rebalance,this error occur again：
{"timestamp":"2026-08-11T16:14:13.022540334+08:00","level":"ERROR","message":"Scanner cycle completed without a durable data usage snapshot","event":"scanner_persist_state","component":"scanner","subsystem":"runtime","cycle":33604,"outcome":"NoUpdate","state":"usage_not_durable","target":"rustfs::scanner","filename":"crates/scanner/src/scanner.rs","line_number":2957,"threadName":"rustfs-worker","threadId":"ThreadId(10)"}

**version**
v1.0.0-rc.1

marshawcoco:
## Related Issues

Fixes #5953

## Summary of Changes

- Classify scanner cycles deferred by active rebalance or decommission work as superseded instead of failed.
- Keep usage publication fail-closed during data movement, preserving the last authoritative snapshot and existing dirty state while reusing the bounded superseded-cycle backoff.
- Add a two-pool ECStore regression test that activates rebalance metadata and verifies that no usage snapshot is published.

## Verification

- `CARGO_TARGET_DIR=/data/rustfs-issue-5953-target /root/.local/bin/rustfs-cargo-safe test -p rustfs-scanner --lib scanner_cycle_is_superseded_while_rebalance_is_active -- --nocapture --test-threads=1`
- `CARGO_TARGET_DIR=/data/rustfs-issue-5953-target /root/.local/bin/rustfs-cargo-safe test -p rustfs-scanner --lib test_scanner_cycle_completion_prioritizes_persist_failure -- --nocapture --test-threads=1`
- `CARGO_TARGET_DIR=/data/rustfs-issue-5953-target /root/.local/bin/rustfs-cargo-safe clippy -p rustfs-scanner --lib --tests -- -D warnings`
- `/root/.local/bin/rustfs-cargo-safe fmt --all --check`
- `git diff --check`

## Impact

Rebalance and decommission no longer produce repeated `usage_not_durable` hard-failure events solely because the scanner intentionally deferred publication. No API, configuration, on-disk format, or wire-format changes are introduced.

## Additional Notes

This change does not publish partial usage or claim that usage is current while data movement is active. A complete scanner cycle still has to run and persist after data movement ends before a new authoritative snapshot is published.


terem42:
I ran your change's premise against a real multi-node cluster and hit a
second producer of the same error that this PR does not cover. Sharing it here
rather than on #5953 because it is about the fix's scope.

### What I ran

Stock **`1.0.0-rc.1`** (the published artifact, `rustfs 1.0.0-rc.1` @`778f1dfa2`),
3 nodes × 4 disks = **one pool of 12**, EC:4, ~800 MB fixture, on real block
devices. **No corruption injected, cluster clean.** The only action was a
**staged rolling restart** — one node at a time, verify it is serving, 120 s
pause, next node.

That produced **12 × `usage_not_durable`** on the node that held the scanner
lease, all of them:

- `"outcome":"NoUpdate"`
- `"line_number":2957` — i.e. the classifier site, not the
  `DataUsagePersistOutcome::Failed` site above it

### The route is `cycle_activity_baseline_failed`, not data movement

In the same window there were exactly **12 × `cycle_activity_baseline_failed`**
— a 1:1 match, and the paired records carry the same cycle number. Each baseline
failure names a peer that was mid-restart:

| `error` | count |
|---|---|
| `Io error: peer http://rustfs-n2:9000 is temporarily offline` | 5 |
| `Io error: peer http://rustfs-n3:9000 is temporarily offline` | 5 |
| `Io error: code: 'The service is currently unavailable'` | 2 |

```json
{"timestamp":"2026-08-12T04:38:13.573346918Z","level":"WARN",
 "message":"Scanner cycle skipped because cluster activity could not be baselined",
 "event":"scanner_set_state","state":"cycle_activity_baseline_failed",
 "error":"Io error: peer http://rustfs-n3:9000 is temporarily offline",
 "filename":"crates/scanner/src/scanner_io.rs","line_number":2313,
 "span":{"leader_epoch":2,"scan_mode":"Deep","want_cycle":199,
         "name":"nsscanner_with_status"}}
```

…followed **38 µs** later by the ERROR this PR is about, carrying the same
cycle number:

```json
{"timestamp":"2026-08-12T04:38:13.573384976Z","level":"ERROR",
 "message":"Scanner cycle completed without a durable data usage snapshot",
 "event":"scanner_persist_state","cycle":199,"outcome":"NoUpdate",
 "state":"usage_not_durable","filename":"crates/scanner/src/scanner.rs",
 "line_number":2957}
```

So the path is:

```
probe_scanner_activity() fails (peer restarting)
  -> scanner_io.rs:2313  return Incomplete          <-- NOT the line this PR changes
  -> usage_persist_outcome = NoUpdate
  -> scanner_cycle_completion_outcome(Incomplete, NoUpdate) => Failed
  -> scanner.rs:2957     usage_not_durable  (ERROR)
```

which is the same `(Incomplete, NoUpdate) => Failed` arm you identified, reached
from the early return **twelve lines above** the one you rewrote.

### Why this PR cannot cover it

Two independent reasons:

**The accounting closes.** Every one of the 12 ERRORs pairs with a visible WARN
naming a peer-offline cause — same count, same cycle numbers, microseconds
apart. If any ERROR had come through the data-movement branch instead, the
ERRORs would outnumber the WARN causes. (I can't show the absence of
`cycle_data_movement_active` directly — that record is `debug!` and these nodes
log at `info` — which is why I'm arguing from the pairing, not from a zero
count.)

**The branch is unreachable in this topology.** `scanner_data_movement_active()`
is `is_decommission_running() || is_rebalance_started()`, and both read state
that only an explicit pool operation creates (`rebalance_meta`,
`pool_meta.decommission`). Neither was ever requested here — and on one pool
neither *can* be: rebalance start returns `NotImplemented` when
`store.pools.len() == 1` (`rustfs/src/admin/handlers/rebalance.rs:454`), and
decommission refuses with *"single pool deployments do not support
decommission"* (`crates/ecstore/src/core/pools.rs:1077`).

I therefore did **not** test a PR-head binary: on a single-pool cluster it is
behaviourally identical to its parent for this path, by construction.

`classify_nsscanner_cycle` and the two sibling early returns
(`cycle_activity_baseline_failed`, `no_disk_sets`) are all still `Incomplete`
producers after this PR, and `scanner_cycle_completion_outcome` is byte-identical
between `1.0.0-rc.1` and `main` — so reclassifying only the data-movement branch
leaves the others reaching the same arm.

### The bit that may matter for the wording

The ERROR says *"Scanner cycle completed without a durable data usage
snapshot"*, but in this path the cycle **never started** — it returned before
listing buckets, and no snapshot was attempted. A transient peer restart is
normal operation (every rolling upgrade does it), so an operator-facing ERROR
here reads as data loss when nothing was lost.

Whether that is best fixed the same way you fixed the data-movement branch, or
in `scanner_cycle_completion_outcome` so that "no snapshot attempted" stops
being a hard failure generally, is your call — the second would cover
`no_disk_sets` and the `classify_nsscanner_cycle` producers at the same time.

### Scope — what this is not

- **This is not #5950's 20-day loop.** Here it is transient and self-clearing:
  12 events across 2m11s during the restarts, then **zero in the following ~70
  minutes** while the scanner ran normally. I am not claiming the sustained
  case.
- Only the node running cycles during the peer outage emitted anything. The node
  restarted last ran **3166 of 3167 cycles to completion with zero errors** — so
  the trigger really is "a peer is away while this node probes", not a broken
  node.
- Deviations from both reporters' setups: 3 nodes/12 disks vs 4/48, ~800 MB vs
  2.87 TB, and I ran `RUSTFS_SCANNER_SPEED=fastest` with
  `RUSTFS_SCANNER_BITROT_CYCLE_SECS=180` as accelerants. The trigger here is a
  rolling restart, not rebalance.

marshawcoco:
Thank you @terem42 for the detailed reproduction and the precise cycle-to-log correlation. I reviewed the current control flow and confirmed your conclusion: `cycle_activity_baseline_failed` is an independent producer of `Incomplete + NoUpdate`, and the current PR does not cover it.

I also agree that `usage_not_durable` is misleading when the scan was deferred before bucket listing and no snapshot was attempted. However, globally reclassifying every `Incomplete + NoUpdate` would be unsafe because the same combination can represent real topology, bucket-scan, namespace, or disk-set failures.

I will revise the fix to model known pre-scan deferrals explicitly, covering both active data movement and an unavailable activity baseline. Deferred cycles will preserve the last durable snapshot and dirty work, publish no partial snapshot, use bounded retry behavior, and avoid advancing the scanner cycle as if scanning had occurred. Structural conditions such as `no_disk_sets` and incomplete topology will remain failures. I will also add regression coverage for the rolling-restart path and negative tests ensuring genuine failures are not downgraded.

PR #5989, opened 2026-08-12
feat(table-catalog): add atomic table rename

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Adds the standard Iceberg REST table rename operation with an atomic durable catalog update.

- Adds `POST /v1/{prefix}/tables/rename` for the primary and compatibility catalog prefixes with strict request parsing, bounded request bodies, route inventory coverage, and Iceberg-compatible status and error responses.
- Authorizes both source and destination table resources with the same authenticated principal before changing catalog state.
- Implements rename as one durable-strong snapshot CAS while preserving table ID, table UUID, warehouse location, current metadata pointer, version token, generation, properties, timestamps, and existing object keys.
- Preserves the original metadata root for commits, snapshot validation, conflict checks, recovery, and manifest traversal after rename.
- Keeps the source identifier reusable after rename while isolating the recreated table by table identity, warehouse, and current metadata pointer.
- Recovers ambiguous snapshot-write outcomes by verifying the durable source and destination table identities.
- Advertises rename only when durable-strong backing is active; object-backed mode returns an unsupported-operation response because it cannot atomically rename multi-object catalog state.
- Adds positive, negative, restart, failure-injection, collision, authorization, route, serde, and post-rename commit coverage.

## Verification

- `cargo fmt --all --check`
- `git diff --check`

## Impact

Durable-strong table catalogs gain standard same-namespace and cross-namespace table rename support. Existing table data and metadata objects are not moved, and subsequent commits continue in the table's pre-rename metadata root. Object-backed catalogs continue to operate unchanged and explicitly reject rename with HTTP 406.

## Additional Notes

Atomic rename is intentionally tied to durable-strong backing. Implementing it for object-backed mode would require a recoverable journal and read fencing across multiple catalog objects.

### DIFF
--- a/rustfs/src/admin/handlers/table_catalog/mod.rs
+++ b/rustfs/src/admin/handlers/table_catalog/mod.rs
@@ -76,6 +76,8 @@ const MIN_TABLE_CATALOG_CREDENTIAL_TTL_SECONDS: i64 = 60;
 const MAX_TABLE_CATALOG_CREDENTIAL_TTL_SECONDS: i64 = 60 * 60;
 const NAMESPACE_REQUEST_BODY_MAX_SIZE: usize = MAX_ADMIN_REQUEST_BODY_SIZE;
 const NAMESPACE_REQUEST_BODY_TIMEOUT: StdDuration = StdDuration::from_secs(10);
+const RENAME_TABLE_BODY_MAX_SIZE: usize = 16 * 1024;
+const RENAME_TABLE_BODY_TIMEOUT: StdDuration = StdDuration::from_secs(10);
 const WAREHOUSE_PROPERTY: &str = "warehouse";
 const PREFIX_PROPERTY: &str = "prefix";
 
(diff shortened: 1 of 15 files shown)


rustfs/src/admin/handlers/table_catalog/mod.rs
```diff
@@ -5168,38 +5142,21 @@ where
...
+            err => catalog_store_error(err),
+        })?;
+    validate_table_snapshot_commit_conflicts(metadata_backend, bucket, &current, &expected_metadata, &request.updates).await?;
     validate_metadata_matches_current_metadata(&expected_metadata, &next_metadata)?;
     let (commit_id, metadata_file_token) = standard_commit_ids(request.commit_id.or_else(|| request.idempotency_key.clone()));
     let next_generation = current.generation.saturating_add(1);
-    let next_metadata_location = crate::table_catalog::default_table_metadata_file_path(
-        namespace,
-        &table_name,
+    let next_metadata_location = crate::table_catalog::table_metadata_file_path_for_entry(
```
overtrue, line 5155: After a rename, this keeps writing under the old identifier's metadata directory, but that identifier can immediately be recreated. Both tables can then be at generation 1 and accept the same valid commit ID, so each targets the same `00002-<id>.metadata.json`; the second commit returns 409 after the first `IfAbsent` write. I reproduced this on this head with rename → recreate source → commit both using the same ID. Please keep generated metadata paths disjoint by immutable table identity and add this regression case.

overtrue (changes requested):
One metadata-isolation bug remains after source reuse.

PR #5990, opened 2026-08-12
fix(ecstore): scope UploadPart commit lock per part number

### LINKED ISSUE #5961: [Bug] Concurrent UploadPart requests for different part numbers are serialized by an upload-ID-scoped write lock and return 503
## Describe the bug

RustFS `1.0.0-rc.1` returns `503 ServiceUnavailable` when multiple
`UploadPart` requests for the same upload ID, but different part numbers,
reach the commit phase concurrently.

All RustFS nodes were healthy. The failures consistently completed after
approximately five seconds, matching RustFS's default namespace-lock
acquisition timeout.

Concurrent uploads of different multipart part numbers are valid S3 client
behavior and should not serialize on one exclusive upload-level lock.

## Environment

- RustFS: `1.0.0-rc.1`, commit `778f1dfa2155cbbc61ad54e6896de9e29d2c4d8d`
- Deployment: 4 RustFS nodes, `EC:2`
- Storage: Ceph RBD-backed Kubernetes volumes
- Client: Apache Flink 1.20.5 with `flink-s3-fs-presto`
- Workload: Flink checkpoints using concurrent multipart uploads
- Checkpoint state size: approximately 2.921 GiB
- Individual state objects: approximately 195-221 MiB

The same locking behavior is still present on `main` at
`a206a0779efa49a78990f1ceba624fedc1fbe04b`.

## Observed behavior

During checkpoint 27570:

1. Several `UploadPart` requests used the same upload ID.
2. The requests had different part numbers.
3. They reached all four RustFS replicas concurrently.
4. Requests returned `503 ServiceUnavailable` after approximately
   `5.04-5.29s`.

The following checkpoint then reported client-side errors including
`AwsChunkedStreamError`, `Broken pipe`, and aborted connections. Those appear
to be consequences of the failed part uploads rather than the initial cause.

## Root-cause hypothesis

RustFS defaults the object lock acquisition timeout to five seconds:

https://github.com/rustfs/rustfs/blob/778f1dfa2155cbbc61ad54e6896de9e29d2c4d8d/crates/config/src/constants/object.rs#L362-L372

During `put_object_part`, the commit phase takes an exclusive write lock on
`upload_id_path`:

https://github.com/rustfs/rustfs/blob/778f1dfa2155cbbc61ad54e6896de9e29d2c4d8d/crates/ecstore/src/set_disk/ops/multipart.rs#L1072-L1091

Therefore, commits for all part numbers within the same upload ID are
serialized. When one commit holds the lock long enough, concurrent commits for
other part numbers exceed the five-second acquisition timeout.

Lock and namespace-lock quorum failures are then mapped to S3
`ServiceUnavailable`:

https://github.com/rustfs/rustfs/blob/778f1dfa2155cbbc61ad54e6896de9e29d2c4d8d/rustfs/src/error.rs#L290-L297

Serializing retries of the same part number is necessary to prevent
mixed-generation shards. Serializing different part numbers does not appear
necessary because they commit to different final part paths.

## MinIO reference design

MinIO uses two locks during `PutObjectPart`:

- An exclusive write lock scoped to `object/uploadID/partID`
- A shared read lock scoped to `object/uploadID`

https://github.com/minio/minio/blob/7aac2a2c5b7c882e68c1ce017d8256be2feea27f/cmd/erasure-multipart.go#L743-L760

This allows different part numbers to commit concurrently, while retries for
the same part number remain serialized. The upload-ID read lock also prevents
the commit from racing with completion or abort operations that take the
upload-ID write lock.

MinIO additionally uses a dynamic general operation timeout initialized at
10 minutes with a 5-minute minimum:

https://github.com/minio/minio/blob/7aac2a2c5b7c882e68c1ce017d8256be2feea27f/cmd/globals.go#L320-L321

However, the main behavioral difference is the lock scope, not only the
timeout value.

## Expected behavior

- Concurrent commits for different part numbers under one upload ID should
  proceed independently.
- Concurrent retries for the same part number should remain serialized.
- `CompleteMultipartUpload` and `AbortMultipartUpload` must wait for active
  part commits and must not race with them.
- Normal multipart concurrency should not return 503 due to upload-level lock
  contention.

## Suggested fix

Use locking similar to MinIO:

1. Take an upload-ID shared/read lock while committing a part.
2. Take a part-specific exclusive/write lock using the upload ID plus part
   number or final part path.
3. Keep `CompleteMultipartUpload` and `AbortMultipartUpload` on the upload-ID
   exclusive/write lock.
4. Add concurrency tests covering:
   - Different part numbers committing concurrently
   - Retries of the same part number
   - Completion racing with active part uploads
   - No lock timeout under normal multipart concurrency

Increasing `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT` may reduce the immediate 503
rate, but it only extends the wait and retains upload-level head-of-line
blocking.

## Impact

Large-object and checkpoint workloads commonly upload multiple parts in
parallel. Under this locking scheme, storage latency can turn normal multipart
concurrency into deterministic 503 responses, failed checkpoints, client
retries, and connection churn.

## Related issue

Related but not a duplicate:

- https://github.com/rustfs/rustfs/issues/5839

That issue concerns `UploadPart` behavior during node loss. This report occurs
with all nodes healthy and is caused by normal concurrent commits for different
part numbers under the same upload ID.

overtrue:
## Related Issues

Fixes #5961.

## Summary of Changes

`put_object_part` previously took an exclusive write lock on the whole `upload_id_path` namespace for its commit phase, so concurrent `UploadPart` commits for **different part numbers** of the same upload ID serialized behind one lock. When one commit held the lock longer than the 5s default lock-acquire timeout (slow storage, many parallel parts), the queued commits failed and surfaced as `503 ServiceUnavailable` — deterministic checkpoint failures for clients like Flink that upload parts in parallel.

This PR adopts MinIO's `PutObjectPart` lock scope (minio/minio `erasure-multipart.go`):

- The part commit now takes a **shared read lock on `upload_id_path`** plus an **exclusive write lock on `{upload_id_path}/part.{part_number}`**, acquired in that order (upload → part), scoped to the commit only.
- Different part numbers commit concurrently: they rename onto disjoint final part paths, so the mixed-generation shard hazard (backlog#853) does not apply across part numbers.
- Retries of the **same** part number still serialize on the part-scoped write lock, keeping each same-part commit atomic across disks (last committer wins consistently).
- `CompleteMultipartUpload` / `AbortMultipartUpload` are unchanged: they take the upload-ID **write** lock, so they still mutually exclude every in-flight part commit (the shared read locks block them), and completion still takes the object lock before the upload lock.
- The lock-loss commit fence now covers both guards (extracted into `fence_commit_on_lock_loss`): losing refresh quorum on either the upload read lock or the part write lock aborts the commit before `rename_part`.

## Verification

- New regression test `put_object_part_different_part_numbers_commit_concurrently` reproduces the issue deterministically with no wall-clock bound: a two-arrival commit barrier releases only once BOTH part commits are simultaneously paused inside their commit sections holding their locks. On `main` the second commit blocks on the upload-wide write lock and errors at the 5s acquire timeout, so the barrier is never reached and the test fails. The test then completes the upload and byte-verifies both parts round-trip.
- New test `put_object_part_same_part_retries_serialize_on_part_lock` pins backlog#853 from both directions: while the first commit is paused holding its locks, (a) a shared probe on the part key must time out — pinning that the part lock is exclusive, not shared; (b) a same-part retry must block on the part key before committing; the retry's generation is then completed and byte-verified as the served content.
- New test `put_object_part_fences_part_lock_loss_before_rename` pins the new part-guard half of the lock-loss fence, asserting the failure names the part lock key (mirrors the existing upload-guard fence test).
- Existing lock-linearization tests pass unchanged against the new scheme (54 passed in the multipart suite): `put_object_part_holds_upload_lock_through_rename` (abort waits for an in-flight part commit), `complete_validates_parts_after_an_inflight_upload_part_commit` (completion waits and revalidates), `put_object_part_rechecks_upload_after_commit_lock`, `put_object_part_fences_upload_lock_loss_before_rename`, and the complete/abort linearization suite.
- `cargo test -p rustfs-ecstore --lib set_disk::ops::multipart::tests`
- `make pre-pr` (fmt + arch checks + clippy + tests): all gates green; 13316/13316 workspace tests passed. One pre-existing test unrelated to this diff, `rustfs-utils net::test::test_resolve_domain_preserves_system_resolver_error_provenance`, was excluded locally because this development machine's fake-IP DNS proxy resolves the RFC 6761 `.invalid` TLD to 198.18.x.x, which breaks the test in any branch; it is expected to pass in CI.

Adversarial Validation (high-risk tier, all seven roles):

- Correctness: no blocker/major; lock key is exact-match and collision-free for legit IDs; failure ordering and 503 mapping verified. Pre-existing input-validation gap (crafted base64 uploadId can decode to a `/`-containing path and alias lock keys) recorded for follow-up — independent of this change.
- Concurrency/durability: no blocker/major; lock hierarchy lifecycle→object→upload→part is acyclic across all production sites; both guards heartbeat and fence independently; `recover_part_transactions` only ever runs under the upload write lock or per-part under the part lock.
- Compatibility: no format/wire change; rolling upgrade verified at key and quorum level — old-node upload write locks and new-node read+part locks mutually exclude in both directions (worst case is the old serialized behavior); `ListParts` concurrency with in-flight commits matches MinIO RLock semantics.
- Security: no new log lines; part number is capped at 10000 before locking, so the lock-key space is bounded; error payload exposes the same opaque hash form as the existing upload-path fence.
- Performance: +1 lock-quorum RTT per UploadPart commit, against removal of upload-wide head-of-line blocking; guard live range otherwise unchanged; diag metrics gated off by default.
- Simplicity: duplicate fence blocks extracted into a helper; inert test scaffolding removed.
- Test-coverage skeptic: revert-detection matrix complete for all four behavior hunks; the write→read weakening of the part lock is now pinned by the exclusivity probe; the wall-clock bound flagged as flaky was replaced by the two-arrival barrier design.

## Impact

- Behavior: normal S3 multipart concurrency (parallel `UploadPart` for different part numbers of one upload) no longer serializes on an upload-wide lock and no longer returns 503 under storage latency; same-part retry semantics and complete/abort linearization are unchanged.
- Compatibility: no on-disk, metadata, or wire format change. During a rolling upgrade, an old node's upload-wide write lock and a new node's read+part locks still mutually exclude, degrading to the old serialized behavior at worst.
- Performance: one extra namespace lock acquisition per `UploadPart` commit, in exchange for removing upload-wide head-of-line blocking.

## Additional Notes

Lock ordering stays acyclic: part commits acquire `upload(read) → part(write)`; completion acquires `object(write) → upload(write)`; abort acquires `upload(write)` only. No path acquires the part lock before the upload lock, and no path holding the part lock waits on the object lock. Holding the upload read lock while waiting on the part lock mirrors MinIO's nesting (MinIO holds the uploadID RLock across the entire `PutObjectPart`, including its partID lock wait; this change confines that window to the commit phase only), and the local lock shard is writer-preferring, so completion cannot be starved by a stream of new part commits.

### DIFF
--- a/crates/ecstore/src/set_disk/ops/multipart.rs
+++ b/crates/ecstore/src/set_disk/ops/multipart.rs
@@ -162,6 +162,22 @@ fn map_upload_id_metadata_error(bucket: &str, object: &str, upload_id: &str, err
     err.into()
 }
 
+/// Abort a multipart commit when the guard's refresh heartbeat has observed a
+/// refresh-quorum loss (backlog#899 Phase 2): a stale holder must not race a
+/// concurrent committer past its fenced commit point.
+fn fence_commit_on_lock_loss(guard: Option<&ObjectLockDiagGuard>, mode: &'static str, lock_path: &str) -> Result<()> {
+    if guard.is_some_and(|guard| guard.is_lock_lost()) {
+        return Err(StorageError::NamespaceLockQuorumUnavailable {
+            mode,
+            bucket: RUSTFS_META_MULTIPART_BUCKET.to_string(),
+            object: lock_path.to_string(),
+            required: 1,
+            achieved: 0,
+        });
+    }
+    Ok(())
+}
+
 fn multipart_bucket_incarnation_id(metadata: &HashMap<String, String>) -> Result<Option<Uuid>> {
     let Some(value) = rustfs_utils::http::metadata_compat::get_consistent_str(metadata, SUFFIX_BUCKET_INCARNATION_ID) else {
         if rustfs_utils::http::metadata_compat::contains_key_str(metadata, SUFFIX_BUCKET_INCARNATION_ID) {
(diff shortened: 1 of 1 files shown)


overtrue (commented):

PR #5991, opened 2026-08-12
perf(ecstore): retain remote shard HTTP chunks

GatewayJ:
## Related Issues

N/A

## Summary of Changes

Retain remote HTTP body chunks through bitrot verification for eligible remote shard reads. The path hashes fragmented `Bytes` without concatenating the full encoded shard into scratch space and appends verified data directly to the decoder buffer.

Bound each bitrot block to 64 retained chunk descriptors. The first 64 chunks remain ownership-preserving; only a fragmented tail is coalesced into the reader's reusable scratch buffer. This prevents tiny-frame responses from growing descriptor memory and hash-call counts with the number of HTTP frames.

Restore the synchronous in-memory block path before probing the chunk capability, preserving the existing local/inline behavior without a fragment allocation. `HttpChunkReader` now returns at most one owned chunk per poll, uses one state machine for both `AsyncRead` and chunk reads, and fails closed with a sticky `InvalidData` error after 64 consecutive empty body chunks. The consecutive count survives `Pending` boundaries and resets only after non-empty data.

## Verification

- `make pre-pr` — passed; 13,405 tests passed and 151 were skipped by configuration
- `cargo test -p rustfs-rio --lib` — 127 tests passed, including the cross-`Pending` empty-frame regression
- `cargo test -p rustfs-ecstore --lib chunked`
- `cargo test -p rustfs-ecstore --lib streaming_fallback_reuses_initialized_scratch`
- `cargo test -p rustfs-ecstore --lib in_memory_fast_path_fires_and_matches_the_streaming_path`
- `cargo test -p rustfs-utils --features hash hash_encode_slices_matches_contiguous_for_all_algorithms --lib`
- `cargo check -p rustfs-ecstore --lib --features hotpath`
- `cargo clippy -p rustfs-rio -p rustfs-ecstore --lib --tests -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`
- `git range-diff` confirmed both PR commits remained patch-equivalent after the final clean rebase onto `origin/main`; the rio and ecstore focused tests above passed again on that base

## Impact

No S3 API, internode HTTP wire format, authentication, on-disk bitrot format, configuration, or migration change. The optional owned-chunk transport and disk capabilities retain default `None` fallbacks, so existing implementations continue through the established `AsyncRead` path; capability errors remain fail-closed.

Normal remote reads retain the intended copy reduction: data is copied once into the decoder output instead of first copying the full `[hash][data]` block into scratch. Local in-memory reads use the synchronous block path without a fragment heap allocation. Pathological fragmentation is bounded to 64 retained `Bytes` handles plus one reusable tail buffer. The reusable chunk vector adds a fixed empty `Vec` handle to each `BitrotReader`; eliminating that fixed metadata without adding remote per-block allocations would require a broader state-layout redesign and is intentionally outside this focused PR.

## Additional Notes

The fast path remains limited to eligible remote reads. Local reads, stage-instrumented paths, unsupported transports, and existing fallback readers retain their established behavior. A deployed distributed throughput A/B benchmark was not run; the performance assessment used copy/allocation/poll counts plus production-sized fragmentation tests.

- Correctness: attacked partial, short, corrupt, oversized, empty, fragmented, fallback, and cross-block reads — no break found.
- Simplicity: the `AsyncRead` implementation delegates to the single bounded chunk state machine; no smaller localized fix preserves the same verification and fallback semantics.
- Security: attacked authenticated tiny/empty-frame amplification, error handling, and verify-before-output behavior — no break found.
- Concurrency/durability: attacked `Pending`, cancellation, timeout/hedge retirement, and reader reuse after failure; the empty-frame count survives scheduler boundaries and the resulting error is sticky — no break found.
- Compatibility: attacked mixed-version transport defaults, wire/auth behavior, all hash algorithms, and `[hash][data]` layout — no break found.
- Performance: normal remote copy reduction and the local synchronous fast path are preserved; descriptor/hash amplification is bounded. The fixed empty-vector metadata tradeoff is noted above.
- Test coverage: 63/64 empty-frame, cross-`Pending` 63+1 empty-frame, non-empty reset, 64/65/66 fragment, production-sized one-byte fragmentation, scratch reuse, EOF, body error, and corruption paths have revert-detecting tests.

---

### DIFF
--- a/crates/ecstore/src/cluster/rpc/internode_data_transport.rs
+++ b/crates/ecstore/src/cluster/rpc/internode_data_transport.rs
@@ -31,7 +31,7 @@ use rustfs_config::{
     DEFAULT_INTERNODE_DATA_TRANSPORT, ENV_RUSTFS_INTERNODE_DATA_TRANSPORT, INTERNODE_DATA_TRANSPORT_TCP,
     KNOWN_INTERNODE_DATA_TRANSPORT_BACKENDS,
 };
-use rustfs_rio::{HttpReader, HttpWriter};
+use rustfs_rio::{ChunkReaderBox, HttpChunkReader, HttpReader, HttpWriter};
 use sha2::{Digest, Sha256};
 use std::collections::HashMap;
 use std::future::Future;
@@ -221,6 +221,11 @@ pub struct NsScannerCapabilityRequest {
 #[async_trait]
 pub trait InternodeDataTransport: Send + Sync + std::fmt::Debug {
     async fn open_read(&self, request: ReadStreamRequest) -> Result<FileReader>;
+    /// Opens an owned-chunk stream when this transport can retain receive-buffer
+    /// ownership. `None` preserves the established `open_read` fallback.
+    async fn open_read_chunks(&self, _request: ReadStreamRequest) -> Result<Option<ChunkReaderBox>> {
+        Ok(None)
+    }
     async fn open_write(&self, request: WriteStreamRequest) -> Result<FileWriter>;
     async fn open_walk_dir(&self, request: WalkDirStreamRequest) -> Result<FileReader>;
     async fn open_ns_scanner(&self, _request: NsScannerStreamRequest) -> Result<FileReader> {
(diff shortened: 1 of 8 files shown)


GatewayJ:
@codex

overtrue (approved):

overtrue (approved):

overtrue (approved):
LGTM

PR #5995, opened 2026-08-12
refactor(kms): consolidate encryption metadata key constants into their shared home

overtrue:
# refactor(kms): consolidate encryption metadata key constants into their shared home

## Summary

Closes the constant-duplication half of rustfs/backlog#1775. The shared module `rustfs_utils::http::object_encryption_keys` is the single source of truth for encryption metadata key names, but three call sites still carried their own copies or bare literals. This PR removes every duplicate; all values are unchanged, so the change is a pure rename that the compiler verifies.

- `crates/kms/src/service.rs`: drops its private `INTERNAL_ENCRYPTION_KEY_ID_HEADER` / `INTERNAL_ENCRYPTION_ALGORITHM_HEADER` copies and its four bare `"x-rustfs-encryption-iv/tag/key/context"` literals (write path in `metadata_to_headers`, read path in `headers_to_metadata`) in favor of the shared constants. The ChaCha20 round-trip doc comment moves to the canonical constant in the shared module.
- `rustfs/src/app/select_object.rs`: drops the six `SELECT_*` copies of the MinIO-internal sealed-key/context/key-id names and the RustFS key-id name.
- `rustfs/src/storage/options.rs`: drops its private `RUSTFS_ENCRYPTION_PREFIX` / `MINIO_ENCRYPTION_PREFIX` copies; `header_compat`'s existing constants become `pub` and are imported instead.
- `crates/utils/src/http/object_encryption_keys.rs` gains `RUSTFS_INTERNAL_ENCRYPTION_PREFIX` so the redaction predicate (`rustfs_filemeta::is_sensitive_metadata_key`) and the replication strip predicate share one definition instead of two hardcoded literals.

## The `X-Rustfs-Internal-Server-Side-Encryption-*` judgment call

rustfs/backlog#1775 asked whether the reader-only `x-rustfs-internal-server-side-encryption-` family is dead code or missing a writer. Verified by sweep: no writer exists anywhere in the repo (the SSE writer persists the MinIO-branded keys verbatim for interop; no `metadata_compat` dual-write call uses an SSE suffix). Two comments claimed the dual-key invariant writes this twin — that was wrong, and both are corrected. The readers are kept, not removed: they are defensive redaction/strip guards on a reserved name family, removing them is risk-asymmetric (cheap to keep, dangerous if a future or third-party writer appears), and the judgment is now recorded on the shared constant's doc comment.

Test literals that pin the wire format (e.g. `service.rs`'s header snapshot assertions) deliberately keep their bare strings so a constant rename cannot silently change the on-disk contract.

## Verification

- `cargo check -p rustfs-utils -p rustfs-kms -p rustfs-filemeta -p rustfs`
- `cargo clippy -p rustfs-utils -p rustfs-kms -p rustfs-filemeta --all-targets -- -D warnings`
- `cargo test -p rustfs-utils --lib`, `cargo test -p rustfs-filemeta --lib fileinfo`, `cargo test -p rustfs-kms --lib service`, `cargo test -p rustfs --lib app::select_object`, `cargo test -p rustfs --lib storage::options`
- `cargo fmt --all`

Refs rustfs/backlog#1775, rustfs/backlog#1562.

### DIFF
--- a/crates/filemeta/src/fileinfo.rs
+++ b/crates/filemeta/src/fileinfo.rs
@@ -277,9 +277,10 @@ pub struct FileInfo {
 /// Values of these keys must never reach logs at any level.
 fn is_sensitive_metadata_key(key: &str) -> bool {
     // `is_encryption_metadata_key` covers the x-minio-internal- SSE prefix but not
-    // its x-rustfs-internal- twin, which the dual-key invariant writes alongside it.
+    // its reserved x-rustfs-internal- twin, which has no writer today but must
+    // stay redacted in case one appears.
     is_encryption_metadata_key(key)
-        || starts_with_ignore_ascii_case(key, "x-rustfs-internal-server-side-encryption-")
+        || starts_with_ignore_ascii_case(key, rustfs_utils::http::RUSTFS_INTERNAL_ENCRYPTION_PREFIX)
         || rustfs_utils::http::REPLICATION_SSE_TRANSPORT_PREFIXES
             .iter()
             .any(|prefix| starts_with_ignore_ascii_case(key, prefix))
--- a/crates/kms/src/service.rs
+++ b/crates/kms/src/service.rs
@@ -27,6 +27,10 @@ use base64::Engine;
 use jiff::Zoned;
 use md5::{Digest as Md5Digest, Md5};
 use rand::random;
+use rustfs_utils::http::object_encryption_keys::{
+    INTERNAL_ENCRYPTION_ALGORITHM_HEADER, INTERNAL_ENCRYPTION_CONTEXT_HEADER, INTERNAL_ENCRYPTION_IV_HEADER,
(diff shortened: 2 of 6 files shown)


overtrue (commented):

overtrue (commented):

overtrue:
Rebased onto main to resolve conflicts with the just-merged #5997/#5998 (which landed the kms constant and select_object `SELECT_*` halves of this change first). Conflict resolution: kept main's versions where they fully covered, kept this PR's additional scope (bare `x-rustfs-encryption-*` literals in `metadata_to_headers`/`headers_to_metadata`, `storage/options.rs` prefixes, `RUSTFS_INTERNAL_ENCRYPTION_PREFIX` + ChaCha doc on the canonical constants, fileinfo predicate sharing). Re-verified after rebase: clippy `-D warnings` on utils/kms/filemeta (all targets) + rustfs (lib+tests) clean; tests green (utils 24, filemeta fileinfo 52, kms service 25, select_object 47, storage::options 67); `make pre-commit` ✅.

PR #5999, opened 2026-08-12
ci(kms): add a nightly live-Vault lane and stop leaking behavior keys

overtrue:
## Summary

Implements both halves of rustfs/backlog#1774: the Vault backends finally get a CI lane, and the behavior harness stops leaking keys into the Vault it runs against.

### 1. Nightly live-Vault lane (`.github/workflows/nightly-gnu.yml`, new `kms-vault-lane` job)

`rg -li vault .github/` used to be zero hits: no workflow ever set `RUSTFS_KMS_VAULT_TOKEN`, so `live_vault_backends()` returned an empty set in every CI run and `behavior_rotation.rs` never asserted the working half of rotate/versioning — this is also why the `#[ignore]` live-Vault tests in rustfs/backlog#1562 had never executed in CI. The new job:

- runs a `vault server -dev` (CLI from HashiCorp's apt repo) with KV v2 + Transit enabled, proving both engines work before the tests start;
- runs `cargo test -p rustfs-kms` with the lane switched on (all `for_each_backend` specs across Local/Static/KV2/Transit);
- runs the dev-Vault `#[ignore]` tests (`backends::contract_tests`, `backends::vault*`, `vault_fault_injection`) — the AWS-credentialed ignored tests stay excluded;
- runs the self-hosting AppRole live script (`scripts/test/vault_approle_kms_live.sh`).

A second job, `kms-vault-ha-failover`, runs `scripts/test/vault_ha_kms_live.sh` (three-node Raft leader failover in Docker, first validated by rustfs/rustfs#5653). It is a separate job so an election-timing flake cannot mask the main lane's verdict, and vice versa.

Runner choice deviates from the issue text deliberately: the issue suggested `sm-standard-4` + docker-in-step, but the banner in `e2e-s3tests.yml` (rustfs/backlog#1149) records that the self-hosted fleet is heterogeneous and burned the last docker-dependent workflow — all 15 runs failed on pods without `docker.sock`. GitHub-hosted `ubuntu-latest` guarantees Docker for the HA script. The job token `dev-only-token` is not a secret (loopback-only ephemeral server, holds only test keys) and its literal value is load-bearing — the dev-Vault fixtures in `backends/vault.rs` hardcode it.

### 2. Behavior harness key cleanup (`crates/kms/tests/common/mod.rs`)

`TestKms::create_key` now records every created key id; `for_each_backend` removes the recorded keys after each Vault-backed case through a fresh manager over the same configuration (fresh because the spec consumes — and may have restarted or stopped — the case's own manager). The cleanup config alone gets `with_immediate_deletion_allowed()`, so the configuration under test keeps the immediate-deletion gate at its production default and gate-refusal specs stay honest. Deletion runs twice per key because the Transit backend parks a key in PendingDeletion on the first `force_immediate` call and destroys it on the second, while KV2 destroys on the first and reports KeyNotFound after. Failures print and continue — a Vault hiccup during cleanup must not turn a green behavior run red.

### 3. Two drive-by findings from validating the lane against a real dev Vault

- `test_vault_cancel_key_deletion_persists_state` (dev-Vault `#[ignore]` test) was broken by construction: it built its config from `Default::default()`, which never picks up the insecure-dev-defaults env override, so the HTTP dev Vault it requires was always refused. Fixed to declare development mode on the config. This had gone unnoticed precisely because the test never ran anywhere — the lane would have caught it on night one.
- **Filed separately, not fixed here**: the Vault Transit backend cannot start against a brand-new (empty) transit engine at all — its health check uses `key::list`, which returns 404 when the engine has no keys, and the service's own probe key is only created after the health check passes. The workflow works around it by pre-creating a probe transit key.

## Verification (against a real `hashicorp/vault:1.17.6` dev server, locally)

- `cargo test -p rustfs-kms` with the lane on: 518 lib + all 12 behavior suites green across all four backends.
- After the run, the Vault holds zero `behavior-*` keys (KV2 metadata list and transit key list both show only the fixed-name probe keys) — the cleanup verified end-to-end.
- `cargo test -p rustfs-kms --lib backends::contract_tests -- --ignored`, `--lib backends::vault -- --ignored` (11/11), `--test vault_fault_injection -- --ignored`: all green.
- `scripts/test/vault_ha_kms_live.sh`: the cluster forms and the failover test executes, but on this macOS machine it times out waiting for the replacement Raft leader under Docker-for-Mac — an environment artifact this script has hit locally before; the script's Linux validation is rustfs/rustfs#5653. This is exactly why the HA lane is a separate job: if it proves flaky on `ubuntu-latest` too, it can be tuned or gated without touching the main lane's signal.
- `actionlint .github/workflows/nightly-gnu.yml`: clean.
- `cargo clippy -p rustfs-kms --all-targets -- -D warnings`, `cargo fmt --all`, `make pre-commit`.

Refs rustfs/backlog#1774, rustfs/backlog#1562.

### DIFF
--- a/crates/kms/tests/common/mod.rs
+++ b/crates/kms/tests/common/mod.rs
@@ -32,14 +32,14 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use std::future::Future;
 use std::path::PathBuf;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use base64::Engine as _;
 use base64::engine::general_purpose::STANDARD as BASE64;
 use rustfs_kms::backends::BackendCapabilities;
 use rustfs_kms::{
-    CreateKeyRequest, KeyUsage, KmsConfig, KmsError, KmsManager, KmsServiceManager, KmsServiceStatus, ObjectEncryptionService,
-    Result,
+    CreateKeyRequest, DeleteKeyRequest, KeyUsage, KmsConfig, KmsError, KmsManager, KmsServiceManager, KmsServiceStatus,
+    ObjectEncryptionService, Result,
 };
 use tempfile::TempDir;
 
@@ -126,6 +126,12 @@ pub struct TestKms {
     manager: Arc<KmsServiceManager>,
     kind: BackendKind,
     config: KmsConfig,
+    /// Ids of the keys [`TestKms::create_key`] created, so a run against a
+    /// persistent Vault can remove them afterwards instead of accumulating
+    /// `behavior-*` keys forever (rustfs/backlog#1774). Shared through an Arc
+    /// because the harness instance is consumed by the spec while the cleanup
+    /// runs after it.
+    created_keys: Arc<Mutex<Vec<String>>>,
     /// Held for the harness lifetime so the local key directory outlives a
     /// simulated process restart.
     _dir: Option<TempDir>,
@@ -147,6 +153,7 @@ impl TestKms {
             manager,
             kind: BackendKind::Local,
             config,
+            created_keys: Arc::new(Mutex::new(Vec::new())),
             _dir: Some(dir),
         }
     }
@@ -166,6 +173,7 @@ impl TestKms {
             manager,
             kind: BackendKind::VaultKv2,
             config,
+            created_keys: Arc::new(Mutex::new(Vec::new())),
             _dir: None,
         }
     }
@@ -180,6 +188,7 @@ impl TestKms {
             manager,
             kind: BackendKind::VaultTransit,
             config,
+            created_keys: Arc::new(Mutex::new(Vec::new())),
             _dir: None,
         }
     }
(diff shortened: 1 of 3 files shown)


crates/kms/tests/common/mod.rs
```diff
@@ -253,8 +263,75 @@ impl TestKms {
...
+pub async fn cleanup_vault_keys(kind: BackendKind, config: &KmsConfig, created_keys: Arc<Mutex<Vec<String>>>) {
+    if !kind.is_vault() {
+        return;
+    }
+    let key_ids: Vec<String> = created_keys.lock().expect("created-keys lock").drain(..).collect();
+    if key_ids.is_empty() {
+        return;
+    }
+    let config = config.clone().with_immediate_deletion_allowed();
+    let manager = start_manager(&config).await;
```
overtrue, line 307: This still panics if Vault becomes unavailable after the behavior assertions pass: `start_manager` uses `expect` for configure/start. I reproduced it against Vault 1.17.6 by dropping the endpoint after the spec's final successful request; cleanup then exited 101 instead of reporting the failure. Please make cleanup startup/get-manager fallible and add a focused unreachable-endpoint regression.

overtrue (commented):
Cleanup can still fail a passing behavior test.

PR #6000, opened 2026-08-12
test(e2e): fold nine identical POST-policy rejection tests into one table

overtrue:
## What

PR1 of rustfs/backlog#1838. The nine `*_missing_from_policy_conditions` tests in `multipart_auth_test.rs` were body-identical after literal normalization (verified by hashing normalized function bodies): policy pins bucket + key + `content-length-range` only, the form smuggles one extra field the policy never declared, and the upload must be rejected with 403 AccessDenied naming the field. Each test booted its own full server.

- New `run_post_object_policy_case` helper: bucket, key, policy conditions, extra form fields, file body, expected status/code/mention, plus a `case` prefix on every assertion message so a failing row is identifiable at a glance. The signature is deliberately general enough for the `policy_mismatch` and sse-kms groups planned as PR2/PR3.
- The nine tests fold into one table-driven test with nine rows. **Every row keeps its original test's exact literals** — bucket, key, field name/value, body bytes, and expected error strings, including the two rows that asserted the stronger `<Code>AccessDenied</Code>` form — so case-row count equals folded-test count (9) and no poison value is lost.
- Folded: cache-control, content-language, content-encoding, website-redirect-location, expires, tagging, metadata, metadata-new-key, content-type. Not folded (different bodies): `accepts_sse_s3_...`, `storage_class`, `checksum_algorithm`, `checksum_auxiliary_fields`, `form_fields`, `success_action_redirect`, `object_lock_*`.

## Inventory

`docs/testing/e2e-suite-inventory.md` is updated in the same diff per the issue's hard constraint. Note the old row said 109 while `cargo nextest list` on main actually reports 111 for this module (0 `#[ignore]`s) — the row was stale by two before this change. It now carries the measured post-fold count: **103** (111 − 9 + 1).

## Verification

- `cargo nextest run -p e2e_test -E 'test(test_anonymous_post_object_rejects_fields_missing_from_policy_conditions)'` — PASS in 23.8s (nine real server round-trips)
- `cargo nextest list -p e2e_test -E 'test(/multipart_auth_test::/)'` — 103 tests, matches the inventory row
- `make pre-commit` — ✅

Ref rustfs/backlog#1838.

### DIFF
--- a/docs/testing/e2e-suite-inventory.md
+++ b/docs/testing/e2e-suite-inventory.md
@@ -63,7 +63,7 @@
 | list_objects_v2_metadata_extension_test | 1 |  |
 | list_objects_v2_pagination_test | 12 | ✅ |
 | mc_mirror_small_bucket_test | 1 |  |
-| multipart_auth_test | 109 |  |
+| multipart_auth_test | 103 |  |
 | multipart_storage_class_test | 3 | ✅ |
 | namespace_lock_quorum_test | 2 |  |
 | negative_sigv4_test | 6 | ✅ |
--- a/crates/e2e_test/src/multipart_auth_test.rs
+++ b/crates/e2e_test/src/multipart_auth_test.rs
@@ -285,6 +285,198 @@ async fn allow_anonymous_put_object(
     Ok(())
 }
 
+/// One rejected POST Object upload driven end-to-end (backlog#1838): starts a
+/// fresh server, allows anonymous PutObject on `bucket`, posts an anonymous
+/// POST Object form whose policy carries `policy_conditions` and whose form
+/// carries `form_fields` on top of the mandatory key+policy fields, then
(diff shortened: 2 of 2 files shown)


docs/testing/e2e-suite-inventory.md
```diff
@@ -63,7 +63,7 @@
 | list_objects_v2_metadata_extension_test | 1 |  |
 | list_objects_v2_pagination_test | 12 | ✅ |
 | mc_mirror_small_bucket_test | 1 |  |
-| multipart_auth_test | 109 |  |
+| multipart_auth_test | 103 |  |
```
overtrue, line 66: This row makes the table total 524, but the footer still says 530. Since this PR is updating the authoritative inventory, please update the footer total in the same diff.

overtrue (commented):
One inventory count needs correction.

PR #6003, opened 2026-08-12
feat(kms): observe unknown fields in the last three silent persisted formats

overtrue:
## Summary

PR-1 of rustfs/backlog#1641: the last three KMS persisted formats that silently ignored unknown fields now count and warn about them, closing their violation of the repository rule that formats too compatibility-bound for `deny_unknown_fields` must at least log unknown fields at warn (AGENTS.md). The `DataKeyEnvelope` and Local `StoredMasterKey` formats already had this machinery (`persisted_observability`); this PR extends the same pattern — bounded field names, power-of-two log sampling, the `rustfs_kms_persisted_unknown_fields_total{record_kind}` counter, never a value — to the remaining three:

- **`VaultKeyData`** (KV2 key record) and **`TransitKeyMetadataPersisted`** (Transit metadata record): hand-written `Deserialize` following the exact `StoredMasterKey`/`DataKeyEnvelope` house pattern. Hand-written rather than wrapped because these records deserialize inside `vaultrs::kv2::read` where the deserializer is not ours to wrap, and the struct-level impl keeps error behavior (missing/duplicate field messages, the record-corruption classification from #5764) byte-identical to the derive. New records kinds: `vault-kv2-key`, `vault-transit-key-metadata`.
- **`KmsConfig`** (persisted `config/kms_config.json`): a new observing loader `kms_config_from_persisted_json` wraps the deserializer with `serde_ignored` and reports every ignored path — including fields nested inside the backend variant, which the externally tagged `BackendConfig` enum exposes to the observer. The admin load site (`decode_persisted_kms_config`) now uses it. The deliberate unknown-field tolerance (rolling upgrades write fields older builds don't know; pinned by an existing regression test) is unchanged — tolerated just no longer means invisible. Record kind: `kms-config`.

`serde_ignored` (dtolnay, Apache-2.0/MIT, no transitive deps beyond serde) is a new workspace dependency; `cargo deny check licenses` passes. It is used only at the config load site — the two Vault records cannot use it for the reason above.

Field values are never logged anywhere: a mistyped field name can sit next to key material or a secret, so only bounded, truncation-marked field names/paths reach the log, matching the #5764 rule that record-level messages must not carry stored scalars.

## Tests

Each format gets a test in the established shape (unknown field → record stays readable, warn emitted naming the field, value string asserted absent from logs, counter asserted via `DebuggingRecorder`): `vault_key_data_unknown_fields_remain_readable_and_are_observed`, `transit_key_metadata_unknown_fields_remain_readable_and_are_observed`, `persisted_config_unknown_fields_remain_readable_and_are_observed` (top-level + nested-in-variant paths, plus a clean-config-observes-nothing half).

## Verification

- `cargo test -p rustfs-kms` offline and with a live dev Vault (`hashicorp/vault:1.17.6`, KV2 + Transit): 16/16 suites green, including the dev-Vault `#[ignore]` tests exercising real KV2/Transit record round-trips through the new deserializers (`backends::vault -- --ignored` 10/11 — the one failure is `test_vault_cancel_key_deletion_persists_state`, broken on main independently of this PR and fixed in #5999).
- `cargo clippy -p rustfs-kms --all-targets -- -D warnings`, `cargo check -p rustfs`, `cargo test -p rustfs --lib admin::handlers::kms_dynamic`, `cargo deny check licenses`, `cargo fmt --all`, `make pre-commit`.

Not in scope (documented in rustfs/backlog#1641): `VaultKeyVersionRecord` (immutable, create-only, never rewritten by newer builds against older readers in place), the `.master-key.salt` header question, and the object-metadata key set (MinIO compatibility constraints).

Refs rustfs/backlog#1641 (PR-1), rustfs/backlog#1562.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9737,6 +9737,7 @@ dependencies = [
  "rustfs-utils",
  "rustify",
  "serde",
+ "serde_ignored",
  "serde_json",
  "sha2 0.11.0",
  "subtle",
@@ -10937,6 +10938,16 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 3.0.3",
+]
+
+[[package]]
+name = "serde_ignored"
+version = "0.1.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "115dffd5f3853e06e746965a20dcbae6ee747ae30b543d91b0e089668bb07798"
+dependencies = [
+ "serde",
+ "serde_core",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -182,6 +182,7 @@ quick-xml = "0.41.0"
 rmp = { version = "0.8.15" }
 rmp-serde = { version = "1.3.1" }
 serde = { version = "1.0.229" }
+serde_ignored = { version = "0.1" }
 serde_json = { version = "1.0.151" }
 serde_urlencoded = "0.7.1"
 
--- a/crates/kms/Cargo.toml
+++ b/crates/kms/Cargo.toml
@@ -35,6 +35,10 @@ tokio = { workspace = true, features = ["fs", "io-util", "macros", "rt-multi-thr
 uuid = { workspace = true, features = ["serde", "v4", "fast-rng", "macro-diagnostics"] }
 jiff = { workspace = true, features = ["serde"] }
 serde = { workspace = true, features = ["derive"] }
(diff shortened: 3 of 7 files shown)


overtrue (commented):
The new observers run on the exact Vault KV2 and persisted-config production readers, preserve tolerant decoding, and keep ignored values out of logs and labels.

PR #6005, opened 2026-08-12
feat(sse): report the wrapping master-key version on S3 audit entries

overtrue:
## Summary

Closes the last unchecked acceptance item of rustfs/backlog#1583 (PR-C): S3 audit entries for SSE-KMS/SSE-S3 requests now carry the KMS key *version* alongside the key id and outcome they already carried.

The `kmsKeyVersion` audit tag existed but was never populated — the single production call site passed `None`, justified by a comment claiming the version "is not observable on the data path". That claim was stale: the stored data-key envelope (`x-rustfs-encryption-key`, `DataKeyEnvelope` JSON) records `master_key_version` for every KV2-wrapped DEK, and on the write path the freshly generated envelope is sitting in `EncryptionMaterial.encrypted_data_key`. This PR extracts it:

- **Write path** (`apply_managed_encryption_material` success): version read from the envelope the request just wrapped.
- **Read path** (`apply_managed_decryption_material` success and failure, and the copy-source `kms:Decrypt` denial in `authorize_sse_kms_object_read`): version read from the stored envelope, through `normalize_managed_metadata` so both the RustFS-branded and MinIO-branded stored keys resolve (the dual internal metadata key rule).
- `None` remains the honest answer everywhere else: opaque Transit/AWS ciphertext is not an envelope, Local records no version, and pre-versioning envelopes never carried the field — nothing is fabricated.

Two deliberate implementation choices:

- The version is read through a `serde_json::Value` probe guarded by `is_data_key_envelope`, not a full `DataKeyEnvelope` parse: the audit path must not double-count the envelope's unknown-field observability (rustfs/rustfs#6003) and needs exactly one field.
- Extraction is lazy (`impl FnOnce() -> Option<u32>`): internal callers — replication, lifecycle, heal — have no S3 audit entry to attach to, and their hot paths must not pay a base64 decode + JSON parse for a value that is discarded.

## Tests

- `envelope_master_key_version_reads_only_true_envelopes`: versioned envelope → `Some`, pre-versioning envelope → `None`, opaque Transit-style ciphertext → `None`, non-envelope JSON carrying a `master_key_version` field → `None`.
- `stored_envelope_master_key_version_reads_both_metadata_families`: RustFS-branded and MinIO-branded stored keys both resolve; empty metadata → `None`.
- `recorded_key_versions_reach_the_audit_tags`: a recorded version appears as the `kmsKeyVersion` tag with `kmsOutcome=success`.

## Verification

- `cargo test -p rustfs --lib storage::sse` (108 passed)
- `cargo fmt --all`, `make pre-commit`

Refs rustfs/backlog#1583 (final acceptance item), rustfs/backlog#1562 (P1.2).

### DIFF
--- a/rustfs/src/storage/sse.rs
+++ b/rustfs/src/storage/sse.rs
@@ -1052,7 +1052,13 @@ pub async fn authorize_sse_kms_object_read(
     // Only a denial is recorded here: an allowed read goes on to unwrap the key,
     // and that operation reports its own outcome.
     if let Err(error) = &result {
-        record_managed_kms_outcome(principal, sse_type, Some(&key_id), Err(error));
+        record_managed_kms_outcome(
+            principal,
+            sse_type,
+            Some(&key_id),
+            || stored_envelope_master_key_version(metadata),
+            Err(error),
+        );
     }
 
     result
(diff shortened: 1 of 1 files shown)


rustfs/src/storage/sse.rs
```diff
@@ -1290,22 +1296,53 @@ impl std::error::Error for KmsDataPlaneFailure {
...
+/// parse, so the audit path cannot double-count the envelope's unknown-field
+/// observability and touches nothing else in the envelope.
+fn envelope_master_key_version(envelope_bytes: &[u8]) -> Option<u32> {
+    if !is_data_key_envelope(envelope_bytes) {
+        return None;
+    }
+    u32::try_from(
+        serde_json::from_slice::<Value>(envelope_bytes)
+            .ok()?
+            .get("master_key_version")?
```
overtrue, line 1333: Vault Transit deliberately leaves `master_key_version` empty and stores the wrapping version in the nested `encrypted_key` value (`vault:vN:`). The existing scripted Transit test proves this exact production envelope shape, so write, read, and denial audit entries still omit `kmsKeyVersion` for a supported rotating backend. Please dispatch the extraction for Transit and add a production-path audit regression.

overtrue (commented):
Vault Transit still has an observable wrapping version that this extractor misses.

PR #6008, opened 2026-08-12
chore(io-metrics): remove the zero-consumer config module (391 lines)

overtrue:
## What

PR4 of rustfs/backlog#1833. `crates/io-metrics/src/config.rs` (391 lines) was a near-copy of io-core's Backpressure/Deadlock configuration whose field names had already drifted from the canonical owner (`high_watermark` here vs `high_water_mark` in `crates/io-core/src/backpressure.rs`). Verified zero consumers: the only user of the exported `IoConfig`/`*Settings` types anywhere in the repo was the crate's own example.

- Delete `crates/io-metrics/src/config.rs`, the `pub mod config;` declaration, and the `pub use config::{...}` re-export block.
- Drop the example's `unified_config_example` section and its imports.
- Settle the ARCHITECTURE.md invariant-3 ledger line that explicitly tracked this copy's removal under backlog#1833 (the doc-paths guard enforces the file reference, so the line is updated in the same diff).

## Verification

- `rg` sweep: no `rustfs_io_metrics::{IoConfig,*Settings,DEFAULT_*}` consumers outside the crate
- `cargo clippy -p rustfs-io-metrics --all-targets -- -D warnings` — clean (includes the example)
- `cargo test -p rustfs-io-metrics --lib` — 158 passed
- `make pre-commit` — ✅ (doc-paths guard green after the ledger update)

Ref rustfs/backlog#1833.


crates/io-metrics/src/lib.rs
```diff
@@ -173,7 +173,6 @@ pub mod backpressure_metrics;
 pub mod cache_config;
 pub mod capacity_metrics;
 pub mod collector;
-pub mod config;
 pub mod deadlock_metrics;
```
overtrue, line 176: Please remove or rewrite the unified-configuration examples and config.rs module-tree entries in both io-metrics READMEs. README.md is the package readme and still tells users to import the deleted IoConfig and settings types, so the documented example no longer compiles.

overtrue (commented):
The removed configuration API is still presented as supported in the package documentation.

PR #6016, opened 2026-08-12
test(e2e): fold seven identical POST-policy exact-mismatch tests into one table

overtrue:
## What

PR2 of rustfs/backlog#1838, continuing the fold started in #6000. The seven `*_policy_mismatch` tests in `multipart_auth_test.rs` were body-identical after literal normalization: the policy pins one field to an exact value, the form sends a different value, and the upload must be rejected with **400 InvalidPolicyDocument** naming the field. Each booted its own full server.

- Folded into one table-driven test on the existing `run_post_object_policy_case` helper (no helper changes needed — PR1 reserved the policy-conditions parameter for exactly this). The pinned condition is built with an explicit `serde_json::Map` because the field name is now a table parameter.
- **Every row keeps its original test's exact literals** — bucket, key, field, policy value, mismatched form value, body bytes, and expected error strings, including the three rows asserting the stronger `<Code>InvalidPolicyDocument</Code>` form. Case-row count equals folded-test count (7).
- Folded: content-disposition, content-language, content-encoding, website-redirect-location, metadata-uuid-exact, sigv4-algorithm, sigv4-credential. The remaining smaller groups (two sse-kms triples, cache-control/expires/tagging mismatch triple, and four pairs) are PR3.

## Inventory

`docs/testing/e2e-suite-inventory.md` updated in the same diff: 103 → **97**, matching `cargo nextest list`.

## Verification

- `cargo nextest run -p e2e_test -E 'test(test_anonymous_post_object_rejects_exact_condition_policy_mismatches)'` — PASS in 20.8s (seven real server round-trips)
- `cargo nextest list` — 97 tests for this module, matches the inventory row
- `make pre-commit` — ✅

Ref rustfs/backlog#1838.

### DIFF
--- a/crates/e2e_test/src/multipart_auth_test.rs
+++ b/crates/e2e_test/src/multipart_auth_test.rs
@@ -579,6 +579,83 @@ async fn test_anonymous_post_object_rejects_exact_condition_policy_mismatches()
             "<Code>InvalidPolicyDocument</Code>",
             "x-amz-credential",
         ),
+        (
+            "cache-control",
+            "anon-post-policy-cache-control-reject",
+            "uploads/cache-control-reject.txt",
+            "Cache-Control",
+            "max-age=60",
+            "max-age=120",
+            b"post-policy-cache-control-mismatch",
+            "InvalidPolicyDocument",
+            "cache-control",
+        ),
+        (
+            "expires",
+            "anon-post-policy-expires-reject",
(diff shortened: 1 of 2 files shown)


overtrue (commented):
The seven-row fold preserves every original request literal and assertion, including the three stronger XML error-code checks. The helper still runs each case against an isolated real server.

PR #6017, opened 2026-08-12
refactor(sse): sink managed-SSE attribution into the shared encryption-keys module

overtrue:
## Summary

PR-B0 of rustfs/backlog#1643 (key reference inventory): the managed-SSE attribution logic moves out of the `rustfs` binary crate into the shared `rustfs_utils::http::object_encryption_keys` module, so `crates/scanner` — which must never depend on the binary crate — can attribute encrypted objects in PR-B without growing a second classifier (the divergence rustfs/backlog#1436 already paid for once). Pure behavior-preserving refactor; no scanner changes, no dependency changes.

**Stacked on #5995** (the constants consolidation this builds on). Merge that first; GitHub will retarget this to main automatically.

- `stored_managed_encryption_key`, `contains_managed_encryption_metadata`, `normalize_managed_metadata` and the `SSEType` enum move to the shared module; `storage/sse.rs` re-exports `SSEType` (and the predicate crate-internally) so every existing path keeps compiling.
- The one piece that cannot move verbatim: the MinIO KMS-context recode (base64-wrapped JSON → plain JSON) needs `base64`/`serde_json`, which `rustfs-utils` does not have and does not gain. The shared normalizer takes an injected `Option<fn(&str) -> Option<String>>` recoder; `sse.rs` supplies the old inline chain verbatim (`recode_minio_kms_context`), including the historical silent skip on decode failure. `stored_managed_encryption_key` passes `None` — provably identical output, since the context mapping only inserts the context key, which the key-id lookup never reads.
- Every metadata lookup keeps its case-SENSITIVE exact match; a new test pins that a lowercased MinIO twin does not normalize (the `header_compat` lowercase-model trap documented in rustfs/backlog#1775).
- New source-scan guard test `managed_sse_classifier_has_exactly_one_definition` (modeled on the `kms_key_lifecycle` scan) asserts `sse.rs` defines none of the three function bodies and the shared module defines each exactly once — the acceptance criterion of rustfs/backlog#1643 PR-B0.
- 5 new unit tests in the shared module pin the predicate, the MinIO→RustFS mapping, key-id precedence (rustfs > minio > amz > "default"), the case-sensitivity trap, and audit labels.

## Verification

- `cargo test -p rustfs --lib storage::sse` (106 passed, existing assertions untouched)
- `cargo test -p rustfs-utils --lib` and `--features http` (all 9 object_encryption_keys tests green)
- `cargo clippy -p rustfs --all-targets -- -D warnings` (pass); `cargo clippy -p rustfs-utils --all-targets` fails only on the pre-existing `hash_hotpath_benchmark` feature gap, verified present on the base commit (CI's workspace clippy is unaffected via feature unification); `--features full` passes
- `bash scripts/check_layer_dependencies.sh` (pass), `cargo fmt --all`

Merge-order note: #6005 adds a 1-arg `normalize_managed_metadata` call in `stored_envelope_master_key_version`; whichever of the two merges second resolves the conflict by passing `None` as the recoder there (the version lookup never reads the context key).

Refs rustfs/backlog#1643 (PR-B0), rustfs/backlog#1562.


overtrue (commented):
The helper move preserves the managed-metadata predicate, key precedence, and MinIO context conversion across every production caller. The code is ready, but this branch is currently conflicting and cannot be merged until its stack is refreshed.

PR #6018, opened 2026-08-12
test(e2e): fold the remaining POST-policy duplicate groups (15 tests)

overtrue:
## What

PR3 of rustfs/backlog#1838, completing the multipart_auth fold. **Stacked on #6016 (PR2)** — base is set to that branch and GitHub will retarget to main when it merges.

The six remaining body-identical groups collapse onto the shared `run_post_object_policy_case` helper (still unchanged since #6000):

- **+7 rows** into PR2's exact-condition mismatch table: cache-control, expires, tagging, storage-class, content-type, success_action_status, metadata-field-exact — all the same policy-pins-one-field / form-mismatches shape.
- **object-lock table (2 rows)**: policy pins mode + retain-until-date; one form field mismatches; 400 InvalidPolicyDocument naming that field.
- **SSE-KMS mismatch table (3 rows)**: policy pins the SSE mode plus one KMS parameter (key-id / context / bucket-key-enabled); form differs; 400 with the strong `<Code>InvalidPolicyDocument</Code>` form as in the originals.
- **SSE-KMS outside-policy table (3 rows)**: pins the distinct contract that an undeclared KMS parameter does **not** fail policy validation — the request is rejected at runtime with **501 `<Code>NotImplemented</Code>`** (SSE-KMS POST uploads unimplemented), preserving the originals' "policy validation must not reject first" semantics.

Every row keeps its original test's exact literals (bucket, key, field names, policy/form values, body bytes, expected status/code/mention). 15 tests fold into 3 new tables + 7 rows; module count 97 → **85**, matching `cargo nextest list`, inventory updated in the same diff.

## Verification

- `cargo nextest run` on the four touched table tests — 4 passed (22 real server round-trips, 19.6s wall)
- `cargo nextest list` — 85, matches inventory
- `make pre-commit` — ✅

Ref rustfs/backlog#1838.

### DIFF
--- a/crates/e2e_test/src/multipart_auth_test.rs
+++ b/crates/e2e_test/src/multipart_auth_test.rs
@@ -579,6 +579,83 @@ async fn test_anonymous_post_object_rejects_exact_condition_policy_mismatches()
             "<Code>InvalidPolicyDocument</Code>",
             "x-amz-credential",
         ),
+        (
+            "cache-control",
+            "anon-post-policy-cache-control-reject",
+            "uploads/cache-control-reject.txt",
+            "Cache-Control",
+            "max-age=60",
+            "max-age=120",
+            b"post-policy-cache-control-mismatch",
+            "InvalidPolicyDocument",
+            "cache-control",
+        ),
+        (
+            "expires",
+            "anon-post-policy-expires-reject",
+            "uploads/expires-reject-object.txt",
+            "Expires",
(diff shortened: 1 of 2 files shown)


overtrue (commented):
Checked the 15 removed cases against the four tables on this exact stacked head. The literals and policy/runtime outcomes are preserved, and the tests still exercise the pinned production POST-policy path.

PR #6019, opened 2026-08-12
feat(kms): reserve and expose KV2 wrap-budget consumption

overtrue:
## Summary

PR-2 of rustfs/backlog#1636: an approximate, reservation-based counter of KEK wrap operations for the Vault KV2 backend — the only backend where RustFS both performs the AES-256-GCM wrap locally AND can rotate to reset the NIST SP 800-38D 2^32 random-nonce ceiling. Local/Static get documentation only (landed in #5992); Transit/AWS wrap externally and are not applicable.

**Mechanism.** `VaultKeyData` gains `#[serde(default)] wrap_budget_reserved: u64`. A node reserves budget in blocks of 1,000,000 via the existing CAS helper on the key record and consumes from an in-process per-key grant; a crash discards unused in-memory budget, so the persisted value only ever overestimates actual wraps — the safe direction for a ceiling. Exactly three wrap choke points consume the budget (`generate_data_key`, `encrypt`, the destination re-wrap in `rewrap_data_key`); decrypt paths are untouched, and nothing is ever written to Vault per wrap.

**Failure and rotation semantics.**

- A failed reservation never blocks the data path: the wrap proceeds against an unpersisted grant carried as debt, the next successful reservation persists `block + debt`, and the structured warn fires at most once per million wraps by construction.
- The in-memory grant is version-tagged (a correctness requirement found during implementation, within the design's latitude): without it, a node's leftover grant from before a rotation would be spent against fresh material whose persisted counter was just reset — an undercount. Stale in-flight wraps racing a rotation count against the newer grant, keeping the overestimate.
- `rotate_key`'s CAS commit resets the counter to 0 — fresh material, fresh nonce budget — without touching any rotation invariant (the full rotation regression suite passes, including live).
- Mixed-version caveat documented: an old build rewriting the record drops the field (the documented old-build behavior), the only underestimate risk; registered in the mixed-version section of `docs/operations/kms-backend-security.md`.

**Exposure.** Aggregate gauge `rustfs_kms_max_key_wrap_operations` (max across usable keys, published by the deletion worker's census under its existing whole-key-set republish discipline; no per-key labels per the #1584 rule). `describe_key` carries the per-key value internally; it stays off the admin wire (`#[serde(skip)]`, pinned by the existing insta snapshot) because exposing it there is a contract decision out of scope here.

## Tests

New scripted tests pin: block reservation shape (one CAS write of B for N<B wraps; persisted ≥ actual after a simulated crash), concurrent reservations through the CAS-conflict path ending at exactly 2B, reservation failure not failing the wrap (5xx scripted; warn captured; no key material in the log), rotation reset (0 on the post-rotation record), post-rotation wraps reserving a fresh block (the version-tag pin), and the gauge's label hygiene. Existing scripted request-sequence tests were extended for the new reservation traffic, which pins the reservation wire shape itself.

## Verification

- `cargo test -p rustfs-kms`: all 16 targets green offline, and **all 16 green again with a live `hashicorp/vault:1.17.6` dev server** (KV2 + Transit lane on).
- Live ignored rotation suite `backends::vault -- --ignored`: 10/11 — the one failure is `test_vault_cancel_key_deletion_persists_state`, broken by construction on main independently of this PR and fixed in #5999.
- `cargo clippy -p rustfs-kms --all-targets -- -D warnings`, `bash scripts/check_logging_guardrails.sh`, `cargo fmt --all`, `make pre-commit`, `cargo check -p rustfs --all-targets`.

Refs rustfs/backlog#1636 (PR-2), rustfs/backlog#1562.


overtrue (commented):
The reservation accounting has reachable undercount paths.

crates/kms/src/backends/vault.rs
```diff
@@ -282,9 +333,78 @@ impl VaultKmsClient {
...
+                    Ok(CasMutation::Write(()))
+                })
+                .await;
+            match reserved {
+                Ok(_) => {
+                    budget.available = WRAP_BUDGET_BLOCK;
+                    budget.unpersisted = 0;
+                }
+                Err(error) => {
+                    budget.available = WRAP_BUDGET_BLOCK;
```
overtrue, line 399: Granting this block after the CAS write fails loses the count across a restart. The exact-head failure test proves the wrap still succeeds; if the node exits before a later reservation, the in-memory debt disappears while Vault stays unchanged, so a real wrap can leave the durable gauge at 0. This is also reachable when data/metadata reads are allowed but the update is denied, or on a transient 5xx. Please persist before granting (or make the debt durable) and add a failure-plus-restart regression.

crates/kms/src/backends/vault.rs
```diff
@@ -282,9 +333,78 @@ impl VaultKmsClient {
...
+        let budget = {
+            let mut budgets = self.wrap_budgets.lock().expect("wrap budget map lock poisoned");
+            match budgets.get(key_id) {
+                // Fast path spares the per-wrap key allocation `entry` needs.
+                Some(budget) => Arc::clone(budget),
+                None => Arc::clone(budgets.entry(key_id.to_string()).or_default()),
+            }
+        };
+        let mut budget = budget.lock().await;
+        if key_version > budget.version {
```
overtrue, line 373: Version alone cannot identify a grant after key deletion and recreation. A key can be wrapped under v1, physically deleted, then recreated under the same id at v1 with fresh material; the retained map entry still has the old grant and this branch spends it without reserving against the new record. Please clear the entry on physical deletion or bind it to an incarnation that changes on recreation, with a delete/recreate regression.

crates/kms/src/deletion_worker.rs
```diff
@@ -154,6 +172,11 @@ fn record_sweep(report: &SweepReport, census: Option<KeyCensus>) {
     metrics::gauge!(METRIC_PENDING_DELETION_KEYS).set(census.pending_deletion as f64);
     metrics::gauge!(METRIC_TOMBSTONE_KEYS).set(census.tombstones as f64);
     metrics::gauge!(METRIC_OLDEST_ROTATION_AGE_SECONDS).set(census.oldest_rotation_age_seconds);
+    // Only emitted when a usable key reported a count: backends that do not
+    // count wraps must not publish a `0` that reads as "no wraps consumed".
+    if let Some(max_wrap_operations) = census.max_wrap_operations {
```
overtrue, line 177: After a complete sweep previously publishes (for example) 3M, a later complete sweep with no usable counted key skips this write and the cumulative last-value exporter keeps reporting 3M. That leaves a deleted or departing key pinning the gauge, contrary to the new runbook. Please explicitly clear the previously published value when the counted set becomes empty and cover two consecutive sweeps under one recorder.

overtrue (commented):
The aggregate gauge also needs an explicit empty-key-set update.

PR #6020, opened 2026-08-12
refactor(admin): add authorize_admin_request and fold four local wrappers

overtrue:
## Related Issues

rustfs/backlog#1829 (T1)

## Summary of Changes

Adds one shared admin gate, `authorize_admin_request(req, actions) -> S3Result<Credentials>`, to `rustfs/src/admin/auth.rs`, composed strictly from the existing `authenticate_request` + `validate_admin_request` (no auth logic is reimplemented), and folds the four byte-near-identical per-handler wrappers onto it: `validate_batch_job_admin_request` (batch_job.rs), `validate_replication_admin_request` (replication.rs), `validate_site_replication_admin_request` (site_replication.rs), and `validate_config_admin_request` (config_admin.rs).

Semantic preservation was treated as the acceptance bar: `config_admin.rs` historically rejects credential-less requests with "missing credentials" while the shared gate reports "get cred failed" — a pre-check keeps that message byte-identical, pinned by a new test (`config_admin_request_without_credentials_keeps_historical_message`). A second new test pins the shared gate's credential-less rejection (`authorize_admin_request_without_credentials_is_rejected`, exact code + message). `deny_only` stays `false` for all folded call sites, matching their previous behavior; the ~100 remaining inline preambles and the deny-aware `user.rs` sites are follow-up tasks (T2-T5) in rustfs/backlog#1829.

Adversarial review (standard tier + security role): correctness/security compared the pre- and post-fold request paths for all four endpoints — same credential extraction, same session-token handling, same RemoteAddr propagation, same action lists, byte-identical error responses; simplicity confirmed the helper only composes existing functions; test-coverage skeptic added the two pinning tests above. No source-grep (`include_str!`) test matches the replaced wrapper patterns (checked per rustfs/backlog#1839).

## Verification

- `cargo nextest run -p rustfs -E 'test(/admin/)'` → 1295 passed
- `cargo fmt --all --check`
- Compilation coverage via the nextest build of the rustfs test targets

## Impact

No behavior change intended for any endpoint; error codes and messages are pinned by tests. This is the foundation PR for the mechanical sweeps of the remaining 100+ inline auth preambles (rustfs/backlog#1829 T2-T4).

## Additional Notes

Part of the pre-1.0 cleanup program tracked in rustfs/backlog#1820.

### DIFF
--- a/rustfs/src/admin/auth.rs
+++ b/rustfs/src/admin/auth.rs
@@ -13,13 +13,14 @@
 // limitations under the License.
 
 use crate::auth::get_condition_values;
+use crate::server::RemoteAddr;
 use http::HeaderMap;
 use http::Uri;
 use rustfs_credentials::Credentials;
 use rustfs_iam::store::Store;
 use rustfs_iam::sys::IamSys;
 use rustfs_policy::policy::{Args, action::Action};
-use s3s::{S3Result, s3_error};
+use s3s::{Body, S3Request, S3Result, s3_error};
 use std::sync::Arc;
 use tracing::debug;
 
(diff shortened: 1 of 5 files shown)


rustfs/src/admin/auth.rs
```diff
@@ -292,6 +293,25 @@ pub async fn authenticate_request(
...
+/// `actions` ([`validate_admin_request`], allowing on the first permitted
+/// action). Returns the authenticated credentials for handlers that need the
+/// caller identity. `deny_only` stays `false`: every caller performs a full
+/// allow check.
+pub async fn authorize_admin_request(req: &S3Request<Body>, actions: Vec<Action>) -> S3Result<Credentials> {
+    let Some(input_cred) = req.credentials.as_ref() else {
+        return Err(s3_error!(InvalidRequest, "get cred failed"));
+    };
+
+    let (cred, owner) = authenticate_request(&req.headers, &req.uri, input_cred).await?;
```
overtrue, line 307: These wrappers previously called `check_key_valid` directly, which masks rejected access keys. Routing them through `authenticate_request` now adds a duplicate WARN with the full attacker-controlled access key for every invalid, disabled, or expired admin credential. Please mask or remove that shared log before expanding this gate.

overtrue (commented):
The shared gate changes authentication-failure logging on the migrated routes.

PR #6021, opened 2026-08-12
test(io-metrics): assert metric emission in six modules of record_* smoke tests

overtrue:
## What

PR3 of rustfs/backlog#1836, in two parts:

**1. The census tool** — `scripts/find_assertless_tests.py`: queues `#[test]`/`#[tokio::test]` functions whose bodies carry no verification signal (no assert/panic/expect/unwrap/`?`/snapshot/proptest), excluding the issue's known false-positive classes: `#[test_case]`-driven functions and suite-delegation bodies (calls into `*assert*`/`*verify*`/`*check*`/`run_case`-style helpers). It is a heuristic review queue, not a lint — exit code is always 0. Current queue: **114 repo-wide**.

**2. First cleanup wave: io-metrics' six per-module smoke clusters (34 tests).** Every `record_*` smoke test in adaptive_ttl, backpressure_metrics, deadlock_metrics, io_metrics, lock_metrics, and timeout_metrics called its helper and asserted nothing — with no recorder installed the calls are no-ops, so a typo'd or silently-dropped metric name could never fail them. Each cluster is replaced by one recorder-backed test following the crate's existing `DebuggingRecorder` idiom (`s3_api_metrics.rs`, `lib.rs`): the same calls with the same literals (both branches kept where the original exercised them, e.g. TTL extension+reduction, spin success+failure) run under `metrics::with_local_recorder`, and **every metric name the helpers own must appear in the snapshot** — 55 pinned metric names across the six modules.

Not touched in this wave, per the issue's wave plan: the `lib.rs` cluster (~30), `list_objects_metrics`' deliberate edge probes, the sampler smoke pair, and one internode smoke that documents itself as such (io-metrics queue now 36, was 70).

## Flake note

One unattributed test failure appeared on the very first `cargo test -p rustfs-io-metrics --lib` run after the change (134/135) and never reproduced: 30 subsequent in-process runs and 3 nextest runs (CI's runner, per-process isolation) are all green. The new tests use thread-local recorders and touch no global state.

## Verification

- `cargo nextest run -p rustfs-io-metrics` — 135 passed (×3)
- `cargo test -p rustfs-io-metrics --lib` — 135 passed (×30)
- `cargo clippy -p rustfs-io-metrics --all-targets -- -D warnings` — clean
- `make pre-commit` — ✅

Ref rustfs/backlog#1836.

### DIFF
--- a/scripts/find_assertless_tests.py
+++ b/scripts/find_assertless_tests.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python3
+# Copyright 2024 RustFS Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Census of assertion-less tests (rustfs/backlog#1836 PR3).
+
+Flags `#[test]` / `#[tokio::test]` functions whose bodies contain no
+verification signal: no assert!/assert_eq!/assert_ne!/panic! macro, no
+`.expect(`/`.unwrap(`, no `?` operator, no `#[should_panic]`, and no
+`insta` snapshot / proptest / matches! usage. Such a test is green no
+matter what the code under test does.
+
+This is a heuristic REVIEW QUEUE, not a lint: a hit still needs human
+reading before it is fixed or deleted, because assertions may live in a
+called helper. Known false-positive classes are excluded up front:
+
+- `#[test_case(...)]`-driven functions (the values are the assertion's
+  parameters; the assert lives in the shared body — still scanned, but a
+  body that asserts is not flagged anyway; the exclusion covers wrappers
+  that only delegate to a suite runner).
+- Functions whose body calls a helper with `assert`, `verify`, `check`,
+  `expect`, `run_` or `_case` in its name (suite-delegation pattern).
+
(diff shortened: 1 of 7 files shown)


scripts/find_assertless_tests.py
```diff
@@ -0,0 +1,112 @@
...
+            attrs.append(lines[j])
+            j += 1
+        if j >= len(lines):
+            break
+        m = FN_LINE.match(lines[j])
+        if not m:
+            i = j + 1
+            continue
+        name = m.group(1)
+        if any(TEST_CASE_ATTR.search(a) for a in attrs):
```
overtrue, line 77: `should_panic` is listed as a verification signal, but attributes are excluded only for `test_case` and the body scan never sees it. Running this exact script still queues five valid panic tests (`erasure.rs` lines 1350/1356 and `runtime/instance.rs` lines 618/650/693). Please recognize `should_panic` in the collected attributes and pin it with a scanner fixture.

scripts/find_assertless_tests.py
```diff
@@ -0,0 +1,112 @@
...
+
+import re
+import sys
+from pathlib import Path
+
+VERIFY_SIGNALS = re.compile(
+    r"assert!|assert_eq!|assert_ne!|debug_assert|panic!\(|\.expect\(|\.unwrap\(|"
+    r"unreachable!|matches!\(|insta::|proptest!|\.await\?|\)\?|\?;|should_panic"
+)
+DELEGATION = re.compile(r"\b[a-z0-9_]*(?:assert|verify|check|expect|run_case|_case|harness|round_trip|roundtrip)[a-z0-9_]*\s*\(")
```
overtrue, line 48: This does not cover the documented `run_` delegation class: the exact census queues four `run_test_suite().await` drivers in the KMS, policy, and protocols suites, even though each returns `Err` on a suite failure. Please cover that advertised exclusion and add a scanner fixture.

overtrue (commented):
The census still has two documented false-positive gaps.

Issue #6025, opened 2026-08-12
Transitioned managed-SSE (SSE-KMS/SSE-S3) object GET silently returns corrupt bytes


overtrue:
## Summary

A GET on a managed-SSE object (SSE-KMS via bucket default encryption, Local KMS backend) that has been transitioned to a remote RustFS tier silently returns **wrong bytes**: the response succeeds with the correct Content-Length (the plaintext length), but the body is not the plaintext. No error is surfaced to the client — this is silent data corruption on read-through, strictly worse than a failed request.

## Reproduction

Single-node hot server (Local KMS, `--kms-default-key-id`), second single-node RustFS as a `rustfs`-type remote tier via the admin tier API. Bucket with SSE-KMS default encryption; PUT a 78-byte object (HeadObject confirms `x-amz-server-side-encryption: aws:kms`); Days=1 transition rule; drive the transition with the manual transition-run admin endpoint (`/rustfs/admin/v3/ilm/transition/run`) until HeadObject reports the tier's storage class; then plain GET.

Expected: the 78-byte plaintext. Actual: 78 bytes of garbage (starts `[0, 99, 0, 0, 112, 253, ...]`) — the length matches the plaintext, so the decryption sizing ran, but the content does not.

An executable repro exists as `ilm_transition_on_sse_kms_bucket_under_enforcement_reads_back` in `crates/e2e_test/src/kms/kms_ilm_sse_kms_test.rs` on branch `overtrue/kms-1582-ilm-sse-kms-e2e` (shipped `#[ignore]`d, pinning this issue). It was found while adding the ILM half of rustfs/backlog#1582's e2e coverage.

## Not an authorization regression

`RUSTFS_KMS_ENFORCE_SSE_KEY_POLICY` is irrelevant: the read-back fails byte-identically with the enforcement flag on and off (control run performed). Unencrypted objects transition and read back fine in the existing tiering e2e (`crates/e2e_test/src/reliant/tiering.rs`), so the break is specific to the managed-SSE read-through path after transition.

## Suspects

[...]

PR #6027, opened 2026-08-12
test(kms): pin ILM behavior on SSE-KMS buckets under key-policy enforcement

overtrue:
## Summary

The last unchecked acceptance item of rustfs/backlog#1582 asked for end-to-end proof that replication AND ILM keep working on SSE-KMS buckets under `RUSTFS_KMS_ENFORCE_SSE_KEY_POLICY=true`. Replication has had a fail-closed contract test since #5588; ILM had zero e2e coverage in `crates/e2e_test/src/kms/`. This adds `kms_ilm_sse_kms_test.rs` with both halves — and writing the second half surfaced a real, pre-existing data-corruption bug.

- **`ilm_expiration_on_sse_kms_bucket_under_enforcement` (active)**: Local-KMS server with enforcement on, SSE-KMS bucket default encryption, Days=1 expiration accelerated by `RUSTFS_ILM_DEBUG_DAY_SECS=2` (the `reliant/lifecycle.rs` pattern), asserting the object actually expires. If enforcement ever starts applying to the lifecycle scanner's internal operations on encrypted buckets, this goes red.
- **`ilm_transition_on_sse_kms_bucket_under_enforcement_reads_back` (`#[ignore]`, pinning rustfs/rustfs#6025)**: full transition topology — a second RustFS wired as a `rustfs`-type remote tier (the `reliant/tiering.rs` pattern), a manual transition run driven through the admin endpoint, storage-class observation, then a GET asserting byte-identical plaintext. The final assertion fails on main today: **the GET on a transitioned managed-SSE object silently returns corrupt bytes of the correct plaintext length**. A control run with the enforcement flag off fails identically, so this is a pre-existing tiering/SSE read-through defect, not an authorization regression — filed as #6025 with the analysis. The test ships ignored with the issue reference so the fix can un-ignore it as its regression test.

## Verification

- `cargo build --bin rustfs` before every e2e run (the harness reuses a stale debug binary otherwise).
- `cargo test -p e2e_test kms_ilm -- --test-threads=1`: expiration test green (repeatedly, ~11-22s), transition test red-on-main as described, now ignored.
- `cargo clippy -p e2e_test --all-targets -- -D warnings`, `cargo fmt --all`.

Refs rustfs/backlog#1582 (final acceptance item), rustfs/rustfs#6025, rustfs/backlog#1562.

### DIFF
--- a/crates/e2e_test/src/kms/kms_ilm_sse_kms_test.rs
+++ b/crates/e2e_test/src/kms/kms_ilm_sse_kms_test.rs
@@ -1,0 +1,612 @@
+// Copyright 2024 RustFS Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! ILM on SSE-KMS buckets while per-key SSE authorization is enforced (backlog#1582).
+//!
+//! Per-key KMS authorization (`RUSTFS_KMS_ENFORCE_SSE_KEY_POLICY=true`) scopes the
+//! SSE-KMS data path to the requesting principal's `kms:GenerateDataKey` /
+//! `kms:Decrypt` grants. Internal callers — the lifecycle scanner's expiry deletes
+//! and the tier transition worker's reads — carry no request principal, and
+//! `authorize_sse_kms_key` (rustfs/src/storage/sse.rs) exempts a `None` principal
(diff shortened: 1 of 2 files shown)


crates/e2e_test/src/kms/kms_ilm_sse_kms_test.rs
```diff
@@ -0,0 +1,612 @@
...
+        "local",
+        "--kms-key-dir",
+        key_dir.as_str(),
+        "--kms-default-key-id",
+        SSE_KEY,
+    ];
+
+    let envs = [
+        ("RUSTFS_KMS_ALLOW_INSECURE_DEV_DEFAULTS", "true"),
+        ("RUSTFS_KMS_ENFORCE_SSE_KEY_POLICY", "false"),
```
overtrue, line 100: This helper explicitly disables `RUSTFS_KMS_ENFORCE_SSE_KEY_POLICY`, and the spawned server consumes this exact value. The active expiration test can therefore pass even if the system-principal exemption is broken. Please set it to `true` and rerun the active E2E against a fresh binary.

overtrue (commented):
The active test currently bypasses the policy branch it is meant to pin.

PR #6030, opened 2026-08-12
fix(iam): remove eight dead error variants and make Clone variant-preserving

overtrue:
## What

PR1+PR2 of rustfs/backlog#1831 (same 90-line file, tightly coupled — PR2's Clone rewrite touches the arms PR1 deletes, so they ship as two commits on one PR).

**Commit 1 — eight dead variants removed.** `iam::Error` mirrors `policy::Error` variant-for-variant (31 twins + a 31-arm hand-written `From`), and eight twins had zero construction and zero match sites anywhere in the workspace: `InvalidServiceType`, `ErrCredMalformed`, `CredNotInitialized`, `JWTError`, `NoAccessKey`, `InvalidToken`, `InvalidAccessKey`, `InvalidExpiration` (each re-verified per the issue's takeover instruction; `InvalidToken` hits elsewhere are `KeystoneError`'s unrelated variant). The `From<policy::Error>` match stays exhaustive: orphaned arms route through one grouped binding to `Error::StringError(err.to_string())`, preserving the rendered message — nothing could observe the old discriminants.

**Commit 2 — Clone keeps variant identity.** The hand-written Clone demoted `PolicyError`/`CryptoError` to `StringError` (non-cloneable payloads) — a clone changed the variant. No production clone site exists today (the issue's refuter confirmed: preventive hardening, not a live bug). The two payloads are now `Arc`-wrapped: Clone is a reference bump that keeps the variant, Display strings are unchanged, `#[from]` becomes manual `From` impls. Trade-off, stated: `source()` is no longer forwarded for these two variants (`Arc<E>` doesn't implement `std::error::Error`); nothing in the workspace consumed it. A regression test pins discriminant + rendered message across clone.

## Verification

- `cargo test -p rustfs-iam` — 249 + 1 passed (incl. the new clone-identity regression test)
- `cargo clippy -p rustfs-iam --all-targets -- -D warnings` — clean
- `cargo check -p rustfs` — clean
- `make pre-commit` — ✅

Ref rustfs/backlog#1831.

### DIFF
--- a/crates/iam/src/error.rs
+++ b/crates/iam/src/error.rs
@@ -58,15 +58,6 @@ pub enum Error {
     #[error("not initialized")]
     IamSysNotInitialized,
 
-    #[error("invalid service type: {0}")]
-    InvalidServiceType(String),
-
-    #[error("malformed credential")]
-    ErrCredMalformed,
-
-    #[error("CredNotInitialized")]
-    CredNotInitialized,
-
     #[error("invalid access key length")]
     InvalidAccessKeyLength,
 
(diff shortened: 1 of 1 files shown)


crates/iam/src/error.rs
```diff
@@ -14,19 +14,23 @@
...
+    // not forwarded (Arc<E> does not implement std::error::Error).
+    #[error("{0}")]
+    PolicyError(Arc<PolicyError>),
 
     #[error("{0}")]
     StringError(String),
 
     #[error("crypto: {0}")]
-    CryptoError(#[from] rustfs_crypto::Error),
+    CryptoError(Arc<rustfs_crypto::Error>),
```
overtrue, line 33: The stored crypto error no longer appears in Error::source(): this field lost its former #[from] source marker, while the new field has neither #[source] nor a manual source implementation. Corrupt or non-decryptable persisted IAM data reaches this variant through decrypt_data_with_source, so downstream diagnostics now stop at the IAM wrapper. Arc<T: Error> implements Error on the pinned toolchain, so please preserve the chain with #[source] (or equivalent) and add a focused source/downcast regression.

PR #6043, opened 2026-08-12
fix(kms): tell an empty KV2 prefix from a missing mount on Vault's 404

overtrue:
## What

The Vault KV2 KMS backend treated every HTTP 404 from Vault as "empty listing", so a KMS whose configured `kv_mount` does not exist reported itself healthy at startup and then answered every key listing with "you have no keys". This teaches it Vault's two distinct 404 shapes, on the typed error rather than on rendered text.

Vault answers a LIST of a path that holds nothing with a 404 whose `errors` array is empty — the path was routed and there is nothing under it. A path with no mount behind it answers with the same status but carries `no handler for route ...`. `vaultrs`'s `ClientError::APIError` keeps that array, while its `Display` renders both as a bare `(status code 404)`, which is why `health_check` matching `e.to_string().contains("status code 404")` could not tell them apart and why the check is now made before anything turns the error into a string.

Verified against a real Vault 1.17.6, not inferred:

| request | answer |
| --- | --- |
| LIST an empty prefix in a mounted KV2 engine | `404 {"errors":[]}` |
| LIST after the last key under the prefix is deleted | `404 {"errors":[]}` |
| LIST a path with no mount behind it | `404 {"errors":["no handler for route ..."]}` |
| read a nonexistent secret in a mounted engine | `404 {"errors":[]}` |
| token without `list`, or a bad token | `403` |
| sealed Vault | `503` |

Both KV2 `LIST` call sites take the discriminator. The per-secret read paths deliberately do not: there an empty-`errors` 404 genuinely means "this record is absent", which is what they already report.

`list_key_version_records` matters for a sharper reason than symmetry — `delete_key` purges the version records it returns before removing the key itself, so an unrouted path read as "no versions" would turn the purge into a no-op.

## Behavior change worth a release note

A deployment whose `kv_mount` does not exist previously started with a KMS that could not serve a single key operation; it now fails the health check that gates KMS startup, with a message naming the mount and carrying Vault's own explanation. The two startup paths differ, as they did before this change: `--kms-enable` propagates the failure and aborts server startup, while a persisted KMS configuration logs a warning and continues without KMS.

Two supporting changes keep that from turning a working deployment into a crash loop:

- `KmsConfig::validate` now rejects an empty `kv_mount` (the Transit backend already rejected its own empty mounts), so the misconfiguration is named as a configuration error instead of spending a round-trip to report an unroutable path.
- The Helm chart had no way to set the KV2 mount at all: `vault_mount_path` was emitted only for `vault-transit`, so every chart-installed KV2 deployment silently used the `secret` default, which only a dev-mode Vault has. The chart now maps `vault_mount_path` to `RUSTFS_KMS_VAULT_KV_MOUNT` for the KV2 backend, and only when it is set, so an unset value keeps the previous default.

## Tests

- Scripted regression pairs on both listing paths and the health check: the empty-`errors` 404 still lists empty and still passes the check (a first-ever deployment must start), the message-bearing 404 fails both.
- A 404 whose body is not a Vault error at all — a proxy's own page — fails closed and still names the mount. `vaultrs` only builds an `APIError` from a body it could parse, so that shape arrives as a different variant.
- Two `#[ignore]` live-Vault tests pin both wire shapes against a real Vault. They are picked up by the existing nightly `--lib backends::vault -- --ignored` filter with no workflow change, and were run green against Vault 1.17.6 along with the rest of that suite (13 passed).
- A test double was wrong and this change exposed it: `versions_listing` modelled "no version records" as a message-bearing 404, which real Vault never sends for an empty listing.

Verification: `cargo test -p rustfs-kms` (529 + all integration targets), `cargo clippy -p rustfs-kms --all-targets -- -D warnings`, `cargo fmt --all`, `make pre-commit`, `helm lint` plus rendering the chart for all three cases (transit, KV2 with a mount, KV2 without one).

## Follow-ups filed separately

Found while reviewing this change, all pre-existing and reachable today through any KMS start failure, so none of them are regressions from this PR:

- SSE falls back to the local environment master key when a configured KMS is down while still reporting `aws:kms`.
- `KmsServiceStatus` is a one-way latch: a single failed health check poisons it and nothing restores it, which can wedge a node's readiness endpoint.
- The KMS restore dry-run reports a target whose KV mount does not exist as clean and restorable.
- The nightly lane installs an unpinned Vault, and no document states a supported Vault version range.

`is_empty_vault_list` is named for Vault's LIST contract rather than for KV2 so the pending Transit first-boot fix, which needs exactly this discriminator for an empty transit engine, can reuse it instead of adding a twin.

### DIFF
--- a/helm/rustfs/templates/configmap.yaml
+++ b/helm/rustfs/templates/configmap.yaml
@@ -142,6 +142,11 @@ data:
   RUSTFS_KMS_DEFAULT_KEY_ID: {{ .default_key | quote }}
         {{- if eq .vault_backend "vault-transit" }}
   RUSTFS_KMS_VAULT_MOUNT_PATH: {{ .vault_mount_path | quote }}
+        {{- else if .vault_mount_path }}
+  {{- /* The KV2 backend never calls the Transit engine: its mount is the KV2
+         one, under a different variable. Emitted only when set, so an unset
+         value keeps falling back to the "secret" default. */}}
+  RUSTFS_KMS_VAULT_KV_MOUNT: {{ .vault_mount_path | quote }}
         {{- end }}
       {{- end }}
     {{- end }}
--- a/crates/kms/src/backends/scripted_vault.rs
+++ b/crates/kms/src/backends/scripted_vault.rs
@@ -57,6 +57,17 @@ impl ScriptedResponse {
         }
     }
 
+    /// The 404 Vault answers a LIST of an empty path with: something routed the
+    /// request and found nothing under it, so the `errors` array comes back
+    /// empty. [`ScriptedResponse::error`] cannot stand in — it always fills
+    /// `errors`, which is what marks a 404 as an unrouted path instead.
+    pub(crate) fn empty_list_404() -> Self {
+        Self::Http {
+            status: 404,
+            body: serde_json::json!({ "errors": [] }).to_string(),
+        }
+    }
+
     /// Close the connection after consuming a request without sending an HTTP response.
     pub(crate) fn close() -> Self {
         Self::Close
--- a/crates/kms/src/backends/vault.rs
+++ b/crates/kms/src/backends/vault.rs
@@ -373,6 +373,47 @@ fn is_cas_conflict(error: &ClientError) -> bool {
     )
 }
 
+/// Whether a Vault LIST failed with the 404 that means "the path was routed,
+/// and there is nothing under it".
+///
+/// Vault answers a LIST of a path holding no entries with a 404 whose `errors`
+/// array is empty. A path with no mount behind it answers with the same status
+/// but carries a "no handler for route" message, so the empty `errors` array is
+/// what separates "reachable but empty" from "nothing mounted there".
+/// `ClientError`'s `Display` renders both as a bare "(status code 404)", so the
+/// distinction survives only on the typed error.
+///
+/// This separates a routed path from an unrouted one, not a correct mount from
+/// a wrong one. Two configurations still read as empty: a `kv_mount` pointing at
+/// a KV v1 engine, which routes the KV2 metadata path and finds nothing under
+/// it, and (on OSS Vault) a `namespace` that does not exist. Telling those apart
+/// needs a `sys/mounts` read the KMS token is not required to be allowed to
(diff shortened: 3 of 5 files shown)


overtrue:
Coordination note with #6040, which fixes the same 404 confusion on the Transit side: that PR adds `is_empty_transit_list` in `vault_transit.rs` with a body byte-identical to `is_empty_vault_list` here. The predicate is a property of Vault's LIST 404, not of either engine, which is why this one is named for Vault rather than for KV2.

Whichever lands second should collapse the two rather than leave twins. `vault_transit.rs` already imports `map_key_record_read_error` from `vault.rs`, so the edge exists: make `is_empty_vault_list` `pub(super)` and drop the Transit copy along with its duplicate unit test. It is private here only because this tree has no second caller yet.

Worth keeping separate as they are: the message helper stays KV2-specific (it names `kv_mount`, and Transit's mount is a different config field), and #6040 leaves `metadata_kv_mount` on the old blanket-404 pattern, so a Transit deployment with a wrong metadata mount still starts healthy and reads every rotation-metadata record as absent — that one is untouched by either PR.

helm/rustfs/templates/configmap.yaml
```diff
@@ -142,6 +142,11 @@ data:
   RUSTFS_KMS_DEFAULT_KEY_ID: {{ .default_key | quote }}
         {{- if eq .vault_backend "vault-transit" }}
   RUSTFS_KMS_VAULT_MOUNT_PATH: {{ .vault_mount_path | quote }}
+        {{- else if .vault_mount_path }}
+  {{- /* The KV2 backend never calls the Transit engine: its mount is the KV2
+         one, under a different variable. Emitted only when set, so an unset
+         value keeps falling back to the "secret" default. */}}
+  RUSTFS_KMS_VAULT_KV_MOUNT: {{ .vault_mount_path | quote }}
```
overtrue, line 149: This value never reaches the Vault KV2 client. Server startup parses only `RUSTFS_KMS_VAULT_MOUNT_PATH`, then `build_vault_kms_config` hardcodes `kv_mount` to `secret`, so a chart configured with a non-default KV2 mount still probes and uses `secret`. Please carry this setting through the server startup config and cover that production path.

PR #6044, opened 2026-08-12
fix(storage): restore multipart disk compression and make the legacy decompressor resumable

### LINKED ISSUE #5957: RUSTFS_COMPRESS_ENABLE only compresses HTTP responses; objects are stored uncompressed (1.0.0-rc.1)
**Describe the bug**

Setting `RUSTFS_COMPRESS_ENABLE=true` (and/or `RUSTFS_COMPRESSION_ENABLED=true`) does not
compress object data at rest. Debug logs show these variables are consumed by the HTTP
**response** compression middleware (`rustfs::server::compress`, `rustfs/src/server/compress.rs`)
rather than by the storage layer, so on-disk usage grows by the full object size.

This looks like the same underlying problem as #1419, which was closed via PR #1534 but never
confirmed by the reporter (labels `S-confirming`, `inactive`). Reproduced here on 1.0.0-rc.1.

**To Reproduce**

1. Configure `/etc/default/rustfs` (SNSD, single XFS volume):

`RUSTFS_VOLUMES="/mnt/rustfs/data"`
`RUSTFS_COMPRESS_ENABLE=true`
`RUSTFS_COMPRESSION_ENABLED=true`
`RUSTFS_COMPRESS_EXTENSIONS=".txt,.csv,.json,.xml,.log,.md,.html,.js,.css"`
`RUSTFS_COMPRESS_MIME_TYPES="text/*,application/json,application/xml"`
`RUSTFS_OBS_LOGGER_LEVEL=debug`

and restart the service.
2. Create a highly compressible 100 MB file:
`yes "repeated line for compression test" | head -c 100M > comp-test.log`
3. Record baseline disk usage: `du -sb /mnt/rustfs/data`
4. Upload to a bucket with versioning enabled and SSE disabled:
`aws s3 cp comp-test.log s3://test01/comp-test.log --content-type text/plain`
5. Record disk usage again: `du -sb /mnt/rustfs/data`
6. Observed growth: **104,862,423 bytes** — the full uncompressed size.

**Expected behavior**

The object is stored compressed and on-disk usage grows by far less than the original size,
matching the behavior of the `MINIO_COMPRESS_*` variables these are documented as mapping from
(see "Legacy compatibility" in the environment-variables reference, which lists `COMPRESS_ENABLE`,
`COMPRESS_EXTENSIONS`, `COMPRESS_MIME_TYPES` among the mapped `MINIO_*` names, where they control
object compression at rest).

**Screenshots**

Not applicable. Relevant debug log lines instead — note the word **"response"** in both messages.

During each multipart PUT carrying the object data:

`"message":"Skipping compression: response does not match any configured extension or MIME type",`
`"target":"rustfs::server::compress",`
`"filename":"rustfs/src/server/compress.rs","line_number":390,`
`"method":"PUT","uri":"/test01/comp-test.log?uploadId=...&partNumber=1"`

On multipart completion, the middleware *does* compress — but only the 1884-byte XML API response,
not the stored object:

`"message":"Compressing response: Content-Type 'application/xml' matches configured MIME pattern",`
`"target":"rustfs::server::compress",`
`"filename":"rustfs/src/server/compress.rs","line_number":368,`
`"content_length":"1884","method":"POST","uri":"/test01/comp-test.log?uploadId=..."`

 - OS: Ubuntu 26.04 LTS (kernel 7.0.0-29-generic, glibc 2.43)
 - RustFS version: 1.0.0-rc.1 (`rustfs-linux-x86_64-gnu-v1.0.0-rc.1.zip`)
 - Deployment mode: SNSD, single volume, XFS (`isize=512`, `ftype=1`, `rmapbt=1`, `reflink=1`), mounted `noatime`
 - Client: aws-cli 2.36.20, path-style addressing, TLS enabled
 - Bucket: versioning enabled, SSE-S3 disabled

Not applicable — server-side issue.

**Additional context**

The binary also contains storage-layer compression symbols that appear unrelated to the HTTP
middleware, e.g. `STANDARD_EXCLUDE_COMPRESS_EXTENSIONS` and
`RUSTFS_ADDED_EXCLUDE_COMPRESS_EXTENSIONS` (referenced from `ecstore`). These names do not appear
in the environment-variable reference.

Questions:

1. Are `RUSTFS_COMPRESS_*` intended to control HTTP response compression only?
2. If so, which variables enable the storage-level compression implemented in `ecstore`?
3. If the two subsystems are distinct by design, could the documentation state this explicitly?
   Users migrating from MinIO will read `COMPRESS_ENABLE` as object compression at rest, which is
   what led to the confusion in #1419 and here.

Happy to run further tests or provide full debug logs.

overtrue:
## Summary

Multipart uploads have bypassed disk compression entirely since #5169 removed the session compression marker: with `RUSTFS_COMPRESSION_ENABLED=true`, any object uploaded via multipart (which `aws s3 cp` uses by default above 8 MiB) was stored at full size (#5957). #5169 was a correct stopgap for real GET failures (`ERROR_frameType_unknown`, unexpected EOF), but the actual root cause was never in the multipart layout — it was a resumability bug in the legacy `DecompressReader`. This PR fixes that root cause and restores multipart disk compression.

Root cause: `DecompressReader::poll_read` was not `Poll::Pending`-safe. After parsing a block header it reset `compressed_read = 0` on every re-entry, so when the inner reader (a duplex fed by the erasure writer — routinely `Pending` under load) suspended mid-payload, the resumed poll discarded the bytes already consumed and refilled the payload from the wrong stream position. The desynchronized framing then surfaced as `LZ4 error: ERROR_frameType_unknown` mid-GET — exactly the failure #5169 observed under compression-enabled multipart pressure, and reproduced deterministically by the new high-ratio-payload e2e test before the fix. The old header loop additionally returned `Poll::Pending` after a successful partial header read without a registered waker (a latent hang), and decoded stale buffer contents when the stream ended mid-block; the rewritten state machine removes both.

Multipart parts were never the problem: each part is an independent self-delimiting block stream with no end marker, so the concatenated stream decodes across part boundaries by construction, and the per-part index machinery (`get_compressed_offsets`) already handles range and partNumber seeks. This is verified end to end by the new unit and e2e coverage, including objects written in the pre-#5169 on-disk shape — legacy compressed multipart objects (from ≤ beta.11 or mixed-version windows) become readable again with no migration.

## Changes

- `crates/rio/src/compress_reader.rs`: rewrite `DecompressReader::poll_read` as a resumable state machine (`compressed_len > 0` marks an in-flight payload; the header phase can no longer re-run and reset consumption state). Truncated headers/payloads now fail with `UnexpectedEof` instead of decoding garbage or stale buffer tails; unknown block types and zero-length blocks are rejected at the header phase; blocks that decode to zero bytes are rejected (they would surface as a silent EOF); stream errors are sticky (`poisoned`) so a retrying caller cannot convert a truncation into a clean short body.
- `rustfs/src/app/multipart_usecase.rs`: `create_multipart_upload` advertises the session compression marker again, gated by `should_advertise_session_compression`: the staged-rollout switch `RUSTFS_COMPRESSION_MULTIPART_ENABLED` (default off) must be set in addition to the regular compression gates, and SSE-C ciphertext-passthrough replication sessions never compress. The default-off switch means a rolling upgrade cannot create new compressed multipart objects while pre-fix nodes may still serve reads; operators enable it once the fleet has converged, and the `multipart-compression-default-off-window` entry in docs/architecture/compat-cleanup-register.md pins the condition for flipping the default. This re-activates the existing, unchanged compression branches in `upload_part` / `upload_part_copy`.
- `rustfs/src/storage/s3_api/multipart.rs`: ListParts now reports the logical (uploaded) part size, matching AWS semantics, the existing GetObjectAttributes ObjectParts logic, and the server's own min-part-size validation. Note this is also a wire-visible correction for SSE multipart uploads (previously ciphertext size).

## Compatibility

- On-disk format: zero drift. `CompressReader` is untouched; the decoder rewrite is byte-identical on well-formed streams (verified hunk by hunk and by the fixture matrix under both compression backends).
- Rolling upgrade: multipart compression is dark by default (`RUSTFS_COMPRESSION_MULTIPART_ENABLED=false`), so an upgrade window creates no new compressed multipart objects. Pre-fix nodes still carry the `Poll::Pending` decoder bug for already-existing compressed objects (single-PUT, and multipart from ≤ beta.11); those reads are transient retryable errors, never damage, and succeed once the node restarts on the fixed build.
- Mixed-feature builds: a cluster must not mix `rio-v2` and default-feature binaries — multipart now allows different nodes to compress different parts of one object, and the two backends write different stream formats. The release binary ships one backend, so this is a build-discipline invariant, not a user-facing setting.
- ETags are unchanged: part ETags remain plaintext MD5s and the completed object keeps the standard `md5-of-md5s-N` shape.

## Testing

- `crates/rio` unit tests pin the root cause: a `PendingChunkReader` that suspends between every poll and caps read sizes drives the decoder across header/payload boundaries (byte-exact for chunk sizes 1..8192, single and concatenated streams); truncation, sticky-error, unknown-type, zero-length and empty-block cases are covered. The old implementation fails these tests.
- `crates/ecstore` readers unit tests cover compressed multipart full GET, boundary-crossing ranges, deep seeks with and without part indexes, partNumber GETs, suffix ranges, and the SSE-C + compression composition — all byte-exact, running under both compression backends in CI.
- e2e (`compression_test`): multipart roundtrip with physical-size assertions, high-ratio binary payload (the deterministic reproducer for the root cause), UploadPartCopy from a compressed source, three-part partNumber matrix, and SSE-S3 + compression roundtrip. The module joins the `e2e-smoke` PR lane.
- Four-node e2e: `four_node_multipart_ignores_disk_compression_fallback` is rewritten as `four_node_multipart_disk_compression_roundtrip` (reverting the restore fails it); the two other compression-enabled cluster tests now expect the `compressed` reader class, and the ILM-transition variant switches its multipart key to an excluded extension because the warm-tier read path does not decode compression (pre-existing issue, tracked separately).
- Full local battery green: rio 134 tests, ecstore readers 51, rebalance 257, rustfs multipart 56, e2e compression suite, plus `cargo clippy --all-targets` on the touched crates and the rio-v2 feature lane.

## Known adjacent issues intentionally not addressed here

- Compressed objects that ILM-transition to a warm tier are served without decoding by the tier read path (`new_getobjectreader` hardcodes `is_compressed = false`); this predates this PR and affects single-PUT compression equally. Under separate investigation.
- Under `rio-v2` (non-default), a range GET on a large compressed object schedules a background drain of the remaining stream (`RangedDecompressReader::new_draining`); restoring multipart compression widens the population of large compressed objects. Worth a byte cap before rio-v2 becomes the default.
- The rebalance migration read used to decode compressed and encrypted objects while `migrate_object` rewrote the stored representation verbatim, corrupting the migrated copy. That forward fix landed separately on main in #6057 and is present on this branch's base, so it is **not** part of this diff. Restoring multipart compression widens the population of objects that would have been exposed, which is why it is noted here. Auditing objects already damaged by the pre-fix path is tracked in backlog #1850.

Fixes #5957. Internal tracking: backlog #1848.


rustfs/src/app/multipart_usecase.rs
```diff
@@ -842,8 +858,13 @@ impl DefaultMultipartUsecase {
             None => (None, None),
         };
 
-        // Multipart parts are independent physical streams. Advertising object-level
-        // compression here would make GET decode the completed object as one stream.
+        if should_advertise_session_compression(ciphertext_passthrough, is_disk_compressible(&req.headers, &key)) {
```
overtrue: This re-enables multipart compression as soon as any upgraded node coordinates the upload, but an old node in the supported rolling-upgrade window still has the non-resumable decoder and deterministically fails a mid-payload suspension with LZ4 ERROR_frameType_unknown. Please gate this marker on fleet capability/convergence, or establish the same explicit no-mixed-version upgrade exception used for incompatible writes.

overtrue: Gating the marker on fleet capability would not close the window it aims to close, because the failing component is the reader, not the writer. A not-yet-upgraded node in the same window hits the identical non-resumable decoder on every already-existing compressed object — single-PUT objects compressed since beta, and any multipart objects from ≤ beta.11 — so suppressing the multipart marker only shrinks the exposed object set while the node itself remains the hazard. The only true fix for the window is the one rolling upgrades already provide: the node picks up the resumable decoder when it restarts on the new build.

Risk class also differs from the incompatible-write case: there is zero on-disk format drift here, the old decoder CAN read these objects (the failure needs a mid-payload suspension, i.e. it is load-dependent and transient), and a failed GET is a retryable 5xx against a self-consistent object — never corruption or a permanently unreadable write. Compression is additionally an explicit opt-in (RUSTFS_COMPRESSION_ENABLED defaults to off), so operators who need a zero-surprise window can simply enable it after the rollout completes, which the PR description calls out as the operational guidance.

If you want this pinned as an explicit exception rather than prose, I can add a behavioral entry to docs/architecture/compat-cleanup-register.md (e.g. multipart-compression-rolling-read-window: transient compressed-object read errors on pre-fix peers during the upgrade window; retires once the minimum supported direct-upgrade release ships the resumable decoder). A general fleet-capability negotiation facility would help more than this one feature and deserves its own issue rather than blocking a data-path fix. Happy to go either way — say the word and I will push the register entry.

overtrue: The existing single-PUT exposure does not make the new multipart exposure safe: this head newly creates more objects that an old serving node can fail to read, while the rolling-restart contract promises continued traffic and compression-enabled clusters are supported. A register entry would document the regression but not prevent it, so the blocker remains.

overtrue: Accepted — prevention it is. Pushed 4be344f11: the session marker is now additionally gated on a staged-rollout switch, RUSTFS_COMPRESSION_MULTIPART_ENABLED, defaulting to off. A rolling upgrade therefore creates no new compressed multipart objects while pre-fix nodes can still serve reads; operators flip the switch once the fleet has converged on a fixed build. This follows the repo's existing convergence-switch precedent (msgpack_only) and is what the tracking issue's open question 3 anticipated for this restoration. The retirement path is pinned as multipart-compression-default-off-window in docs/architecture/compat-cleanup-register.md: the default flips (and the extra switch retires) once the minimum supported direct-upgrade release ships the resumable decoder. The guard's truth table covers the new gate, and all compression e2e suites (single-node and four-node) run with the switch enabled.

overtrue:
CI note: the round-3 failures (E0063 missing field wrap_budget_reserved in VaultKeyData, plus fail-fast cancellations) come from the PR merge-ref inheriting a breakage on main — main's own CI fails identically at fc927caad (runs 31654473783 / 31653813030), and #6055/#6056 are already open to fix it. Nothing in this PR touches KMS. Will re-run checks once main is green.

reatang (changes requested):
Reviewed with a focus on the root-cause claim, the decoder rewrite, and the blast radius of restoring multipart compression. The core diagnosis and the `rio` fix hold up under independent checking — one blocking issue below is about the PR description, not the code.

## Verified independently

**The format claim is correct, and it invalidates the earlier root-cause note.** The container is a self-delimiting 8-byte block header (type + 24-bit length + crc32), and `COMPRESS_TYPE_END` is referenced only by the decoder — `CompressReader` never emits it, so a stream terminates on inner EOF. Concatenated part streams therefore decode across boundaries by construction. The analysis recorded earlier on #5957 and backlog#1848 (multipart parts as independent frames that cannot be concatenated) was wrong, and treated `CompressionAlgorithm::Lz4` as an LZ4 *frame* container rather than the per-block codec. Those two write-ups need correcting so the wrong model does not get reused; I will handle that.

**The `Poll::Pending` root cause reproduces as described.** With `header_done` already true the header loop is skipped, but the block at `compress_reader.rs:265-286` still re-executes unconditionally and resets `compressed_read` to 0. A `Pending` mid-payload therefore discards the bytes already consumed while the inner stream has moved on — block framing desynchronizes, surfacing as `ERROR_frameType_unknown`.

**The new tests genuinely catch the old bug — they do more than fail.** I ported `PendingChunkReader` plus the two byte-exactness tests onto the *unfixed* decoder. The run does not fail; it **hangs** (600 s timeout, process at ~0% CPU). That independently confirms the second defect noted in the description: after a partial header read the old code returns `Poll::Pending` while `inner.poll_read` returned `Ready`, so no waker is ever registered and the task is never woken. Good catch, and the test shape (alternating `Pending`, capped chunk sizes, 1..8192) is the right one to pin it.

The state machine itself reads correctly: `compressed_len == 0` as the "no in-flight payload" sentinel is what makes the header phase unrepeatable, and the EOF handling (`header_read == 0 && n == 0` as the only clean termination, `UnexpectedEof` otherwise) is a real improvement over the previous silent `break` into a stale-buffer decode.

## Blocking: the description documents changes that are not in this PR

The summary devotes a bullet and a testing line to a rebalance fix — `crates/ecstore/src/services/rebalance/migration.rs` setting `data_movement + raw_data_movement_read`, and `test_migrate_entry_version_reads_raw_stored_bytes` pinning it via a spy backend.

That change is not in this PR. The changed-file list has 14 entries and `migration.rs` is not among them; the diff contains zero occurrences of "rebalance".

The underlying defect is real, which is why this matters:

- the decode bypass at `object_api/readers.rs:502` keys on `raw_data_movement_read` **only**;
- decommission sets both flags (`core/pools.rs:2291`);
- rebalance sets `data_movement` alone (`services/rebalance/migration.rs:30` and `:48`), so the migration read yields decompressed/decrypted plaintext while `migrate_object` rewrites the stored representation verbatim.

Restoring multipart compression widens the population of objects exposed to that path. Leaving the description as-is would tell a reviewer or release manager the risk is closed when it is not. Either land the code and its test here, or drop those two passages and carry the defect as a named known-issue with its own tracking item.

## Non-blocking questions

1. **Default-off means #5957 is not fixed by default.** `RUSTFS_COMPRESSION_MULTIPART_ENABLED=false` is the right rollout shape, but a user following the guidance on #5957 (`RUSTFS_COMPRESSION_ENABLED=true`) still gets uncompressed multipart objects after this merges. Worth settling explicitly: the release-note wording, the follow-up comment on #5957, and which release flips the default.

2. **The ListParts change is two fixes in one.** Reporting logical size is required by compression; the SSE side is an independent, wire-visible correction of pre-existing behavior (previously ciphertext size). The new behavior matches AWS semantics and the fallback logic is right (`actual_size > 0`, else stored size, so legacy `0`/`-1` parts are safe), but it deserves its own release-note line, or its own PR.

3. **The mixed-feature invariant has no runtime guard.** With multipart compression restored, different nodes may compress different parts of one object, and `rio-v2` vs default write different stream formats. Build discipline is the stated control; a startup-time warning would make a mis-deployed cluster loud instead of silently corrupting.

4. **`e2e-smoke` lane cost.** The whole `compression` module joins the PR lane alongside ~650 new e2e lines — a measured before/after lane duration would help.

5. Nits: the poison guard opens with `if *this.poisoned { *this.poisoned = true; ... }` (self-assignment); and the `COMPRESS_TYPE_END` branch is now dead on the write side — worth a line saying it is kept defensively.

## Suggested addition: pin that range reads stay bounded

The e2e coverage proves compressed multipart reads are byte-exact, but nothing fails if a range GET regresses into decoding the object from byte zero — the bytes would still be correct, only the read amplification would explode. That is the property most at risk if the seek path is refactored later.

This is observable without touching production code: `rustfs_io_get_object_shard_read_observed_bytes_total` is already emitted per shard read by the erasure layer (`erasure/coding/decode.rs`). On a four-node cluster with the reader-metric env already used by `inline_fast_path_cluster_test`, request only the tail of a two-part object and diff the counter across the request.

Measured on the existing `put_two_part_multipart` layout: a bounded tail read costs **~16 KiB** of physical reads against a **~5.2 MB** object, so a whole-object decode is ~320x over the bounded case and a budget at half the object size sits far from both.

Two traps worth inheriting if you pick this up — both cost me a debugging cycle:

- **Settle across export intervals.** With `RUSTFS_OBS_METER_INTERVAL=1`, treating a single unchanged sample as "settled" measures a delta of 0, because the range read's counter has not been exported yet. Requiring several consecutive equal samples fixes it.
- **Assert a lower bound too.** My first version asserted only `read_bytes < budget` and passed with `read_bytes == 0` — a vacuous green. `assert!(read_bytes > 0)` turns "measured nothing" into a failure instead of a pass.

Happy to hand over the working version of this test rather than have you rebuild it.

overtrue:
The rebalance raw-read fix is already present on this head through merged #6057, including the production-path flag assertions. The relevant files match the current base, so no code patch is needed here; the PR description is stale.

overtrue:
Rebased onto latest main (conflicts resolved), and the blocking item plus the nits are addressed.

**Blocking — description claimed a change that is not in this diff.** You are right, and thank you for catching it. The rebalance fix *was* in this branch originally; when #6057 landed the same fix on main, an earlier rebase resolved that file to main's version and the change left my diff while the description kept describing it. The description now drops both passages and states it accurately under known issues: the forward fix landed in #6057 and is present on this branch's base (`rebalance_object_migration_read_opts` sets both flags), it is not part of this diff, and restoring multipart compression widens the population of objects that path would have exposed. Damaged-object auditing stays tracked in backlog #1850.

One correction on the defect's current state, since it changes what a release manager should conclude: `migration.rs:30` and `:48` are `rebalance_delete_marker_opts` and `rebalance_remote_tiered_opts` — the delete-marker and remote-tier paths, which do not do an object body read. The object migration read is `rebalance_object_migration_read_opts` (`migration.rs:70-80` on current main), and it sets `data_movement: true` **and** `raw_data_movement_read: true`. So the read-path defect is closed on main today; what remains open is only the audit of objects damaged before that fix.

**Nits (5):** the self-assignment in the poison guard is gone, and the `COMPRESS_TYPE_END` branch now carries a line saying the writer never emits an end block — that absence is precisely what makes concatenated per-part streams decode as one, so the branch is kept for streams that do carry it.

**Independent confirmation of the hang:** porting the tests onto the unfixed decoder and getting a 600 s timeout at ~0% CPU rather than a failure is a better demonstration of the lost-wakeup defect than anything in my write-up. Noted in the description as a second defect the rewrite removes.

**Non-blocking items** — taking them in order:

1. **Default-off vs #5957.** Agreed this needs settling explicitly rather than implicitly. My read: the alternative is shipping a default that can break reads on not-yet-upgraded nodes, so the flag should stay off for one release, with #5957 getting a follow-up comment saying the capability is restored but gated, and the default flip riding the `multipart-compression-default-off-window` register entry. Happy to align on whichever release you want to name.
2. **ListParts as two fixes.** Fair. The SSE half is genuinely independent of compression. I would rather not split it out now and leave the wire inconsistent between the two halves, so my preference is one release-note line calling out the SSE-visible change explicitly — but if you would rather see it as its own PR I will pull it out.
3. **Mixed-feature invariant has no runtime guard.** Agreed that build discipline is a weak control for something that corrupts silently. Worth its own issue: a startup warning needs a way for a node to observe its peers' backend, which is a capability-handshake question rather than a compression one.
4. **e2e-smoke lane cost.** Reasonable ask; I do not have a trustworthy before/after from this machine (its four-node runs fail environmentally). I will take the number from the first green CI run on this head rather than quote a local figure.
5. **Suggested range-bound test — yes please, I will take the handover.** It pins the property most at risk in a later seek-path refactor, and it is exactly the gap I flagged internally: my `read_compressed_multipart` helper plans the window and then feeds that same window back, so it cannot catch a plan that is self-consistent but wrong. A physical-bytes assertion closes that from the outside. Both traps noted — settling across export intervals, and the `read_bytes > 0` lower bound so "measured nothing" fails instead of passing green.

CI note: the arch-guard failures on this head come from main's own `rio-v2-dormant-variant` register entry (#6036), not from this PR. I diffed the guard output against a pristine main worktree: the failure sets are byte-identical, zero new entries from this branch.

rustfs/src/app/multipart_usecase.rs
overtrue: Addressed: the session marker is now gated on `RUSTFS_COMPRESSION_MULTIPART_ENABLED` (default off), so a rolling upgrade creates no new compressed multipart objects until the operator confirms fleet convergence; retirement is pinned as `multipart-compression-default-off-window` in the cleanup register.

overtrue:
Correcting my earlier CI note with a sharper finding, because it affects more than this PR.

I previously said the arch-guard failures here were inherited from main and left it there. That is still true — I diffed this head's 15 failures against a pristine main worktree and they are byte-identical, zero from this branch. But the follow-up matters: **the guard is not merely red on this PR, it is broken for every PR whose checks run from now on**, and it has not been noticed yet only because of timing.

`scripts/check_architecture_migration_rules.sh` runs as an unconditional step in Quick Checks (`ci.yml:138`) and does a whole-repo scan (no diff scoping — no `git diff`/base-ref logic in it). Since #6036 merged at 00:19:50Z it fails 15 times on main itself. #6088 looked like a counter-example because its Quick Checks passed, but its run started at 00:16:09Z — before #6036 landed — so its merge ref predated the breakage, and it merged at 00:57:12Z on those already-green checks. So this is not maintainers tolerating a red guard; it is that nobody has hit it yet.

Root cause is a formatting collision, not a real architectural violation. The `rio-v2-dormant-variant` entry at `compat-cleanup-register.md:37` trips two rules at once: the PR-type rule fires only on lines containing the word "PR" (the entry says "Per-PR CI keeps only…") and then treats every backticked lowercase-kebab token on that line as a PR type, flagging `rio-v2`, `test-and-lint-rio-v2`, `build-rustfs-debug-binary-rio-v2`, `e2e-tests-rio-v2`; and the source-marker rule requires a `RUSTFS_COMPAT_TODO(<id>)` for each backticked kebab identifier in Open Items. `minio-file-format-compat.md:279` and `:284` fail for the same reason. Other register entries have multiple backticks (`.usage.json`, `%1F`, `usage_snapshot_complete`) and pass because none of those match the lowercase-kebab pattern.

The minimal fix is to un-backtick the non-identifier tokens in that entry and give the slug a source marker (`crates/rio-v2/src/lib.rs` would be the natural home, and it must spell out a "Remove after …" condition or it trips a third rule). I have not pushed that here: it is #6036's text, and the alternative fix — loosening the PR-type rule so it does not apply to prose — is a repo-wide guard semantics change that should be its owner's call, not something I slip into a compression PR.

Flagging it so whoever owns #6036 can unblock the queue before the next PR runs into it.

overtrue:
The `Test and Lint (swift)` and `Test and Lint (rio-v2)` failures on this PR are inherited from main's #6058, not from this branch. Traced to the exact line:

Both lanes fail on one test, `storage::rpc::node_service::tests::snapshot_lease_acquire_and_renew_handlers_fail_closed`, with

```
left:  Some(Error { code: 36, error_info: "io error cannot find disk" })
right: Some(Error { code: 7,  error_info: "drive does not support O_DIRECT" })
```

#6058 (`d60a77b75`, fix(quota): enforce durable hard quota reservations) rewrote the snapshot-lease handlers in `node_service/disk.rs`: it deleted `snapshot_lease_disabled_response()` — which produced the `UnsupportedDisk` the test asserts — and replaced it with a real implementation whose miss path is `find_disk(..) => None => Err(DiskError::other("cannot find disk"))`. It changed zero lines matching `snapshot_lease` in `node_service.rs`, so the test still asserts the retired value.

The reason it surfaced here first is timing. GitHub tests `refs/pull/6044/merge`, i.e. this branch merged with current main, so this PR picked up `d60a77b75` immediately. main's own last two green swift runs were `c48a6330d` (03:39Z) and `307f50ee1` (05:46Z), both before #6058 landed.

Confirmation on main itself: run 31776268304 on `d60a77b75` is `conclusion=failure` with `Test and Lint (swift): failure` and `Test and Lint (rio-v2): failure` — the same single test, 1 failed out of 2943 run. Nothing in this branch touches `node_service.rs`, `node_service/disk.rs`, or the disk layer; the 14 changed files are compression, multipart, ListParts, e2e and docs.

Worth noting for whoever picks up #6058: the test being wrong here is arguably the smaller half. These nine node_service tests were un-ignored only yesterday by #6047 and had never run in any CI lane before, so #6058 was written against a file whose tests had no CI history. The fix is either to update the assertion to the new miss error or to keep a fail-closed path returning `UnsupportedDisk`, and that choice belongs to #6058's author since it decides what the handler should actually do for an unknown disk.

PR #6056, opened 2026-08-13
fix(kms): teach the observing Vault deserializer the wrap-budget field

overtrue:
## Urgent: main does not compile

Merge race: #6019 added `wrap_budget_reserved` to `VaultKeyData` (with `#[serde(default)]`), then #6040 — green against a base **without** that field — landed its unknown-field-observing manual `Deserialize` whose `VaultKeyData { .. }` initializer doesn't mention it. `cargo check -p rustfs-kms` fails on main at `crates/kms/src/backends/vault.rs:311` (E0063).

## Fix

Not just the missing initializer line: the visitor would also have **misclassified `wrap_budget_reserved` as an unknown field and dropped its value on every read**, silently regressing the persisted wrap-budget count that #6019's design exists to preserve (its doc comment names exactly this old-build-rewrites hazard). The field is now a first-class visitor member — parsed when present, duplicate-checked, `unwrap_or(0)` when absent, mirroring the struct's serde default for records written before the field existed.

## Verification

- `cargo check -p rustfs-kms` — fails on main, passes here
- `cargo test -p rustfs-kms --lib vault` — 178 passed
- `cargo clippy -p rustfs-kms --all-targets -- -D warnings` — clean


overtrue:
Superseded by #6055, which shipped the same fix (field as a first-class visitor member with serde-default mirroring, plus a regression test) minutes earlier from the other working session. Closing; no unique content remains here.

PR #6057, opened 2026-08-13
fix(rebalance): converge multipart data movement retries

### LINKED ISSUE #5948: rebalance fail by Precondition failed
**Describe the bug**
when I click rebalance, this logs occur

{"timestamp":"2026-08-11T11:41:10.852482075+08:00","level":"ERROR","message":"Rebalance pool failed","event":"rebalance_state","component":"ecstore","subsystem":"rebalance","pool_index":0,"state":"pool_failed","error":"Io error: rebalance bucket rustfs-perf-bench failed for pool 0: Io error: rebalance entry write_target failed for rustfs-perf-bench/bench/upload/20260805T064802Z/obj-00000000: Io error: rebalance_object: complete_multipart_upload failed for rustfs-perf-bench/bench/upload/20260805T064802Z/obj-00000000: Precondition failed","target":"rustfs_ecstore::services::rebalance::runtime","filename":"crates/ecstore/src/services/rebalance/runtime.rs","line_number":158,"threadName":"rustfs-worker","threadId":"ThreadId(13)"}

cxymds:
## Related Issues

Fixes #5948.

## Summary of Changes

- Make multipart data-movement retries converge when a target was committed before part-checksum persistence was enabled, while continuing to reject unowned, partially populated, empty, corrupt, or conflicting checksum state.
- Isolate data-movement multipart uploads under an internal namespace so external multipart APIs cannot enumerate or mutate migration-owned uploads.
- Coordinate stale multipart cleanup with upload locks, quorum metadata rechecks, ownership checks, and bucket lifecycle rules; active migration uploads remain available to the migration path while default stale-age cleanup remains in place.
- Preserve exact dual RustFS/MinIO ownership and tag proofs, and revoke migration ownership on client metadata/tag mutations so later stale replacement cannot overwrite user changes.

## Verification

- `make pre-commit` (passed)
- `cargo fmt --all --check` (passed)
- `cargo test -p rustfs-ecstore data_movement_ -- --nocapture` (passed: 124 tests)
- `cargo test -p rustfs-ecstore stale_multipart_cleanup_ -- --nocapture` (passed: 7 tests)
- `cargo nextest run -p rustfs-ecstore --status-level all --final-status-level all --failure-output immediate` (passed: 3,975 tests)
- Real store regression `data_movement_conflicts_and_multipart_retries_converge_safely` (passed; gates disabled for the first commit, then both checksum gates enabled for retry)
- Real stale-target regression `data_movement_multipart_replaces_only_unlocked_owned_generation` (passed)
- Real lifecycle regression `stale_multipart_cleanup_excludes_data_movement_from_abort_lifecycle` (passed)
- `cargo clippy --workspace --all-targets -- -D warnings` (passed)
- `cargo test -p rustfs-filemeta` (passed: 249 unit tests and 2 property tests)
- `./scripts/check_logging_guardrails.sh` (passed)
- `git diff --check` (passed)
- Workspace `make test` was attempted but could not complete because the macOS runner exhausted disk space while compiling unrelated test targets; the affected `rustfs-ecstore` package suite passed independently. Build artifacts were cleaned afterward.

## Impact

Rebalance and other data-movement operations no longer fail permanently with `PreconditionFailed` when retrying a multipart object across a rolling upgrade that enables checksum-sidecar persistence. Replacement remains fail-closed for client-owned objects, newer generations, retention, legal hold, lifecycle or namespace fence loss, and malformed or conflicting checksum metadata. Internal multipart namespace and metadata changes are not exposed through the public S3 API. No public API or required configuration changes are introduced.

## Additional Notes

- Correctness: attacked pre-gate to post-gate retry convergence, exact-generation matching, partial/empty/conflicting checksum state, lifecycle age boundaries, and cleanup rechecks; no remaining break found.
- Simplicity: verified the ownership proof helper has two callers and prevents proof drift; no smaller safe implementation was identified.
- Security: attacked unowned and forged marker combinations, tag-proof changes, client metadata mutation, internal upload IDs, and checksum sidecar corruption; fail-closed behavior remained intact.
- Concurrency/durability: attacked PutPart/Complete/Abort versus stale cleanup, lock loss, quorum rechecks, and retry idempotency; no committed-data deletion or half-visible object path found.
- Compatibility: attacked rolling upgrades, RustFS/MinIO metadata aliases, public multipart listing, and legacy sidecar-less targets; no public API or on-disk identity break found.
- Performance: checked per-object metadata/part scans, retry I/O, cleanup lock hold time, and allocations; added work is bounded and limited to migration/error or stale-candidate paths.
- Test coverage: the new unit and real-store tests fail when the compatibility branch or lifecycle exclusion is reverted; partial, empty, corrupt, conflicting, and unowned boundaries are covered.


crates/ecstore/src/data_movement/mod.rs
```diff
@@ -763,6 +1148,24 @@ fn resolve_data_movement_overwrite_resume_result(
...
     src_pool_idx: usize,
     target_pool_idx: usize,
+) -> Result<bool> {
+    resolve_data_movement_overwrite_resume_result_for(
+        err,
+        target_result,
+        source,
+        src_pool_idx,
+        target_pool_idx,
+        data_movement_part_checksum_writer_enabled(),
```
overtrue, line 1182: A target committed while part-checksum persistence is disabled cannot converge after both documented gates are enabled. The new retry compares part checksums, rejects that sidecar-less target, and stale replacement also rejects its equal mod_time; the real store path fails with PreconditionFailed. Please keep pre-gate targets compatible during retry or add a safe upgrade path.

overtrue (changes requested):

crates/ecstore/src/data_movement/mod.rs
```diff
@@ -763,6 +1159,24 @@ fn resolve_data_movement_overwrite_resume_result(
...
```
overtrue, line 1182: This still keys retry equivalence to the current checksum gate. On this exact head, the real store test still commits a sidecar-less target with both gates off and then fails retry with PreconditionFailed after both gates are enabled. Please preserve compatibility for targets produced before gate enablement or migrate them safely.

overtrue (changes requested):

crates/ecstore/src/data_movement/mod.rs
cxymds: Fixed in 23fd5487d. Retry now accepts a sidecar-less target only when it has the exact dual RustFS/MinIO data-movement ownership and tag proof, every target part checksum is absent, no checksum sidecar metadata exists, and all remaining object identity matches. Unowned, partially missing, empty, corrupt, and conflicting checksum state remains rejected. Added unit boundary coverage and a real store regression that commits with both gates disabled, enables both gates, retries, and verifies the target data directory and multipart topology are unchanged.

crates/ecstore/src/data_movement/mod.rs
cxymds: This is fixed in 23fd5487d as well. The retry compatibility branch is independent of the current checksum gate only for a fully legacy sidecar-less target with exact dual ownership/tag proof and full object identity equivalence; partial, empty, corrupt, conflicting, or unowned checksum state remains rejected. The real store regression now runs the disabled-gate commit, enables both gates, retries, and verifies the existing target data directory and parts are preserved.

overtrue (approved):
LGTM

PR #6059, opened 2026-08-13
feat(kms): report a key as due for rotation once its wrap budget is spent

overtrue:
## Summary

Closes the last unimplemented acceptance criterion of rustfs/backlog#1636. #6019 landed the wrap accounting (per-key reserved count, aggregate gauge) but never fed it into the per-key `rotation_due` verdict, so the criterion "wraps 超阈值 → true/wraps" had no `Wraps` reason and no threshold to compare against.

- `RUSTFS_KMS_ROTATION_MAX_WRAPS` is a second, independent threshold, parsed with the same discipline as the age one: unset or unparsable leaves the verdict unreported rather than inventing a policy, `0` disables, and values below 1,000,000 are raised to it — wraps are reserved in blocks of that size, so a smaller threshold would trip on the first reservation regardless of how many wraps actually happened.
- The wrap check runs **before** the age check, so a key crossing both reports `wraps`. The AES-GCM random-nonce ceiling (NIST SP 800-38D) is a cryptographic bound an operator cannot negotiate; the age period is a policy they chose, and it is the non-negotiable one they need to see.
- Backends that report no count leave the wrap half silent instead of guessing: Transit and AWS wrap externally, and records written before wrap accounting carry nothing. A backend that cannot rotate at all is still never reported as due (`unsupported` wins first, unchanged).

Documented in the rotation section of `docs/operations/kms-backend-security.md` and in the `KmsKeyRotationOverdue` runbook entry, which previously listed only three reasons.

## Tests

- `rotation_readiness_reports_an_exhausted_wrap_budget`: at / above / below the threshold; `wraps` outranks `age` when both are crossed; no threshold configured reports nothing even at `u64::MAX`; a backend reporting no count stays silent; a non-rotating backend is never told to rotate whatever it wrapped.
- `rotation_wrap_threshold_parsing_refuses_to_guess`: unset / unparsable / empty / negative / zero all disable; below-floor clamps to the floor; an above-floor value is honored verbatim.

## Verification

- `cargo test -p rustfs-kms --lib` (537+ passed), `cargo clippy -p rustfs-kms --all-targets -- -D warnings`, `cargo fmt --all`, `bash scripts/check_doc_paths.sh`

Based on rustfs/rustfs#6055 (the hotfix restoring `main`'s build); its commit is the branch point, so this PR shows only the wraps change once #6055 merges.

Refs rustfs/backlog#1636, rustfs/backlog#1562.


crates/kms/src/manager.rs
```diff
@@ -68,6 +75,33 @@ fn parse_rotation_max_age(value: Option<&str>) -> Option<Duration> {
...
+        warn!(
+            variable = ENV_KMS_ROTATION_MAX_WRAPS,
+            "ignoring unparsable KMS rotation wrap budget; rotation readiness stays unreported"
+        );
+        return None;
+    };
+    if wraps == 0 {
+        return None;
+    }
+    Some(wraps.max(MIN_ROTATION_MAX_WRAPS))
```
overtrue, line 102: This accepts configured limits above the AES-GCM 2^32 random-nonce ceiling. For example, a value of 5000000000 reaches the persisted wrap counter and keeps the admin rotation verdict false after 2^32 wraps. Please cap enabled values at 2^32 and add the upper-bound regression.

overtrue, line 102: This still accepts wrap limits above the AES-GCM 2^32 ceiling. With a 5000000000 limit, KV2 can persist 4295000000 reserved wraps while the current key material has already reached 2^32 actual wraps, yet the admin verdict remains false. Please cap enabled values at 2^32 and add the upper-bound regression.

PR #6061, opened 2026-08-13
refactor(rustfs): move layer-neutral shared types out of server

overtrue:
## What

PR4 of rustfs/backlog#1834. **Re-opened**: the previous #6060 was stacked on the #6056 hotfix branch and got merged into that branch moments before #6056 was closed as superseded by #6055 — so the change never reached main. This is the same commit rebased onto current main (which carries #6055).

`RemoteAddr`, the `DependencyReadiness` family (`DependencyReadiness`, `ReadinessDegradedReason`, `DependencyReadinessReport`), and `convert_ecstore_object_info` (+ its `offset_date_time_to_timestamp` helper) are consumed across app, infra, and interface layers but lived under `server` — every lower-layer import was an upward edge the layer guard had to baseline.

- New layer-neutral `rustfs/src/shared_types.rs` (classified **infra** by the guard, so all consumer imports become downward/lateral). Pure move — no type, impl, or behavior change.
- `server::readiness` / `server::event` re-export for their own internals; eight consumer sites (app usecases, cluster_snapshot, storage access/helper, admin handler tests) import from the new home.
- **Baseline: −8 lines, +0** — the ratchet's intended direction.
- The two remaining readiness baseline entries (`collect_/snapshot_dependency_readiness_report`) need the collection machinery extracted from `server/readiness.rs` and stay in the issue's PR5 scope.

## Verification (re-run after rebase onto main)

- `bash scripts/check_layer_dependencies.sh` — passes; baseline diff is exactly 8 deletions
- `cargo check -p rustfs` — clean
- Earlier on the identical tree: `cargo test -p rustfs --lib server` (314) / `cluster_snapshot` (15) / `server::event` (10) passed; `cargo clippy --lib --tests -- -D warnings` clean; `make pre-commit` ✅

Ref rustfs/backlog#1834.


rustfs/src/lib.rs
```diff
@@ -94,6 +94,7 @@ pub mod protocols;
 pub mod runtime_capabilities;
 pub(crate) mod runtime_sources;
 pub mod server;
+pub mod shared_types;
```
overtrue, line 97: This makes the readiness types a new public Rust API even though they were intentionally crate-private (`c2e58917e`). The move is otherwise internal, so please use `pub(crate) mod shared_types;`; `server::RemoteAddr` can remain the public re-export.

overtrue (commented):
Please preserve the existing visibility.

PR #6069, opened 2026-08-13
perf(ecstore): guard inline data-read metadata early-stop

houseme:
## Related Issues

Related to https://github.com/rustfs/backlog/issues/1647

## Summary of Changes

Add a default-off inline-only data-read metadata early-stop gate for small GET reads. Before cancelling pending metadata requests, the gate verifies the expected distinct erasure shards, shard identity, inline payload length, bitrot, erasure distribution, and full plaintext body; if any condition cannot be confirmed, the read falls back to full metadata fanout.

Keep prepared reads, Range, partNumber, explicit version/free-version, data-movement, restore, non-inline, transformed, remote, purge-pending, corrupt, mixed-generation, and unsafe metadata shapes on full fanout. The change also records ReadVersion fanout lifecycle metrics for scheduled, completed, and cancelled tasks on the normal-return path, with legacy_duplex and internal_meta path attribution.

This PR intentionally keeps the optimization behind the existing opt-in gates and does not enable it by default.

## Verification

- `cargo test -p rustfs-ecstore metadata_early_stop -- --nocapture`
- `cargo test -p rustfs-ecstore data_read_early_stop -- --nocapture`
- `cargo test -p rustfs-ecstore metadata_fanout_lifecycle -- --nocapture`
- `cargo test -p rustfs-ecstore direct_memory -- --nocapture`
- `cargo test -p rustfs-ecstore inline_data_shards_direct -- --nocapture`
- `cargo test -p rustfs-io-metrics metadata_fanout_lifecycle_records_named_histograms -- --nocapture`
- `cargo fmt --all --check`
- `git diff --check`

No additional local verification was rerun after the final commit per request.

## Impact

The runtime behavior is unchanged unless the relevant metadata early-stop feature gates are enabled. When enabled for eligible whole latest plain inline GET reads, the reader may return after verified quorum metadata instead of waiting for every metadata response. Non-eligible or unsafe shapes continue to use full fanout.

New get-object stage metrics expose metadata fanout lifecycle counts for scheduled, completed, and cancelled ReadVersion tasks. These metrics cover fanout calls that return normally; caller cancellation or outer request drops remain outside this lifecycle sample.

## Additional Notes

The implementation preserves full fanout for prepared reads and request shapes where response length, version identity, transformation, or fallback semantics are not yet proven equivalent. Healthy bounded 12+4 paths may still schedule a spare via the existing single-pending hedge; the lifecycle metrics make that distinction visible.

### DIFF
--- a/crates/ecstore/src/cluster/rpc/internode_data_transport.rs
+++ b/crates/ecstore/src/cluster/rpc/internode_data_transport.rs
@@ -31,7 +31,7 @@ use rustfs_config::{
     DEFAULT_INTERNODE_DATA_TRANSPORT, ENV_RUSTFS_INTERNODE_DATA_TRANSPORT, INTERNODE_DATA_TRANSPORT_TCP,
     KNOWN_INTERNODE_DATA_TRANSPORT_BACKENDS,
 };
-use rustfs_rio::{HttpReader, HttpWriter};
+use rustfs_rio::{ChunkReaderBox, HttpChunkReader, HttpReader, HttpWriter};
 use sha2::{Digest, Sha256};
 use std::collections::HashMap;
 use std::future::Future;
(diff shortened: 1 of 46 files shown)


overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #6071, opened 2026-08-13
test(table-catalog): share the stateful object backend across both test files

overtrue:
## What

PR2 of rustfs/backlog#1837 (both halves), continuing #6066.

**Commit 1 — store side moves.** The store tests' `TestCatalogObjectBackend` cluster (state/record/locks/pause types, the seed/fail/pause instrumented inherent impl, the trait impl, and the `BlockingObjectPublication`/`UnserializedTestPublication` commit-publication fakes — 544 lines) moves verbatim into `test_support.rs` with `pub(crate)` visibility where tests reach in. This also covers the publish/put pause structures the issue listed under PR1.

**Commit 2 — admin side folds in.** The admin handler tests' `TestTableCatalogObjectBackend` (201 lines) is deleted and its instrumentation folded into the shared fake, which the admin tests take through a type alias so no call site is renamed. Per the issue's ruling this is the union of both instruments, not a replacement of one by the other.

### Two behavioral differences the suites caught

Neither was assumed away; both are preserved with comments:

1. **Lock observability** — the admin fake implemented only `acquire_write_lock`, so the trait's default `acquire_read_lock -> acquire_write_lock` delegation made *read* acquisitions visible in `lock_attempts`. The shared fake implements both independently, so five fence/lock tests timed out until the read path also records attempts.
2. **Etag generation** — the admin fake used content-addressed sha256 etags (its tests observe an etag and expect rewriting identical bytes to reproduce it); the store fake uses an incrementing counter. Rather than silently picking one, the union carries a `content_addressed_etags` flag: the 80 admin construction sites go through `TestCatalogObjectBackend::content_addressed()`, and the store tests keep counter semantics unchanged.

The six one-shot path-keyed knobs (fail/corrupt put, missing/fail read, put barrier) run *before* the store fake's attempt-indexed injection maps, matching each fake's original ordering.

## Verification

- `cargo test -p rustfs --lib table_catalog` — 481 passed
- `cargo test -p rustfs --lib admin::handlers::table_catalog` — 167 passed
- `cargo clippy -p rustfs --lib --tests -- -D warnings` — clean
- `make pre-commit` — ✅

PR3 (store mocks + the `row_level_conflict` table-driven fold) is the remaining item.

Ref rustfs/backlog#1837.


rustfs/src/table_catalog/test_support.rs
```diff
@@ -226,3 +235,707 @@ pub(crate) fn manifest_avro_bytes_with_nullable_sequences(files: &[(&str, i32, i
...
+        if result.as_ref().is_some_and(|object| object.data.len() > max_size) {
+            return Err(TableCatalogStoreError::Invalid(format!(
+                "catalog object {bucket}/{object} exceeds the maximum size of {max_size} bytes"
+            )));
+        }
+        Ok(result)
+    }
+
+    async fn object_metadata(&self, bucket: &str, object: &str) -> TableCatalogStoreResult<Option<TableCatalogObjectMetadata>> {
+        let mut state = self.state.lock().await;
```
overtrue, line 661: This override bypasses the admin fake’s existing one-shot missing and failing read injections. A focused publication-path test observes the object, arms the missing-read injection, then prepares publication: the previous fake returns Conflict, but this code succeeds and leaves the injection armed. Please preserve those injections for content_addressed() and cover both missing and error cases.

PR #6090, opened 2026-08-14
fix(ecstore): preserve CopyObject producer errors

cxymds:
## Related Issues
Fixes #1854. Related upstream report: rustfs/rustfs#6028. Fresh-connection resume remains out of scope and is tracked separately in backlog #1855. #1850 is intentionally not included.

## Summary of Changes
Preserve the terminal result of the legacy EC CopyObject duplex producer so producer-side decode, bitrot, transport, cancellation, panic, or channel-close failures cannot be converted into a false clean EOF. Whole-object and explicit full-range compressed/encrypted reads now validate the transformed EOF and then observe the producer terminal result, while partial ranges retain the existing bounded-consumption behavior. The reader releases the duplex source before awaiting the producer terminal result so corrupt physical tails fail closed without a backpressure deadlock.

Fix `HardLimitReader` to respect the `AsyncRead` contract: it never writes more than the declared remaining length into the caller's buffer before reporting excess data, reuses a scratch buffer for the overflow probe, and treats zero-capacity `ReadBuf` calls as no-ops. The legacy duplex reader applies the same zero-capacity rule.

Sanitize `StorageError::Io` messages mapped to S3 `InternalError` while retaining the typed storage error as the internal source. Existing request-body `UnexpectedEof` to `IncompleteBody` behavior remains unchanged. Internode HTTP and chunk-reader stall timeouts continue to expose a typed `BodyStalled` source.

## Verification
- `cargo fmt --all --check` - passed.
- `git diff --check` - passed.
- `cargo test -p rustfs-ecstore --lib legacy_duplex_producer_reader_tests -- --nocapture` - 9 passed.
- `cargo test -p rustfs-rio --lib hardlimit_reader::tests -- --nocapture` - 8 passed.
- `cargo test -p rustfs-ecstore --lib object_api::readers::tests::test_get_object_reader -- --nocapture` - 6 passed.
- `cargo test -p rustfs --lib storage_io_internal_error_redacts_public_message_and_retains_source -- --nocapture` - 1 passed.
- `target/debug/deps/rustfs_rio-3ae5253aaf70c006 stall_timeout --nocapture --test-threads=1` - 3 passed.
- Existing `IncompleteBody` mapping tests - 2 passed.
- `make pre-pr` - in progress locally; guard scripts have passed and workspace Clippy is compiling.

## Impact
[...]

Issue #6028, opened 2026-08-12
Erasure decode stalls with all shards present when server-side copying a multi-GiB object


stevapple:
**Describe the bug**

On a 4-node erasure-coded cluster (1.0.0-beta.11), a Docker registry (Harbor) pushing a ~26 GiB blob fails at commit, reproducibly (3/3 attempts, ~16 min each). The blob body uploads fine every time (multipart, ~2,650 parts, ~13.5 min). At commit the registry requests a server-side copy of the staged upload (`.../_uploads/<uuid>/data`) to the final object path; rustfs reads the staged object back in 32 MiB ranges, and those reads stall at the erasure decode stage — median ~30 s, max 131 s — **with every slow decode logging `available_shards: 12, missing_shards: 0`**. Reads that outlast the 30 s `RUSTFS_OBJECT_DISK_READ_TIMEOUT` fail quorum, and the client receives `500 InternalError: Io error: Incomplete body: 33554432 bytes were still expected` — a misleading message, since nothing is missing: no disk errors, no kernel I/O errors, multi-TB free per drive, load 24–46 on 80-CPU machines. The failure follows whichever node coordinates the request (reproduced with two different coordinators), so it is not a bad node.

The stall signature is:

```
Erasure decode: decode stage is slow or failed ... available_shards: 12, missing_shards: 0, bytes_written: 0, elapsed_ms: 30002
```

with `bitrot reader read error: HttpReader stall timeout: no data received before deadline` on a subset of shard streams, and the quorum failure / client 500 at the same instant. Tens of seconds to decode a 2 MiB slice from fully available, healthy shards on idle disks.

**Context: we already run internode HTTP/2 window overrides, which fixed this signature for smaller objects**

On beta.10 we saw the identical signature on plain GETs: 256 MiB objects failed 0/8 with all shards available, failure durations quantized at multiples of the read-stall deadline, on idle machines. The behavior was consistent with HTTP/2 flow-control starvation between lockstep-consumed shard streams: fast streams fill their windows with read-ahead the decoder can't consume yet, that read-ahead exhausts the shared connection-level window, and the slow streams the decoder is actually waiting on get zero window until a deadline fires. Setting explicit client windows —

```
RUSTFS_INTERNODE_HTTP2_INITIAL_STREAM_WINDOW_SIZE=2097152
RUSTFS_INTERNODE_HTTP2_INITIAL_CONNECTION_WINDOW_SIZE=67108864
[...]

PR #1855, opened 2026-02-17
fix: uncompress -> compress

cuiweixie:
## Type of Change
- [ ] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [ ] Other:

## Checklist
- [ ] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [ ] Passed `make pre-commit`
- [ ] Added/updated necessary tests
- [ ] Documentation updated (if needed)
- [ ] CI/CD passed (if applicable)

## Impact
- [ ] Breaking change (compatibility)
- [ ] Requires doc/config/deployment update
- [x] Other impact:

## Additional Notes

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

PR #6091, opened 2026-08-14
fix(ecstore): resume remote shard reads once

cxymds:
## Related Issues
Fixes rustfs/backlog#1855. Upstream report: rustfs/rustfs#6028. Depends on #1854 / PR #6090 for the typed body-stall and producer-error contract. #1850 is intentionally out of scope.

## Summary of Changes
Add stream-level recovery for remote shard reads after one typed stall/reset. The AsyncRead and ChunkReader paths track only bytes already delivered to BitrotReader, compute the resumed request with checked offset arithmetic, open one fresh connection, and stop after one retry. Fresh TCP clients preserve the current CA, mTLS identity, proxy policy, HTTP/2 tuning, timeout, and request signature configuration. Non-retryable errors and a second stall remain terminal.

## Verification
- `cargo test -p rustfs-ecstore --lib cluster::rpc::remote_disk::tests::remote_reader_resumes_from_emitted_bytes_without_duplicates -- --nocapture --test-threads=1` — passed.
- `remote_chunk_reader_resumes_from_emitted_bytes_without_duplicates` — passed.
- `remote_reader_retries_at_most_once_and_preserves_non_retryable_errors` — passed.
- `resumed_read_request_checks_large_offsets` — passed.
- `cargo test -p rustfs-rio stall_timeout -- --nocapture --test-threads=1` — 3 passed.
- `cargo clippy -p rustfs-rio -p rustfs-ecstore --all-targets -- -D warnings` — passed.

## Impact
No wire or on-disk format changes. Normal reads keep the existing pooled transport. Only one failing remote shard stream is re-opened, and the retry request starts at the exact emitted byte count, preventing duplicate or missing data. A transport that cannot provide a fresh chunk reader fails closed.

## Additional Notes
This is a stacked draft PR on PR #6090 so the dependency remains visible and the diff is limited to backlog #1855. The implementation deliberately does not change global H2 read-ahead, re-run CopyObject, clear connection caches, or modify rebalance state. Release acceptance still requires the four-node TLS/H2 canary with 20 GiB and 26 GiB CopyObject and Harbor commit workloads, including connection-count, RSS, checksum, and retry-bound checks.

---


crates/ecstore/src/cluster/rpc/remote_disk.rs
```diff
@@ -214,6 +216,231 @@ where
...
+    }
+
+    fn start_resume(&mut self) -> io::Result<()> {
+        if self.request.length != 0 && self.emitted >= self.request.length {
+            self.reader = None;
+            return Ok(());
+        }
+        let request = resumed_read_request(&self.request, self.emitted)?;
+        let transport = Arc::clone(&self.transport);
+        self.resume = Some(tokio::spawn(async move { transport.open_read_fresh(request).await }));
```
overtrue: This fresh open has no usable recovery budget on the production shard path. `ParallelReader` wraps the read with the same disk-read timeout used by the HTTP body-stall timer, so when the stall fires the outer timeout is already due and retires this reader before the spawned open can run. Dropping the Tokio 1.53 `JoinHandle` also detaches the pending request. Please give the retry a bounded, cancellation-bound budget and add a BodyStalled-through-ParallelReader regression for both reader paths.

overtrue (changes requested):

crates/ecstore/src/cluster/rpc/remote_disk.rs
cxymds: Implemented in 68070e3b9. Remote body-stall reads now reserve a three-part budget: body stall T/3, initial-reader cutoff at 2T/3, and fresh-open recovery capped at T/3. The initial cutoff starts on the first poll and races the fresh-open so progress from the original reader still wins; pending fresh opens are cancelled when the reader is dropped. Added regular and chunk regression coverage through RemoteDisk and ParallelReader, including body-stall recovery, bounded fresh-open timeout, cutoff races, exact-length EOF, and cancellation.

PR #6093, opened 2026-08-14
fix(ci): restore the architecture migration guard on main

overtrue:
## main is red

The `Quick Checks` job's architecture migration check fails on `main` at eb41f4517 with **six failures**, and every open PR inherits them. Reproduced on a clean detached worktree at `upstream/main`, so it is not PR-specific.

> The `main` CI run that looks green in the branch list is a `CLA Check`. The actual `push` CI run — `31759060637` — is a failure.

#6036's `rio-v2` lifecycle entry trips **two** of the guard's parsing rules.

## Rule 1 — any line containing "PR" is scanned for PR types

The guard treats any documentation line matching `\bPR\b` as a PR-type reference line and validates every backtick-quoted token on it. The entry reads:

> Per-**PR** CI keeps only `` `test-and-lint-rio-v2` `` …

so the CI job names on that line are checked as PR types and reported unknown. Same on two lines of the `rio-v2` lifecycle section in `minio-file-format-compat.md`.

## Rule 2 — every backtick token in "## Open Items" is a register id

The guard treats every backtick-quoted token inside the register's `## Open Items` section as a cleanup-register id requiring a `RUSTFS_COMPAT_TODO` source marker. The entry backticks the crate name, the feature name, the CI job names, and **the words `RUSTFS_COMPAT_TODO` themselves** — which is why the failure list contains an entry literally named `RUSTFS_COMPAT_TODO`.

Every pre-existing entry in that file backticks **only its own id** and leaves names like `RUSTFS_ENCRYPTED_RANGE_SEEK` or `RUSTFS_INTERNODE_RPC_BODY_DIGEST_STRICT` as plain text.

## The fix follows the guard's conventions rather than loosening it

- Drop the backticks from everything except the entry id.
- Say "per pull request" where the prose meant pull requests rather than the PR-type vocabulary.

That leaves the entry id itself, which the guard requires to have a source marker. The entry argued none applies, because the temporary surface is CI workflow YAML plus a feature-gated crate. The workflow YAML is indeed outside the marker convention's Rust scope — **but the candidate crate is not**, so the marker goes at the top of `crates/rio-v2/src/lib.rs`, naming the same removal condition the register entry states.

If maintainers would rather teach the guard to exempt marker-less entries, that changes the guard's invariant and belongs in its own PR.

## Verification

- `make architecture-migration-check` — six failures → **zero**
- `make pre-commit` — ✅ (exit 0)
- `cargo check -p rustfs-rio-v2` — clean


overtrue:
已由 #6092 修好——维护者选择删除该 register 条目与 26 行 lifecycle 文档,而不是让文档符合守卫约定(去反引号 + 补 source marker)。干净 main 上 `make architecture-migration-check` 现已通过,本 PR 冗余,关闭。

若日后想恢复 rio-v2 的 lifecycle 文档,注意守卫的两条解析规则:含 `\bPR\b` 的行会把行内所有反引号 token 当 PR type 校验;register 的 `## Open Items` 段里所有反引号 token 都被当作需要 source marker 的条目 id。该文档既有惯例是只给条目 id 加反引号。

Issue #5945, opened 2026-08-11
Track object-path wedge root causes after restart churn


cxymds:
## Context

#5921 documented production nodes whose object requests wedged while readiness stayed green. This tracking issue coordinated the remaining root-cause work across readiness, disk-handle recovery, distributed lock leases, exclusive-lock behavior, and same-key contention.

## Resolution

- [x] #5936 - readiness now reflects bounded foreground object-progress stalls, preventing a wedged node from remaining eligible for load-balancer traffic.
- [x] #5938 / #6096 - RemoteDisk recovery is single-flight and health-state transitions are published atomically, allowing transient faulty handles to recover without requiring a process restart.
- [x] #5939 / #6094 / #6230 - expired local lease guards are reclaimed periodically, and `top/locks` reports the authoritative heartbeat-refreshed lease TTL instead of the original fast-lock timeout.
- [x] #5940 - the long-lived exclusive-lock investigation is closed.
- [x] #5670 - same-key fast-lock waiters are notified on release and can acquire past registered waiter counters.

## Outcome

The issues tracked here are resolved and their implementation PRs are merged into `main`. The fixes cover the identified failure classes:

- stalled foreground object traffic withdraws readiness;
- transient RemoteDisk faults can recover without restarting the affected process;
- expired lease guards no longer depend on future same-resource contention for cleanup;
- lock diagnostics expose refreshed lease state accurately;
- same-key contention no longer leaves waiters sleeping past a released lock.

## Verification

The linked PRs include focused regression coverage for readiness recovery, RemoteDisk failure/recovery races, lease expiry and periodic reclaim, lock TTL diagnostics, timeout/cancellation behavior, and same-key waiter notification. Their required CI checks passed before merge.

## Closure

All linked investigations are closed and the corresponding fixes are on `main`. Any newly observed restart-churn wedge should be reported as a new issue with current-version evidence and linked back here for historical context.

cxymds:
Closing this tracking issue as completed.

The remaining work has landed on `main`: #6094 adds periodic expired-lease guard reclamation, #6096 makes RemoteDisk recovery single-flight with atomic health-state publication, and #6230 reports the authoritative refreshed lease TTL in `top/locks`. Together with #5936, #5940, and #5670, all linked investigations are now closed.

If restart churn produces a new wedge on a current build, please open a new issue with the affected version, per-node lock and disk-health snapshots, and the restart/rejoin timeline, and reference #5945.

Issue #5939, opened 2026-08-10
Distributed lock entries never expire (ttl_secs: 0)


stevapple:
**Describe the bug**

Operating a 4-node erasure-coded cluster (1.0.0-beta.10 and 1.0.0-beta.11) we have been monitoring the lock tables via `GET /rustfs/admin/v3/top/locks`. Every entry reports `ttl_secs: 0`, and entries whose owner is gone are never removed: after any disturbance (a node outage, restarts, a crash-looping peer) the tables gain batches of entries with identical `elapsed_secs` — mass leaks from a single moment — which then age unbounded. We have observed entries at 5+ days. Normal operations on this cluster hold locks for milliseconds to seconds.

Consequences we traced to these orphaned entries:

- Requests touching an object with a stale lock hang until client timeout — on one occasion a single object hung on exactly one node for 20+ minutes while the other three nodes served it instantly.
- When a leak batch landed on shared system objects (`.rustfs.sys/config/config.json`, IAM config, bucket `.metadata.bin` objects), **every node's** console login and IAM admin operations timed out — a cluster-wide outage caused by a handful of stale READ entries, and resolved without restarting anything by force-unlocking exactly those entries.

**What we observed**

A sample from one node's table while a peer was crash-looping (redacted; every entry `ttl_secs: 0`):

```
locks=224 stale(>600s)=26
  WRITE 448679s .rustfs.sys/leader.lock
  READ  429397s bucket-b/chunks/0/412/412398_2_4194304
  WRITE 287279s bucket-a/chunks/342/342584/342584652_0_4096
  WRITE 287279s bucket-a/chunks/342/342584/342584628_0_4096   <- identical age: one leak event
  READ   62687s .rustfs.sys/buckets/bucket-a/.metadata.bin
  READ    3669s .rustfs.sys/config/config.json
```

During churn we also repeatedly see `distributed unlock failed on client: Internal error: lock not found for release: ...` warnings — release attempts against replicas that no longer exist — which is plausibly one way partial/orphaned entries arise.

`POST /rustfs/admin/v3/force-unlock` removes stale entries and immediately unblocks the affected resources; a node restart wipes that node's whole table. We now run a periodic force-unlock sweep (threshold: entries older than 10 minutes) as self-defense, and it has prevented recurrences — but it is an operator workaround for state the server should be garbage-collecting itself.

**To Reproduce**

1. Run a multi-node cluster under steady write load and watch `top/locks` on each node.
[...]

PR #6094, opened 2026-08-14
fix(lock): reap expired local lease guards

cxymds:
## Related Issues

Fixes backlog #1853. Tracks upstream rustfs/rustfs#5939 and root-cause rustfs/rustfs#5945.

## Summary of Changes

Adds a lazily started LocalClient reaper that scans the existing 64 guard shards at the fast-lock cleanup interval. Lease liveness now uses a monotonic Tokio deadline while SystemTime remains diagnostic/API metadata. Same-resource eager reclaim shares the extraction/release helper with periodic all-resource reclaim; shard write locks are released before FastLockGuard release, and periodic logging is aggregated without object keys or owners. Acquire retries now use one absolute deadline so eager reclaim cannot double the caller's timeout. No wire fields, cross-node force-unlock, fast-lock manager semantics, or scanner fencing are changed.

## Verification

- `cargo test -p rustfs-lock --lib -- --test-threads=1` (126 passed)
- `cargo clippy -p rustfs-lock --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`
- `make pre-pr` (stopped at pre-existing architecture migration guardrail failures for unregistered `rio-v2` document entries; no affected docs are in this diff)

Added regression coverage for no-reacquire orphan cleanup, sibling LocalClient/shared manager behavior, refresh-vs-reaper linearization, zero/oversized TTL, absolute acquire deadlines, four-node failed-release convergence without replica repair, and stale-quorum deadline behavior.

## Impact

No on-wire, on-disk, or public API changes. New nodes reclaim only entries owned by their own LocalClient after a monotonic local lease deadline; mixed-version deployments retain existing same-resource eager reclaim, with full unattended cleanup after all nodes are upgraded. The reaper is one low-frequency task per active LocalClient and performs bounded shard scans with no per-entry warning log.

## Additional Notes

High-risk review covered correctness, simplicity, security, concurrency/durability, compatibility, performance, and test coverage. Release canary remains required: four nodes × 12 NVMe with failed release, restart, and TTL plus one sweep interval convergence checks, alongside the #5922 rolling-upgrade validation.

---


crates/lock/src/client/local.rs
```diff
@@ -26,43 +28,51 @@ use crate::{
...
-        self.expires_at <= SystemTime::now()
+        self.deadline <= Instant::now()
     }
 
     fn refresh(&mut self) {
-        self.expires_at = SystemTime::now() + self.ttl;
+        let now = SystemTime::now();
+        let monotonic_now = Instant::now();
+        self.expires_at = now.checked_add(self.ttl).unwrap_or(now);
+        self.deadline = monotonic_now.checked_add(self.ttl).unwrap_or(monotonic_now);
```
overtrue, line 75: This revives an already expired guard if a delayed refresh arrives before the periodic sweep. Please check the monotonic deadline while holding the shard write lock, remove and release expired entries outside that lock, and return false; add a paused-time refresh-after-expiry regression.

crates/lock/src/client/local.rs
```diff
@@ -175,28 +206,36 @@ impl Default for LocalClient {
...
             LockType::Shared => crate::ObjectLockRequest::new_read(request.resource.clone(), request.owner.clone())
-                .with_acquire_timeout(request.acquire_timeout),
+                .with_acquire_timeout(acquire_timeout),
         };
 
         let mut retried_after_reclaim = reclaimed_before_acquire > 0;
         loop {
-            match lock_manager.acquire_lock(build_lock_request()).await {
+            let remaining = acquire_deadline.saturating_duration_since(Instant::now());
+            if remaining.is_zero() {
```
overtrue, line 228: This changes acquire_timeout=0 from a try-once fast path into an unconditional failure. Please allow the first manager attempt with a zero remaining budget, then enforce the deadline before any retry; cover free and contended zero-timeout requests.

overtrue (changes requested):

crates/lock/src/client/local.rs
overtrue: I could not confirm a supported production path: distributed callers reject a zero total timeout before LocalClient, and no production caller passes zero here. Treating this as non-blocking.

crates/lock/src/client/local.rs
cxymds: Addressed in the follow-up commits on this PR. refresh now checks the monotonic deadline while holding the shard write lock, removes expired entries, and releases the guard after leaving the lock scope. The paused-time regression refresh_after_expiry_releases_guard_without_reviving_it asserts refresh returns false, the expired entry is removed, and a new owner can reacquire the resource.

overtrue (approved):

PR #6095, opened 2026-08-14
fix(kms): resolve Vault auth from the environment at startup and add Kubernetes auth

### LINKED ISSUE #5934: KMS configuration with Hashicorp Vault and Kubernetes authentication
Hi,

Environment:
kubernetes 1.36
hashicorp vault: 1.12.0
rustfs operator: latest
rustfs: 1.0.0-beta.12

I would like to encrypt my RustFS instance with a key managed by Hashicorp Vault. I am using kubernetes auth backend and I would like RustFS to authenticate against Hashicorp Vault with its service account. So far I've done the following:

``` tf
locals {
  tenant_name = "example"
}

resource "vault_kubernetes_auth_backend_role" "rustfs" {
  backend = vault_auth_backend.kubernetes.path

  role_name = "rustfs"

  bound_service_account_names = [
    "${local.tenant_name}-sa"
  ]

  bound_service_account_namespaces = [
    "rustfs"
  ]

  token_ttl = 3600

  token_policies = [
    vault_policy.rustfs.name
  ]
}

resource "vault_mount" "rustfs" {
  path = "rustfs"
  type = "transit"
}

resource "vault_transit_secret_backend_key" "rustfs" {
  backend = vault_mount.rustfs.path
  name    = local.tenant_name
  type    = "aes256-gcm96"

  deletion_allowed = false
}

resource "vault_policy" "rustfs" {
  name = local.tenant_name

  policy = <<-EOT
    path "transit/encrypt/${local.tenant_name}" {
      capabilities = ["update"]
    }

    path "transit/decrypt/${local.tenant_name}" {
      capabilities = ["update"]
    }

    path "transit/keys/${local.tenant_name}" {
      capabilities = ["read"]
    }

    path "${local.tenant_name}/metadata/rustfs/kms/keys" {
      capabilities = ["list", "read"]
    }

    path "${local.tenant_name}/metadata/rustfs/kms/keys/*" {
      capabilities = ["list", "read", "delete"]
    }

    path "${local.tenant_name}/data/rustfs/kms/keys/*" {
      capabilities = ["create", "update", "read", "delete", "list"]
    }
  EOT
}
```

``` yaml
apiVersion: rustfs.com/v1alpha1
kind: Tenant
metadata:
  name: example
  namespace: rustfs
spec:
  buckets:
  - name: example
  credsSecret:
    name: rustfs
  env:
    - name: RUSTFS_KMS_ENABLE
      value: "true"
    - name: RUSTFS_KMS_BACKEND
      value: "vault-transit"
    - name: RUSTFS_KMS_VAULT_ADDRESS
      value: "http://vault.vault.svc.cluster.local:8200"
    - name: RUSTFS_KMS_VAULT_MOUNT_PATH
      value: "transit"
  image: harbor.example.com/docker.io/rustfs/rustfs:1.0.0-beta.12
  mountPath: /data
  pools:
  - name: primary
    persistence:
      volumeClaimTemplate:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 64Gi
      volumesPerServer: 1
    servers: 1
```

The documentation only states the environment variable to specify the token (RUSTFS_KMS_VAULT_TOKEN). I want to use kubernetes auth backend, if possible. The console says: "Vault AppRole is not supported in the first version of this console.". But I want to do it via yaml definition of the tenant. It does not seem to be documented. I wonder if someone can help me (and update the documentation if possible).

Thanks in advance,
Davide

reatang:
## Related Issues

Fixes #5934.

## Summary of Changes

Two changes, both on the Vault KMS authentication path.

**1. The startup path ignored most Vault environment variables.**

`init_kms_system` built the Vault backend config field by field from the
command-line struct, hardcoding `VaultAuthMethod::Token` and failing with
"Vault token is required" when no token was set. `KmsConfig::from_env()`
already resolved AppRole and token-file auth, namespace, TLS and the KV/mount
settings — but it was never called outside tests, so on the
`RUSTFS_KMS_ENABLE=true` path all of those variables were silently dropped.
The AppRole and Vault Agent deployments documented in
`docs/operations/vault-kms-authentication.md` could not start at all.

The assembly moves into `vault_kv2_config_from_env` /
`vault_transit_config_from_env` in the KMS crate; `from_env()` and `init.rs`
both route through them, with the command line supplying only the values it
owns (address, token, mount path). One implementation now serves both entry
points, so they cannot drift apart again. The address stays required at the
startup path so a missing one is still named rather than falling back to the
crate's localhost default.

**2. Native Kubernetes authentication.**

`VaultAuthMethod::Kubernetes { role, mount, jwt_path, refresh_safety_window_secs }`
exchanges the pod's projected ServiceAccount token for a lease-bound Vault
token and renews it exactly as AppRole does. Selected with
`RUSTFS_KMS_VAULT_KUBERNETES_ROLE`, or through
`POST /rustfs/admin/v3/kms/configure`. This removes the Vault Agent sidecar
requirement on Kubernetes and leaves no credential to distribute or rotate.

Two deliberate decisions worth a reviewer's attention:

- The JWT is re-read on every login rather than cached, because the kubelet
  rotates a projected token well inside the pod's lifetime.
- Unlike `TokenFileSource`, the file mode is **not** checked. The kubelet owns
  the projected token and mounts it world-readable by default, so rejecting
  group/other bits would refuse every standard pod instead of catching a
  deployment error.

`VaultCliOverrides` deliberately does not derive `Debug` — it carries the raw
token, so denying the derive turns a future `{overrides:?}` into a compile
error rather than a leak.

Auth-method precedence is unchanged: a leftover `RUSTFS_KMS_VAULT_TOKEN`
alongside a configured login method stays tolerated and ignored (pinned by
`test_from_env_selects_approle_when_role_id_is_set`). Kubernetes is slotted
into the same shape — mutually exclusive with AppRole and with the token file,
tolerant of a stale token.

## Verification

- `make pre-commit`
- `cargo clippy -p rustfs-kms --all-targets -- -D warnings`
- `cargo test -p rustfs-kms` (550 unit tests + 15 behavior binaries, all green)
- `cargo test -p rustfs --lib init::tests`

Not run: the live-Vault lanes (`vault_approle_live`, `vault_ha_failover_live`)
and a real in-cluster Kubernetes login, which need infrastructure not available
here. The Kubernetes login exchange itself is covered only at the
token-source and configuration level.

New regression tests:

| Test | Pins |
| --- | --- |
| `build_vault_transit_kms_config_resolves_auth_and_mounts_from_env` | auth method, namespace and metadata mounts reach the startup path |
| `build_vault_kms_config_resolves_kv_mount_and_prefix_from_env` | the KV2 builder's key location (separate builder, separate risk) |
| `build_vault_transit_kms_config_refuses_skip_tls_verify_without_opt_in` | the newly honoured skip-verify flag fails closed |
| `test_skip_tls_verify_reaches_every_vault_client_generation` | the flag reaches `client.settings.verify` on both the authenticated and login clients |
| `test_vaultrs_skip_verify_env_cannot_override_the_configured_setting` | a stray `VAULT_SKIP_VERIFY` cannot disable verification |
| `build_vault_transit_kms_config_selects_kubernetes_auth` | Kubernetes selectable from the environment at startup |
| `build_vault_kms_config_refuses_two_auth_methods` | ambiguous identity refuses to start |
| `test_from_env_selects_kubernetes` / `..._is_mutually_exclusive_with_other_auth` | env resolution and exclusivity |
| `test_validate_rejects_bad_kubernetes_settings` | empty role / mount / token path |
| `test_kubernetes_login_rereads_and_trims_the_service_account_token` | rotation is picked up without a restart |
| `test_kubernetes_login_rejects_an_unusable_service_account_token` | missing/empty token fails the attempt without reaching Vault |
| `test_deserialize_vault_configure_request_accepts_kubernetes_auth` | admin API shape, and that unknown auth fields are rejected |
| `test_current_refuses_rather_than_panics_on_an_unrepresentable_safety_window` | an oversized `refresh_safety_window_secs` fails closed instead of panicking |
| `test_an_unrepresentable_lease_is_treated_as_non_expiring` | an oversized server-sent `lease_duration` does not panic either |

## Impact

**Environment variables that were previously ignored now take effect** when
`RUSTFS_KMS_ENABLE=true`. This is the point of the fix, but it changes the
effective configuration of any deployment that already sets them:

- `RUSTFS_KMS_VAULT_KV_MOUNT`, `RUSTFS_KMS_VAULT_KEY_PREFIX`,
  `RUSTFS_KMS_VAULT_TRANSIT_METADATA_KV_MOUNT`,
  `RUSTFS_KMS_VAULT_TRANSIT_METADATA_PREFIX` — RustFS will now read and write
  keys at the configured location instead of the hardcoded default. An
  operator who set these while running on this path has keys under the old
  default; after upgrading, those keys will appear absent. Either unset the
  variables or migrate the key material.
- `RUSTFS_KMS_VAULT_SKIP_TLS_VERIFY=true` — previously dropped twice over (it
  never reached the startup config, and the config never reached the HTTP
  client). It is now honoured end to end, and therefore rejected at startup
  without `RUSTFS_KMS_ALLOW_INSECURE_DEV_DEFAULTS=true`.
- `VAULT_SKIP_VERIFY` — vaultrs' own variable, which previously disabled Vault
  certificate verification without passing any RustFS gate. RustFS now sets
  `verify` explicitly, so this variable no longer has any effect. A deployment
  that (knowingly or not) relied on it to reach a self-signed Vault must switch
  to `RUSTFS_KMS_VAULT_SKIP_TLS_VERIFY` plus the insecure-defaults opt-in.
- `RUSTFS_KMS_VAULT_NAMESPACE` — now applied to Vault requests. Same
  keys-appear-absent hazard as the mount variables: a namespace that was
  set-and-ignored redirects every read to a different Vault namespace.
- Setting `RUSTFS_KMS_VAULT_TOKEN_FILE` together with `RUSTFS_KMS_VAULT_TOKEN`
  now fails at startup instead of using the token, matching the behaviour
  `from_env()` already had.

Both directions fail closed rather than silently downgrading, but a deployment
that relied on the broken behaviour will need its configuration corrected.

**Downgrade:** a cluster configured with Kubernetes auth and then rolled back
to a binary without this change cannot deserialize the persisted
`Kubernetes` auth variant; `load_kms_config()` returns `None` and the KMS
service stays unconfigured rather than starting with a wrong identity. Same
shape as when AppRole and TokenFile were added.

No on-disk or on-wire format change, no S3 API surface change, no change to
the default (static token) configuration.

## Additional Notes

Reviewed under the high-risk adversarial tier (KMS/auth). Three findings from
those passes are folded into this diff: the `VaultCliOverrides` `Debug` derive
over a raw token, and the two missing coverage areas (the KV2 builder's key
location, and the newly honoured skip-TLS-verify flag).

Review follow-up (e4781e763): @overtrue caught that my skip-verify test only
pinned the config struct — the flag still never reached the Vault HTTP client,
because `VaultConnectionSettings` dropped `VaultConfig::tls` and `build_client`
never called `verify`. Fixed and now covered at the client level. The remaining
`TlsConfig` fields (`ca_cert_path`, `client_cert_path`, `client_key_path`) are
still unused, but no supported input can set them — every constructor leaves
them `None` — so wiring them is left out of this diff.

Review follow-up (aa9f77f0f): @overtrue also caught an unbounded
`refresh_safety_window_secs` reaching `Instant::now() + safety_window`, which
panicked the request path after any lease-bearing login — reachable from the
admin configure API for every login-based auth method, so it predates this PR
for AppRole and TokenFile. Both operands of that comparison now use checked
arithmetic: an unrepresentable window collapses to "refuse" (fail closed), and
an unrepresentable lease TTL — built from the server's unvalidated
`lease_duration` u64 — collapses to "no expiry", matching the zero-lease tokens
Vault already issues. Fixing only the window would have left the same panic
reachable through the lease.

`docs/operations/vault-kms-authentication.md` gains a Kubernetes section, the
matching troubleshooting rows, and a note that both the startup and admin-API
paths now resolve auth identically.

### DIFF
--- a/crates/kms/src/api_types.rs
+++ b/crates/kms/src/api_types.rs
@@ -293,6 +293,15 @@ enum StrictVaultAuthMethod {
         #[serde(default)]
         refresh_safety_window_secs: Option<u64>,
     },
+    Kubernetes {
+        role: String,
+        #[serde(default)]
+        mount: Option<String>,
+        #[serde(default)]
+        jwt_path: Option<std::path::PathBuf>,
+        #[serde(default)]
+        refresh_safety_window_secs: Option<u64>,
+    },
     TokenFile {
         path: std::path::PathBuf,
         #[serde(default)]
@@ -317,6 +326,17 @@ impl From<StrictVaultAuthMethod> for VaultAuthMethod {
                 secret_id,
                 secret_id_file,
                 mount: mount.unwrap_or_else(|| crate::config::DEFAULT_VAULT_APPROLE_MOUNT.to_string()),
+                refresh_safety_window_secs,
+            },
+            StrictVaultAuthMethod::Kubernetes {
+                role,
+                mount,
+                jwt_path,
+                refresh_safety_window_secs,
+            } => Self::Kubernetes {
+                role,
+                mount: mount.unwrap_or_else(|| crate::config::DEFAULT_VAULT_KUBERNETES_MOUNT.to_string()),
+                jwt_path: jwt_path.unwrap_or_else(|| std::path::PathBuf::from(crate::config::DEFAULT_VAULT_KUBERNETES_JWT_PATH)),
                 refresh_safety_window_secs,
             },
             StrictVaultAuthMethod::TokenFile {
@@ -499,6 +519,7 @@ impl From<&KmsConfig> for KmsConfigSummary {
                 auth_method_type: match &vault_config.auth_method {
                     VaultAuthMethod::Token { .. } => "token".to_string(),
                     VaultAuthMethod::AppRole { .. } => "approle".to_string(),
+                    VaultAuthMethod::Kubernetes { .. } => "kubernetes".to_string(),
                     VaultAuthMethod::TokenFile { .. } => "token_file".to_string(),
                 },
                 has_stored_credentials: true,
@@ -513,6 +534,7 @@ impl From<&KmsConfig> for KmsConfigSummary {
                 auth_method_type: match &vault_config.auth_method {
                     VaultAuthMethod::Token { .. } => "token".to_string(),
                     VaultAuthMethod::AppRole { .. } => "approle".to_string(),
+                    VaultAuthMethod::Kubernetes { .. } => "kubernetes".to_string(),
                     VaultAuthMethod::TokenFile { .. } => "token_file".to_string(),
                 },
                 has_stored_credentials: true,
@@ -899,6 +921,42 @@ mod tests {
         });
         let request: ConfigureKmsRequest = serde_json::from_value(vault_opt_in_raw).expect("vault request should deserialize");
         assert!(request.to_kms_config().validate().is_ok());
+    }
+
+    /// The admin API reaches Kubernetes auth with the role alone; the mount and
+    /// the projected token path fall back to the cluster defaults, so a Tenant
+    /// manifest carries no credential and no cluster-specific paths.
+    #[test]
+    fn test_deserialize_vault_configure_request_accepts_kubernetes_auth() {
+        let raw = serde_json::json!({
+            "backend_type": "vault-transit",
+            "address": "https://vault.example.com:8200",
+            "mount_path": "rustfs",
+            "auth_method": { "Kubernetes": { "role": "rustfs" } }
+        });
+
+        let request: ConfigureKmsRequest = serde_json::from_value(raw).expect("kubernetes auth should deserialize");
+        let config = request.to_kms_config();
+        config.validate().expect("kubernetes auth must validate");
+
+        let vault = config.vault_transit_config().expect("vault transit backend config");
+        let VaultAuthMethod::Kubernetes {
+            role, mount, jwt_path, ..
+        } = &vault.auth_method
+        else {
+            panic!("expected Kubernetes auth, got {:?}", vault.auth_method);
+        };
+        assert_eq!(role, "rustfs");
+        assert_eq!(mount, crate::config::DEFAULT_VAULT_KUBERNETES_MOUNT);
+        assert_eq!(jwt_path, std::path::Path::new(crate::config::DEFAULT_VAULT_KUBERNETES_JWT_PATH));
+
+        let unknown_field = serde_json::json!({
+            "backend_type": "vault-transit",
+            "address": "https://vault.example.com:8200",
+            "auth_method": { "Kubernetes": { "role": "rustfs", "service_account": "rustfs" } }
+        });
+        serde_json::from_value::<ConfigureKmsRequest>(unknown_field)
+            .expect_err("an unknown auth field must be rejected rather than silently dropped");
     }
 
     #[test]
--- a/crates/kms/src/config.rs
+++ b/crates/kms/src/config.rs
@@ -25,6 +25,10 @@ use url::Url;
 
 pub const ENV_KMS_ALLOW_INSECURE_DEV_DEFAULTS: &str = "RUSTFS_KMS_ALLOW_INSECURE_DEV_DEFAULTS";
 pub const ENV_KMS_ALLOW_IMMEDIATE_DELETION: &str = "RUSTFS_KMS_ALLOW_IMMEDIATE_DELETION";
+pub const ENV_KMS_VAULT_ADDRESS: &str = "RUSTFS_KMS_VAULT_ADDRESS";
+pub const ENV_KMS_VAULT_TOKEN: &str = "RUSTFS_KMS_VAULT_TOKEN";
+pub const ENV_KMS_VAULT_NAMESPACE: &str = "RUSTFS_KMS_VAULT_NAMESPACE";
+pub const ENV_KMS_VAULT_MOUNT_PATH: &str = "RUSTFS_KMS_VAULT_MOUNT_PATH";
 pub const ENV_KMS_VAULT_SKIP_TLS_VERIFY: &str = "RUSTFS_KMS_VAULT_SKIP_TLS_VERIFY";
 pub const ENV_KMS_VAULT_TRANSIT_METADATA_KV_MOUNT: &str = "RUSTFS_KMS_VAULT_TRANSIT_METADATA_KV_MOUNT";
 pub const ENV_KMS_VAULT_TRANSIT_METADATA_PREFIX: &str = "RUSTFS_KMS_VAULT_TRANSIT_METADATA_PREFIX";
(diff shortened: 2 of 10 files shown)


crates/kms/src/config.rs
```diff
@@ -1201,34 +1222,138 @@ fn is_under_temp_dir(path: &Path) -> bool {
...
+    };
+
+    Ok(VaultConfig {
+        address: vault_address_from_env(overrides.address),
+        auth_method: vault_auth_method_from_env(overrides.token)?,
+        namespace: get_env_opt_str(ENV_KMS_VAULT_NAMESPACE),
+        mount_path,
+        kv_mount: get_env_str("RUSTFS_KMS_VAULT_KV_MOUNT", "secret"),
+        key_path_prefix: get_env_str("RUSTFS_KMS_VAULT_KEY_PREFIX", "rustfs/kms/keys"),
+        tls: vault_tls_config(get_env_bool(ENV_KMS_VAULT_SKIP_TLS_VERIFY, false)),
```
overtrue, line 1268: This flag still never reaches the Vault HTTP client. Both backend constructors drop config.tls when building VaultConnectionSettings, and its build_client leaves pinned vaultrs 0.8.0 at verify=true unless the unrelated VAULT_SKIP_VERIFY variable is set. With the two RustFS opt-ins and a self-signed HTTPS Vault, startup still fails with UnknownIssuer. Please thread skip_verify into every Vault client generation and cover the actual client setting or handshake.

overtrue (changes requested):

crates/kms/src/config.rs
reatang: Confirmed and fixed in e4781e763: `VaultConnectionSettings` now carries `skip_tls_verify`, both backend constructors pass `VaultConfig::tls` into it, and `build_client` sets `verify` explicitly on every generation (authenticated and login) — setting it unconditionally also stops a stray `VAULT_SKIP_VERIFY` from disabling verification behind the insecure-defaults gate; covered by `test_skip_tls_verify_reaches_every_vault_client_generation` (asserts `client.settings.verify` for both generations) and `test_vaultrs_skip_verify_env_cannot_override_the_configured_setting`, and the restore path pins verification on since `VaultRestoreTarget` has no TLS input.

crates/kms/src/api_types.rs
```diff
@@ -293,6 +293,15 @@ enum StrictVaultAuthMethod {
...
         refresh_safety_window_secs: Option<u64>,
     },
+    Kubernetes {
+        role: String,
+        #[serde(default)]
+        mount: Option<String>,
+        #[serde(default)]
+        jwt_path: Option<std::path::PathBuf>,
+        #[serde(default)]
+        refresh_safety_window_secs: Option<u64>,
```
overtrue, line 303: `refresh_safety_window_secs` is unbounded here. Supplying `u64::MAX` passes validation, but after a successful lease-bearing Kubernetes login the request path evaluates `Instant::now() + safety_window` and panics (`overflow when adding duration to instant`). Please reject out-of-range values or use checked arithmetic, with a focused boundary test.

overtrue (changes requested):
Please fix the overflow panic above.

crates/kms/src/api_types.rs
reatang: Fixed in aa9f77f0f: the comparison now uses `checked_add` and collapses an unrepresentable window to "refuse" — a window that wide means every token is always inside it, so that is the fail-closed answer the arithmetic was reaching for; I chose checked arithmetic over a validation bound because the value arrives from persisted config that may predate any bound. The gate moved into one `inside_safety_window` helper so `current()` and `record_credential_gauges()` cannot drift. I also fixed the other operand of the same comparison — `expires_at`/`renew_at` add a TTL built from the server's unvalidated `lease_duration` u64, so leaving it unchecked kept the identical panic reachable via the lease; it now collapses to `None`, matching the no-expiry case Vault already produces for zero-lease tokens. Boundary tests `test_current_refuses_rather_than_panics_on_an_unrepresentable_safety_window` and `test_an_unrepresentable_lease_is_treated_as_non_expiring` both reproduce the exact `overflow when adding duration to instant` panic when the guards are reverted. Shout if you consider the lease-side fix out of scope and I will split it out.

overtrue (approved):
LGTM

Issue #5938, opened 2026-08-10
Disk handles stuck in FaultyDisk state never recover without a process restart


stevapple:
**Describe the bug**

Over three weeks of operating a 4-node erasure-coded cluster (1.0.0-beta.10 and 1.0.0-beta.11) we have repeatedly seen a node degrade after restart activity somewhere in the cluster — its own restart, or a peer's:

- Requests touching certain erasure sets hang or fail on that node, while every other node serves the identical objects in milliseconds. In the worst cases the node is effectively wedged (its health endpoints stay green throughout — #5921).
- Other nodes' storageinfo/peer-health views sometimes report a **healthy** peer's disks as `offline` — in one incident, three of four nodes all reported the same peer's 12 disks offline while that peer was serving traffic normally.
- In one case the namespace scanner ran at zero effective progress for days.

Each time, restarting the affected node's process fully cured it, with no change to any disk or peer. We have hit this in at least five distinct incidents, so we are reporting it together with our investigation notes.

**What we observed — case A (scanner at zero for days)**

The scanner's counted rate dropped to 0 objects/min. Its log showed a continuous stream (2–3/second, for days) of failed metadata reads, all on one local drive:

```json
{"timestamp":"2026-08-07T13:00:00.007697068+08:00","level":"WARN","message":"Scanner folder failed to get object size","event":"scanner_folder_state","component":"scanner","subsystem":"folder","path":"/mnt/rustfs5/bucket-a/pool/chunks/320/320543/320543021_0_4096/xl.meta","failed_objects":1,"state":"get_size_failed","error":"Io error: scanner metadata transient: failed to read metadata: drive is faulty, bucket=bucket-a, object_path=pool/chunks/320/320543/320543021_0_4096/xl.meta","target":"rustfs::scanner::folder","filename":"crates/scanner/src/scanner_folder.rs","line_number":2065,"threadName":"rustfs-worker","threadId":"ThreadId(132)"}
```

We checked the drive directly while this was ongoing: the same `xl.meta` read perfectly from a shell, the XFS filesystem was mounted rw, the kernel log had zero I/O errors since boot, and the NVMe device was healthy. Restarting the RustFS process cleared the condition instantly and the same drive scanned fine afterwards.

**What we observed — case B (degraded node after restart churn)**

After two nodes restarted within about a minute of each other, the one that came up first started emitting `FaultyDisk` errors (78 within minutes in one occurrence):

```json
[...]

PR #6096, opened 2026-08-14
fix(ecstore): single-flight remote disk recovery

cxymds:
## Related Issues

Fixes backlog #1852. Tracks upstream rustfs/rustfs#5938 and root-cause rustfs/rustfs#5945.

## Summary of Changes

Adds a stable per-RemoteDisk handle identity to recovery spans and faulty short-circuit diagnostics, and makes recovery monitoring single-flight per handle. The monitor's active flag is released on success or cancellation and is re-armed when a new failure arrives during monitor teardown, preventing both duplicate probes and a permanently faulty handle after a race. DiskHealthTracker now publishes the runtime state and external faulty status as one atomic snapshot while retaining compatibility mirrors for existing metrics/tests; failure, recovery, reset, and peer-health mutations are serialized through one transition lock. Existing probe count, authentication, transport, quorum, and replacement protocols are unchanged.

## Verification

- `cargo test -p rustfs-ecstore --lib concurrent_failure_and_recovery_publish_one_health_snapshot -- --nocapture --test-threads=1`
- `cargo test -p rustfs-ecstore --lib faulty_handle_runs_only_one_recovery_monitor -- --nocapture --test-threads=1`
- `cargo test -p rustfs-ecstore --lib runtime_state_transitions -- --nocapture --test-threads=1`
- `cargo test -p rustfs-ecstore --lib test_remote_disk_recovery_requires_disk_rpc_readiness -- --nocapture --test-threads=1`
- `cargo test -p rustfs-ecstore --lib disk_health -- --test-threads=1` (12 passed)
- `cargo fmt --all --check`
- `git diff --check`

## Impact

No wire, on-disk, or public protocol changes. Recovery behavior remains three authenticated probes and the existing active-check/recovery intervals. Each RemoteDisk handle owns at most one recovery task; replacement handles get a distinct identity, and existing TLS/H2/authentication and quorum semantics are preserved. A four-node TLS/H2 peer restart and real metadata/object read canary remains required before release.

## Additional Notes

High-risk review covered correctness, simplicity, security, concurrency/durability, compatibility, performance, and test coverage. The added tests cover concurrent failure/recovery snapshot legality, duplicate monitor suppression, cancellation cleanup, and the existing authenticated probe readiness path. The release canary must additionally prove `peer down -> FaultyDisk -> three successful disk_info probes -> Online -> real metadata/object read` without a process restart, and verify old handles are not used after replacement.

---


crates/ecstore/src/cluster/rpc/remote_disk.rs
```diff
@@ -4382,6 +4441,45 @@ mod tests {
...
+        .await
+        .expect("remote disk should construct");
+        if !disk.health_check {
+            return;
+        }
+
+        disk.force_runtime_state_for_test(RuntimeDriveHealthState::Offline);
+        disk.spawn_recovery_monitor_if_needed();
+        disk.spawn_recovery_monitor_if_needed();
+        assert!(disk.recovery_monitor_is_active(), "only one recovery monitor should own the handle");
```
overtrue, line 4785: This assertion only observes the active flag, so it stays green if the failed compare-exchange path starts a second monitor. The linked backlog acceptance also requires the positive three authenticated probes to Online to real-read path and replacement-handle routing, but this PR only covers the negative readiness case. Please count monitor starts and add those required production-path regressions.

overtrue (changes requested):
Please complete the linked acceptance tests before merging.

crates/ecstore/src/cluster/rpc/remote_disk.rs
```diff
@@ -4397,6 +4690,152 @@ mod tests {
...
+
+                assert_eq!(
+                    peer.peer.disk_info_calls(),
+                    3,
+                    "RemoteDisk recovery requires the configured three successful disk_info probes"
+                );
+                let recovered_read = disk.read_all("bucket", "object").await.expect("recovered handle should read");
+                assert_eq!(recovered_read, Bytes::from_static(b"replacement-data"));
+
+                let replacement = RemoteDisk::new(
```
overtrue: This constructs and reads the replacement directly, so it stays green if SetDisks::renew_disk never publishes the new handle or fresh reads keep using the old slot. Please renew through the owning SetDisks, retain or poison the old handle, then fetch and read through the production slot.

overtrue (changes requested):

PR #6097, opened 2026-08-14
fix(site-replication): route every state RMW through the locked transaction

reatang:
## Related Issues

Closes rustfs/backlog#1796 (P1-15 PR2, batch B2 of rustfs/backlog#1675). Follow-up promised by #5882.

## Summary of Changes

PR1 (#5882) established the site-replication state transaction but left ~26 read-modify-write call sites on `config/site-replication/state.json` in the old shape: a transitional process-local mutex around `load` / mutate / `save`, with no distributed lock across the sequence — so two nodes of one site could still lose each other's updates, and the old shape stayed available to copy.

- **Every remaining RMW now runs inside `update_site_replication_state`** (distributed state-object write lock). Read-only sites call `load_site_replication_state()`, whose object read takes the object-level read lock. `SITE_REPLICATION_STATE_LOCK` and `site_replication_state_process_guard()` are removed, together with their architecture-guard allowlist entry and `global-state-inventory.md` row.
- **Multi-stage flows (add / edit / peer join / peer edit / remove / rotate) keep their `updated_at` + pending-id CAS, but the CAS runs inside the committing transaction**, against the state that transaction loaded. Peer probes, IAM work and fan-outs run between transactions holding no lock — the add no longer blocks every writer of the site across its peer join round trips. The add also re-checks the precondition right after the capability probes, before the IAM write and the join fan-out, so the common race is rejected before any remote side effect; a commit-time CAS failure tells the operator the peers may already be joined and that re-running the add reconverges.
- **The add adopts only the fields it computed** via an exhaustive destructure — adding a `SiteReplicationState` field becomes a compile error until it is classified; fields owned by writers that do not bump `updated_at` (retry events, peer-edit generations, resync progress, pending rotation/removal acks) keep their freshly loaded values.
- **Peer-edit delivery ordering rests on the PR1 generation fence**, not on a guard that could never order two nodes: the add's finalize fan-out now carries the generation allocated in its commit. An accepted peer join **preserves** the applied-generation high-water marks — join fan-outs are routine (adds and rotations both deliver `SRPeerJoin` to existing peers), so wiping them would let stalled older edits land after any join.
- **`SRRotateServiceAccountHandler` takes the lifecycle guard**: the background service-account reconciler repairs under a lifecycle try-acquire, and its pending-rotation precheck is only sound if a rotation cannot start mid-repair — an exclusion the removed process mutex used to provide as a side effect.
- **`update_site_replication_state_when_changed` adds persist-or-skip**: ack markers and pending-clearing paths stop rewriting the object on a miss. Load-bearing, because the shared persist helper clears the whole object for a ≤1-peer pending-free state — a "harmless" no-op persist would delete it.
- **`save_site_replication_state` is now `#[cfg(test)]`** — the pre-P1-15 shape can no longer be written in production code at all.

No on-disk format change.

### Adversarial review

Three independent adversarial passes (correctness/concurrency, security/compatibility, simplicity/test-coverage) ran over the final diff; confirmed findings were folded in: restored the add's post-probe re-check, kept the join from wiping fence marks, added the rotation lifecycle guard, replaced a carry-forward helper with the exhaustive destructure, made the CAS error messages name the actual stage, and added revert-detecting tests for persist-or-skip and rotation finalization. Response shapes of every peer-handler branch were byte-compared against main; admin authz order and on-disk/wire formats verified unchanged.

Known limitations (pre-existing, documented in code): a site removed while unreachable that later rejoins still misfences its stamped edits until the fence carries an epoch (follow-up task filed); `PendingEndpointRefresh` has no TTL, so a cancelled refresh wedges add/remove/rotate until the edit is re-run.

## Verification

- `cargo nextest run -p rustfs -E 'test(/admin::handlers::site_replication::/)'` — 181 passed (re-run after rebasing onto `main`).
- Site-replication dual/three-node e2e — 13 passed (`replication_extension_test::test_site_replication_*`, with a freshly built `target/debug/rustfs`).
- Mutation checks: dropping the state-object lock from the boundary reds the separate-node concurrency tests; flipping a persist-or-skip miss to a persist reds `test_missed_pending_clear_must_not_rewrite_the_state_object`.
- `cargo clippy -p rustfs --all-targets -- -D warnings`; `make pre-commit` (fmt + arch/layer/logging/doc guards + quick check) — all green.

## Impact

- Cross-node concurrent site-replication admin operations no longer lose updates; no S3 API surface change; no on-disk or wire format change (the add's finalize fan-out reuses the existing `?editOrigin=&editGeneration=` query params, which MinIO and older RustFS peers already ignore or handle).
- Two client-visible error behaviors are new, both replacing silent last-writer-wins overwrites: a concurrent topology change now fails `replicate add` / `replicate update` with an explicit retryable error instead of being overwritten.
- On a ≤1-peer pending-free state, all commits now uniformly clear the state object (previously only add/join/remove did); such a state is unconfigured (`enabled() == false`), and MinIO rejects edits on unconfigured sites.

## Additional Notes

The 30-lock-point map and call graph from the #5882 PR description locate every migrated call site. B3's P1-3 (retry-queue auto drain) builds on this transaction boundary.


rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -8776,23 +8888,32 @@ impl Operation for SRPeerJoinHandler {
...
-        persist_site_replication_state(&state).await?;
-        // Committed; release the state lock before the reverse-reachability
-        // probe and bucket back-fill — their transport helpers' retry-event
-        // bookkeeping re-enters the state transaction (P1-15).
-        drop(_state_guard);
+        // The join is applied to the state loaded inside the transaction, not
+        // to the snapshot the IAM work above ran against: only the freshly
+        // loaded state can decide whether a newer join beat this one here.
+        let incoming_updated_at = join_req.updated_at;
+        let committed = update_site_replication_state_when_changed(move |state| {
```
overtrue: Move the authoritative staleness check before changing IAM. Join A can load the old snapshot and pause while reading its body; join B then writes secret B and commits T2; when A resumes, it writes secret A and this transaction returns Superseded. The persisted T2 state now advertises B while IAM only accepts A, so peer requests fail. A gated-body handler test reproducing that ordering would catch this.

overtrue (changes requested):

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in 13a5e9d90: the join admission (fresh staleness check → IAM upsert → commit) now runs under the lifecycle guard, with the check taken against a guarded load BEFORE the IAM write; the gated-IAM test `test_peer_join_admission_serializes_iam_apply_against_a_newer_join` reproduces your ordering and is mutation-verified (removing the guard turns it red). Cross-node joins are arbitrated by the state-object lock, same reach as the old process mutex.

rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -8697,102 +8804,157 @@ impl Operation for SiteReplicationNetPerfHandler {
...
+    if join_request_is_superseded(&fresh, join_req.updated_at) {
+        let peer = fresh
+            .peers
+            .get(&fresh_local_peer.deployment_id)
+            .cloned()
+            .unwrap_or(fresh_local_peer);
+        return Ok(PeerJoinOutcome::Superseded(peer));
+    }
+
+    apply_iam(join_req.clone()).await?;
```
overtrue: This still serializes joins only within one process. Node A can pass the check for older T1, node B can write IAM B and commit newer T2, then A can write IAM A and have its state commit return Superseded; state advertises T2 while IAM accepts A. The distributed state lock is acquired only after this write. Please add cross-node fencing or reconverge IAM on supersession, plus a separate-node regression test.

overtrue (changes requested):
Please close the cross-node admission race above.

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in 937b4df24: the admission (staleness check → IAM upsert → commit) now also runs under a distributed join-admission lock (namespace-lock key, repair-execution-lock pattern, lease expiry from the lock subsystem), with the check taken against a load inside the lock before IAM changes anything — a superseded join exits without touching IAM. `test_peer_join_admission_serializes_across_separate_nodes` drives `admit_peer_join_across_nodes` (the admission minus the process-local guard, i.e. what a second node runs) with A gated mid-IAM and is mutation-verified: removing the admission lock turns it red. Remaining window: old-version nodes during a rolling upgrade do not take the lock and keep today's behavior until the upgrade completes.

overtrue (approved):

PR #6099, opened 2026-08-14
fix(scanner): add context to corrupt metadata logs

marshawcoco:
## Related Issues

Related to #5950.

## Summary of Changes

- Stop emitting a generic XL2 parser error before the typed corruption error reaches scanner context.
- Remove the duplicate intermediate scanner error and emit one scanner-owned `scanner_metadata_corrupt` event at the existing first-16/every-1024 failed-object sampling point.
- Include drive, bucket, object, metadata path, failure count, and scanner state in the contextual event while preserving corruption classification and heal admission.

## Verification

- `cargo test -p rustfs-scanner --lib corrupt -- --nocapture --test-threads=1`
- `cargo fmt --all --check`

## Impact

Corrupt XL2 metadata produces one actionable, sampled scanner error instead of duplicate context-free errors. Metadata formats, read behavior, heal classification, and repair behavior are unchanged.

## Additional Notes

This change improves diagnosis only. It does not delete, rewrite, or otherwise repair corrupt `xl.meta` files.


overtrue (approved):
LGTM

PR #6104, opened 2026-08-14
perf(ecstore): borrow rename metadata during commit fanout

houseme:
## Related Issues
rustfs/backlog#1792

## Summary of Changes
This PR reduces PUT commit fanout metadata cloning in `SetDisks::rename_data` by adding an owned commit path that keeps the per-disk `FileInfo` vector inside the cancellation-resistant coordinator and borrows each entry for the disk rename call. The existing slice-based `rename_data` entry remains as a compatibility wrapper for older call shapes and focused tests.

The disk `rename_data` API now accepts `&FileInfo`; local disk clones only when it must mutate metadata for the durable local commit, while remote disk can encode the borrowed `FileInfo` directly. Ordinary PUT and CompleteMultipartUpload assign per-slot erasure indexes before the owned rename call and consume the returned committed `FileInfo` for response metadata instead of keeping the full vector alive past commit.

The wire payload, `xl.meta` format, bitrot framing, quorum decisions, rollback, fsync ordering, and RPC request/response schema are unchanged.

## Verification
- `cargo test -p rustfs-ecstore --lib set_disk::core::io_primitives::tests::rename_data -- --nocapture`
- `cargo test -p rustfs-ecstore --lib inline_put_direct_commit -- --nocapture`
- `cargo fmt --all --check`
- `git diff --check`

## Impact
No user-facing API, configuration, deployment, or on-disk compatibility change is expected. The intended impact is lower per-object allocation/clone pressure on PUT and multipart commit fanout, especially when the coordinator targets remote disks.

High-risk validation summary: correctness attacked quorum-1 rollback, delete marker metadata, inline exact quorum and post-encode rename failures with focused tests; concurrency/durability attacked cancellation-resistant coordinator ownership, rollback and post-commit cleanup ordering with no break found; compatibility attacked RPC wire shape, `RenameDataResp`, `FileInfo` encoding, `xl.meta`, and bitrot layout with no format change found; security attacked secret/log/auth/path surfaces and found no new untrusted input or secret exposure; performance attacked per-disk `FileInfo` cloning and confirmed the remote fanout clone is removed while local durable commit still keeps its required local mutable copy; test coverage maps the changed path to existing rename_data and inline PUT commit tests.

## Additional Notes
This does not claim an end-to-end Warp throughput improvement by itself; it is a narrow hot-path allocation reduction. Fleet validation should still use the same allocator, stage-metrics gate, and ABBA discipline as the ongoing small PUT investigation.

### DIFF
--- a/crates/ecstore/src/disk/mod.rs
+++ b/crates/ecstore/src/disk/mod.rs
@@ -430,7 +430,7 @@ impl DiskAPI for Disk {
         &self,
         src_volume: &str,
         src_path: &str,
-        fi: FileInfo,
+        fi: &FileInfo,
         dst_volume: &str,
         dst_path: &str,
     ) -> Result<RenameDataResp> {
@@ -664,6 +664,20 @@ impl DiskAPI for Disk {
             Disk::Local(local_disk) => local_disk.read_metadata(volume, path).await,
             Disk::Remote(remote_disk) => remote_disk.read_metadata(volume, path).await,
         }
+    }
+}
+
+#[cfg(test)]
+impl Disk {
+    async fn rename_data(
+        &self,
+        src_volume: &str,
+        src_path: &str,
+        fi: FileInfo,
+        dst_volume: &str,
+        dst_path: &str,
+    ) -> Result<RenameDataResp> {
+        <Self as DiskAPI>::rename_data(self, src_volume, src_path, &fi, dst_volume, dst_path).await
     }
 }
 
@@ -904,7 +918,7 @@ pub trait DiskAPI: Debug + Send + Sync + 'static {
         &self,
         src_volume: &str,
         src_path: &str,
-        file_info: FileInfo,
+        file_info: &FileInfo,
         dst_volume: &str,
         dst_path: &str,
     ) -> Result<RenameDataResp>;
--- a/crates/ecstore/src/disk/mod.rs
+++ b/crates/ecstore/src/disk/mod.rs
@@ -55,6 +55,7 @@ pub fn part_transaction_path(part_path: &str) -> String {
 
 use crate::cluster::rpc::RemoteDisk;
 use crate::cluster::rpc::build_internode_data_transport_from_env;
+use crate::disk::disk_store::DiskStoreRenameDataExt;
 use crate::disk::disk_store::LocalDiskWrapper;
 use crate::disk::health_state::RuntimeDriveHealthState;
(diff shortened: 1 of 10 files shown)


crates/ecstore/src/disk/mod.rs
```diff
@@ -904,7 +918,7 @@ pub trait DiskAPI: Debug + Send + Sync + 'static {
         &self,
         src_volume: &str,
         src_path: &str,
-        file_info: FileInfo,
+        file_info: &FileInfo,
```
overtrue: `DiskAPI` is publicly re-exported through `rustfs_ecstore::api::disk`, which is documented as a compatibility surface. This signature change makes existing downstream calls and implementations fail to compile. Please keep the owned public method and add an internal borrowed fast path for built-in disks.

crates/ecstore/src/set_disk/ops/object.rs
```diff
@@ -2236,11 +2237,13 @@ impl SetDisks {
                     return Err(err);
                 }
 
-                let rename_result = SetDisks::rename_data(
+                Self::assign_rename_data_indexes(&mut parts_metadatas);
+                let fallback_file_info = parts_metadatas.get(response_metadata_slot).cloned();
```
overtrue: This eager clone makes an all-local PUT perform N local `FileInfo` clones plus this fallback clone, versus N clones before this PR. A successful quorum always yields a committed slot below, so please remove the success-path fallback clone and make that invariant explicit.

overtrue (changes requested):

crates/ecstore/src/disk/mod.rs
```diff
@@ -667,6 +667,20 @@ impl DiskAPI for Disk {
     }
 }
 
+#[cfg(test)]
+impl Disk {
+    async fn rename_data(
+        &self,
+        src_volume: &str,
+        src_path: &str,
+        fi: FileInfo,
```
overtrue: Under `cfg(test)`, this same-name inherent method shadows the borrowed trait method. Normal crate code calls `self.disk.rename_data(..., fi, ...)` with `&FileInfo`, so unit-test compilation resolves here and fails with E0308 instead of falling back to the trait. Please use a differently named borrowed path and remove this shim.

overtrue (changes requested):

overtrue (approved):
LGTM

zhi22915 (approved):

PR #6119, opened 2026-08-14
fix(site-replication): lift a rejoined site's restarted edit counter over stale marks

reatang:
## Related Issues
N/A (follow-up to the peer-edit ordering fence from the P1-15 series, most recently #6097; closes the unilateral-removal rejoin gap called out in `apply_peer_join`'s doc comment).

## Summary of Changes
A site removed while it was unreachable never leaves the receiver's peer map, so `parse_site_replication_state`'s load-time mark pruning never fires. When that site later rejoins, its recreated state object restarts `edit_generation` at zero, and the receiver's surviving high-water mark silently fences out (acks without applying) every stamped delivery from it — peer edits and the add finalize fan-out alike — until the restarted counter catches up.

This PR allocates the generation as a hybrid logical clock instead: `max(wall clock in unix nanoseconds, previous + 1)`, still inside the state transaction under the distributed state-object lock. Every value a state-object lifetime hands out is capped by the wall clock at its own allocation, so a recreated lifetime's first allocation exceeds them all and clears the stale mark, while a pre-removal delivery still in flight stays below the new floor and remains correctly fenced. The `previous + 1` half keeps allocations strictly increasing across same-tick allocations and mid-lifetime clock regressions.

Nothing changes on the wire (`editGeneration` stays the single fence param) or in the persisted schema (`edit_generation` stays the single counter field): pre-hybrid receivers get the fix as soon as the sender upgrades, old binaries preserve the field across rolling up/downgrades, and marks recorded by plain-counter receivers are cleared by any wall-clock allocation. A clock that regresses across a delete/recreate degrades to a fence that self-heals once real time passes the previous lifetime's last allocation, and introduces no rollback window beyond what the plain counter already had.

An epoch-based design (additive `editEpoch` wire param plus per-origin epoch marks) was fully built first and rejected under adversarial review: old binaries rewriting the state object drop the unknown epoch fields, which both disarms the fix mid-rolling-upgrade and — because epoch adoption lowers the generation mark — reopens the pre-restart rollback the fence exists to prevent; a backwards clock also fences an origin permanently instead of self-healing. The hybrid clock has none of these modes and is an order of magnitude less machinery.

**Release-ordering note:** the plain-counter fence exists only on unreleased `main` (since the fence landed; no tag contains it). No tag should be cut from that range without this fix: a site that runs the plain-counter fence, downgrades there after running the hybrid clock, and then recreates its state object would be near-permanently fenced by nanosecond-scale marks. Published releases (<= 1.0.0-rc.1) predate the fence entirely and are unaffected in every mixed-version direction.

## Verification
- `cargo test -p rustfs --lib -- site_replication` — 207 tests pass, including four new regression tests: the unilateral-removal rejoin unit test, the plain-counter-receiver mark test, the strict-monotonicity-under-clock-stall test, and an end-to-end state-object lifecycle test (clear at <= 1 peers -> recreate -> restarted counter clears the old mark, straggler stays fenced).
- Mutation checks run for real: dropping the wall-clock floor reds 3 tests; dropping the `previous + 1` half reds the clock-stall test (deterministically — the test pins the counter above any wall-clock value).
- `make pre-pr` (fmt + arch checks + clippy + full test suite).
- Adversarial validation (high-risk tier), one-line verdicts:
  - Correctness: attacked the wall-clock-cap induction, cross-lifetime equal-value collisions, u64 extremes, and the deletion-avoidance conjunction — no regression; the only note was doc phrasing ("never a rollback" weakened to "no new rollback window"), applied.
  - Concurrency/durability: attacked racing first allocations under the distributed lock, cross-node skew, crash windows, and the pending-op deletion exception — null report; counter continuation is monotone-safe in every interleaving.
  - Compatibility: attacked every mixed-version direction (old/new sender nodes, pre-fence releases, MinIO peers, retry-event persistence, JS-float exposure) — no break; produced the release-ordering note above.
  - Security: no authz, wire, serde, or logging surface touched; the pre-existing spoofing surface (authenticated peer stamping a huge generation) is unchanged — null report.
  - Test coverage: every commit-message claim maps to a red-on-revert assertion; all clock-sensitive assertions are equal-tolerant, strict assertions are clock-independent — near-null (two acceptable doc-level gaps where the wall clock has no injection seam).
  - Simplicity: this design IS the reviewer-proposed minimal alternative to the epoch scheme; branch necessity and replacement completeness verified.
  - Performance: change is confined to the admin site-replication control plane (operator-frequency allocations); no data-path impact.

## Impact
- No wire or on-disk format change; `editGeneration` values become nanosecond-scale (u64, exact in serde_json; not exposed to any JS-float-limited consumer — verified repo-wide).
- Mixed-version rolling upgrades and downgrades against published releases are unaffected; see the release-ordering note for the unreleased plain-counter range.
- A malicious authenticated peer could already permanently poison a victim origin's mark with `editGeneration=u64::MAX` before this change; this PR neither widens nor narrows that pre-existing trust surface (follow-up hardening tracked separately: bind `editOrigin` to the authenticated peer identity).

## Additional Notes
The first implementation of this fix (epoch design) is preserved in this branch's reflog for reference. The adversarial review that rejected it is summarized in the commit message.

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -1067,9 +1067,13 @@ fn parse_site_replication_state(data: &[u8]) -> S3Result<SiteReplicationState> {
     state.peers = normalize_peer_map_by_identity(state.peers);
     // A peer-edit high-water mark only fences a CURRENT peer. A site that
     // leaves drops below two peers, which clears its own state object and
-    // restarts its generation counter at zero — a mark left over from the
-    // previous membership would then reject every edit it sends after it
-    // rejoins. Dropping departed origins on load also keeps the map bounded.
+    // restarts its generation counter — a mark left over from the previous
+    // membership must not reject the edits it sends after it rejoins. This
+    // pruning covers departures THIS site observed; an origin removed
+    // unilaterally elsewhere stays in this peer map with its mark, and the
+    // wall-clock floor in `next_peer_edit_generation` is what lifts its
+    // restarted counter over that mark. Dropping departed origins on load
+    // also keeps the map bounded.
     state
         .applied_edit_generations
         .retain(|origin, _| state.peers.contains_key(origin));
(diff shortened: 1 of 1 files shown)


overtrue (approved):
LGTM

PR #6123, opened 2026-08-15
fix(site-replication): admit only verifiable peer-edit fences

reatang:
## Related Issues

N/A (hardening follow-up to the peer-edit ordering fence; builds on #6119)

## Summary of Changes

The peer-edit ordering fence trusted its own wire input: `SRPeerEditHandler` accepted whatever `editOrigin` / `editGeneration` an authenticated caller stamped on the delivery. Every site authenticates peer traffic with the **shared** `site-replicator-0` service account, so the receiver cannot tell which site actually sent a request — a compromised peer could stamp `editOrigin=<victim-deployment-id>` with `editGeneration=u64::MAX` and permanently poison the receiver's `applied_edit_generations` high-water mark for that victim: every genuine delivery from the victim would then be silently acked-and-dropped until the victim left the receiver's peer map.

Because the shared credential makes strict sender-identity binding impossible, the fence is instead held to what the receiving state can vouch for, checked inside the same transaction that applies the delivery (`peer_edit_fence_is_admissible`):

- **Membership**: the claimed origin must be a site the receiver currently replicates with (`state.peers`, the exact rule the load-time mark pruning applies, so every recorded mark is one a reload would keep) and must not be the receiver itself, which never delivers edits to itself.
- **Generation ceiling**: the generation must not exceed the receiver's wall clock (reusing `edit_generation_wall_clock()` from #6119) plus a 24h cross-site skew allowance. Genuine generations are wall-clock-floored hybrid values since #6119, so a forged near-ceiling mark now decays within the allowance instead of lasting forever.

An inadmissible fence is **ignored, not fatal**: the delivery applies exactly as an unstamped (pre-fence) delivery would, no high-water mark is read or written, and a structured `warn!` records the rejection reason. Both legitimate fence senders (the edit fan-out and the add finalize fan-out) stamp their own deployment id, which the receiver keys its peer map by (same `current_deployment_id()` source at join time), so no legitimate flow changes behavior.

## Verification

- `make pre-pr` — pass (full gate: fmt, arch checks, clippy `-D warnings`, workspace tests)
- `cargo test -p rustfs --lib site_replication` — 209 passed (re-run after rebasing onto #6119 and applying review fixes)
- `cargo clippy -p rustfs --lib --tests` — 0 warnings (post-rebase)
- `cargo fmt --all --check`, `./scripts/check_logging_guardrails.sh` — pass (post-rebase)
- Mutation spot-checks (applied, observed red, reverted): `>` → `>=` on the ceiling comparison fails `peer_edit_fence_ceiling_admits_counter_and_clock_seeded_generations`; skew allowance → `0` fails the same test's ahead-of-clock case; dropping the `contains_key` membership clause or inverting the self-origin clause fails `forged_peer_edit_fences_cannot_poison_the_high_water_marks`.

Post-rebase re-verification was scoped to the touched file's package (tests + clippy + fmt + guardrails) because the later edits were confined to the same handler file already covered by the full gate.

## Impact

- No wire or persisted-schema change: same query params, same JSON bodies, same responses, no new state fields. Unstamped deliveries (older peers, MinIO) behave exactly as before and log nothing.
- A sender whose clock is more than 24h ahead of the receiver loses fence protection for its deliveries (they still apply, unstamped) and the receiver logs `fence_generation_beyond_ceiling` — observable, self-healing misconfiguration.
- Residual risk (documented in code): a peer with the shared credential can still stamp a *current* peer's origin with a generation up to the ceiling. Since #6119, genuine generations are wall-clock-floored, so such a mark self-heals within 24h. Only an origin still running a pre-#6119 plain-counter binary cannot outrun a near-ceiling mark until it upgrades.

## Additional Notes

High-risk adversarial validation (all seven roles, per AGENTS.md):

- **Correctness**: null — add-finalize/edit fan-out, retry replays, endpoint-refresh, unstamped deliveries, and boundary values (0, ==ceiling, empty origin, pre-epoch clock) all traced; no legitimate flow changes outcome.
- **Simplicity**: one finding (Option-in/out helper → bool predicate + `Option::filter`) — applied; downstream fence use-sites reverted to their pre-diff shape.
- **Security**: one finding (original doc overstated the day-bound decay under pre-#6119 plain counters) — resolved by rebasing onto #6119 and rewording the doc to name the pre-hybrid caveat; warn-flood concern rebutted (requires valid credentials; the handler's own locked state RMW dwarfs one log line).
- **Concurrency/durability**: null — admission, stale-check, apply, and mark-write share one locked transaction (closure is `FnOnce`, runs once); ceiling computed pre-lock only lags, never wrongly admits; retry queue never replays fenced URLs.
- **Compatibility**: null — origin string and peer-map key share one source (`current_deployment_id()` at join); membership rule is byte-identical to load-time pruning; old→new, new→old, and MinIO unstamped paths unchanged.
- **Performance**: null — admin-plane only; one clock read outside the lock, O(log peers) lookup inside; warn fires only on forged/misconfigured input.
- **Test-coverage**: two real gaps found by live mutation (ceiling boundary and skew allowance unconstrained) — fixed with the `==ceiling` and ahead-of-clock assertions; both mutants re-run and confirmed killed. Handler wiring is pinned by a source-anchor assertion, consistent with the file's existing P1-15 anchors.

Advisory for a possible follow-up (out of scope here): with #6119's HLC, a single far-future clock excursion pins a site's generation there permanently (`max(nanos, prev+1)` never decreases); once it is more than 24h past real time, that site's fences are ignored (warn-only) until wall clock catches up. Clamping the HLC seed would remove that mode.


rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -9576,6 +9634,11 @@ impl Operation for SRPeerEditHandler {
         let outcome = update_site_replication_state_when_changed(move |state| {
             let mut incoming = incoming;
             let local_peer = local_peer_at_endpoint(commit_endpoint, state);
+            // The fence is self-reported — the shared service account means
+            // the sender cannot be identified — so it is honoured only after
+            // the admissibility check, against the same state it will gate.
+            let commit_fence = commit_fence
+                .filter(|fence| peer_edit_fence_is_admissible(state, &local_peer.deployment_id, fence, fence_generation_ceiling));
```
overtrue: Filtering a genuine HLC above the receiver ceiling to None drops its ordering fence. After a temporary >24h sender clock jump, #6119 persists that future HLC and continues with previous + 1 after the clock is corrected. Two later edits can arrive as g+1 then g; both skip the stale check here, so g commits last and rolls back g+1. Please preserve ordering for this persisted state and add the reverse-delivery regression.

overtrue (changes requested):

rustfs/src/admin/handlers/site_replication.rs
reatang: Confirmed — the ceiling regressed exactly this case: once `max(now, prev + 1)` persists an excursion, every genuine fence from that origin exceeds `now + 24h` forever, the filter downgrades them all to unstamped, and the reverse delivery of a later race (g after g+1) commits last. Thanks for the concrete interleaving.

Reworked to bound the fence by **distance** instead of by an absolute clock ceiling:

- Admissibility keeps only the membership half (origin must be a current remote peer, mirroring the load-time mark pruning); the generation is no longer clamped, so an excursion-regime origin records real marks and its races stay ordered wherever its clock operates.
- `peer_edit_delivery_is_stale` now fences only when `mark > generation && mark - generation <= 24h`. A genuine superseded delivery trails its origin's mark by delivery latency at most — retransmissions re-run the sender flow and mint a fresh generation, the retry queue keys on the bare path and never replays a fenced URL, and the new reconcile-tick drain (#6131) likewise allocates a fresh generation per drained peer-edit rather than resending the stored path — so the window keeps every real race ordered (including in the persisted-excursion regime, where g trails g+1 by nanoseconds). A mark further above than the window cannot be explained by any genuine race (one documented exception: a plain-counter straggler landing after its origin's first hybrid-clock edit applies unfenced, once, at upgrade time); fencing on distant marks would silently drop the origin's real edits, so they are ignored (warn-logged) and the origin merely runs unfenced until its generations catch up.

This also strengthens the original poisoning fix: a forged `u64::MAX` mark is now inert immediately (the victim's deliveries keep applying) instead of the previous ceiling design's 24h drop window, and a hybrid-clock → plain-counter downgrade is no longer permanently fenced by leftover nanosecond marks. Residual, documented: an attacker with the shared service-account credential can still plant a mark within `victim_generation + 24h` (bounded 24h fencing, same as the ceiling design) or a distant mark that downgrades the victim to pre-fence unfenced delivery — with the shared credential it can already deliver arbitrary edit content directly, so neither adds capability beyond baseline.

Regressions added in `peer_edit_fence_orders_a_persisted_future_clock_and_defuses_distant_marks`: the reverse delivery (g fenced after g+1 landed, ten years ahead of the receiver's clock), equal/next generation still applying, the exact window boundary on both sides, and the distant-mark/downgrade cases; `forged_peer_edit_fences_cannot_poison_the_high_water_marks` now proves a planted `u64::MAX` mark cannot silence a genuine hybrid-clock delivery.

overtrue (approved):

PR #6126, opened 2026-08-15
fix(admin): stream madmin ReplicationMRF documents from /v3/replication/mrf

reatang:
## Summary

Part of the MinIO replication compatibility review (rustfs/backlog#1675, **P1-13 mrf half**; the diff half shipped in #5799 and this mirrors its pattern).

madmin's `BucketReplicationMRF` decodes the response body one `ReplicationMRF` document at a time. The aggregate envelope RustFS returned decoded as exactly one phantom row (`"Bucket"` case-insensitively fills `ReplicationMRF.Bucket`, everything else is an unknown key), so `mc replicate backlog` printed a single row with an empty object instead of the backlog.

- Default response is now a bare stream of `ReplicationMRF` documents (exact madmin json tags; `Size`/`TargetARNs` as ignored extension keys) built from the durable backlog ledger; an empty backlog renders an empty body (zero rows, no phantom).
- The aggregate counter envelope moves behind `?aggregate=true` (RustFS extension) and now truthfully advertises `PerObjectEntriesAvailable` whenever the ledger is readable.
- An unreadable ledger is signalled out-of-band via `x-rustfs-replication-mrf-backlog-unavailable` (mirrors the diff truncation header) plus a warn event.
- The madmin `node` parameter is accepted but documented as a no-op: the ledger is cluster-shared with no per-node attribution.
- Delete-marker purge entries fall back to the marker version id so those rows keep a version identity.

## Testing

Red→green TDD (red commit pins the stream contract and fails against the envelope):
- unit: bare stream shape / empty body / aggregate envelope + 3 existing aggregate tests updated
- `make pre-commit`, clippy clean; no e2e or console consumer depends on the old envelope (grepped)

### DIFF
--- a/rustfs/src/admin/handlers/replication.rs
+++ b/rustfs/src/admin/handlers/replication.rs
@@ -1304,7 +1304,9 @@ fn mrf_entry_documents(
 ///
 /// `aggregate=true` (RustFS extension) keeps the enveloped counter shape;
 /// backlog-source health (`RuntimeStatsAvailable`/`DurableBacklogAvailable`)
-/// is only representable there and is signalled out-of-band for the stream.
+/// is only representable there — an unreadable ledger fails the stream
+/// request outright in the handler (madmin only decodes the body of a 200,
+/// so an empty stream would read as a healthy zero-row backlog).
 fn render_mrf_backlog(
     response: &MrfResponse,
     durable: &crate::admin::storage_api::replication::DurableMrfBacklog,
@@ -1381,24 +1383,24 @@ impl Operation for ReplicationMrfHandler {
         let response = build_mrf_response(bucket, &bucket_stats, &durable);
 
         if !durable.available && !aggregate {
-            // The madmin stream has no envelope to carry source health; an
-            // unreadable backlog would otherwise be indistinguishable from an
-            // empty (healthy) one.
+            // The madmin stream has no envelope to carry source health, and
+            // madmin only decodes the body of a 200 — an empty stream would
+            // read as a clean, healthy zero-row backlog. Fail loudly instead;
+            // aggregate mode still reports the availability fields.
             tracing::warn!(
                 bucket = %response.bucket,
-                "durable MRF backlog is unreadable; stream response is empty — use aggregate=true to see source health"
+                "durable MRF backlog is unreadable; failing the stream request — use aggregate=true to see source health"
             );
+            return Err(S3Error::with_message(
+                S3ErrorCode::ServiceUnavailable,
+                "durable MRF backlog is unreadable; retry, or use aggregate=true for source health".to_string(),
+            ));
         }
 
         let data = render_mrf_backlog(&response, &durable, aggregate)
             .map_err(|e| S3Error::with_message(S3ErrorCode::InternalError, format!("serialize failed: {e}")))?;
         let mut headers = HeaderMap::new();
         headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));
-        // Out-of-band source-health signal for the bare stream (madmin/mc
-        // ignore unknown headers), mirroring x-rustfs-replication-diff-truncated.
-        if !durable.available {
-            headers.insert("x-rustfs-replication-mrf-backlog-unavailable", HeaderValue::from_static("true"));
-        }
         Ok(S3Response::with_headers((StatusCode::OK, Body::from(data)), headers))
     }
 }
--- a/rustfs/src/admin/handlers/replication.rs
+++ b/rustfs/src/admin/handlers/replication.rs
@@ -1266,8 +1266,8 @@ struct MrfEntryDocument {
 /// Upper bound on the number of documents one stream response emits. The
 /// durable ledger is not bounded by the in-memory pending cap (recovery can
 /// persist far larger generations), and the body is buffered before send, so
-/// an unbounded read could stage hundreds of MB per request. A truncated
-/// stream is signalled via `x-rustfs-replication-mrf-truncated`.
+/// an unbounded read could stage hundreds of MB per request. The handler
+/// rejects a response beyond this bound instead of returning a partial 200.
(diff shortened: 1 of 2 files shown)


rustfs/src/admin/handlers/replication.rs
```diff
@@ -1275,14 +1368,37 @@ impl Operation for ReplicationMrfHandler {
...
+        }
+
         let durable = crate::admin::storage_api::replication::read_durable_mrf_backlog(store).await;
         let bucket_stats = cluster_replication_stats(&bucket, app_context_from_req(&req)).await;
-        let response = build_mrf_response(bucket, &bucket_stats, durable);
+        let aggregate = queries.get("aggregate").map(String::as_str) == Some("true");
+        let response = build_mrf_response(bucket, &bucket_stats, &durable);
 
-        let data = serde_json::to_vec(&response)
+        if !durable.available && !aggregate {
```
overtrue, line 1434: madmin ignores this header. On a 200 response it only decodes the body, so an unreadable ledger becomes clean EOF and clients report zero rows. Please return a non-2xx error in stream mode; aggregate mode can keep the availability field.

overtrue (changes requested):

rustfs/src/admin/handlers/replication.rs
reatang: Fixed in 98a96776c: stream mode now returns 503 (`ServiceUnavailable`) when the durable ledger is unreadable, instead of an empty 200 + header; aggregate mode keeps the `RuntimeStatsAvailable`/`DurableBacklogAvailable` fields. The out-of-band header is removed since no client could act on it.

rustfs/src/admin/handlers/replication.rs
```diff
@@ -1235,15 +1238,107 @@ fn build_mrf_response(
...
+#[derive(Debug, Serialize)]
+struct MrfEntryDocument {
+    /// The durable backlog is a cluster-shared ledger with no per-node
+    /// attribution, so the madmin `nodeName` tag is always empty.
+    #[serde(rename = "nodeName")]
+    node_name: String,
+    #[serde(rename = "bucket")]
+    bucket: String,
+    #[serde(rename = "object")]
+    object: String,
```
overtrue, line 1255: This stream now exposes object names and version IDs, but the handler still requires only admin:GetReplicationMetrics. RustFS treats that and admin:ReplicationDiff as separate exact actions, skips resource matching for this admin action, and performs no later bucket authorization, so a metrics-only IAM principal can enumerate keys in any replicated bucket it can name. MinIO gates this endpoint with ReplicationDiff; please require that action for stream mode (aggregate metrics can keep GetReplicationMetrics).

rustfs/src/admin/handlers/replication.rs
```diff
@@ -1235,15 +1238,107 @@ fn build_mrf_response(
...
+        .map(|entry| MrfEntryDocument {
+            node_name: String::new(),
+            bucket: entry.bucket.clone(),
+            object: entry.object.clone(),
+            // Delete-marker purge entries track the marker version separately;
+            // fall back to it so those rows still carry a version identity.
+            version_id: entry
+                .version_id
+                .or(entry.delete_marker_version_id)
+                .map(|v| v.to_string())
```
overtrue: Uuid::nil() is RustFS's in-memory null-version sentinel, but this serializes it as 00000000-0000-0000-0000-000000000000 instead of the S3 wire token null. A pre-existing unversioned object becomes Some(nil) when scanned after versioning and existing-object replication are enabled, and queue backpressure can persist that ID into this ledger. A focused stream test fails on this head; please map nil to NULL_VERSION_ID and cover it.

rustfs/src/admin/handlers/replication.rs
```diff
@@ -1235,15 +1238,107 @@ fn build_mrf_response(
...
+                .unwrap_or_default(),
+            retry_count: entry.retry_count,
+            size: entry.size,
+            target_arns: entry.target_arns.clone(),
+        })
+        .collect()
```
overtrue: The durable ledger is not bounded by MRF_PENDING_CAP: the existing persistence test records a valid 200,001-entry generation, and this endpoint reads it without a byte limit. This collect clones every matching object and target string, then render_mrf_backlog buffers the complete JSON too; valid 1,024-byte keys can add hundreds of MB per request before a byte is sent. Please stream documents as MinIO does, or enforce documented pagination or bounds.

overtrue (changes requested):

rustfs/src/admin/handlers/replication.rs
reatang: Fixed in 4609e78fc: the default stream now requires `admin:ReplicationDiff` (route policy entry updated to match, with the aggregate relaxation documented there); `?aggregate=true` carries no object identities and keeps `admin:GetReplicationMetrics`.

rustfs/src/admin/handlers/replication.rs
reatang: Fixed in 4609e78fc: the nil UUID now maps to the S3 wire token `null` (`NULL_VERSION_ID`) in the stream documents, pinned by `mrf_stream_maps_nil_version_to_null_token`.

rustfs/src/admin/handlers/replication.rs
reatang: Fixed in 4609e78fc via documented bounds: the stream stops at 10,000 documents (`REPLICATION_MRF_MAX_STREAM_ENTRIES`) and signals truncation out-of-band via `x-rustfs-replication-mrf-truncated` + a warn event, mirroring the diff endpoint's cap/truncation precedent; the projection also stops cloning past the bound. True chunked streaming (MinIO's encoder loop) needs a streaming admin body and is left as the noted upgrade path.

overtrue (approved):
LGTM

PR #6127, opened 2026-08-15
fix(admin): serialize replication metrics in minio-go wire shapes

reatang:
## Summary

Part of the MinIO replication compatibility review (rustfs/backlog#1675, **P1-11**).

`?replication-metrics[=2]` and the admin `replicationmetrics` endpoint serialized the internal snake_case `BucketStats` family straight onto the wire. minio-go's `replication.Metrics`/`MetricsV2` expect camelCase tags (`currStats`/`queueStats`/`replicaCount`/`queued`/…); Go's decoder is case-insensitive but does not ignore underscores, so `mc replicate status` showed all zeros with no error.

The internal structs cannot be renamed — they are the intra-cluster peer-RPC wire format (`rmp_serde::to_vec_named` in node_service.rs), now pinned by a regression test that fails if anyone 'fixes' the interop by renaming them.

- New `admin/replication_metrics_wire.rs`: Serialize-only projections with the exact minio-go json tags. Per-target `failed` becomes the `TimedErrStats` envelope fed from the `FailStats` rolling window; the queue peak is dual-emitted as `max` (MinIO server tag) and `peak` (minio-go tag) to cover the upstream drift.
- `queueStats` synthesizes one node from the bucket queue snapshot — the aggregation path leaves `queue_stats.nodes` empty and mc treats an empty node list as 'no data' — and carries Large/Small/Total transfer summaries derived from the per-target xfer rates.
- Both endpoints share the DTOs; RustFS source-health extension keys ride along (ignored by Go decoders).
- ecstore `replication_stats_boundary` re-exports widened (`BucketReplicationStat`/`InQueueMetric`/`XferStats`) so the admin facade chain can name the projected types.

## Testing

Red→green TDD; wire-shape unit tests (v1/v2/node synthesis) + rmp_serde named round-trip pin; `make pre-commit` + clippy clean. Note: `mrf`/reset shapes are separate PRs (#6126, #5799).


rustfs/src/admin/replication_metrics_wire.rs
```diff
@@ -0,0 +1,511 @@
...
+    #[serde(rename = "failed")]
+    pub failed: TimedErrStatsWire,
+    #[serde(rename = "failedReplicationSize")]
+    pub failed_size: i64,
+    #[serde(rename = "failedReplicationCount")]
+    pub failed_count: i64,
+}
+
+fn target_timed_err_stats(stat: &InternalReplicationStat) -> TimedErrStatsWire {
+    let last_minute = stat.fail_stats.recent_since(Duration::from_secs(60));
```
overtrue: Both metrics endpoints aggregate first, and FailStats::merge clears recent samples, so this makes lastMinute and lastHour report zero even immediately after a failure while totals stays nonzero. Please preserve the rolling-window data through cluster aggregation and add a regression test that exercises aggregation before serialization.

overtrue (changes requested):

rustfs/src/admin/replication_metrics_wire.rs
reatang: Fixed in 1c660362d: `FailStats` now carries serializable `last_minute`/`last_hour` window snapshots — recomputed on every `add_size`, re-stamped at the per-node collection point (`get_latest_replication_stats`, which also serves the peer RPC), and summed by `merge` (which still drops the process-local samples; they are serde-skipped and never crossed the wire anyway). The DTO takes the component-wise max of the live samples and the snapshot. `failure_windows_survive_aggregation_before_serialization` drives a stat through an rmp round trip + merge before serialization, per your request. The wire change is compatible in both directions (serde default on read, unknown fields ignored). This commit also fixes the -D warnings CI lanes: the new module declaration had slipped between `#[allow(dead_code)]` and `route_policy`.

rustfs/src/admin/replication_metrics_wire.rs
```diff
@@ -0,0 +1,563 @@
...
+/// `transferSummary` map keyed by minio-go `MetricName` (Large/Small/Total).
+type XferSummaryWire = HashMap<&'static str, XferStatsWire>;
+/// `tgtTransferStats` map keyed by target ARN.
+type TargetXferSummaryWire = HashMap<String, XferSummaryWire>;
+
+fn transfer_summaries(stats: &InternalReplicationStats) -> (XferSummaryWire, TargetXferSummaryWire) {
+    let mut summary: XferSummaryWire = HashMap::new();
+    let mut per_target: TargetXferSummaryWire = HashMap::new();
+    for (arn, stat) in &stats.stats {
+        let large = XferStatsWire::from(&stat.xfer_rate_lrg);
```
overtrue, line 389: `xfer_rate_lrg` is populated for every object above 1 MiB (`stats.rs:667`), but this new wire label means >=128 MiB in minio-go and RustFS already has the same 128 MiB cutoff. A successful 2 MiB replication therefore appears under Large and leaves Small at zero. Please align the producer with `MIN_LARGE_OBJ_SIZE`, including the 128 MiB boundary, and add a focused MetricsV2 assertion for 2 MiB and 128 MiB.

crates/replication/src/stats.rs
```diff
@@ -535,6 +543,16 @@ impl FailStats {
         self.size = self.size.saturating_add(size);
         self.recent.push_back(FailureSample { observed_at, size });
         self.prune(observed_at);
+        self.refresh_windows();
```
overtrue: This runs two full scans of the one-hour failure deque for every failed object while `ReplicationStats::update` holds the bucket-stats write lock. With n fresh failures the burst does n(n+1) visits; a 30,000-event direct benchmark took 2.1 s here versus under 1 ms before. The collection path already refreshes before peer serialization, so keep this scan off the failure hot path and make the regression drive collection through RPC and aggregation.

overtrue (changes requested):

rustfs/src/admin/replication_metrics_wire.rs
reatang: Fixed in 0200519d0: `update_xfer_rate` now bins on `MIN_LARGE_OBJ_SIZE` (the same 128 MiB boundary as the worker-pool split), and `transfer_summary_bins_on_the_128_mib_boundary` pins 2 MiB / 127 MiB as Small and exactly 128 MiB as Large through the MetricsV2 body.

crates/replication/src/stats.rs
reatang: Fixed in 0200519d0: `add_size` no longer touches the windows — they are recomputed only at the collection point (`get_latest_replication_stats`, which serves both the local leg and the peer RPC), so the failure hot path is back to O(1) per event. The aggregation regression now stamps the windows via `refresh_windows()` (collection semantics) before the rmp round trip + merge, per your note.

overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #6129, opened 2026-08-15
fix(replication): transport and persist LWW timestamps for tag, retention, and legal hold

reatang:
## Summary

Part of the MinIO replication compatibility review (rustfs/backlog#1675, **P1-6**).

The three `X-Minio-Source-Replication-{Tagging,Retention,LegalHold}-Timestamp` headers were neither sent nor parsed: the boundary filled the option fields but `PutObjectOptions::header()` never serialized them (dead fields), the receiver had no `ObjectOptions` slots for them, and the tagging timestamp had no local truth source at all (the internal key was never written). Active-active LWW for tag/retention/legal-hold modifications was therefore not degraded but zeroed — both sides fell back to identical mod_times.

- Sender: suffix constants added (`insert_header` dual-writes `x-rustfs-`/`x-minio-`, wire names pinned byte-for-byte against MinIO's headers), `header()` emits non-epoch timestamps, `AdvancedPutOptions` defaults move from `now_utc()` to `UNIX_EPOCH` so 'unset' is never sent as 'modified right now'.
- Local truth: PutObjectTagging / DeleteObjectTagging now stamp the tagging-timestamp internal key (lock timestamps already had writers).
- Receiver: authorized replication requests (PUT + multipart complete) parse the headers into three new `ObjectOptions` fields and persist them into the internal metadata keys, closing the loop with the outbound boundary read points. Unauthorized clients cannot forge them (pinned by gated-header tests, following #5625).

## Not in this PR

Receiver-side LWW comparison (stored-newer-wins per category) needs the set-layer metadata merge point and is left as a marked TODO; this PR delivers bidirectional transport + persistence, which P1-5 (GET/HEAD proxy) also depends on.

Note: `crates/replication/src/http.rs` intentionally does not duplicate the three constants — that wire-contract copy only mirrors values it consumes, and unused constants fail CI's dead_code gate.

## Testing

Red→green TDD: header emission (present + epoch-omitted), gated persistence (authorized vs forged), fake_s3_target header-set extension with self-tests; `make pre-commit` + clippy `-D warnings` on the three touched crates; nextest sweeps (ecstore replication 220, storage::options 69, utils http 38, replication crate 161) all green.

### DIFF
--- a/crates/ecstore/src/bucket/bucket_target_sys.rs
+++ b/crates/ecstore/src/bucket/bucket_target_sys.rs
@@ -2842,6 +2842,57 @@ mod tests {
         );
     }
 
+    #[test]
+    fn put_object_headers_carry_replication_timestamp_headers() {
+        // MinIO receivers resolve concurrent tag/retention/legal-hold edits by
+        // last-writer-wins on these headers (object-api-options.go parses them
+        // as RFC3339); a replica without them loses every conflict resolution.
+        let mut opts = PutObjectOptions::default();
+        opts.internal.replication_request = true;
+        let tagging = OffsetDateTime::from_unix_timestamp(1_700_000_001).expect("valid timestamp");
+        let retention = OffsetDateTime::from_unix_timestamp(1_700_000_002).expect("valid timestamp");
+        let legalhold = OffsetDateTime::from_unix_timestamp(1_700_000_003).expect("valid timestamp");
+        opts.internal.tagging_timestamp = tagging;
+        opts.internal.retention_timestamp = retention;
+        opts.internal.legalhold_timestamp = legalhold;
+
+        let header = opts.header();
+        for (suffix, expected) in [
+            ("source-replication-tagging-timestamp", tagging),
+            ("source-replication-retention-timestamp", retention),
+            ("source-replication-legalhold-timestamp", legalhold),
+        ] {
+            assert_eq!(
+                rustfs_utils::http::get_header(&header, suffix).as_deref(),
+                Some(expected.format(&Rfc3339).expect("RFC3339 timestamp").as_str()),
+                "replication put requests must carry the {suffix} header"
+            );
+        }
+    }
+
+    #[test]
+    fn put_object_headers_omit_unset_replication_timestamps() {
+        // UNIX_EPOCH means "never modified on the source"; sending it would
+        // make the receiver treat an unset category as a fresh modification.
+        let mut opts = PutObjectOptions::default();
+        opts.internal.replication_request = true;
+        opts.internal.tagging_timestamp = OffsetDateTime::UNIX_EPOCH;
+        opts.internal.retention_timestamp = OffsetDateTime::UNIX_EPOCH;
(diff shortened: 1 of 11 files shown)


crates/ecstore/src/bucket/bucket_target_sys.rs
```diff
@@ -1675,6 +1680,16 @@ impl PutObjectOptions {
             );
         }
 
+        for (suffix, timestamp) in [
+            (SUFFIX_SOURCE_REPLICATION_TAGGING_TIMESTAMP, self.internal.tagging_timestamp),
```
overtrue, line 1684: DeleteObjectTagging now persists this timestamp, but the outbound mapper only loads it when user_tags is nonempty. After a tag deletion the internal key is stripped from generic metadata, tagging_timestamp stays at the epoch, and this loop omits the header, so the deletion timestamp never reaches or persists on the replica. Please load the stored tagging timestamp independently of whether any tags remain and add a deletion-path regression test.

overtrue (changes requested):

crates/ecstore/src/bucket/bucket_target_sys.rs
reatang: Fixed in 77a536093: the stored tagging timestamp is now loaded unconditionally in `replication_put_object_options` (the mod_time fallback still applies only while tags exist, MinIO parity; a tagless object without the key keeps the epoch default so no header is sent). Added `replication_put_options_carry_tagging_timestamp_after_tag_deletion` covering both the deletion path and the never-tagged path.

overtrue, line 1684: An ordinary PutObject can currently persist x-amz-meta-x-rustfs-source-replication-tagging-timestamp as a bare user-metadata key because the ingest namespacing guard does not reserve these new transport names. For a tagless object its real timestamp stays at epoch, so this loop does not overwrite the attacker value; header() forwards it with the server-authorized replication request, and the receiver persists it as trusted internal LWW state. The focused ingress test fails on this head. Please reserve the RustFS and MinIO source-replication timestamp keys and cover the forged user-metadata path.

overtrue (changes requested):

crates/ecstore/src/bucket/bucket_target_sys.rs
reatang: Fixed in 74b118901: the ingest namespacing guard now reserves the whole `x-rustfs-source-` / `x-minio-source-` families — the three new timestamp names and their pre-existing siblings (source-mtime/-etag/-version-id/-replication-request), which were forgeable the same way — folding forged keys back under `x-amz-meta-`. `test_replication_transport_names_cannot_be_forged_via_user_metadata` covers both prefixes, the x-rustfs-meta- variant, and a sibling. (The E2E lane failure was the compression multipart test; it passes locally on this branch — 13.7s, fresh debug binary — so it looks like a lane flake; the new push re-runs it.)

overtrue (approved):
LGTM

overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #6130, opened 2026-08-15
fix(site-replication): merge incoming ILM expiry documents instead of overwriting

reatang:
## Summary

Part of the MinIO replication compatibility review (rustfs/backlog#1675, **P1-1**).

The lc-config receiver replaced the whole local lifecycle config with the peer's document (and deleted it wholesale on peer delete), so an expiry-only update erased the receiver's local tier/transition rules, and a peer's transition rules were installed across sites.

**Receiver** — `apply_bucket_meta_item` now merges via `merge_incoming_lifecycle_config` (MinIO `mergeWithCurrentLCConfig` parity plus a trust-boundary hardening: incoming site-local fields — transitions, abort-mpu, del-marker — are discarded outright; the local side of a same-id rule is authoritative for them). A peer delete merges with the empty set: pure-expiry rules go, site-local rules survive. Staleness moves to the expiry axis (`expiry_updated_at`), falling back to the whole-config write time when the config is absent/legacy — the fallback survives deletion in bucket metadata, so a delayed stale broadcast cannot resurrect deleted rules. A receiver-side `replicateILMExpiry` gate mirrors the sender hook. MinIO's zero-rule tombstone documents (rejected by the strict deserializer) are recognized as delete semantics.

**Sender** — both admin choke points (bucket-meta hook; SRInfo bucket entry feeding bootstrap/repair) emit only the expiry subset: MinIO receivers install incoming rules verbatim, so transition/abort-mpu/del-marker fields must never leave the site. `PutBucketLifecycle` stamps `expiry_updated_at` only when the expiry subset can have changed (MinIO parity) — a transition-only edit no longer inflates the axis and can no longer shadow/roll back newer peer expiry edits; the hook item and repair items travel with the config's expiry axis instead of the wall clock.

The merge is order-deterministic (local order, incoming-new appended), so re-delivery is byte-stable and does not rewrite bucket metadata per broadcast.

## Follow-ups (documented in-code)

- Receiver-side validation of merged configs (object-lock / tier constraints; MinIO runs `finalLcCfg.Validate`).
- A two-site e2e with a real tier backend to pin transition-rule preservation end to end.

## Testing

Red→green TDD: 11 merge-contract unit tests (transition preservation, same-id authority, trust-boundary strip, delete-with-empty, disabled rules, del-marker/abort-mpu locality, idempotence, axis stamping) + subset/sniff tests; full nextest sweep over site_replication / lifecycle / bucket_usecase / bucket_meta (339 tests); `make pre-commit` + clippy clean. An adversarial review pass ran on the diff; all four confirmed findings are fixed in the final commit.

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -2331,6 +2331,8 @@ fn append_bootstrap_bucket_items(
             bucket.expiry_lc_config_updated_at,
             |item, value| {
                 item.expiry_lc_config = Some(value);
+                // `updated_at` here is the entry's expiry axis (see the
+                // SRBucketInfo construction), not the wall clock.
                 item.expiry_updated_at = item.updated_at;
                 Ok(())
             },
@@ -4343,7 +4345,14 @@ async fn build_sr_info(state: &SiteReplicationState, local_peer: &PeerInfo) -> S
             entry.versioning_config_updated_at = maybe_time(metadata.versioning_config_updated_at);
             entry.replication_config_updated_at = maybe_time(metadata.replication_config_updated_at);
             entry.quota_config_updated_at = maybe_time(metadata.quota_config_updated_at);
-            entry.expiry_lc_config_updated_at = maybe_time(metadata.lifecycle_config_updated_at);
+            // The expiry axis, not the whole-config write time: local
+            // transition-only edits inflate the latter, and a repair item
+            // stamped with it could out-rank a newer real expiry edit on a
+            // third site. Legacy configs without the axis fall back to the
+            // whole-config time (bounded staleness, same as the receiver).
+            entry.expiry_lc_config_updated_at = lifecycle_expiry_updated_at(&metadata.lifecycle_config_xml)
+                .map(Some)
+                .unwrap_or_else(|| maybe_time(metadata.lifecycle_config_updated_at));
             entry.cors_config_updated_at = maybe_time(metadata.cors_config_updated_at);
             entry.replication_targets_online =
                 Some(site_replication_targets_online(&bucket.name, &metadata.replication_config_xml).await);
@@ -6865,23 +6874,22 @@ fn merge_incoming_replication_config(
 /// Merge a peer's ILM expiry document into the local lifecycle config.
 ///
 /// Mirrors MinIO's `mergeWithCurrentLCConfig` with one hardening: incoming
-/// transition fields are discarded outright (trust boundary — a peer must
-/// never install tier/transition rules on this site, whatever it sends).
-/// Local transition rules always survive; a delete (`incoming == None`)
-/// therefore merges with the empty set instead of dropping the whole config.
+/// site-local fields (transitions, abort-multipart, del-marker expiration —
+/// exactly what MinIO's `CloneNonTransition` sender never emits) are
+/// discarded outright at the trust boundary, whatever the peer sends. Local
+/// site-local fields always survive; a delete (`incoming == None`) therefore
+/// merges with the empty set instead of dropping the whole config.
 fn merge_incoming_lifecycle_config(
     incoming: Option<s3s::dto::BucketLifecycleConfiguration>,
     local: Option<s3s::dto::BucketLifecycleConfiguration>,
     updated_at: Option<OffsetDateTime>,
 ) -> Option<s3s::dto::BucketLifecycleConfiguration> {
-    // Incoming rules reduced to their expiry side (trust boundary). Rules
-    // that carry no expiry semantics after the strip — transition-only or
-    // abort-multipart-only — are site-local and are not installed.
+    // Incoming rules reduced to their traveling expiry side. Rules with no
+    // expiry semantics after the strip are not installed.
     let mut incoming_by_id: HashMap<String, s3s::dto::LifecycleRule> = HashMap::new();
     let mut incoming_order: Vec<String> = Vec::new();
     for mut rule in incoming.into_iter().flat_map(|config| config.rules) {
-        rule.transitions = None;
-        rule.noncurrent_version_transitions = None;
+        strip_site_local_lifecycle_fields(&mut rule);
         if !lifecycle_rule_has_expiry(&rule) {
             continue;
         }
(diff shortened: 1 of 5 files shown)


rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -4319,7 +4331,12 @@ async fn build_sr_info(state: &SiteReplicationState, local_peer: &PeerInfo) -> S
             entry.sse_config = raw_config_to_base64(&metadata.encryption_config_xml);
             entry.replication_config = raw_config_to_base64(&metadata.replication_config_xml);
             entry.quota_config = raw_config_to_base64(&metadata.quota_config_json);
-            entry.expiry_lc_config = raw_config_to_base64(&metadata.lifecycle_config_xml);
+            // Expiry subset only: this entry feeds both the bootstrap/repair
+            // plan (peers must not receive transition rules) and cross-site
+            // consistency views (transition rules are site-local and would
+            // read as false mismatches).
+            entry.expiry_lc_config =
```
overtrue: An expiry deletion has no config body but still has an expiry timestamp. append_bootstrap_bucket_item only emits present values, so repair sends no lc-config item and a peer that missed the live delete keeps stale expiry rules. Please emit a timestamped zero-rule or explicit delete item and cover missed-delete repair.

rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -7970,6 +8173,71 @@ async fn apply_bucket_meta_item(item: SRBucketMeta) -> S3Result<()> {
...
+
+        let incoming = match item.expiry_lc_config.as_ref() {
+            Some(raw) => {
+                let data = decode_bucket_meta_wire_value(raw);
+                match deserialize::<s3s::dto::BucketLifecycleConfiguration>(&data) {
+                    Ok(config) => Some(config),
+                    // MinIO's delete tombstone / transition-only state is a
+                    // zero-rule document the strict deserializer rejects; it
+                    // means "no expiry rules here" (delete semantics).
+                    Err(_) if !lifecycle_document_has_rules(&data) => None,
```
overtrue: This classifies every parse failure without the literal <Rule as a delete, including arbitrary malformed or truncated bytes, so a bad peer payload can erase local expiry rules. Please recognize only a structurally valid zero-rule tombstone and reject other malformed documents.

rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -7970,6 +8173,71 @@ async fn apply_bucket_meta_item(item: SRBucketMeta) -> S3Result<()> {
...
+        // Staleness axis: the config's expiry_updated_at when present. When
+        // the config is absent (deleted) or predates the axis field, fall
+        // back to the whole-config write time — it survives deletion in
+        // bucket metadata and is the deletion's lower bound, so a delayed
+        // stale broadcast cannot resurrect deleted expiry rules.
+        let whole_config_axis = metadata_sys::get(&item.bucket)
+            .await
+            .map(|bucket_meta| bucket_meta.lifecycle_config_updated_at)
+            .unwrap_or(OffsetDateTime::UNIX_EPOCH);
+        let local_expiry_updated_at = local
```
overtrue: A present transition-only config has no expiry axis but still advances lifecycle_config_updated_at. Falling back to that timestamp rejects an independent peer expiry update and repair with the same expiry timestamp remains blocked. Please reserve the whole-config fallback for deleted or legacy expiry state and add a delayed-delivery regression test.

overtrue (changes requested):

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in 4786463ad: `lifecycle_expiry_statement` now distinguishes a deleted expiry state (whole-config write time advanced past the created-time backfill) from never-configured buckets, and the bootstrap plan emits an explicit lc-config delete item stamped with the deletion axis. Covered by `test_site_replication_bootstrap_plan_emits_timestamped_lifecycle_delete` and the statement matrix test.

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in 4786463ad: replaced the `<Rule` sniff with `is_zero_rule_lifecycle_tombstone`, which requires a well-delimited `<LifecycleConfiguration>` document (optional XML prolog, attributes allowed, matching close tag or self-closing root) with no `<Rule` child. Truncated/foreign/garbage payloads now surface `InvalidRequest` instead of mapping to delete semantics. Negative cases pinned in `test_zero_rule_lifecycle_tombstone_recognition`.

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in 4786463ad: the fallback is now scoped by `local_lifecycle_staleness_axis` — whole-config time only for deleted state and legacy configs that actually contain expiry rules; a present transition-only config without the axis compares at epoch, so it can neither shadow an independent peer expiry update nor block a same-timestamp repair. Pinned by `test_local_lifecycle_staleness_axis_selection` (the delayed-delivery orderings are exercised through the axis matrix + the existing is_stale semantics).

rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -6847,6 +6881,262 @@ fn merge_incoming_replication_config(
...
+    }
+    // The root open tag may carry attributes (xmlns); it must close, and the
+    // document must end with the matching close tag.
+    let Some(open_end) = rest.find('>') else {
+        return false;
+    };
+    let Some(body) = rest[open_end + 1..].trim_end().strip_suffix("</LifecycleConfiguration>") else {
+        return false;
+    };
+    !body.contains("<Rule")
```
overtrue: This fallback still accepts structurally malformed documents as tombstones. For example, `<LifecycleConfiguration><ExpiryUpdatedAt></LifecycleConfiguration>` fails strict deserialization but passes this check and is then applied as a delete, erasing local expiry state. Please validate the exact empty-root document and cover this malformed-child case.

overtrue (changes requested):

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in 0b9405d7e: the tombstone body is now validated structurally — a sequence of well-formed simple children only (matching open/close or self-closing, no nested markup, no stray text, none named Rule). Your `<ExpiryUpdatedAt>` dangling-open example and the stray-text / nested-markup variants are pinned as rejected in `test_zero_rule_lifecycle_tombstone_recognition`.

overtrue (approved):
LGTM

PR #6131, opened 2026-08-15
feat(site-replication): drain the retry queue from the reconcile tick

reatang:
## Summary

Part of the MinIO replication compatibility review (rustfs/backlog#1675, **P1-3**, scoped per review: drive the existing retry-queue ledger rather than porting MinIO's full periodic heal).

The retry queue recorded every failed peer delivery but had no consumer — missed IAM/bucket metadata updates drifted until an operator ran a manual repair. The 600s reconcile tick now drains the queue behind the existing lifecycle guard and pending_* gates.

**Replay discipline** (pinned by red-light tests):
- IAM / bucket-meta entries collapse per (peer, path) with no persisted body — the only faithful replay is the current bootstrap-plan snapshot (reusing `SiteReplicationRepairTask`), sent to the affected peer.
- `make-with-versioning` / `configure-replication` are re-derived from the *current* plan for their bucket (never the recorded path, whose query can carry an expired one-shot bootstrap token); an op whose bucket left the plan is provably stale and is settled.
- Peer edits are re-sent as the current peer records under a freshly allocated generation inside one state transaction — the recorded generation is stale by definition and the receiver would fence it.
- Destructive bucket ops and `internal:` marker records (the pending-endpoint-refresh backup store) are never background-replayed.
- Exponential backoff (600s·2^(n-1), 24h ceiling) gates each attempt; transport failures re-queue so backoff still advances for unreachable peers.

**Concurrency**: the drain takes the repair-execution config lock (lock order matches repair: lifecycle guard → repair execution lock → state object lock); settlement reuses the generation-fenced upsert/settle semantics from #5882/#6097. The lock does *not* close the repair dry-run→execute window (dry-run takes no lock); that window fails safe via the preflight token, which hashes the replayable entries — the comment says exactly that.

**Hardening from the adversarial review pass** (final commit):
- Conditional settlement for collapsed entries: a hook failure stamped during the (potentially long) snapshot delivery window belongs to a newer local commit the snapshot did not contain and now survives the snapshot's success.
- pending_* gates re-checked on the freshly loaded state; no upsert for peers that already left the state; `(Some settled, None failed)` generation-settle combination pinned by a test.

## Scope note

The drain settles 'delivered once and failed' entries. A hook that never fired (crash between local commit and send) leaves no entry; a low-frequency plan-diff catch-all remains follow-up work and manual repair stays authoritative.

## Testing

Red→green TDD (classification / backoff / actionable-filter specs first); site-replication suite 212 green; `make pre-commit` + clippy clean.

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -6119,6 +6119,7 @@ async fn migrate_collapsed_retry_queue_paths() -> S3Result<()> {
     .await
 }
 
+#[cfg(test)]
 fn dequeue_site_replication_retry_events(queue: &mut Vec<SiteReplicationRetryEvent>, peer: &PeerInfo, path: &str) -> usize {
     settle_site_replication_retry_events(queue, peer, path, None)
 }
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -14437,12 +14437,12 @@ mod tests {
     }
 
     #[test]
-    fn test_retry_event_dequeue_preserves_collapsed_path_across_peer_identity() {
+    fn test_retry_event_dequeue_matches_deployment_id_or_endpoint() {
         let peer = PeerInfo {
             deployment_id: "current-dep".to_string(),
             ..peer("remote", "https://remote.example.com")
         };
-        let path = "/rustfs/admin/v3/site-replication/peer/iam-item";
+        let path = SITE_REPLICATION_PEER_EDIT_PATH;
         let mut queue = vec![
             SiteReplicationRetryEvent {
                 id: "same-endpoint".to_string(),
@@ -14462,8 +14462,9 @@ mod tests {
 
         let removed = dequeue_site_replication_retry_events(&mut queue, &peer, path);
 
-        assert_eq!(removed, 0);
-        assert_eq!(queue.len(), 2);
+        assert_eq!(removed, 1);
+        assert_eq!(queue.len(), 1);
+        assert_eq!(queue[0].id, "different-path");
     }
 
     #[test]
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -2973,6 +2973,18 @@ fn reconcile_site_replication_wiring() -> std::pin::Pin<Box<dyn std::future::Fut
             return;
         };
 
+        if let Err(err) = migrate_collapsed_retry_queue_paths().await {
+            warn!(
+                event = EVENT_ADMIN_SITE_REPLICATION_STATE,
+                component = LOG_COMPONENT_ADMIN,
+                subsystem = LOG_SUBSYSTEM_SITE_REPLICATION,
+                result = "retry_queue_migration_failed",
+                error = ?err,
+                "admin site replication state"
+            );
+            return;
+        }
+
         match load_site_replication_state().await {
             Ok(state) => {
                 if state.pending_endpoint_refresh.is_some() || state.pending_remove.is_some() || state.pending_rotation.is_some()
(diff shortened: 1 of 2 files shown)


rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -6171,6 +6179,396 @@ fn retry_event_replayed_by_bootstrap(event: &SiteReplicationRetryEvent) -> bool
...
+                if let Err(err) = task.send(transport, access_key, secret_key).await {
+                    enqueue_site_replication_retry_event(peer, &event.path, &err).await;
+                    return Err(err);
+                }
+            }
+            // Conditional settlement: iam-item / bucket-meta entries collapse
+            // per (peer, path), so a hook failure recorded while this snapshot
+            // was in flight belongs to a commit the snapshot did not contain
+            // and must survive this success.
+            dequeue_site_replication_retry_event_up_to(peer, &event.path, event.updated_at).await;
```
overtrue: The current snapshot cannot replay deletions: deleted IAM entities and absent bucket configs produce no task, yet this clears the collapsed retry event after the remaining snapshot succeeds. A failed user or config delete is therefore lost and the peer keeps stale state permanently. Please persist or reconstruct a tombstone, or keep these events until remote absence is proven.

overtrue (changes requested):

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in 971addca6, taking the 'keep these events until remote absence is proven' route: after a successful snapshot resend the collapsed iam-item / bucket-meta entry is now **escalated** (failed=true + a marker last_error), not cleared — it stays operator-visible in retry_stats until a later full delivery or a manual repair settles it. Classification skips marked entries so the snapshot is sent once per failure episode, not daily; a newer hook failure overwrites the marker and re-arms the drain. Escalation is conditional on the snapshot timestamp (the in-flight-failure guarantee from the previous round is preserved). Bucket-op and peer-edit replays are faithful and keep auto-settle. Reconstructing deletion tombstones (bucket-meta could derive them from the per-config updated_at that survives deletion) is noted as the follow-up that would restore auto-settle.

rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -6171,6 +6179,420 @@ fn retry_event_replayed_by_bootstrap(event: &SiteReplicationRetryEvent) -> bool
...
+            (Some(current), Some(seen)) => current > seen,
+            (Some(_), None) => true,
+            (None, _) => false,
+        };
+        if newer_failure_recorded {
+            continue;
+        }
+        event.failed = true;
+        event.retry_count = event.retry_count.max(SITE_REPLICATION_RETRY_FAILED_AFTER);
+        event.last_error = SITE_REPLICATION_RETRY_SNAPSHOT_REPLAYED_MARKER.to_string();
```
overtrue: This marker is still removed by any later successful single-item delivery: every IAM item and bucket-meta update shares a constant path, and the success path dequeues by `(peer, path)`. A failed Alice deletion can be marked here and then cleared by a successful Bob update while Alice still exists remotely. Please give the marker entity/tombstone identity, or isolate it from ordinary success dequeue, and add that regression.

overtrue (changes requested):

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in af802756c, via the isolation route: ordinary settlement (`settle_site_replication_retry_events`, i.e. every hook/transport success dequeue) now skips escalated entries — a successful Bob update no longer clears the marker recording a possibly-unreplayed Alice deletion. Only the repair path clears them through the new `dequeue_site_replication_retry_events_including_escalated` (the operator's explicit accountability transfer). A new hook failure still overwrites the marker and re-arms the drain. The escalate regression now pins survive-ordinary-dequeue + repair-clears. Entity/tombstone identity for the collapsed paths (which would restore full auto-settle) stays noted as the follow-up.

overtrue (approved):
LGTM

overtrue (approved):
LGTM

reatang:
Verified @overtrue's follow-up commits (96a2f9c67…7e071dc0f) locally on this branch: site-replication suite 215/215 green, `make pre-commit` clean. The internal `retry-snapshot:*` liability paths + startup migration + snapshot stability loop with concurrent-deletion tombstones close the entity-identity gap properly — thanks for taking it further than the isolation fix. Both earlier CI runs on these commits were cancelled before running anything; re-triggered run 31911766396 (attempt 2), now in progress.

reatang:
CI is green on attempt 2 (all lanes passed, mergeState CLEAN) — the two earlier cancellations never ran anything. @overtrue your follow-up commits are locally verified on my side (215/215 site-replication suite + pre-commit, see comment above); from my end this is ready to merge.

PR #6140, opened 2026-08-16
perf(ecstore): enable inline data read early-stop by default

houseme:
## Related Issues
rustfs/backlog#1647

## Summary of Changes
Enable the verified inline data-read metadata early-stop path by default while keeping bounded fanout opt-in. This follows the testing1 ABBA result from rustfs/backlog#1647 where the post-#6137 binary showed a stable small-object GET uplift with `data_read_inline_missing_shard=0`, but preserves the #5917 slow-tail prerequisite for enabling bounded scheduling by default.

The default still keeps the narrow safety envelope: explicit version GETs remain behind the independent version-aware gate, bounded fanout remains disabled unless explicitly opted in, and unsafe request shapes or metadata states still fall back to full-wait fanout. The production reader default test now proves whole/latest inline GET returns the exact body under default env settings, while prepared metadata, Range, partNumber, and non-inline paths continue to require full fanout.

This PR also boxes the large ignored ILM `test_transition_and_restore_flows` future after CI exposed a Linux runtime stack overflow in that test process. The change is test-harness-only and keeps the same test behavior.

## Verification
- `cargo test -p rustfs-ecstore --lib metadata_early_stop -- --nocapture`
- `cargo test -p rustfs-ecstore --lib inline_data_read_early_stop_defaults_return_exact_body -- --nocapture`
- `cargo test -p rustfs-ecstore --lib prepared_metadata_uses_full_fanout_even_when_data_read_early_stop_is_enabled -- --nocapture`
- `cargo test -p rustfs-ecstore --lib data_read_early_stop_request_shapes_full_wait_in_production_reader -- --nocapture`
- `cargo test -p rustfs-ecstore --lib metadata_early_stop_permitted_respects_caller_opt_out -- --nocapture`
- `cargo test -p rustfs-ecstore --lib --features rio-v2 missing_internal_meta_reader_records_internal_meta_path -- --nocapture`
- `cargo nextest run -j1 --run-ignored ignored-only -p rustfs-scanner -E 'test(test_transition_and_restore_flows)'`
- `cargo fmt --all --check`
- `cargo clippy -p rustfs-ecstore --lib --tests --features rio-v2 -- -D warnings`
- `cargo clippy -p rustfs-scanner --tests -- -D warnings`
- `git diff --check`

## Impact
Whole/latest plaintext inline small-object GETs can now use verified data-read metadata early-stop without setting `RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE=true`. Bounded fanout remains opt-in via `RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT=true`, so cold or non-inline GETs do not default to the two-wave metadata scheduling path. Deployments can disable data-read early-stop with `RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE=false`.

No on-disk format, wire format, S3 API shape, authorization, or mixed-version protocol behavior changes are introduced by this PR.

## Additional Notes
Adversarial validation summary:
- Correctness: default early-stop remains guarded by the inline verifier; non-inline, transformed, remote, deleted, malformed, mixed-generation, copied-shard, corrupt, Range, partNumber, prepared, and explicit-version-adjacent paths retain full-wait or independent gates.
- Simplicity: production diff is limited to default values and explanatory comments; tests were updated to be default/revert-sensitive rather than adding new behavior.
- Security: no authn/authz surface, secrets, path trust boundary, or protocol parsing changes.
- Concurrency/durability: early-stop task draining and fallback behavior are unchanged; default tests pin exact body and lifecycle behavior.
- Compatibility: explicit version early-stop remains disabled by default; request-shape tests preserve full-wait for Range and partNumber.
- Performance: enables the verified inline small-object GET early-stop path by default while keeping bounded scheduling behind explicit slow-tail validation; no additional hot-path allocations or syscall-path changes are introduced.
- Test coverage: focused tests kill reverts of the new defaults, production call-site wiring, prepared full-wait, Range/partNumber full-wait, non-inline full-wait fallback, and the ILM test stack-harness regression.

### DIFF
--- a/crates/ecstore/src/set_disk/mod.rs
+++ b/crates/ecstore/src/set_disk/mod.rs
@@ -692,8 +692,9 @@ const DEFAULT_RUSTFS_GET_SMALL_OBJECT_DIRECT_MEMORY_THRESHOLD: usize = 128 * 102
 const ENV_RUSTFS_GET_METADATA_EARLY_STOP_ENABLE: &str = "RUSTFS_GET_METADATA_EARLY_STOP_ENABLE";
 // Enabled by default (backlog#872): the early-stop path only engages for
 // requests `should_allow_metadata_early_stop` classifies as safe (latest-version
-// metadata-only reads by default, without version_id / healing / free-version
-// needs) and still requires a full read-quorum agreement before stopping. Set
+// reads by default, without version_id / healing / free-version needs) and still
+// requires a full read-quorum agreement before stopping. Data-read requests add
+// a separate inline-shard verifier before cancelling the remaining fanout. Set
 // the env var to `false` to fall back to full-wait metadata fanout.
 const DEFAULT_RUSTFS_GET_METADATA_EARLY_STOP_ENABLE: bool = true;
 
@@ -704,10 +705,10 @@ const ENV_RUSTFS_GET_METADATA_VERSION_EARLY_STOP_ENABLE: &str = "RUSTFS_GET_META
 const DEFAULT_RUSTFS_GET_METADATA_VERSION_EARLY_STOP_ENABLE: bool = false;
 
 const ENV_RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE: &str = "RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE";
-const DEFAULT_RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE: bool = false;
+const DEFAULT_RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE: bool = true;
 
 const ENV_RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT: &str = "RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT";
-const DEFAULT_RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT: bool = false;
+const DEFAULT_RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT: bool = true;
 
 // --- Multipart Reader-Setup Prefetch Configuration (backlog#870) ---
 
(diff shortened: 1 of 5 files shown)


crates/ecstore/src/set_disk/mod.rs
```diff
@@ -704,10 +705,10 @@ const ENV_RUSTFS_GET_METADATA_VERSION_EARLY_STOP_ENABLE: &str = "RUSTFS_GET_META
 const DEFAULT_RUSTFS_GET_METADATA_VERSION_EARLY_STOP_ENABLE: bool = false;
 
 const ENV_RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE: &str = "RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE";
-const DEFAULT_RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE: bool = false;
+const DEFAULT_RUSTFS_GET_METADATA_DATA_READ_EARLY_STOP_ENABLE: bool = true;
 
 const ENV_RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT: &str = "RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT";
-const DEFAULT_RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT: bool = false;
+const DEFAULT_RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT: bool = true;
```
overtrue: This also changes cold whole/latest non-inline GETs. The default now starts only the write quorum and schedules the remaining disks after that quorum reports data_read_inline_not_inline, serializing two metadata waves on the normal GET path. #5917 explicitly left p50/p90/p99 slow-tail validation as the prerequisite for enabling bounded fanout by default, but this PR only measures 1 KiB inline GETs. Please keep this default off until the non-inline path preserves eager fanout or has the required slow-tail evidence.

overtrue (changes requested):

crates/ecstore/src/set_disk/mod.rs
houseme: Adopted. Bounded fanout now remains default-off and opt-in only; the PR only defaults the verified inline data-read early-stop gate. The default inline reader test now uses eager full fanout with a paused parity metadata response, and non-inline/default paths no longer enter bounded two-wave scheduling without explicit opt-in.

overtrue (approved):
LGTM

PR #6143, opened 2026-08-16
fix(storage): skip table guards for multipart parts

### LINKED ISSUE #6133: 1.0.0-rc.2 regression: concurrent multipart uploads fail with 500 (table-catalog publication lock)
## Summary

`1.0.0-rc.2` adds a table-catalog *publication guard* to the S3 data plane. For every content-mutating request — including every `UploadPart` — it acquires a namespace lock on one per-bucket resource, `<bucket>/.rustfs-table/warehouses/default/publication.lock`, **before** the code checks whether the bucket is a table bucket at all. Under concurrent multipart load the acquisition tail crosses the default 5 s `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT` budget and uploads fail with `500 InternalError` naming a subsystem the workload never touched.

At the default settings, 21–23 of 24 concurrent uploads into one ordinary bucket fail. The path does not exist at `1.0.0-rc.1`; it is introduced by `5e3010c6b` (#5779, *fix(table-catalog): harden commit publication*). Reproduced on a single node with one volume, and on a 3-node EC:4 cluster.

## Environment

Binary: official release asset `rustfs-linux-x86_64-musl-v1.0.0-rc.2.zip` (asset sha256 `dbe7b71892b7…`, inner binary sha256 `d286e5f8939b…`, `--version` banner `rustfs 1.0.0-rc.2`, git commit `e9f53180273b`).

Two rigs, same binary:

1. **Standalone** — the binary run directly on a Linux host (Fedora, kernel 7.1.x), ONE local volume, `RUSTFS_ADDRESS=127.0.0.1:9100`. No env overrides relevant to locking (`RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT` unset → default 5 s; no `RUSTFS_TABLE_CATALOG_*` set). One arm pinned to 2 CPUs with `taskset -c 0,1`; a control arm unconstrained on 16 cores.
2. **3-node cluster** — KVM, 3 nodes × 4 vCPU / 8 GiB (Alpine 3.24), 4 disks per node, one erasure set, EC:4, plaintext HTTP/1.1 internode. Only `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT` varied between runs.

Load: boto3, N uploads in parallel, the parts of **each** upload uploaded by 16 concurrent workers (the AWS SDK / TransferManager default behaviour), 8 MiB parts. Bucket created by a plain `CreateBucket`; nothing table-related anywhere.

### Standalone, stock defaults (single node, one volume)

| CPUs | load | result |
|---|---|---|
| 2 (`taskset -c 0,1`) | 24 uploads × 16 part-workers, 400 MiB objects | **3/24 completed — 53 × `500 InternalError`** (+3 connection-closed), 18.8 s |
| 16 (control) | same | 24/24 clean, 12.2 s |
| 16 (control) | 8 × 16, 200 MiB | 8/8 clean, 2.4 s |

Every failure is `UploadPart` returning:

```
500 InternalError: failed to acquire table publication guard: table catalog
store error: failed to acquire catalog migration lock: Lock acquisition
timeout for resource
'mplock/.rustfs-table/warehouses/default/publication.lock@latest' after 5s
```

The server's own response log shows the failing requests at `duration: 5.001032701s` — the acquire timeout, verbatim. The clean 16-core control shows this is a saturation threshold, not a hard concurrency ceiling — the same shape #6064 had.

### 3-node EC:4 cluster — only the lock-timeout env varied

| load (uploads × part-workers, object size) | `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT=5` (default) | `=30` |
|---|---|---|
| 8 × 16, 200 MiB | 8/8 | 8/8 |
| 16 × 16, 200 MiB | 16/16 | 16/16 |
| 16 × 16, 400 MiB | 16/16 | 16/16 |
| 24 × 16, 400 MiB | **3/24, 48 × 500** (run 1); **1/24, 78 × 500** (run 2) | **24/24, 0 errors** |

Same error text on every failure, resource always the one publication.lock object of the target bucket.

### Control that isolates the mechanism

Plain `PutObject` at higher concurrency into one bucket, same default 5 s timeout, same cluster: **384 objects × 8 MiB with 256 concurrent workers — 0 errors** (and 256 × 128 workers — 0 errors). So this is not "the lock service is slow in general": the guard's acquisition only becomes the victim when it queues behind the multipart path's own per-upload/per-part lock traffic on a loaded box. What makes it a regression is that it now *fails the request* of a workload that has nothing to do with table catalogs.

`.rustfs-table/` was confirmed **absent on disk in every bucket on every node** after these runs — the locked path is synthetic; no catalog objects exist.

## Mechanism (code, at the `1.0.0-rc.2` tag)

1. `authorize_table_data_plane_if_needed` ([rustfs/src/storage/access.rs#L1604](https://github.com/rustfs/rustfs/blob/1.0.0-rc.2/rustfs/src/storage/access.rs#L1604)) runs for every S3 object action — `table_data_plane_admin_action` (L1385) maps essentially all object reads/writes. For content mutations (`PutObject`, `DeleteObject`, replication ops — L1423) it calls `retain_table_bucket_publication_guard(req, bucket)` at **L1617**, and only on the *next line* asks `table_bucket_enabled_for_data_plane` whether this bucket is a table bucket. The lock is taken first, the relevance check second. (Same ordering in `deny_anonymous_table_data_plane_if_needed`, L1679, for anonymous requests.)
2. The guard ([access.rs#L1460](https://github.com/rustfs/rustfs/blob/1.0.0-rc.2/rustfs/src/storage/access.rs#L1460)) calls `TableCatalogObjectBackend::acquire_read_lock(bucket, ".rustfs-table/warehouses/default/publication.lock")` ([rustfs/src/table_catalog/store/mod.rs#L1702](https://github.com/rustfs/rustfs/blob/1.0.0-rc.2/rustfs/src/table_catalog/store/mod.rs#L1702)), i.e. `new_ns_lock(...).get_read_lock(get_lock_acquire_timeout())` — a real namespace-lock acquisition per request (deduplicated within one request via `req.extensions`, held for the request's lifetime). On a multi-node deployment that acquisition is distributed; standalone it is local — both reproduce.
3. `get_lock_acquire_timeout()` is `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT`, default **5 s** ([crates/ecstore/src/set_disk/mod.rs#L1467](https://github.com/rustfs/rustfs/blob/1.0.0-rc.2/crates/ecstore/src/set_disk/mod.rs#L1467), [crates/config/src/constants/object.rs#L403](https://github.com/rustfs/rustfs/blob/1.0.0-rc.2/crates/config/src/constants/object.rs#L403)).
4. The resource is **one object per bucket** — every concurrent mutating request in a bucket funnels through the same entry. That is a strictly worse contention shape than #6064's per-upload-id lock (fixed by #5990, which is in rc.2 and works — see below): there, only the parts of one upload contended; here, **all uploads into the bucket** do.
5. No configuration disables the data-plane guard: rc.2 exposes `RUSTFS_TABLE_CATALOG_BACKING` / `_CREDENTIAL_VENDING` / `_STRONG_SNAPSHOT_V2` etc., none of which gate this path.

## Regression boundary

* `retain_table_data_plane_publication_guard` does not exist anywhere at the `1.0.0-rc.1` tag.
* It is introduced by `5e3010c6b` — *fix(table-catalog): harden commit publication (#5779)*.
* `git merge-base --is-ancestor 5e3010c6b 1.0.0-rc.1` → no; `… 1.0.0-rc.2` → yes.
* For contrast, `1021d7228` (#5990, per-part commit lock) is in rc.2 and visibly effective: at a 30 s budget this cluster now passes a rung that rc.1-era builds failed *even at 30 s*. The new publication guard is what re-introduces upload failures at default settings.

## Impact

Any rc.2 deployment at default settings running backup/archive-style traffic — large objects, concurrent parts, one bucket — starts failing `UploadPart` with a non-obvious `500 InternalError` about table catalogs once the box is under load. The per-bucket keying means spreading across uploads does not help; only lowering per-upload part concurrency, spreading across *buckets*, or raising the timeout does.

Workaround measured: `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT=30` → 24/24 clean at the same load (3-node; not re-tested standalone). It masks the failure rather than removing the contention — under deeper saturation the failures would return with 30 s tails.

## Suggested fix directions

Any one of these would remove the regression for non-table buckets:

1. **Check first, lock second.** `table_bucket_enabled_for_data_plane` is a bucket-metadata flag read (L1522, from the metadata cache) and is already evaluated on the very next line — evaluating it *before* `retain_table_bucket_publication_guard` would mean ordinary buckets never touch the publication lock.
2. Cache the negative verdict per bucket so repeated mutations on a non-table bucket skip the acquisition entirely.
3. If the guard must stay for table buckets, make publication fencing something other than a per-request namespace-lock acquisition on one shared resource (e.g. a per-node leased guard refreshed out of band, or fencing at catalog commit time only).

We can re-run any of the above loads against a proposed fix — the rigs and load generator are kept.

overtrue:
## Related Issues

Fixes #6133

## Summary of Changes

`UploadPart` and the destination side of `UploadPartCopy` authorize as `PutObject`, so they also retained the table publication lock for the full staged-part request. Under concurrent multipart traffic, that created unnecessary publication-lock pressure even though parts are not visible yet.

This keeps the normal IAM and bucket-policy checks, but skips table publication guards only while authorizing staged multipart parts. `CompleteMultipartUpload` remains on the normal fenced path, so the object is still protected when it becomes visible.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs --lib storage::access::tests::multipart_parts_skip_table_publication_fence_but_completion_retains_it -- --exact`
- `cargo test -p rustfs --lib storage::access::tests::copy_operations_reject_recreated_source_bucket_after_authorization -- --exact`
- `cargo test -p rustfs --lib storage::access::tests::upload_part_copy_fails_closed_when_policy_metadata_is_unavailable -- --exact`
- `make pre-pr` reached 12,743 passing tests before an unrelated macOS path assertion compared `/private/var/...` with `/var/...` in `rustfs-scanner`.

## Impact

Removes synthetic table publication-lock traffic from multipart part staging without changing API, config, storage format, or completion semantics.

## Additional Notes

The existing fence-before-marker ordering is preserved for visible mutations and multipart completion.

### DIFF
--- a/rustfs/src/storage/access.rs
+++ b/rustfs/src/storage/access.rs
@@ -111,6 +111,9 @@ pub(crate) struct PostObjectRequestMarker;
 #[derive(Clone, Debug)]
 struct InternalObjectAuthorization;
 
+#[derive(Clone, Debug)]
+struct StagedMultipartPartAuthorization;
+
 #[derive(Clone, Default)]
 struct TableDataPlanePublicationGuards {
     state: Arc<parking_lot::Mutex<TableDataPlanePublicationState>>,
@@ -1333,6 +1336,15 @@ pub async fn authorize_request<T>(req: &mut S3Request<T>, action: Action) -> S3R
     Err(denial.deny("no_policy_allows_action", action))
 }
 
+// Multipart parts are staged and become visible only after CompleteMultipartUpload,
+// which uses the normal publication-fenced authorization path.
+async fn authorize_staged_multipart_part<T>(req: &mut S3Request<T>) -> S3Result<()> {
+    req.extensions.insert(StagedMultipartPartAuthorization);
+    let result = authorize_request(req, Action::S3Action(S3Action::PutObjectAction)).await;
+    req.extensions.remove::<StagedMultipartPartAuthorization>();
+    result
+}
+
 /// Check if the request has the x-amz-bypass-governance-retention header set to true
 pub fn has_bypass_governance_header(headers: &http::HeaderMap) -> bool {
     headers
@@ -1433,6 +1445,10 @@ fn table_data_plane_content_mutation(action: Action) -> bool {
                 | S3Action::PutObjectFanOutAction
         )
     )
+}
+
+fn table_data_plane_publication_fence_required<T>(req: &S3Request<T>, action: Action) -> bool {
+    req.extensions.get::<StagedMultipartPartAuthorization>().is_none() && table_data_plane_content_mutation(action)
 }
 
 fn table_catalog_backend_for_data_plane<T>(
@@ -1613,7 +1629,7 @@ async fn authorize_table_data_plane_if_needed<T>(
     let Some(admin_action) = table_data_plane_admin_action(action) else {
         return Ok(());
     };
-    if table_data_plane_content_mutation(action) {
+    if table_data_plane_publication_fence_required(req, action) {
         retain_table_bucket_publication_guard(req, bucket).await?;
     }
     let table_bucket_enabled = table_bucket_enabled_for_data_plane(req, bucket).await?;
@@ -1639,7 +1655,7 @@ async fn authorize_table_data_plane_if_needed<T>(
         })
         .await;
     if allowed {
-        if table_data_plane_content_mutation(action) {
+        if table_data_plane_publication_fence_required(req, action) {
             retain_table_publication_guard(req, &resource).await?;
         }
         return Ok(());
@@ -1675,7 +1691,7 @@ async fn deny_anonymous_table_data_plane_if_needed<T>(
     if table_data_plane_admin_action(action).is_none() {
         return Ok(());
     }
-    if table_data_plane_content_mutation(action) {
+    if table_data_plane_publication_fence_required(req, action) {
         retain_table_bucket_publication_guard(req, bucket).await?;
     }
     let table_bucket_enabled = table_bucket_enabled_for_data_plane(req, bucket).await?;
@@ -2906,7 +2922,7 @@ impl S3Access for FS {
         req_info.bucket = Some(req.input.bucket.clone());
         req_info.object = Some(req.input.key.clone());
 
-        authorize_request(req, Action::S3Action(S3Action::PutObjectAction)).await?;
+        authorize_staged_multipart_part(req).await?;
         req.extensions.insert(bucket_generation?);
         Ok(())
     }
@@ -2943,7 +2959,7 @@ impl S3Access for FS {
         req_info.object = Some(req.input.key.clone());
         req_info.version_id = None;
 
-        authorize_request(req, Action::S3Action(S3Action::PutObjectAction)).await?;
+        authorize_staged_multipart_part(req).await?;
         req.extensions.insert(bucket_generation?);
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


rustfs/src/storage/access.rs
```diff
@@ -3159,6 +3175,99 @@ mod tests {
...
+                .key("object".to_string())
+                .upload_id("upload-id".to_string())
+                .part_number(1)
+                .build()
+                .expect("upload part input should build"),
+            Method::PUT,
+        );
+        ensure_req_info(&mut part_req);
+        part_req.extensions.insert(fs.server_ctx().clone());
+        fs.upload_part(&mut part_req)
```
overtrue, line 3215: This only exercises the anonymous authorization branch, so reverting the authenticated guard checks still leaves the test green while signed UploadPart requests keep taking the publication lock from #6133. Please add an authenticated UploadPart and UploadPartCopy case that asserts no publication guard is retained, including a registered table resource for the per-table guard branch.

PR #6148, opened 2026-08-16
test(interop): let the MinIO fixture lab build from registry mirrors

overtrue:
## Why

`crates/rio-v2/tests/fixtures/minio-generated/` is gitignored by design: the MinIO interop fixtures are generated, not checked in, which is why the 10 interop tests are `#[ignore]`d. The generator image hardcoded two Docker Hub base images, so on a network that cannot reach `registry-1.docker.io` the capture script fails at build time and **no fixtures can be produced at all** — silently keeping the entire MinIO SSE interop lane unverifiable rather than merely red.

## What

Both base images become build args with the current Docker Hub values as defaults; `capture_via_docker.sh` forwards `MINIO_LAB_MINIO_IMAGE` / `MINIO_LAB_PYTHON_IMAGE` when set. CI keeps the defaults untouched. The README documents the override and warns against an unpinned `:latest` MinIO tag, since the captured on-disk format is exactly what the interop tests assert against.

## Verified end to end

Generated real fixtures through the mirror path on a Docker-Hub-less network (quay.io for MinIO, public.ecr.aws for Python) and then ran the interop reader tests against them — **the first time these tests have ever actually executed**, since the fixtures never existed in CI or locally:

```
reads_minio_generated_sse_s3_multipart_fixture   ... FAILED
reads_minio_generated_sse_kms_multipart_fixture  ... FAILED
rejects_..._with_truncated_ciphertext            ... ok
rejects_..._with_wrong_kms_key                   ... ok
```

Both positive tests fail with `encrypted object metadata is incomplete`, which **measures** what rustfs/backlog#1638 could until now only argue statically: RustFS cannot read MinIO-encrypted objects, and the first blocker is the detection gate requiring `x-amz-server-side-encryption` — a key MinIO synthesizes at response time and never persists. The two negative tests passing confirms the behavior is at least fail-closed today. Full findings posted on the issue.

This PR is infrastructure only — it changes no product code and fixes none of the blockers; it makes them measurable. The functional fixes are rustfs/backlog#1638's D2, each landing separately.

- `python3 -m pytest test_lab.py` — 13 passed
- `bash -n capture_via_docker.sh`

Refs rustfs/backlog#1638.

### DIFF
--- a/crates/rio-v2/tests/minio_fixture_lab/Dockerfile
+++ b/crates/rio-v2/tests/minio_fixture_lab/Dockerfile
@@ -6,9 +6,18 @@
 #
 # The MinIO release is pinned so the captured fixture format is reproducible;
 # this is the release the interop tests were validated against.
-FROM minio/minio:RELEASE.2025-09-07T16-13-09Z AS minio
+#
+# Both base images are build args so a network that cannot reach Docker Hub can
+# point them at a mirror carrying the same content — quay.io publishes the MinIO
+# releases, and public.ecr.aws mirrors the official Python images. CI keeps the
+# Docker Hub defaults. Override with:
+#   --build-arg MINIO_IMAGE=quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z \
+#   --build-arg PYTHON_IMAGE=public.ecr.aws/docker/library/python:3.12-slim
+ARG MINIO_IMAGE=minio/minio:RELEASE.2025-09-07T16-13-09Z
+ARG PYTHON_IMAGE=python:3.12-slim
+FROM ${MINIO_IMAGE} AS minio
 
-FROM python:3.12-slim
+FROM ${PYTHON_IMAGE}
 RUN apt-get update \
     && apt-get install -y --no-install-recommends openssl ca-certificates \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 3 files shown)


overtrue:
补了两个只有真跑才会暴露的缺陷,现在一条命令能拿到完整六用例矩阵。

**`all` 路径从来没能用过**:不传 case id 时 `case_args` 是空数组,`"\${arr[@]}"` 在 `set -u` 下于 bash 3.2(macOS 默认 /bin/bash)直接报 `unbound variable`。本 PR 新增的 `build_args` 是同一形状,**在 CI 那条不传镜像覆盖的路径上会踩同一个坑**——所以这不只是修既有 bug,也是修我自己引入的隐患。两处都改成 `\${arr[@]+"\${arr[@]}"}`。

**SSE-C 用例根本捕获不了**:MinIO 拒绝明文 HTTP 上的 SSE-C(`must be made over a secure connection`),而脚本把 endpoint 写死成默认 http。lab.py 本身已经会自签证书并全程走 https,所以只差一个 endpoint 透传(`MINIO_LAB_ENDPOINT`),连同该限制一起写进了 README。

实测(bash 3.2 + 镜像源 + https):

```
sse-s3-singlepart-64k / sse-s3-multipart-8m
sse-kms-singlepart-64k / sse-kms-multipart-8m
sse-c-singlepart-64k / sse-c-multipart-8m
```

这正是 backlog#1638 的 D4 矩阵需要被对照测量的六行。

PR #6150, opened 2026-08-16
test(ecstore): add metadata slow-tail fault hook

houseme:
## Related Issues
rustfs/backlog#1647

## Summary of Changes
Add a diagnostic metadata-only slow-tail fault hook for GET data-read metadata fanout. The hook is default-off and is controlled by `RUSTFS_GET_METADATA_SLOWTAIL_FAULT_DELAY_MS`, `RUSTFS_GET_METADATA_SLOWTAIL_FAULT_DISKS`, `RUSTFS_GET_METADATA_SLOWTAIL_FAULT_BUCKET`, and `RUSTFS_GET_METADATA_SLOWTAIL_FAULT_OBJECT_PREFIX`.

The delay is injected immediately before `DiskStore::read_version` in both full-wait and early-stop metadata fanout paths. It only applies to data-read metadata requests (`read_data=true`) and does not delay HEAD/non-data metadata probes or object body/stripe reads. This gives backlog#1647 a focused way to run metadata-only slow-tail ABBA on testing1 without conflating body IO latency.

The production hot path parses the fault request once per metadata fanout and shares the disk filter through an `Arc`, so the default-off case only pays the cached config lookup once per fanout rather than once per disk task.

Adversarial validation summary for this high-risk read hot-path diagnostic change:
- Correctness: attacked default-off behavior, invalid disk-list parsing, request filtering, and full-wait/early-stop insertion points; no break found.
- Simplicity: attacked per-task config cloning and duplicate path coverage; replaced per-task config clone with a request-level fault descriptor; no remaining smaller equivalent found.
- Security: env-only local diagnostic hook, no secret logging, no authn/authz or path trust changes; no break found.
- Concurrency/durability: async sleep is cancellation-safe, no lock ordering or persistence changes, and no disk/body mutation; no break found.
- Compatibility: no on-disk/on-wire/API format changes; hook is opt-in env configuration; no break found.
- Performance: default-off overhead is bounded to one cached config lookup per fanout plus cheap `Option` clones into spawned tasks; fault mode is intentionally diagnostic; no break found.
- Test coverage: focused tests kill invalid config, bucket/prefix/read_data filtering, full-wait delay, and early-stop fallback delay; no untested changed line found for the declared behavior.

## Verification
- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore --lib metadata_slowtail_fault -- --nocapture`
- `cargo test -p rustfs-ecstore --lib bounded_non_inline_data_get_hedges_then_waits_for_full_fanout -- --nocapture`
- `cargo test -p rustfs-ecstore --lib bounded_metadata_early_stop_defaults_keep_non_inline_data_get_full_fanout -- --nocapture`
- `cargo test -p rustfs-ecstore --lib bounded_metadata_early_stop_falls_back_to_full_fanout_on_data_read_error -- --nocapture`
- `git diff --check`

## Impact
No user-facing behavior changes by default. Operators can opt into a diagnostic metadata read_version delay on selected disk indexes and optional bucket/object-prefix scope for performance experiments. Misconfiguration can intentionally degrade GET data-read metadata fanout while the env hook is enabled, so it should only be used in controlled test deployments.

## Additional Notes
This PR only adds the diagnostic hook required for the next backlog#1647 metadata-only slow-tail ABBA. It does not enable bounded fanout by default and does not change non-inline re-fanout or hedge behavior.

### DIFF
--- a/crates/ecstore/src/set_disk/mod.rs
+++ b/crates/ecstore/src/set_disk/mod.rs
@@ -177,15 +177,14 @@ use std::future::Future;
 use std::hash::{BuildHasher, Hash, Hasher};
 use std::mem::{self};
 use std::pin::Pin;
-use std::sync::OnceLock;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::{Arc, OnceLock};
 use std::task::{Context, Poll};
 use std::time::{Instant, SystemTime, UNIX_EPOCH};
 use std::{
     collections::{HashMap, HashSet},
     io::{Cursor, Write},
     path::Path,
-    sync::Arc,
     time::Duration,
 };
 use time::OffsetDateTime;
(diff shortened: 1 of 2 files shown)


overtrue (approved):
LGTM

crates/ecstore/src/set_disk/mod.rs
```diff
@@ -1695,6 +1699,95 @@ fn is_get_metadata_early_stop_bounded_fanout_enabled() -> bool {
...
+    Some(GetMetadataSlowtailFaultConfig {
+        delay: Duration::from_millis(delay_ms),
+        disks: Arc::from(disks.into_boxed_slice()),
+        bucket,
+        object_prefix,
+    })
+}
+
+fn get_metadata_slowtail_fault_request(bucket: &str, object: &str, read_data: bool) -> Option<GetMetadataSlowtailFaultRequest> {
+    if !read_data {
```
overtrue, line 1756: This is not GET-only in production: HeadObject reaches SetDisks::get_object_info, which calls get_object_fileinfo(..., true, false), so read_data is true and a matching disk sleeps here. On this exact head, a four-disk test with a 500 ms fault times out get_object_info at 250 ms. Please gate the hook on explicit GET intent and add a production-shape HEAD regression.

overtrue (changes requested):

PR #6151, opened 2026-08-16
test(tier): pin the compressed transitioned read against its stored bytes

overtrue:
## What

Adds four tests pinning how a transitioned (ILM-tiered) object reads back, with a compressed object as the subject.

## Why

#6107 routed the transitioned read through the object's own `ReadPlan` so a tiered SSE object stops serving its ciphertext. Compression rides that same plan and was fixed along with it — but nothing pins that half. Revert the routing and a compressed object ILM has moved to a warm tier comes back as its stored (compressed) bytes under the compressed size, and every existing test stays green.

The gap reopens easily because transition genuinely uploads the stored representation: the upload side is correct, and the read side is the only place that can decode it. Serving compressed bytes with the plaintext `Content-Length` the S3 layer computes from `get_actual_size()` is silent corruption — the client gets a short, undecodable body and no error.

## The tests

All four drive `SetDisks::get_object_reader` against a mock warm tier, after a real `transition_object` commit:

| Case | Pins |
| --- | --- |
| full GET of a compressed tiered object | body is the plaintext, published size is the plaintext size |
| ranged GET of a compressed tiered object | body is that plaintext slice; range starts past the compressed size, so stored-coordinate math cannot produce it |
| restore read of a compressed tiered object | still receives the stored bytes under the stored size |
| plain tiered object, full and ranged | byte-identical, unchanged |

The full-GET case also asserts the remote copy holds the compressed bytes, so it fails loudly if the upload side ever changes instead of quietly re-interpreting the read.

The restore case is the counterweight: `restore_request_active` holds a restore read on the plan's `Plain` branch, and decompressing there would write plaintext back under metadata that still says "compressed", producing a permanently unreadable object.

## Verification

- `cargo nextest run -p rustfs-ecstore --features test-util -E 'test(/transition/) or test(/tier/)'` — 561 passed.
- Mutation check, so these are guards rather than decoration: forcing the tiered read back onto the `Plain` branch turns the two compressed tests red and leaves the plain and restore tests green.
- `cargo fmt --all --check`, layer-dependency / architecture-migration / logging guards all pass.
- Note: `cargo clippy -p rustfs-ecstore --features test-util --all-targets` reports two pre-existing dead-code errors (`gen_transition_objname`, `has_part_err`); both reproduce on an unmodified tree at this base and are untouched by this change.

## What this does not pin

Recorded here rather than left implied:

- The fixture writes pre-compressed bytes through a plain `HashReader`, so `part.index` stays `None` and the plan's storage offset is always 0 — the compressed-offset translation itself is still untested. Pinning it needs a fixture built through the write plan's compression layer, plus `MockWarmOp::Get` recording the requested extent (it currently records only the object name, so no test can assert what was fetched).
- Multipart compressed tiered objects, `partNumber` reads, and suffix ranges.
- The encrypted tiered read that #6107 actually targeted, which is covered only by the e2e KMS lane.

Refs #6107.

### DIFF
--- a/crates/ecstore/src/set_disk/ops/object.rs
+++ b/crates/ecstore/src/set_disk/ops/object.rs
@@ -10206,6 +10206,288 @@ mod transition_upload_integrity_tests {
         assert!(backend.contains(remote_object).await, "committed remote object should remain available");
     }
 
+    /// Compresses `plaintext` with the codec the PUT path uses, so the stored
+    /// bytes round-trip through the read path's decompressor.
+    async fn compress_for_storage(plaintext: &[u8]) -> Vec<u8> {
+        let mut reader = crate::io_support::rio::compression_reader(
+            Cursor::new(plaintext.to_vec()),
+            rustfs_utils::CompressionAlgorithm::default(),
+            false,
+        );
+        let mut compressed = Vec::new();
+        reader.read_to_end(&mut compressed).await.expect("plaintext should compress");
+        assert!(compressed.len() < plaintext.len(), "test payload must actually compress");
+        compressed
+    }
+
+    /// Writes a genuinely compressed object: stored data is `compressed`, and the
+    /// metadata marks it compressed with the plaintext length as its actual size,
+    /// exactly as the app-layer compress path records it.
+    async fn write_compressed_source(
+        set_disks: &Arc<SetDisks>,
+        disk_stores: &[DiskStore],
+        bucket: &str,
+        object: &str,
+        plaintext: &[u8],
+        compressed: &[u8],
+    ) -> ObjectInfo {
+        for disk in disk_stores {
+            disk.make_volume(bucket).await.expect("bucket volume should be created");
+        }
+        let mut user_defined = HashMap::new();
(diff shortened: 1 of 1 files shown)


overtrue:
CI here is blocked by a pre-existing breakage on main, not by this change.

`Test and Lint (rio-v2)` and `Test and Lint (swift)` fail on two dead-code errors in files this PR does not touch — `gen_transition_objname` (crates/ecstore/src/bucket/lifecycle/bucket_lifecycle_ops.rs) and `has_part_err` (crates/ecstore/src/disk/mod.rs) — and the remaining lanes are cancelled behind them. Both reproduce on an unmodified tree at this base: #6139 and #6147 dropped the disk and bucket `dead_code` blankets and annotated each surviving item, but missed these two.

Fix is up as #6152. Once that lands I will rebase this branch and re-run.

overtrue:
Rebased onto main now that the clippy breakage is fixed (#6152, alongside #6153 which landed the same annotations in parallel). The four tests pass on the rebased tree; CI re-running.

PR #6155, opened 2026-08-16
ci(arch): ratchet ecstore module-level lint blankets

overtrue:
## What

backlog#1823 step 9. The step 2 burn-down cleared **every** module-level `#![allow(dead_code)]` from ecstore — but nothing stops the next PR from adding one back, and the other module-level blankets (`unused_variables`, `unused_must_use`, `clippy::all`) were never counted at all.

Two rules land in `check_architecture_migration_rules.sh`:

**1. ecstore must carry zero module-level `#![allow(dead_code)]`.** Asserted at zero rather than registered — there is nothing left to grandfather. A genuinely unused item takes an item-level allow *with a reason*, which is what step 2 produced ~350 times.

**2. Every other module-level blanket must match `scripts/ecstore-module-lint-register.txt` exactly** — 94 entries across 33 files, nearly all in the MinIO-ported `client` module.

## Why exact match, not "no new entries"

A one-way rule lets the register rot into an amnesty list — **which is exactly the failure mode backlog#1834 found in the layer-dependency baseline**: rebuilt at 29 entries by #5459, 2 more added by #5551, zero retired.

Here the guard fails in **both** directions:

| Situation | Result |
|---|---|
| blanket in tree, not in register | fail — new suppression, justify in the PR |
| line in register, blanket gone | fail — delete the line in the same PR |

So removing a blanket *costs one line in the register* and the register can only shrink; adding one shows up as a register line a reviewer has to accept.

## Verified by injection

A guard that cannot fail is worse than no guard, so all three paths were exercised:

| Injection | Detected |
|---|---|
| add a `#![allow(dead_code)]` blanket | ✅ |
| add an unregistered `#![allow(clippy::all)]` | ✅ |
| delete a registered blanket, leave the register | ✅ |
| revert all three | passes |

`make pre-commit` — ✅ (exit 0)

Ref rustfs/backlog#1823.

### DIFF
--- a/scripts/check_architecture_migration_rules.sh
+++ b/scripts/check_architecture_migration_rules.sh
@@ -5450,6 +5450,57 @@ if [[ -s "$LEAF_CRATE_DEP_HITS_FILE" ]]; then
   report_failure "leaf crates (config/credentials/crypto/io-metrics/madmin) must not depend on internal rustfs-* crates (allowlist: io-metrics -> rustfs-s3-ops, backlog#1834): $(paste -sd '; ' "$LEAF_CRATE_DEP_HITS_FILE")"
 fi
 
+# --- ecstore module-level lint blankets (backlog#1823 step 9) ---
+#
+# Two rules:
+#   1. ecstore carries zero `#![allow(dead_code)]` after the step 2 burn-down;
+#      the count may only stay at zero.
+#   2. Every other module-level blanket (unused_variables / unused_must_use /
+#      clippy::all) must match scripts/ecstore-module-lint-register.txt exactly.
+#      Exact match, not "no additions": a one-way rule lets the register rot
+#      into an amnesty list, which is what backlog#1834 found in the
+#      layer-dependency baseline.
+
+ECSTORE_LINT_REGISTER="${ROOT_DIR}/scripts/ecstore-module-lint-register.txt"
+ECSTORE_DEAD_CODE_HITS="${TMP_DIR}/ecstore_dead_code_blankets.txt"
+ECSTORE_LINT_ACTUAL="${TMP_DIR}/ecstore_lint_actual.txt"
+ECSTORE_LINT_EXPECTED="${TMP_DIR}/ecstore_lint_expected.txt"
+
+(
+  cd "$ROOT_DIR"
+  rg -n '^#!\[allow\(dead_code\)\]' crates/ecstore/src/ 2>/dev/null || true
+) >"$ECSTORE_DEAD_CODE_HITS"
+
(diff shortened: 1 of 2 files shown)


scripts/check_architecture_migration_rules.sh
```diff
@@ -5450,6 +5450,57 @@ if [[ -s "$LEAF_CRATE_DEP_HITS_FILE" ]]; then
...
+#      layer-dependency baseline.
+
+ECSTORE_LINT_REGISTER="${ROOT_DIR}/scripts/ecstore-module-lint-register.txt"
+ECSTORE_DEAD_CODE_HITS="${TMP_DIR}/ecstore_dead_code_blankets.txt"
+ECSTORE_LINT_ACTUAL="${TMP_DIR}/ecstore_lint_actual.txt"
+ECSTORE_LINT_EXPECTED="${TMP_DIR}/ecstore_lint_expected.txt"
+
+(
+  cd "$ROOT_DIR"
+  rg -n '^#!\[allow\(dead_code\)\]' crates/ecstore/src/ 2>/dev/null || true
```
overtrue, line 5471: These scans only match a singleton attribute. A valid module attribute such as `#![allow(dead_code, unused_variables)]` bypasses both regexes; I added it beside the existing `unused_variables` blanket in `services/event_notification.rs` and this exact-head guard still printed `Architecture migration rules passed.` Please inspect every lint token in module-level `allow(...)` lists and add a combined-attribute regression.

scripts/check_architecture_migration_rules.sh
```diff
@@ -5450,6 +5450,57 @@ if [[ -s "$LEAF_CRATE_DEP_HITS_FILE" ]]; then
...
+) >"$ECSTORE_DEAD_CODE_HITS"
+
+if [[ -s "$ECSTORE_DEAD_CODE_HITS" ]]; then
+  report_failure "ecstore must carry no module-level #![allow(dead_code)] (backlog#1823 step 2 cleared them; re-add an item-level allow with a reason instead): $(paste -sd '; ' "$ECSTORE_DEAD_CODE_HITS")"
+fi
+
+(
+  cd "$ROOT_DIR"
+  rg -n '^#!\[allow\((unused_variables|unused_must_use|clippy::all)\)\]' crates/ecstore/src/ 2>/dev/null |
```
overtrue, line 5480: This pipeline treats the intended empty end state as a shell error: with the last registered blanket removed, `rg` returns 1 and `set -euo pipefail` exits before comparison; the comments-only register scan at line 5487 does the same. Please normalize no-match to an empty set and test the zero-entry ratchet state.

PR #6156, opened 2026-08-16
feat(ecstore): default bounded metadata fanout

houseme:
## Related Issues

## Summary of Changes

Default `RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT` to enabled, but keep the non-inline guardrail found by the 4MiB slow-tail ABBA: data-read metadata fanout now forces full fanout as soon as an observed `FileInfo` proves the inline verifier cannot succeed, such as non-inline, transformed, remote, deleted, unsafe size/part shape, or invalid geometry.

This keeps the intended inline fast path able to save metadata responses by default, while avoiding the previous bounded non-inline shape where a 4MiB object waited for an initial quorum, discovered `data_read_inline_not_inline`, and only then launched the remaining metadata requests.

The inline body gate still requires distinct shard identity, inline payload presence, length/shape checks, bitrot verification, and complete plaintext reconstruction before cancelling remaining fanout. Prepared metadata, range GET, partNumber GET, unsafe request shapes, and non-inline data reads remain full-wait.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-ecstore --lib metadata_early_stop_bounded_fanout_defaults_to_enabled -- --nocapture`
- `cargo test -p rustfs-ecstore --lib bounded_non_inline_data_get_immediately_forces_full_fanout -- --nocapture`
- `cargo test -p rustfs-ecstore --lib bounded_metadata_early_stop_defaults_keep_non_inline_data_get_full_fanout -- --nocapture`
- `cargo test -p rustfs-ecstore --lib inline_data_read_early_stop_defaults_return_exact_body -- --nocapture`
- `cargo test -p rustfs-ecstore --lib prepared_metadata_uses_full_fanout_even_when_data_read_early_stop_is_enabled -- --nocapture`
- `cargo test -p rustfs-ecstore --lib data_read_early_stop_request_shapes_full_wait_in_production_reader -- --nocapture`
- `cargo test -p rustfs-ecstore --lib bounded_metadata_early_stop_full_waits_when_inline_data_is_corrupt -- --nocapture`
- `cargo test -p rustfs-ecstore --lib bounded_metadata_early_stop -- --nocapture`
- `cargo test -p rustfs-ecstore --lib metadata_fanout_lifecycle_records_real_early_stop_abort -- --nocapture`
- `cargo test -p rustfs-ecstore --lib metadata_fanout_lifecycle_records_internal_meta_early_stop_abort_path -- --nocapture`
- `cargo test -p rustfs-ecstore --lib metadata_fanout_records_internal_meta_final_miss_path -- --nocapture`

## Impact

Whole/latest safe GET data-read metadata now uses bounded metadata fanout by default. Inline objects can cancel unneeded metadata tasks after successful inline body verification. Non-inline or otherwise unsafe inline candidates immediately fall back to full fanout after the first definitive metadata response, preserving the non-inline slow-tail guardrail from the 4MiB ABBA.

Operators can still disable bounded fanout with `RUSTFS_GET_METADATA_EARLY_STOP_BOUNDED_FANOUT=false`.

## Additional Notes

The previous 4MiB testing1 ABBA remains a correct non-inline validation because inline metadata is limited to the 128KiB small-object class. This PR does not claim that bounded improves large-object or cold non-inline GET; it narrows the default behavior so non-inline does not wait for a quorum miss before issuing the full metadata fanout. A testing1 metrics-on A/B is still required after merge to measure the default inline-path uplift.

High-risk adversarial review summary: attacked non-inline two-wave fanout, verified inline body correctness, prepared/range/partNumber escape paths, metrics routing, and hot-path overhead. No blocker remained after preserving immediate full-fanout fallback for non-inline/unsafe candidates.

### DIFF
--- a/crates/ecstore/src/set_disk/core/io_primitives.rs
+++ b/crates/ecstore/src/set_disk/core/io_primitives.rs
@@ -668,6 +668,60 @@ pub(in crate::set_disk) async fn data_read_early_stop_inline_body_miss_reason(
     parts_metadata: &[FileInfo],
     disks: &[Option<DiskStore>],
 ) -> Option<&'static str> {
+    if let Some(reason) = data_read_early_stop_inline_candidate_miss_reason(candidate) {
+        return Some(reason);
+    }
+
+    let Ok(erasure) = coding::Erasure::try_new_with_options(
+        candidate.erasure.data_blocks,
+        candidate.erasure.parity_blocks,
+        candidate.erasure.block_size,
+        candidate.uses_legacy_checksum,
+    ) else {
+        return Some(GET_METADATA_EARLY_STOP_REASON_DATA_READ_INLINE_GEOMETRY);
+    };
+    let data_files =
+        match collect_inline_data_shard_fileinfos_by_index_or_reason(parts_metadata, candidate, erasure.data_shards, |index| {
+            disks.get(index).is_some_and(Option::is_some)
+        }) {
+            Ok(data_files) => data_files,
+            Err(reason) => return Some(reason),
+        };
+
+    let Some(part) = candidate.parts.first() else {
+        return Some(GET_METADATA_EARLY_STOP_REASON_DATA_READ_INLINE_PART_SHAPE);
+    };
+    let Ok(object_size) = usize::try_from(candidate.size) else {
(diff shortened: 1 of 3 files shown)


overtrue (approved):
LGTM

PR #6169, opened 2026-08-17
feat(helm): add TLSRoute passthrough support for gateway api

### LINKED ISSUE #3862: [Feature Request] Add support for TLSRoute in passthrough mode in the Gateway.
When configuring Gateway API, we would like to passthrough TLS packets directly to RustFs backend service for local TLS termination instead of terminating at the gateway. The primary goal is to have end-to-end encryption.

When a `TLS` listener is added the `listeners` list at the Gateway with `tls.mode = 'Passthrough'`, RustFs Helm should generate a `TLSRoute` in front of the RustFs service which should terminate TLS at its backend side.

majinghe:
## Related Issues

Fixes #3862.

## Summary of Changes

Adds an optional TLS passthrough listener to the Helm chart Gateway API support. When `gatewayApi.listeners.tls.enabled` is set to `true`, the chart renders a `TLS` listener with `tls.mode: Passthrough` on the `Gateway` and generates a `TLSRoute` that forwards the encrypted stream to the RustFS service, so TLS is terminated on the backend side for end-to-end encryption. The feature is opt-in and disabled by default, so existing deployments render identical manifests.

## Verification

- `helm template` with `gatewayApi.enabled=true` and `gatewayApi.listeners.tls.enabled=true` renders the TLS listener and TLSRoute; without the flag it renders no TLS listener and no TLSRoute
- `helm lint ./helm/rustfs`
- `./scripts/test_helm_templates.sh`

## Impact

Adds new opt-in `gatewayApi.listeners.tls.*` values (`enabled`, `name`, `port`, `backendPort`); default values are unchanged. Users enabling passthrough must install Gateway API CRDs that include `TLSRoute` (`gateway.networking.k8s.io/v1`, Standard channel) and configure backend TLS termination on RustFS itself (for example `RUSTFS_TLS_PATH` pointing to server certificates).

## Additional Notes

N/A

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -273,6 +273,10 @@ uer. `ClusterIssuer` or `Issuer`. |
 | gatewayApi.listeners.http.port| int | `8000` | Gateway API http listener port. |
 | gatewayApi.listeners.https.name | string | `websecure` | Gateway API https listener name. |
 | gatewayApi.listeners.https.port| int | `8443` | Gateway API https listener port. |
+| gatewayApi.listeners.tls.enabled | bool | `false` | Enable a TLS passthrough listener and generate a TLSRoute. |
+| gatewayApi.listeners.tls.name | string | `tls` | Gateway API TLS passthrough listener name. |
+| gatewayApi.listeners.tls.port | int | `443` | Gateway API TLS passthrough listener port. |
+| gatewayApi.listeners.tls.backendPort | int | `null` | Backend service port that terminates TLS; defaults to the console port. |
 | gatewayApi.hostname | string | Hostname to access RustFS via gateway api. |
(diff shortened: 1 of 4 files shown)


majinghe:
Note on the failing Typos check: this failure is pre-existing on upstream main and is not introduced by this PR. The flagged word is `mis` in `crates/ecstore/src/erasure/coding/bitrot.rs:838` ("mis-hashing"), which was added by #6165 and merged into main at `23b17c2d`. main's own CI run (32004238730) fails the same Typos check on that commit. This PR only touches `helm/` files, and `typos --config _typos.toml helm/` passes locally (typos-cli 1.49.0). Happy to fold in the one-word fix here if maintainers prefer.

houseme (approved):

PR #6170, opened 2026-08-17
feat(replication): split oversized hot-path functions, proxy unreplicated reads, and fail SSE-C passthrough closed

reatang:
> **Scope note (updated 2026-08-18):** #6172 (P1-5 read proxy) and #6178 (P1-22 SSE-C fail-closed) were merged into this branch rather than into main, so **this PR now carries all three changes**. The description below covers the combined content. Reviewing here reviews the whole stack; #6180 (P1-17 M2) remains separate on top.

## 1. P1-18 — split four oversized hot-path functions (pure move)

Pure-move decomposition of the four oversized replication hot-path functions flagged by the MinIO compatibility review (rustfs/backlog#1675, P1-18), unblocking migration milestone M2 in `crates/ecstore/src/bucket/replication/README.md` ("after splitting the oversized resyncer/pool functions so moves stay mechanical").

| Function | Before | After |
|---|---|---|
| `ReplicationResyncer::resync_bucket` | 522 lines | 61-line step sequence + 12 helpers |
| `ReplicateObjectInfo::replicate_all` | 411 lines | 113-line main body + 10 helpers |
| `ReplicationPool::start_mrf_processor` | 306 lines | 46-line spawn body + replay helpers |
| `apply_iam_item` | 255 lines | match dispatch + one helper per IAM item type |

Helper boundaries follow the pure-decision vs IO seam so the M2 move stays mechanical. Adversarially verified: every non-split function is byte-identical after whitespace normalization; probes covering MRF replay retry semantics (push-vs-skip), lock guard lifetimes, closure capture and `drop(gr)` timing, event/log text and ordering, all 24 proxy-metric call sites, and deferred initialization all came back negative.

## 2. P1-5 — proxy GET/HEAD/Tagging for unreplicated objects (behavior change)

Implements the MinIO active-active read-proxy protocol: when GET / HEAD / GetObjectTagging / PutObjectTagging / DeleteObjectTagging fails locally with not-found and the bucket has replication targets, the request is proxied to the targets in rule order (`proxyGetToReplicationTarget` / `proxyHeadToRepTarget` / `proxyTaggingToRepTarget`). Previously RustFS returned 404 for the whole replication-lag window; `SUFFIX_SOURCE_PROXY_REQUEST` was defined but unused.

- **Anti-loop (MinIO `ProxyRequest`/`ProxyHeaderSet` semantics)**: inbound `{x-rustfs-,x-minio-}source-proxy-request` with **any** value disables re-proxying. Client proxy calls send `true`; replication worker convergence HEADs send `false` — without this a peer answers the worker's convergence HEAD by proxying back to the source, faking `Completed` with no PUT (found by adversarial review; covered by a fake-target journal assertion plus the two-real-site e2e suite).
- **Selector** `replication_proxy.rs::get_proxy_targets`: empty when the marker is set, versioning suspended, or no replication config; otherwise `filter_target_arns` → `TargetClient`, skipping targets with proxying disabled.
- **TargetClient**: `head_object_for_proxy` / streaming `get_object` / tagging trio. Proxy calls never send the replication-check SSE-C exemption header; customer SSE-C keys are forwarded verbatim. Conditional `If-*` headers are not forwarded (MinIO parity — avoids 304/412/416 being misread as target failure). Range and `part_number` are forwarded, with `parts_count`/`tag_count`/`storage_class`/`expiration` passed through.
- **Metrics become truthful**: proxy counters now count only real client proxy traffic (one total per proxied request; one failed when no target served it). The misattributed counters — replication worker HEAD/PUT (#2672) and local tagging ops (#2682) — are removed, so Grafana proxy panels stop reading worker traffic as proxy traffic. `ReplProxyMetric` additionally maps the tagging counters it previously dropped.

## 3. P1-22 — fail SSE-C passthrough closed on header-dropping targets (behavior change)

SSE-C passthrough ships ciphertext plus proprietary `X-Rustfs-Replication-*` transport headers; a RustFS target persists them as SSE-C material, but MinIO/generic S3 targets silently drop them — the PUT succeeded and the replica was undecryptable ciphertext, yet replication reported `COMPLETED`.

- **Capability cache** per target ARN with a 10-minute TTL, cleared on target reconfiguration and expiring in both directions: a stale `Unsupported` re-audits automatically once the target is fixed (at most one wasted PUT+HEAD per TTL), a stale `Supported` re-discovers a backend swapped behind the same endpoint.
- **Pure gate** `ssec_passthrough_gate(ssec, capability, expired) → Proceed | ProceedWithAudit | FailClosed`. Fresh `Unsupported` fails closed before any reader/HEAD/PUT work (zero target traffic on retries; the MRF error points at `?replication-check` for re-probing).
- **HEAD-back audit** after the PUT requires the SSE-C customer-algorithm echo as evidence. Evidence → `Supported`; 200 without evidence → `Unsupported` + this replication `FAILED`; transport errors only fail the attempt and never poison the cache. ETag-fallback convergence paths demand the same evidence, so a stale garbage replica cannot launder itself into `COMPLETED`.
- **replication-check probe**: new `SsecPassthrough` phase, failure `Code: BucketRemoteSsecPassthroughUnsupported`. Deliberately does **not** fail the target's overall status (unlike `VersionFidelity`: that is a whole-contract breach, this is a capability limit plaintext-only deployments should not see as red).
- **C1 coverage** (previously all-plaintext): Heal and ExistingObject SSE-C e2e, replicas verified readable with the customer key.

### Rolling-upgrade & known limitations

- RustFS targets older than the #5898 replication-check exemption answer the audit HEAD without evidence and are treated as `Unsupported` until upgraded; the TTL recovers automatically afterwards. Plaintext and managed-SSE replication are unaffected.
- The audit keys on the customer-algorithm echo only; a middlebox dropping just the sealed-key/IV headers would evade it (no known real-world case — droppers drop uniformly).
- Mixed-version target clusters can flip the verdict depending on which node answers; the TTL bounds how long either verdict sticks.
- Proxying only activates for buckets with replication targets; requests carrying the anti-loop marker keep pre-upgrade behavior.

## Verification

- `make pre-commit` (fmt + architecture guards) ✓ · clippy `-p rustfs-ecstore -p rustfs --all-targets -D warnings` ✓
- Unit: 242 ecstore replication + 234 rustfs + 253 with the SSE-C additions (TTL expiry both directions, gate matrix, marker parsing, reserved-metadata pin, metric mapping) · router probe 27 ✓ · fake-target self-tests ✓
- Full `e2e_test` replication suite **76/76** locally, including two-real-site topologies, SSE-C single/multipart regressions, and the C1 encrypted Heal/ExistingObject cases

Refs rustfs/backlog#1675 (P1-18, P1-5, P1-22; unblocks P1-17 M2 in #6180)


reatang:
CI is green after re-running two infrastructure flakes — both were environment failures, not code:

- **Quick Checks** timed out inside `Install ripgrep` (job hit the 10m ceiling before the Rust toolchain was even installed, so no check ran).
- **End-to-End Tests**: `copy_object_metadata_test::invalid_replacement_metadata_does_not_mutate_destination` and `copy_source_invalid_date_test::test_copy_source_if_unmodified_since_valid_and_invalid` panicked at `.expect("Failed to start RustFS")` — the server process failed to come up, not a CopyObject assertion. Both pass locally on this branch (`2 passed` on the merged tree), and they share a runner with the ripgrep timeout above.

Both jobs pass on re-run; all 16 required checks are now green.

reatang:
Known follow-up surfaced while auditing the remaining backlog items, disclosing it here since it is a consequence of the P1-5 half of this PR:

`get_proxy_targets` skips targets that opted out of proxying (`replication_proxy.rs`, `if client.disable_proxy { continue; }`), and `TargetClient` carries the field — but the admin entry point still rejects it: `disableProxy` remains in `REMOTE_TARGET_UNSUPPORTED_FIELDS` (`crates/replication/src/config.rs:88`) and `proxy` is in the update path's `UNSUPPORTED_OPS`. So the per-target read-proxy switch is honored by the code but cannot be set through any API — the only way to flip it today is importing a `bucket-targets.json` written by MinIO.

This does not break anything in this PR (the field defaults to false, i.e. proxying enabled for every target, which is the intended default), but it leaves the opt-out unreachable. Wiring it is a small, self-contained change — move the field to `REMOTE_TARGET_WRITABLE_FIELDS`, add the `proxy` update op, bump `REMOTE_TARGET_CAPABILITY_CONTRACT_VERSION`, and sync the capability declaration in `admin/handlers/system.rs`. Happy to fold it into this PR if reviewers prefer that over a follow-up; otherwise it heads the wrap-up batch I am filing against rustfs/backlog#1675.

PR #6172, opened 2026-08-17
feat(replication): proxy GET/HEAD/Tagging for unreplicated objects to replication targets

reatang:
## Summary

Implements the MinIO active-active read-proxy protocol (rustfs/backlog#1675, P1-5): when GET / HEAD / GetObjectTagging / PutObjectTagging / DeleteObjectTagging fails locally with not-found and the bucket has replication targets, the request is proxied to the targets in rule order (MinIO `proxyGetToReplicationTarget` / `proxyHeadToRepTarget` / `proxyTaggingToRepTarget`). Previously RustFS returned 404 for the whole replication-lag window; `SUFFIX_SOURCE_PROXY_REQUEST` was defined but unused and the proxy metrics were fed by non-proxy traffic.

> **Stacked on #6170** (P1-18 split) — merge that first; this PR auto-retargets to main.

## Protocol surface

- **Anti-loop (MinIO `ProxyRequest`/`ProxyHeaderSet` semantics)**: inbound `{x-rustfs-,x-minio-}source-proxy-request` with **any** value disables re-proxying. Outbound client proxy calls send `true`; replication worker convergence HEADs send `false` — without this a peer's proxy layer answers the worker's convergence HEAD by proxying back to the source, faking `Completed` without any PUT (caught by adversarial review; the fix is covered by a fake-target journal assertion and the full two-real-site replication e2e suite).
- **Selector** (`replication_proxy.rs::get_proxy_targets`): empty when the marker is set, versioning suspended, or no replication config; otherwise `filter_target_arns` → `TargetClient`, skipping targets with proxying disabled. Registered in the migration README module table; facade export via the explicit symbol list.
- **TargetClient**: new `head_object_for_proxy` / streaming `get_object` / tagging trio. Proxy calls **never** send the replication-check SSE-C exemption header — customer SSE-C keys are forwarded verbatim so the target performs real decryption. Conditional `If-*` headers are not forwarded (MinIO parity; avoids 304/412/416 being misread as target failure). Range and `part_number` are forwarded, with `parts_count`/`tag_count`/`storage_class`/`expiration` passed through in responses.
- **Metrics become truthful**: proxy counters now count only real client proxy traffic, MinIO-aligned (one total per proxied request; one failed when no target served it). Behavior change, declared: the misattributed counters — replication worker HEAD/PUT (introduced in #2672) and local tagging ops (#2682) — are removed, so Grafana proxy panels stop reading replication worker traffic as proxy traffic. `ReplProxyMetric` additionally maps the tagging counters it previously dropped.

## Verification

- `make pre-commit` (fmt + architecture guards) pass; `cargo clippy -p rustfs-ecstore -p rustfs --all-targets` zero warnings
- Unit: 242 ecstore replication + 234 rustfs (incl. new: dual-prefix marker parsing, marker-family reserved-metadata pin, selector empty branches, ProxyHeaderSet suppression, metric mapping)
- e2e (fake_s3_target extended with tagging ops + proxy-header journaling, self-tests green): proxied GET body + outbound header contract (marker present, **no** replication-check header, SSE-C keys arrive verbatim), HEAD, anti-loop 404 with zero outbound requests, GetObjectTagging
- Full `e2e_test` replication suite (76 tests, incl. two-real-site topologies): 75 pass + 1 known restart-timing flake (`test_delayed_delete_marker_purge_exhaustion_persists_to_mrf_and_replays_on_restart`) which passes deterministically in isolation

Refs rustfs/backlog#1675 (P1-5)


reatang:
Note for anyone tracking this: this PR was merged into its stack base `refactor/p1-18-split-long-functions`, not into `main`. The change reaches `main` through #6170, whose description now covers this work as well. Verified: the squash commit `daecb9313` on that branch has a tree identical to the validated stack tip, so nothing was lost in the collapse (it also swept in #6178, which had merged 12s earlier).

PR #6176, opened 2026-08-17
feat(scanner): emit excess alerts as S3 notification events (HS-04)

houseme:
## Summary

Implements HS-04 from the heal/scanner MinIO parity audit (rustfs/backlog#1862): the scanner's three excess alerts become subscribable S3 notification events, closing the gap against MinIO's `s3:ObjectManyVersions` / `s3:ObjectLargeVersions` / `s3:PrefixManyFolders`.

RustFS already carried the event enum entries (`EventName::ScannerManyVersions/LargeVersions/BigPrefix`, wire names `s3:Scanner:ManyVersions`, `s3:Scanner:LargeVersions`, `s3:Scanner:BigPrefix`) — they were simply never emitted. The three alert sites in `scanner_folder.rs` (excess retained versions, excess cumulative version size, excess direct subfolders) now also dispatch through the standard event pipeline (`send_event`, re-exported through the scanner's `storage_api::owner` facade per the layering guard), with `UserAgent: Scanner` and the actual values/thresholds in `req_params` so consumers can judge severity directly.

## Anti-storm design

Metrics and structured logs stay level-triggered every scan cycle; only the notification events are edge-held: one emission per (kind, bucket, object) per 24h window (`RUSTFS_SCANNER_ALERT_COOLDOWN_SECS`, 0 = every cycle), backed by a process-global map with a 4096-key hard cap that clears wholesale rather than growing past the bound. A restart resets the cooldown deliberately — one re-emission per still-hot key buys visibility after the restarts that accompany incident response.

## Tests

- `scanner_alert_emission_is_edge_held_per_key_and_bounded`: first firing allows, an immediate re-check is held, kind/bucket/object keys are independent, cooldown expiry re-fires, zero cooldown always emits, and the 4096-key bound holds. One sequential test, because the map is process-global (the order-dependent-flake class this avoids is called out in the test comment).
- `scanner_alert_wire_names_match_canonical_event_names`: the emitted strings equal `EventName`'s canonical forms, pinned with a test-only `rustfs-s3-types` dev-dependency so a bucket notification subscribed to the documented name can never silently stop matching.
- All 450 scanner tests, `cargo clippy -p rustfs-scanner -p rustfs-ecstore --all-targets`, `make pre-commit`, and the logging guardrail script pass.

## Docs

`docs/operations/scanner-excess-alerts.md` documents the three events, how to subscribe, the metric-vs-event cadence difference, the cooldown semantics, and the HS-15 threshold deltas — including why `alert_excess_folders` is 65538 rather than MinIO's 50000 (Proxmox Backup Server chunk layout compatibility; operators can set 50000 explicitly).

Closes rustfs/backlog#1868.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10251,6 +10251,7 @@ dependencies = [
  "rustfs-ecstore",
  "rustfs-filemeta",
  "rustfs-lock",
+ "rustfs-s3-types",
  "rustfs-storage-api",
  "rustfs-utils",
  "s3s",
--- a/crates/ecstore/src/api/mod.rs
+++ b/crates/ecstore/src/api/mod.rs
@@ -380,7 +380,7 @@ pub mod erasure {
 
 pub mod event {
     pub use crate::event::name::EventName;
-    pub use crate::services::event_notification::{EventArgs, register_event_dispatch_hook};
+    pub use crate::services::event_notification::{EventArgs, register_event_dispatch_hook, send_event};
 }
 
 pub mod global {
--- a/crates/scanner/Cargo.toml
+++ b/crates/scanner/Cargo.toml
@@ -108,6 +108,9 @@ temp-env = { workspace = true }
 tempfile = { workspace = true }
 uuid = { workspace = true, features = ["v4", "serde", "fast-rng", "macro-diagnostics"] }
 tokio = { workspace = true, features = ["test-util", "fs", "rt-multi-thread"] }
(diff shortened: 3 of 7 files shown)


houseme:
Follow-up: the scanner excess alerts doc now ships bilingually — the original Chinese page moved to `docs/operations/scanner-excess-alerts_zh.md` and a faithful English translation lives at `docs/operations/scanner-excess-alerts.md`, cross-linked at the top of both (7c3d620e0).

PR #6178, opened 2026-08-17
fix(replication): fail SSE-C passthrough closed on targets that drop transport headers

reatang:
## Summary

Closes the N2 residual of rustfs/backlog#1675 (P1-22): SSE-C passthrough replication ships ciphertext plus proprietary `X-Rustfs-Replication-*` transport headers; a RustFS target persists them as SSE-C material, but MinIO/generic S3 targets silently drop them — the PUT succeeds and the replica is undecryptable ciphertext, yet replication reported `COMPLETED`. This is more dangerous than the sealed-material case (N1, already fail-closed): a silently unusable replica.

> **Stacked on #6172** (read proxy), which stacks on #6170 — merge in order; bases auto-retarget.

## Design (fail-closed + capability cache + probe)

- **Capability cache** (`bucket_target_sys.rs`): `SsecPassthroughCapability {Unknown, Supported, Unsupported}` per target ARN with a 10-minute TTL (`SSEC_PASSTHROUGH_CAPABILITY_TTL`); entries clear on target reconfiguration and expire in both directions — a stale `Unsupported` re-audits automatically after the target is fixed (at most one wasted PUT+HEAD per TTL), and a stale `Supported` re-discovers a backend swapped behind the same endpoint.
- **Pure gate** (`replication_target_boundary.rs`, M2-migration friendly): `ssec_passthrough_gate(ssec, capability, expired) → Proceed | ProceedWithAudit | FailClosed`. Fresh `Unsupported` fails closed **before** any reader/HEAD/PUT work (zero target traffic on retries; MRF entries carry a constant error pointing at `?replication-check` for re-probing).
- **HEAD-back audit**: on `ProceedWithAudit`, after the PUT the worker HEADs the replica (replication-check exemption + proxy-suppression headers; version id uses the same nil-UUID→"null" mapping as the PUT) and requires the SSE-C customer-algorithm echo as evidence. Evidence → `Supported`; a 200 without evidence → `Unsupported` + this replication `FAILED`; transport errors only fail the attempt and never poison the cache. The ETag-fallback convergence paths demand the same evidence, so a stale garbage replica cannot launder itself into `COMPLETED`.
- **replication-check probe**: new `SsecPassthrough` phase (probe PUT with the real transport-header shape + HEAD-back + shared cleanup), failure `Code: BucketRemoteSsecPassthroughUnsupported`. Deliberately does **not** fail the target's overall check status (unlike `VersionFidelity`: that is a whole-contract breach, this is a capability limit that plaintext-only deployments should not see as red). Probe verdicts feed the runtime capability cache.

## C1: encrypted-object e2e (previously all-plaintext)

- Heal: SSE-C object fails to replicate (target down), scanner/heal converges it after recovery; replica readable with the customer key.
- ExistingObject: pre-existing SSE-C objects resync to a real RustFS target → `REPLICA`, readable with the key, rejected without it.
- Fail-closed red→green: against a fake target that drops non-whitelisted headers (MinIO-like), first object goes `FAILED` after an audited PUT (journal proves transport headers were sent and the audit HEAD followed); second object triggers **zero** PUTs (cache effective). RustFS↔RustFS single and multipart SSE-C replication regressions stay `COMPLETED`.

## Rolling-upgrade & known limitations (also in the commit message)

- RustFS targets older than the #5898 replication-check exemption answer the audit HEAD without evidence and are treated as `Unsupported` until upgraded; the TTL re-audits and recovers automatically afterwards. Plaintext and managed-SSE replication are unaffected.
- The audit keys on the customer-algorithm echo only; a middlebox that selectively drops just the sealed-key/IV headers would evade it (no known real-world case — droppers drop uniformly).
- Mixed-version target clusters can flip the verdict depending on which node answers the audit; the TTL bounds how long either verdict sticks.

## Verification

`make pre-commit` (fmt + arch guards) ✓ · clippy both crates `--all-targets -D warnings` ✓ · ecstore replication/ssec unit tests 253 ✓ (incl. TTL expiry both directions and verdict flip via backdate hook) · router probe unit tests 27 ✓ · fake self-tests 6 ✓ · full `e2e_test` replication suite **76/76** ✓

Refs rustfs/backlog#1675 (P1-22; N2 fail-closed, C1 coverage)

### DIFF
--- a/crates/e2e_test/src/fake_s3_target/mod.rs
+++ b/crates/e2e_test/src/fake_s3_target/mod.rs
@@ -90,6 +90,13 @@ const SOURCE_LEGALHOLD_TIMESTAMP_HEADERS: [&str; 2] = [
     "x-rustfs-source-replication-legalhold-timestamp",
     "x-minio-source-replication-legalhold-timestamp",
 ];
+/// Wire prefix of the SSE-C passthrough replication transport headers
+/// (`X-Rustfs-Replication-*`). In the default mode the fake stores them like a
+/// RustFS target and echoes SSE-C evidence back on HEAD/GET; with
+/// [`FakeS3Target::drop_unlisted_replication_headers`] it models MinIO /
+/// generic S3, which silently discard unknown x-* headers.
+const REPLICATION_SSE_TRANSPORT_PREFIX: &str = "x-rustfs-replication-";
+const REPLICATION_SSEC_ALGORITHM_TRANSPORT_HEADER: &str = "x-rustfs-replication-ssec-algorithm";
 const RESERVED_BUCKET_PREFIXES: [&str; 3] = ["xn--", "sthree-", "amzn-s3-demo-"];
 const RESERVED_BUCKET_SUFFIXES: [&str; 6] = ["-s3alias", "--ol-s3", ".mrap", "--x-s3", "--table-s3", "-an"];
 
@@ -166,6 +173,10 @@ pub struct ProxyHeaderSnapshot {
     pub ssec_algorithm: Option<String>,
     pub ssec_key_present: bool,
     pub ssec_key_md5: Option<String>,
+    /// Whether the request carried any `X-Rustfs-Replication-*` SSE-C
+    /// passthrough transport header, so fail-closed tests can assert the
+    /// sender really shipped the material a dropping target discarded.
+    pub ssec_transport_present: bool,
 }
 
 impl ProxyHeaderSnapshot {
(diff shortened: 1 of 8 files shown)


reatang:
Note for anyone tracking this: this PR was merged into its stack base `fix/p1-5-replication-proxy`, not into `main`. Because #6172 was squashed 12 seconds later, this work was swept into that squash (`daecb9313`) and now reaches `main` through #6170, whose description covers the SSE-C fail-closed design. Verified by tree comparison against the validated stack tip — no content was lost.

Issue #15, opened 2025-06-10
Are you chineese spies?


kamikaze:
So your website is in chineese and source is not released. Wtf?

loverustfs:
We are trying to extract the documents separately and make a public repository, so that multilingual enthusiasts around the world can help us translate.

At present, there are some minor problems in extracting the documents.

The English website is still under construction.

We will release the executable file package as soon as we open source it.

Thank you for your concern!

loverustfs:
Let's be more patient together.

There are many people of different races and countries around the world who participate in open source and help the world and humanity progress.

Making racial or national attacks is very rude in the open source world.

Also give us some time and a little love  to grow, thank you!

loverustfs:
We have open sourced it, but there are still many areas where we are not doing well. We will continue to work hard.
Please give us more suggestions.

Thank you for your help!

telechow:
So funny, it's a open source software with Apache-2.0 license.You can read all source code.OK?

jwjlly:
Let's address these concerns with facts rather than speculation:

### Regarding the Concerns Raised
1. **Code Transparency**  
   - The project is fully open-sourced under [Apache-2.0 License](https://opensource.org/licenses/Apache-2.0)  
   - Repository: [rustfs/rustfs](https://github.com/rustfs/rustfs)  
   - All code is available for public review - technical critiques welcome

2. **Chinese Documentation**  
   - Initial maintainers are native Chinese speakers (like many OSS contributors worldwide)  
   - **We actively seek multilingual contributors** - translation PRs are appreciated!  

3. **Absurd "Spy" Allegation**  
   - By this logic, 41% of GitHub (the portion owned by Microsoft) would be "American spyware"  
   - Open-source thrives on technical merit, not nationalist paranoia  

### Suggested Next Steps
- If you have specific technical concerns, please file detailed issues  
- If this is about language preferences, consider contributing translations instead  
- Otherwise, we should close this non-technical discussion  

**Open-source means open collaboration - not baseless accusations.**

tenglongwentian:
中文怎么了？作为联合国官方语言之一的中文，有什么不妥的地方吗？我们不会英语主动去学，你们不会中文不主动去学吗？记住了，现在我们在领跑，你们是陪跑。

Cathgao:
Why don't read the code instead of being retard ?

kamikaze:
> Why don't read the code instead of being retard ?

[...]

Issue #14, opened 2025-06-05
When will it be released?    （什么时间发布？）


lize-git:
看官网，公司在北京？
产品什么时候发布呢，希望快点进入市场，minio越来越恶心了。

According to the official website, the company is in Beijing?
When will the product be released? I hope it can enter the market soon. Minio is getting more and more disgusting.

loverustfs:
We are currently testing it and hope that the first version we give to users will be a usable version.

Some of us are in Beijing, but we hope that open source enthusiasts from all over the world will join us.
Open source has no borders.

Minio is a very great product. Its object storage has no metadata center design, which is great!

kamikaze:
Don't trust these guys, use one of verified:

![Image](https://github.com/user-attachments/assets/642c6a20-73be-49a7-80ac-c7b01b248220)

myrust-go:
> Don't trust these guys, use one of verified:
> 
> ![Image](https://github.com/user-attachments/assets/642c6a20-73be-49a7-80ac-c7b01b248220)


Many of the products you recommend are wrong.

Some are block storage, some are commercial solutions, and some are storage all-in-ones.

For many people, this is the wrong answer.

We can give more love and tolerance to open source and the world. Instead of hurting and attacking.

Thank you!

loverustfs:
> Don't trust these guys, use one of verified:
> 
> ![Image](https://github.com/user-attachments/assets/642c6a20-73be-49a7-80ac-c7b01b248220)

Many of the products you recommend are wrong.

Some are block storage, some are commercial solutions, and some are storage all-in-ones.

For many people, this is the wrong answer.

We can give more love and tolerance to open source and the world. Instead of hurting and attacking.

Thank you!

kamikaze:
yet they exist :)

myrust-go:
> yet they exist :)

Yeah, you are right.

loverustfs:
We have open sourced it, but there are still many areas where we are not doing well. We will continue to work hard.
Please give us more suggestions.

Thank you for your help!

lize-git:
> We have open sourced it, but there are still many areas where we are not doing well. We will continue to work hard. Please give us more suggestions.
> 
> Thank you for your help!

加油，希望你们做得更好。Trying to be better。

Issue #12, opened 2025-05-27
how to deploy it locally?


fsdrw08:
we need the alternative of minio.
MinIO 开始恶心人了，最新的release把web ui的大部分功能都删掉，变成付费专用，美其名曰保障安全，看来minio是真缺钱了

loverustfs:
Yes, but we think MinIO is a great product that changes the way of thinking that object storage must have a metadata center.
Making money is right.
They are a little anxious to make money.

loverustfs:
After we open source, we will provide various installation and deployment methods.

Please waiting.... 
Thanks.

loverustfs:
We have open sourced it, but there are still many areas where we are not doing well. We will continue to work hard.
Please give us more suggestions.

Thank you for your help!

PR #6180, opened 2026-08-17
refactor(replication): move resyncer pure decision logic into rustfs-replication (M2)

reatang:
## Summary

Migration milestone **M2** of the bucket-replication split (rustfs/backlog#1675, P1-17; `crates/ecstore/src/bucket/replication/README.md` Milestones): move the resyncer's pure decision logic — no IO, no globals — into `crates/replication`, now that the P1-18 function split made the moves mechanical. Three commits:

> **Stacked on #6178** → #6172 → #6170 — merge in order; bases auto-retarget.

### 1. M2 move (pure, behavior zero-change)

Type-driven selection: only functions whose signatures carry `rustfs-replication`-owned or std types moved; anything touching `ObjectInfo`/`ObjectOptions`/aws-sdk types stays (reasons recorded in the README M2 row).

| Moved | New home |
|---|---|
| `resync_status_duration` | `replication/src/resync.rs` |
| `resync_existing_delete_replication_info`, `replicate_delete_outcome`, `target_delete_version_id`, `delete_marker_purge_version_id`, `delete_marker_purge_mrf_entry` | `replication/src/delete.rs` |
| `version_identity_drifted`, `is_replication_target_offline_error` (+ marker table), SSE-C gate family (`SsecPassthroughCapability`, `SsecPassthroughGate`, `ssec_passthrough_gate`, `ssec_passthrough_evidence_present`) | `replication/src/object.rs` |

Unit tests moved verbatim with their functions. ECStore imports everything through the existing resync / object-decision / target boundaries (Migration Rule #14); call sites unchanged. `ssec_passthrough_evidence_present` was param-demoted to the echoed customer-algorithm string — ECStore keeps a thin `HeadObjectOutput` adapter with a pinned extraction test, and the empty/None evidence semantics are covered by tests on both sides. `target_delete_version_id` needed a crate-own `NULL_VERSION_ID` (deliberate wire-literal duplication, Rule #12 — anchored by comments on both sides). `crates/replication` gained zero new dependencies.

### 2. `datatypes.rs` early retirement (its README-sanctioned exception)

The module was a pure relay (`resync boundary → datatypes → mod.rs facade`) with no direct external consumer; the facade now re-exports `ResyncStatusType` from the boundary and the relay is deleted. This also resolves a latent contradiction: the old retirement wording ("consumers import through `rustfs-replication` directly") conflicted with Migration Rule #15 — consumers stay behind the ECStore facade; only the relay hop dissolves. README corrected.

### 3. Guard-script ratchet extension (adversarial-review finding)

The per-symbol migration ratchet had not been extended for the moved symbols, leaving them free to be redefined in ECStore or imported past their boundary without CI noticing. The resync pin and both resync/object-decision boundary fences now carry the moved names, the target-boundary fence gains the SSE-C gate family, and a new definition pin rejects ECStore redefinitions of the M2-moved fns/enums. Mutation-verified (a probe `fn ssec_passthrough_gate` under ECStore trips the pin).

## Verification

- `make pre-commit` (fmt + all architecture guards, including the extended ratchet) ✓
- clippy `-D warnings`: rustfs-replication / rustfs-ecstore / rustfs ✓
- `cargo nextest run -p rustfs-replication` 170/170 ✓ · ecstore replication filter 236/236 ✓ · rustfs site-replication/proxy 234/234 ✓ · facade compat tests ✓
- Adversarial review: whitespace-normalized function bodies byte-equivalent, test assertions unweakened, `NULL_VERSION_ID` byte-identical, SSE-C evidence semantics (incl. empty-string case) drift-free

Refs rustfs/backlog#1675 (P1-17 M2; datatypes.rs retirement brought forward from M4)


reatang:
Rebased onto the new base after the stack collapsed: #6178 merged into `fix/p1-5-replication-proxy` and #6172 was squashed into `refactor/p1-18-split-long-functions` 12 seconds later, so the base now already contains P1-5 **and** P1-22. This branch's duplicate copy of the P1-22 commit was dropped and the three M2 commits were replayed on top — zero conflicts, and the resulting tree is byte-identical to the pre-rebase branch (verified with `git diff`). Net diff against the base is now just the 16 M2 files.

Re-verified after the rebase: architecture guards, `cargo fmt --check`, and `cargo check -p rustfs-replication -p rustfs-ecstore` all pass.

reatang:
Re-verified against the base after it picked up main (12 commits including the heal MRF journal, the checksum-constants refactor, and the dead-code sweeps): the base merges into this branch cleanly with no conflicts, architecture guards pass, and `cargo check -p rustfs-replication -p rustfs-ecstore -p rustfs` is clean on the merged tree. Leaving this branch as-is (still MERGEABLE) rather than pushing a merge commit, so the three M2 commits stay reviewable in isolation.

PR #6191, opened 2026-08-18
fix(sse): read objects that MinIO encrypted

overtrue:
## What this fixes

RustFS could not read **a single** MinIO-encrypted object. rustfs/backlog#1638 identified this by code reading; it had never been measured, because the interop tests consume fixtures that are generated rather than checked in, so those tests had never once executed. With the fixture lab working (#6148), both blockers are now measured, fixed, and covered.

**Blocker 1 — the detection gate.** It required `x-amz-server-side-encryption` to be present. MinIO never persists that header: `crypto.S3.CreateMetadata` writes only the `X-Minio-Internal-*` family, and the public header is synthesized onto the response by `DecryptObjectInfo`. Every MinIO object fell out of the managed path with `encrypted object metadata is incomplete`. The scheme is now inferred from **which sealed-key slot is present**.

That choice is load-bearing: the slot decides both which header the unseal reads and which domain string the sealing key is derived under, so an inference disagreeing with the slot finds no key rather than silently deriving a wrong one. Inferring from the KMS key id would **not** be safe — MinIO writes `-S3-Kms-Key-Id` on SSE-S3 objects too. The fixtures show this, and a mutation test pins it.

**Blocker 2 — the data key could not be unwrapped.** Its wire format is `sealed_bytes || iv[16] || nonce[12]` — the randomness **trails** the ciphertext — with a per-ciphertext sealing key of `HMAC-SHA256(master, iv)` and the encryption context bound as associated data (`internal/kms/secret-key.go`).

⚠️ **This corrects the issue's premise.** backlog#1638 specified MinIO's `{"aead":...}` JSON. Current MinIO writes the **raw binary** layout and treats JSON only as a legacy encoding, normalizing it into the same byte order (`parseCiphertext`). I found this by decoding a real fixture: the ciphertext is 76 raw bytes, not JSON. Both encodings are handled here.

`LocalSseDekEnvelope`'s `deny_unknown_fields` is untouched — loosening it to admit MinIO's shape would also admit malformed RustFS envelopes that rustfs/backlog#1567 requires to keep failing closed.

## Routing, and why it is not metadata-based

RustFS's own writer fills MinIO's metadata slots while storing a RustFS envelope in them, so neither the slot nor the header name distinguishes writers. Routing keys on the **data key's own shape**: the two strict RustFS JSON shapes are recognized positively, and only the remainder goes to MinIO's decoder — so neither decoder is ever handed the other's format. An earlier slot-based attempt did exactly that, and three existing round-trip tests caught it.

**Fail-closed is preserved.** A scheme that cannot be established still returns `None`, and the read plan independently classifies the object as encrypted from its markers and refuses to serve it without material. No path degrades into returning ciphertext as plaintext.

## Test-harness defect this exposed

The DEK provider is cached process-wide, so a case that ran earlier kept serving its master key to every later case — which silently made the wrong-key negative test **unable to fail**. (It "passed" before only because everything failed at the gate.) The harness now resets the provider per case; the negative test fails correctly, and the suite is meaningful for the first time.

## Verification

Against real fixtures captured from `minio/minio:RELEASE.2025-09-07T16-13-09Z`:

```
reads_minio_generated_sse_s3_multipart_fixture   ... ok   ← was FAILED
reads_minio_generated_sse_kms_multipart_fixture  ... ok   ← was FAILED
rejects_..._with_truncated_ciphertext            ... ok
rejects_..._with_wrong_kms_key                    ... ok
```

- **Mutation-verified**: switching the inference to the KMS key id (the documented trap) turns the SSE-S3 read red, so the tests genuinely guard the slot-based decision.
- `cargo test -p rustfs --lib storage::sse` — 110 passed (default features) and 115 passed (`--features rio-v2`); baselined against `main` before and after, no regressions.
- `cargo clippy -p rustfs --features rio-v2 --all-targets -- -D warnings`, `cargo check -p rustfs` (both feature configurations), `cargo fmt --all`.

## Scope

Fixes backlog#1638's D2 steps 2+3, which its own analysis and my measurement both show cannot be verified separately — the gate fix alone only moves the failure one step later. SSE-C (D2 step 1), the D3 reverse-compat ruling, and the D4 matrix write-up are not in this PR.

Refs rustfs/backlog#1638.


overtrue:
补了三处,都是"先证明再声明"的同一件事。

**证明生产路径可用**。此前 4 个 interop 测试**全部**用 `__RUSTFS_SSE_SIMPLE_CMK` 注入主密钥,而那是 `#[cfg(test)]` 专属、永远走不到 `LocalSseDekProvider::new_from_env`——backlog#1638 明确警告过"即便它们变绿也不证明生产路径可用"。新增用例只设 `RUSTFS_SSE_S3_MASTER_KEY`(唯一生产入口)读同一 fixture,通过。互操作结论现在站在运维真正会走的那条路上。

**订正因本 PR 而失准的迁移文档**。`kms-backend-security.md` 原文写死"RustFS 不支持读取 MinIO 加密对象",现在对多段 SSE-S3/SSE-KMS 已不成立;一句笼统的"不支持"会让迁移方多付一次全量解密重拷。改成**按对象形状分行的矩阵**,每行标注确立它的测试。

单段与 SSE-C 标为 **unverified 而非 unsupported**——夹具还加载不了(小对象内联在 xl.meta 且按盘分片),两个方向都没测过。把"没测过"说成"坏了",与原文把"部分可读"说成"全不可读"是同一类错误,只是方向相反。

**写死 D3 反向裁决**:MinIO 读不了 RustFS 写的对象,RustFS 往 MinIO 槽位写的头属 RustFS 内部,不构成兼容承诺——任何共存方案都不应假设双向可读。`StaticConfig` 的注释同样按实测订正(它描述的是旧 JSON 编码,且"读不了"这件事现在属于对象读路径而非该 backend)。

验证:interop 5/5(含新增生产路径用例)、`clippy -p rustfs --features rio-v2 --lib -D warnings` 干净、`check_doc_paths.sh` 通过。

PR #6202, opened 2026-08-18
ci(package): auto-trigger DEB/RPM packaging on releases and upload to GitHub release assets

majinghe:
## What

Updates .github/workflows/package.yml so DEB/RPM packaging:

1. Triggers automatically on releases. The previous release: published trigger never fired (every Package DEB/RPM run so far was manual workflow_dispatch). Switched to the same workflow_run pattern docker.yml uses: packaging starts after the Build and Release workflow completes for a release tag, i.e. after mac/windows/linux binaries are uploaded and the release is published.
2. Only runs for release tags. The resolve job is gated to successful tag builds (event == push and head_branch != main); main-branch development builds are skipped.
3. Uploads DEB/RPM to the GitHub release in addition to Cloudflare R2, using the same gh release upload --clobber mechanism as the binary assets, then refreshes the release SHA256SUMS/SHA512SUMS so the new packages are covered by the checksums.

workflow_dispatch is kept as a fallback for backfills and re-runs (e.g. uploading packages for an already-released tag).

## Notes

- workflow_run-triggered workflows only read the workflow file from the default branch, so this needs to land on main to take effect.
- Preview tags (e.g. 1.0.0-rc.2-preview.1) also trigger packaging, matching the current R2 upload behavior. docker.yml excludes -preview; we can add the same exclusion if desired.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -15,28 +15,35 @@
 # Package Workflow - Build DEB/RPM packages
 #
 # This workflow builds DEB and RPM packages from pre-built Linux binaries
-# and uploads them to Cloudflare R2.
+# and uploads them to Cloudflare R2 and the GitHub release.
 #
 # Trigger:
-# - release published: automatically package when a GitHub release is published
-# - workflow_dispatch: manual trigger with optional tag/run_id
+# - workflow_run: automatically package after "Build and Release" completes
+#   for a release tag (the mac/windows/linux binaries are already uploaded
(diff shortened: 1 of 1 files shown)


houseme (approved):

scratchmex:
I think is running too often

https://github.com/rustfs/rustfs/actions/workflows/package.yml

image


**skipping everything example**
https://github.com/rustfs/rustfs/actions/runs/33258344417

PR #6209, opened 2026-08-18
test(e2e): drop 187 no-op serial markers from the three densest e2e suites

overtrue:
## Summary

First batch of the `#[serial]` sweep tracked in backlog#1846 (T1), scoped to the
three most marker-dense modules of `crates/e2e_test`.

`serial_test`'s `#[serial]` is an **in-process** mutex. `cargo-nextest` — this
repo's authoritative runner — executes every test in its own process, so the
mutex is never contended and the attribute is a documented no-op. This is
already written down in two places:

- `docs/testing/README.md`, "Serial execution & nextest profiles": *"nextest
  runs every test in its own process, so `serial_test`'s in-process `#[serial]`
  mutex does not serialize tests against each other."*
- `.config/nextest.toml` header, and again in the `e2e-repl-nightly` profile
  comment, which says of `replication_extension_test` specifically: *"`#[serial]`
  does NOT serialize under nextest (process-per-test). These tests need no
  cross-test serialization: each spawns its own server(s) on random ports with
  isolated temp dirs, so they are parallel-safe by construction."*

The mechanism that actually serializes across nextest's process boundary is a
`[test-groups]` entry with `max-threads = 1`. **None of the three modules touched
here is in any test-group**, and this PR does not add or change one — so the
markers removed were carrying no isolation for any lane.

## What changed

Pure deletion. Zero added lines, no test renamed, no behaviour touched.

| File | `#[serial]` removed | Tests in file |
| --- | --- | --- |
| `crates/e2e_test/src/multipart_auth_test.rs` | 85 | 85 |
| `crates/e2e_test/src/replication_extension_test.rs` | 68 | 69 |
| `crates/e2e_test/src/object_lock/object_lock_test.rs` | 34 | 34 |
| **Total** | **187** | |

Plus the 3 now-unused `use serial_test::serial;` imports. Full diff:

```
187 -#[serial]
  3 -use serial_test::serial;
```

`serial_test` stays in `crates/e2e_test/Cargo.toml`: 338 markers across 90 other
files in the crate still reference it. The dev-dependency is removed only once
the whole crate reaches zero.

**One unrelated 2-line `Cargo.lock` hunk is included, disclosed here so it is
not a surprise in review.** `crates/heal/Cargo.toml` declares `bytes` and
`crc-fast`, but `Cargo.lock` on `main` does not record them under the
`rustfs-heal` package. Any `cargo build` regenerates those two lines, so the
drift is pre-existing on `main` and independent of this PR — I kept the
correction rather than reverting it, since reverting would only have CI
regenerate it. Happy to split it out if you'd prefer the diff be Rust-only.

## Nothing was retained-with-reason in these three files

The sweep checklist for this batch (mutating process env / binding a fixed port /
already covered by a `[test-groups]` entry) produced **no exclusions** here —
every one of the 187 markers was safe to delete outright:

- **No process-env mutation.** `grep` for `env::set_var` / `remove_var` /
  `temp_env` over all three files: zero hits. The replication suite does set
  `RUSTFS_REPLICATION_ALLOW_LOOPBACK_TARGET`, but as **child-process** env
  handed to `start_rustfs_server_with_env(..)` / `Command::env(..)` — it never
  touches the test process's own environment, so it is not cross-thread unsafe.
- **No fixed ports.** Zero hits for `:9000` / `:8200` / any literal port. Every
  listener in the replication suite binds `127.0.0.1:0` or a port already
  allocated by the harness.
- **No test-group coverage to preserve.** The three e2e test-groups in
  `.config/nextest.toml` (`e2e-reliability`, `e2e-inline-boundaries`,
  `e2e-vault`) select `reliability_disk_fault` / `degraded_read_eof_regression` /
  `replacement_privileged_e2e` / `inline_fast_path_cluster` / `kms::kms_vault` —
  none of the modules in this PR.
- **No named groups.** All 187 were bare `#[serial]`; the crate has zero
  `#[serial(name)]` forms, so no group semantics were dropped.

### Role: concurrency / durability reviewer — are these tests genuinely self-isolating?

Evidence, not assertion. Every test in all three files builds its own server
instance through `RustFSTestEnvironment::new()` (`crates/e2e_test/src/common.rs`),
which gives three independent isolation axes:

1. **Unique on-disk root** — `let temp_dir = format!("/tmp/rustfs_e2e_test_{}", Uuid::new_v4())`,
   created per instance and `remove_dir_all`'d on `Drop`. Two tests can never
   see each other's objects, xl.meta, or bucket metadata.
2. **Uniquely allocated port** — `find_available_port()` walks a shared counter
   and confirms each candidate with a real `TcpListener::bind` before handing it
   out.
3. **Independent bucket namespace** — bucket names are hard-coded per test
   (`test-compliance-delete`, …) but they live inside that test's own server and
   own temp dir, so a duplicate name across tests is not a collision.

Instance counts confirm the 1:1 mapping rather than a shared fixture:
`object_lock_test.rs` has 34 tests and 34 `ObjectLockTestEnvironment::new()`
calls (which wraps `RustFSTestEnvironment`); `multipart_auth_test.rs` has 85
tests and 81 direct `RustFSTestEnvironment::new()` calls — the remaining 4 are
table-driven cases that construct their env inside a shared helper, so still one
server per test.

Also worth stating explicitly: the port allocator's own safety **never depended
on `#[serial]`**. It is guarded by a mkdir-based **filesystem** mutex
(`/tmp/rustfs_e2e_port_allocator.lock`, with stale-lock reaping) plus a counter
file — i.e. deliberately built for the *cross-process* regime, which is exactly
the regime an in-process `serial_test` mutex cannot reach. Under nextest the
`#[serial]` mutex was uncontended by construction; the filesystem lock was doing
all the work before this PR and continues to do all of it after.

### Role: test skeptic — if one of these actually needed serialization, how would it fail, and how do we know it won't?

The honest failure modes and why each is closed:

- **Port collision → server fails to bind, test fails at startup.** Closed:
  serialization here was already provided by the cross-process filesystem lock
  described above, not by `#[serial]`. Deleting an uncontended in-process mutex
  cannot change contention on a filesystem lock.
- **Shared on-disk state → cross-test data bleed, flaky assertions.** Closed:
  per-instance UUID temp dir; no test in these files writes to any path outside
  it (`grep` for literal `/tmp`, `current_dir`, `home_dir`: zero hits in all
  three files).
- **Process-global singleton (`OnceLock`/`static`) mutated by two tests.** Closed:
  no `static mut`, `lazy_static`, or `OnceLock` in these files. The `AtomicU64` /
  `tokio::sync::Mutex` in `replication_extension_test.rs` are per-test values
  constructed inside the test body (fake-target request counters), not module
  statics.
- **Process env raced across threads.** Closed: no env mutation at all (above).

The strongest structural argument is that **this change cannot regress CI at
all**: under nextest — every lane RustFS runs, `default` / `ci` / `e2e-smoke` /
`e2e-repl-nightly` / `e2e-full` — the removed attribute had no runtime effect,
so pre- and post-PR execution are bit-for-bit identical. The only execution mode
where `#[serial]` did anything is the `RUSTFS_ALLOW_CARGO_TEST_FALLBACK=1`
`cargo test` escape hatch, which `docs/testing/README.md` explicitly declares
**not authoritative** ("serialization semantics differ from CI and
`[test-groups]` do not apply"). Under that fallback these tests remain safe for
the same reason CI is safe: temp-dir + port isolation is real isolation, and the
filesystem port lock works there too.

If a real race is ever surfaced by load, the correct remedy per `ci-10` /
backlog#937 policy is a `[test-groups]` entry or a quarantine with an OPEN issue
link — never re-adding a marker the runner ignores.

## Verification

```
cargo fmt --all --check                       # clean
cargo check -p e2e_test --all-targets         # clean
cargo nextest run -p e2e_test \
  -E 'test(/^(multipart_auth_test|object_lock::object_lock_test)::/)'
cargo nextest run --profile e2e-smoke -p e2e_test \
  -E 'test(/^replication_extension_test::/)'
```

Results on this branch (macOS aarch64, nextest 0.9.140):

| Lane | Result |
| --- | --- |
| `object_lock::object_lock_test::` (default profile) | **34 run, 34 passed** — 27.2s |
| `multipart_auth_test::` (default profile) | 85 run, **84 passed, 1 failed** — the pre-existing failure below |
| `multipart_auth_test::` minus that one test | **84 run, 84 passed** — 26.1s |
| `replication_extension_test::` under `--profile e2e-smoke` | **20 run, 20 passed** — 7.7s |
| `--profile e2e-repl-nightly` (the other 49 replication tests) | **49 run, 49 passed** — 54.2s |

**Pre-existing failure, not caused by this PR:**
`multipart_auth_test::test_signed_put_object_extract_skips_invalid_entry_when_ignore_errors_enabled`
fails with `InvalidArgument: Archive entry path exceeds limit ... length=1126, limit=1024`.
This is **rustfs#4843**, already excluded by name in the `e2e-full` profile's
`default-filter` with that issue cited. It is a product-behaviour failure in
over-limit archive-entry handling and is entirely unrelated to test
attributes — this PR adds no lines at all.

Worth highlighting for the reviewer: the `e2e-repl-nightly` run above executed
**all 49 slow replication tests fully in parallel with no `#[serial]` at all** —
including the 12 `_real_dual_node` and 1 `_real_three_node` site-replication
tests that each spawn 2–3 complete rustfs servers — and finished green in 54
seconds. That is the empirical counterpart to the structural argument: if any of
these had depended on serialization, this is the run that would have exposed it.

## Scope

`crates/e2e_test` only, and only the three densest modules. The remaining 338
markers in the crate (90 files), plus the `ecstore` / `rustfs` crates and
backlog#1846 T2–T9, are out of scope for this PR and follow separately.

Refs: backlog#1846 (T1).

### DIFF
--- a/crates/e2e_test/src/multipart_auth_test.rs
+++ b/crates/e2e_test/src/multipart_auth_test.rs
@@ -30,7 +30,6 @@ use md5::{Digest as Md5Digest, Md5};
 use rustfs_signer::constants::UNSIGNED_PAYLOAD;
 use rustfs_signer::sign_v4;
 use s3s::Body;
-use serial_test::serial;
 use std::collections::HashMap;
 use std::error::Error;
 use std::io::Cursor;
@@ -356,7 +355,6 @@ async fn run_post_object_policy_case(
 /// smuggles one extra field the policy never declared, and the upload must be
 /// rejected with 403 AccessDenied naming the offending field.
 #[tokio::test]
-#[serial]
 async fn test_anonymous_post_object_rejects_fields_missing_from_policy_conditions()
 -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     init_logging();
@@ -484,7 +482,6 @@ async fn test_anonymous_post_object_rejects_fields_missing_from_policy_condition
 /// sends a different one, and the upload must be rejected with 400
 /// InvalidPolicyDocument naming the field.
 #[tokio::test]
-#[serial]
 async fn test_anonymous_post_object_rejects_exact_condition_policy_mismatches()
 -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     init_logging();
@@ -689,7 +686,6 @@ async fn test_anonymous_post_object_rejects_exact_condition_policy_mismatches()
 /// one of them with a different value, and the upload must be rejected with
 /// 400 InvalidPolicyDocument naming the mismatched field.
 #[tokio::test]
-#[serial]
 async fn test_anonymous_post_object_rejects_object_lock_policy_mismatches() -> Result<(), Box<dyn std::error::Error + Send + Sync>>
 {
     init_logging();
@@ -757,7 +753,6 @@ async fn test_anonymous_post_object_rejects_object_lock_policy_mismatches() -> R
 /// exact values, the form sends a different parameter value, and the upload
 /// must be rejected with 400 InvalidPolicyDocument naming the parameter.
 #[tokio::test]
-#[serial]
 async fn test_anonymous_post_object_rejects_sse_kms_policy_mismatches() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     init_logging();
 
@@ -839,7 +834,6 @@ async fn test_anonymous_post_object_rejects_sse_kms_policy_mismatches() -> Resul
 /// NotImplemented (SSE-KMS POST uploads are not implemented), not with a
 /// policy error.
 #[tokio::test]
-#[serial]
 async fn test_anonymous_post_object_rejects_sse_kms_params_outside_policy_conditions()
 -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     init_logging();
@@ -894,7 +888,6 @@ async fn test_anonymous_post_object_rejects_sse_kms_params_outside_policy_condit
 }
 
 #[tokio::test]
-#[serial]
 async fn test_anonymous_multipart_control_apis_require_auth() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     init_logging();
 
@@ -968,7 +961,6 @@ async fn test_anonymous_multipart_control_apis_require_auth() -> Result<(), Box<
 }
 
 #[tokio::test]
-#[serial]
 async fn test_anonymous_post_object_requires_auth() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     init_logging();
 
@@ -1002,7 +994,6 @@ async fn test_anonymous_post_object_requires_auth() -> Result<(), Box<dyn std::e
 }
 
 #[tokio::test]
-#[serial]
 async fn test_anonymous_post_object_honors_success_action_status() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     init_logging();
 
@@ -1066,7 +1057,6 @@ async fn test_anonymous_post_object_honors_success_action_status() -> Result<(),
 }
 
 #[tokio::test]
-#[serial]
 async fn test_anonymous_post_object_honors_success_action_redirect() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     init_logging();
 
@@ -1139,7 +1129,6 @@ async fn test_anonymous_post_object_honors_success_action_redirect() -> Result<(
 }
 
 #[tokio::test]
-#[serial]
 async fn test_anonymous_post_object_defaults_to_no_content() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     init_logging();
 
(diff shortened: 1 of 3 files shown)


overtrue:
Split the unrelated `Cargo.lock` hunk out of this PR — it is now https://github.com/rustfs/rustfs/pull/6211 and this branch is strictly attribute-only (190 deletions, zero additions).

That hunk turned out to be more than lockfile hygiene: `crates/heal/Cargo.toml` gained `bytes` and `crc-fast` in #6189 without a lockfile regeneration, so `cargo metadata --locked` currently fails on `main` (exit 101), which means `.github/workflows/nightly-gnu.yml`'s `cargo build --release --locked` is broken until #6211 lands. Reviewing it on its own merits seemed better than burying a CI fix inside a test-attribute change.

houseme (approved):

PR #6210, opened 2026-08-18
chore(zip): trim the unused extract and create surface

overtrue:
## Related Issues

rustfs/backlog#1847 (T1 only). T2-T5 of that issue stay post-1.0 and are not touched here.

## Summary of Changes

`rustfs-zip` has exactly one workspace consumer, `rustfs`, and it uses only `CompressionFormat::from_extension`, `CompressionFormat::extension`, `CompressionFormat::get_decoder` and `ArchiveLimits`. Everything else in the crate was public API with zero callers anywhere in the repository.

**Why now (semver):** trimming a crate's public API is a semver-major change once the stable tag is cut. Doing it before the tag costs one PR; doing it after costs a major bump. This is the only reason T1 is pulled out of an otherwise post-1.0 issue.

Removed (each verified to have zero callers first): `read_archive_entries`, `read_archive_entries_with_limits`, `decompress`, `extract_tar_entries`, `extract_zip_simple`, `extract_zip_to_path_with_limits`, `extract_zip_with_limits`, `create_zip_simple`, `create_zip_with_options`, `Compressor`, `Decompressor`, and the types that these left orphaned: `ZipEntry`, `ZipExtractSummary`, `ZipWriteOptions`, `CompressionLevel`, `ArchiveKind`, `CompressionCodec`, `ArchiveFormat`. The private helpers behind them (`extract_zip_impl`, `write_small_zip_entry`, `normalize_zip_entry_name`, the four `validate_archive_*` guards, `SharedBuffer`, `zip_method_for_level`, `parent_directories_for`, `ensure_directory`) and the whole `zip_benchmark` bench go with them.

`CompressionFormat::from_extension` previously delegated to `ArchiveFormat::from_extension` and mapped the result back; with `ArchiveFormat` gone the extension table is inlined, arm for arm, so the mapping is unchanged.

Two deviations from the issue text, both because the issue assumed a caller that does not exist:

- `get_encoder` was listed as "keep", but it has no caller outside the deleted `Compressor` and its own tests, and its signature takes `CompressionLevel`, which the same issue lists for deletion. It is removed together with `convert_level` and `ZipError::InvalidCompressionLevel`. Say the word and it can come back.
- `ZipError` is kept as the error type of `get_decoder`, but its variants were trimmed to the only one still constructible, `UnsupportedFormat`. The `Zip(#[from] zip::result::ZipError)` variant was the last thing holding the `zip` crate dependency, which the issue asked to drop if nothing else used it.

Dependencies dropped as a result: `zip`, `astral-tokio-tar`, `tokio-stream`, and the `criterion` / `tempfile` dev-dependencies. `tokio` features narrowed from `["fs", "io-util", "macros", "rt-multi-thread"]` to `["io-util", "macros", "rt"]`. `README.md` and the crate `description` are updated to describe what the crate now is.

Net: 57 insertions, 2115 deletions; `crates/zip/src/lib.rs` goes from 1786 to 189 lines.

## Verification

- `cargo check -p rustfs-zip`
- `cargo check -p rustfs`
- `cargo check -p rustfs --tests` (the only consumer of `CompressionFormat::extension` sits in a `#[cfg(test)]` module, so the plain check does not cover it)
- `cargo nextest run -p rustfs-zip` (4 tests, all pass)
- `cargo fmt --all --check`

Adversarial validation, Mechanical tier (correctness + simplicity):

- **Correctness adversary** — attacked the zero-caller claim with three grep syntaxes per removed item (bare word, `rustfs_zip::`-qualified path, and `.method()` call form) across the whole repository including `fuzz/`, docs, scripts and CI. The only surviving hits are unrelated same-name symbols: `crates/protocols/src/swift/bulk.rs` defines its own private `ArchiveFormat` and `extract_tar_entries`, `crates/utils/src/compress.rs` uses `brotli::Decompressor`, and the `is_supported()` hits in `crates/ecstore` and `rustfs/src/server/readiness.rs` are on peer status, not on a compression format. Also attacked the security angle: the deleted `validate_archive_*` guards were reachable only from the deleted functions, and the PutObject extract path in `rustfs/src/app/object_usecase.rs` enforces its own limits through `validate_put_object_extract_*` on `ArchiveLimits`, which is untouched, so no guard on a live path was removed. The one behavior-carrying edit, inlining the extension table, was checked arm by arm against the two-step mapping it replaces. No break found.
- **Simplicity adversary** — attacked the leftovers: every public item in the trimmed crate now has a real caller (`from_extension`, `extension`, `get_decoder`, `ArchiveLimits`; `Result` and `ZipError` appear in `get_decoder`'s signature), no private helper is unreferenced, and no manifest dependency survives only for deleted code. One residue reported rather than fixed: the `hotpath` dependency and its three feature aliases are unused in this crate's source, but they are wired into `rustfs/Cargo.toml`'s feature lists the same way as every other crate, so removing them belongs to a repo-wide change, not here.

## Impact

Public API of `rustfs-zip` shrinks substantially. This is deliberate and is the point of the change: the crate is pre-1.0, published at `1.0.0-rc.2`, and every item removed had no caller in this repository. After the stable tag the same trim would require a major bump.

No runtime behavior change for RustFS: the archive extract path keeps the same `from_extension` results, the same decoder construction, and the same `ArchiveLimits` defaults and enforcement.

## Additional Notes

`Cargo.lock` also picks up `bytes` and `crc-fast` under `rustfs-heal`. That is pre-existing drift on `main` (`crates/heal/Cargo.toml` already declares both, the lock was never refreshed); cargo synced it while resolving this crate's removed dependencies.


houseme (approved):

houseme:
`cargo shear --fix `

PR #6230, opened 2026-08-19
fix(admin): report refreshed lock lease TTLs

### LINKED ISSUE #5939: Distributed lock entries never expire (ttl_secs: 0)

cxymds:
## Related Issues
Fixes #5939.

## Summary of Changes
`/rustfs/admin/v3/top/locks` now reports the authoritative heartbeat-refreshed lease deadline held by `LocalClient`, instead of the original fast-lock timeout. Direct fast locks remain visible with their existing timeout behavior. Lease data is merged only when the current holder count matches, and exclusive locks also validate acquisition time so released or replaced guards cannot contribute stale TTLs. The existing response shape is unchanged.

## Verification
- `cargo fmt --all`
- `git diff --check`
- `cargo test -p rustfs-lock --lib -- --test-threads=1` (128 passed)
- `cargo test -p rustfs top_locks_prefers_renewable_lease_deadlines`
- `cargo test -p rustfs collect_top_locks_reports_live_lock`
- `cargo test -p rustfs top_locks_handler_rejects_missing_credentials`
- `make pre-pr`
- Seven-role adversarial validation completed for correctness, simplicity, security, concurrency/durability, compatibility, performance, and test coverage.

## Impact
The admin top-locks diagnostics endpoint now shows live renewable lease TTLs, improving operational accuracy for heartbeat-refreshed locks. There is no new JSON field and no configuration or deployment change. The lock client trait gains a default-compatible additive lease snapshot method; clients without renewable lease state continue to return an empty snapshot.

## Additional Notes
The merge retains the existing admin authorization (`TopLocksAdminAction`) and preserves direct fast-lock visibility. Build artifacts generated in the dedicated worktree were removed after verification.

---

### DIFF
--- a/crates/lock/src/client/local.rs
+++ b/crates/lock/src/client/local.rs
@@ -47,6 +47,7 @@ pub struct LocalClient {
 #[derive(Debug)]
 struct LocalGuardEntry {
     guard: FastLockGuard,
+    acquired_at: SystemTime,
     expires_at: SystemTime,
     deadline: Instant,
     ttl: Duration,
@@ -54,11 +55,12 @@ struct LocalGuardEntry {
 
 impl LocalGuardEntry {
     fn new(guard: FastLockGuard, ttl: Duration) -> Self {
-        let now = SystemTime::now();
+        let acquired_at = SystemTime::now();
         let monotonic_now = Instant::now();
         Self {
             guard,
-            expires_at: now.checked_add(ttl).unwrap_or(now),
+            acquired_at,
+            expires_at: acquired_at.checked_add(ttl).unwrap_or(acquired_at),
             deadline: monotonic_now.checked_add(ttl).unwrap_or(monotonic_now),
             ttl,
         }
@@ -231,13 +233,14 @@ impl LockClient for LocalClient {
             match lock_manager.acquire_lock(build_lock_request(remaining)).await {
                 Ok(guard) => {
                     let lock_id = request.lock_id.clone();
-                    let acquired_at = SystemTime::now();
-                    let expires_at = acquired_at.checked_add(request.ttl).unwrap_or(acquired_at);
+                    let entry = LocalGuardEntry::new(guard, request.ttl);
+                    let acquired_at = entry.acquired_at;
+                    let expires_at = entry.expires_at;
 
                     {
                         let shard = self.get_shard(&lock_id);
                         let mut guards = shard.write().await;
-                        guards.insert(lock_id.clone(), LocalGuardEntry::new(guard, request.ttl));
+                        guards.insert(lock_id.clone(), entry);
                     }
 
                     let lock_info = LockInfo {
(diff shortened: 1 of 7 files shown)


houseme:
@codex review

PR #6296, opened 2026-08-20
fix(tier): recover multi-committed mutation intents

### LINKED ISSUE #6204: Tier reload fails when multiple committed coordinator mutation intents exist for the same tier
Summary
After restart, TierConfigMgr::reload_handle() can fail during recovery if multiple coordinator mutation intents are already in Committed state for the same tier, such as COLD. In that case, tier loading stops before load_tier_config() runs, and ListTiers returns an empty result.

Observed behavior

Restart logs show:

Remote tier COLD already has another prepared mutation
TierConfigMgr init error: Io error: Remote tier COLD already has another prepared mutation
load_tier_config() is never reached

ListTiers reports tier_count=0

Expected behavior
Committed coordinator mutation intents should be replayed and cleaned up during recovery. Multiple committed intents for the same tier should not block tier reload.

Root cause
reconcile_committed_mutation_intent_blocks() applies the same conflict logic used for prepared intents. When multiple committed intents exist for the same tier, recovery fails early and the cleanup path is never reached.

Impact
A node restart can leave tier configuration unavailable even though the tier data and config are still present.

cxymds:
## Related Issues

Fixes #6204

## Summary of Changes

Make tier mutation recovery resilient when a tier has multiple committed coordinator intents. Recovery now keeps peer and coordinator intent states aligned, rejects new config updates while pending recovery still needs to converge, and preserves the fence until the durable state is safe to publish. Run the recovery refresh worker on standalone deployments too, so their committed mutation fences converge instead of blocking later tier updates.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-ecstore services::tier::tier::tests -- --nocapture`
- `cargo test -p rustfs-ecstore --features test-util tier_mutation_peer_handler -- --nocapture`
- `cargo test -p rustfs-protos tier_mutation_rpc -- --nocapture`
- `cargo clippy -p rustfs-ecstore -p rustfs-protos -p rustfs --all-targets -- -D warnings`
- `CARGO_TARGET_DIR=... cargo test -p e2e_test reliant::tiering::test_manual_transition_async_tier_failure_reports_terminal_partial -- --nocapture`

## Impact

Standalone and distributed tier configurations now both converge committed mutation fences. No RPC wire-format changes.

## Additional Notes

Adversarial validation completed across correctness, simplicity, security, concurrency/durability, compatibility, performance, and test coverage.

### DIFF
--- a/crates/e2e_test/src/reliant/tiering.rs
+++ b/crates/e2e_test/src/reliant/tiering.rs
@@ -200,11 +200,19 @@ async fn add_rustfs_tier(hot: &RustFSTestEnvironment, cold: &RustFSTestEnvironme
 
 async fn remove_rustfs_tier_force(hot: &RustFSTestEnvironment) -> TestResult {
     let path = format!("/rustfs/admin/v3/tier/{TIER_NAME}?force=true");
-    let (status, resp) = signed_admin_request(&hot.url, Method::DELETE, &path, None, &hot.access_key, &hot.secret_key).await?;
-    if !status.is_success() {
-        return Err(format!("RemoveTier(RustFS) failed: status={status}, body={resp}").into());
+    let deadline = Instant::now() + StdDuration::from_secs(30);
+    loop {
+        let (status, resp) =
+            signed_admin_request(&hot.url, Method::DELETE, &path, None, &hot.access_key, &hot.secret_key).await?;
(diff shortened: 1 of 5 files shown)


cxymds:
@codex review

Issue #925, opened 2025-11-25
UnauthorizedAccess Your account is not signed up2


mrgionsi:
**Describe the bug**

**To Reproduce**
1. Set up a bucket.
2. Connect Longhorn
3. After some times (Some backups are completed) longhorn start reporting: 
```
time="2025-11-25T17:28:32.108953074Z" level=error 
msg="Failed to get info from backup store" func="controller.(*BackupTargetController).reconcile" 
file="backup_target_controller.go:459" 
controller=longhorn-backup-target cred=rustfs-secret 
error="failed to list system backups in s3://talos-prod@us-east-1/: error listing system backup in 
s3://talos-prod@us-east-1/: failed to execute: /var/lib/longhorn/engine-binaries/longhornio-longhorn-engine-v1.10.0/longhorn
 [/var/lib/longhorn/engine-binaries/longhornio-longhorn-engine-v1.10.0/longhorn system-backup list 
s3://talos-prod@us-east-1/], output , stderr time=\"2025-11-25T17:28:32.106725032Z\" level=error 
msg=\"Failed to list s3\" func=\"s3.(*BackupStoreDriver).List\" file=\"s3.go:116\" error=\"failed to 
list objects with param: {\\n  Bucket: \\\"talos-prod\\\",\\n  Delimiter: \\\"/\\\",\\n  Prefix: \\\"/\\\"\\n} error: AWS Error:  UnauthorizedAccess Your account is not signed up2 <nil>\\n401 \\n\" 
pkg=s3\ntime=\"2025-11-25T17:28:32.106987352Z\" level=fatal 
msg=\"Failed to run list system backup command\" func=cmd.SystemBackupCmd.SystemBackupListCmd.func4 file=\"system_backup.go:72\" 
error=\"failed to list objects with 
param: {\\n  Bucket: \\\"talos-prod\\\",\\n  Delimiter: \\\"/\\\",\\n  Prefix: \\\"/\\\"\\n} error: AWS Error:  UnauthorizedAccess Your account is not signed up2 <nil>\\n401 \\n\"\n: exit status 1" interval=5m0s name=default node=talos-prod-worker-1 
url="s3://talos-prod@us-east-1/"
```

**Expected behavior**
No error displayed

**Desktop (please complete the following information):**
 - Docker image i am using the latest one. 
 I saw issue was solved in one of the latest release, but in my case still persist. 

**Additional context**

loverustfs:
[...]

PR #6327, opened 2026-08-21
feat(ecstore): tune fdatasync group wait budget

houseme:
## Related Issues
Related to rustfs/backlog#925 and rustfs/backlog#1856.

## Summary of Changes
Adds a default-off fdatasync group-commit wait budget controlled by `RUSTFS_EXPERIMENTAL_FILE_FDATASYNC_GROUP_COMMIT_WAIT_MICROS`.

The default remains `0`, preserving the current group worker behavior. When the existing experimental file fdatasync group commit is enabled and the wait budget is non-zero, the worker waits briefly before draining the pending batch so late followers can join the leader's fdatasync batch. The value is capped at 1000 microseconds to keep the experiment bounded.

A deterministic regression test covers the intended behavior: a follower enqueued after the leader worker starts but before the wait budget expires must be included in the same fdatasync batch, while both source directories still receive their required fsyncs.

## Verification
- `cargo test -p rustfs-ecstore disk::os::tests::file_fdatasync_group_commit_wait_budget_batches_late_follower -- --exact`
- `cargo test -p rustfs-ecstore disk::os::tests::file_fdatasync_group_commit_ -- --nocapture`
- `cargo fmt --all --check`
- `git diff --check`

## Impact
Default behavior is unchanged unless both experimental file fdatasync group commit and this new wait budget are explicitly enabled.

No S3 API, RPC protocol, on-disk format, metadata, or rename syscall behavior changes are introduced.

This is intended for the existing #925/#1856 4-node clean-data 1MiB PUT c16 probe only; it should help determine whether a small wait window can raise `fdatasync_batch_files` above 1 and whether `file_fdatasync`, `disk_wait`, and `quorum_wait` p90/p99 converge.

## Additional Notes
Adversarial validation summary for this high-risk durability-adjacent diff:

- Correctness: attacked disabled-by-default behavior, late follower batching, failure propagation, and required source-directory fsyncs; no break found in the diff or focused tests.
- Simplicity: attacked unnecessary abstraction and config surface; no smaller safe change found that still supports bounded runtime tuning plus deterministic test override.
- Security: attacked secret/path leakage, untrusted input, and high-cardinality surfaces; no new external input, logging, or labels were added.
- Concurrency/durability: attacked lock-held-across-await, cancellation, overflow, and sync ordering; no lock is held during the wait, existing hard limits remain, and sync ordering after batch drain is unchanged.
- Compatibility: attacked S3 wire/on-disk/mixed-version impact; no protocol, metadata, or format change found, and default-off keeps mixed nodes compatible.
- Performance: attacked hot-path overhead; runtime delay is opt-in, capped to 1000µs, and no new syscall is added.
- Test coverage: the new test should fail if the wait budget is ignored or inverted; existing group-commit tests continue to cover default-off, zero-wait batching, overflow, and failure paths.

Pending after CI: run the same 4-node clean-data 1MiB PUT c16 probe with a small wait-budget sweep, keeping admission and fdatasync/group-commit conclusions separate and continuing not to prioritize renameat.

### DIFF
--- a/crates/ecstore/src/disk/os.rs
+++ b/crates/ecstore/src/disk/os.rs
@@ -23,7 +23,7 @@ use std::{
     io,
     path::{Component, Path, PathBuf},
     sync::{Arc, LazyLock, Weak},
-    time::Instant,
+    time::{Duration, Instant},
 };
 use tokio::fs;
 use tokio::sync::{
@@ -328,6 +328,9 @@ const ENV_DST_DIR_FSYNC_GROUP_COMMIT_ENABLE: &str = "RUSTFS_EXPERIMENTAL_DST_DIR
 const DEFAULT_DST_DIR_FSYNC_GROUP_COMMIT_ENABLE: bool = false;
 const ENV_FILE_FDATASYNC_GROUP_COMMIT_ENABLE: &str = "RUSTFS_EXPERIMENTAL_FILE_FDATASYNC_GROUP_COMMIT_ENABLE";
 const DEFAULT_FILE_FDATASYNC_GROUP_COMMIT_ENABLE: bool = false;
+const ENV_FILE_FDATASYNC_GROUP_COMMIT_WAIT_MICROS: &str = "RUSTFS_EXPERIMENTAL_FILE_FDATASYNC_GROUP_COMMIT_WAIT_MICROS";
+const DEFAULT_FILE_FDATASYNC_GROUP_COMMIT_WAIT_MICROS: u64 = 0;
+const MAX_FILE_FDATASYNC_GROUP_COMMIT_WAIT_MICROS: u64 = 1_000;
 #[cfg(not(test))]
 const MAX_DST_DIR_FSYNC_GROUPS: usize = 1024;
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


crates/ecstore/src/disk/os.rs
```diff
@@ -6141,12 +6183,83 @@ mod tests {
         assert_eq!(file_fdatasync_group_commit_counts_for_test(), (0, 0, 0));
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[serial_test::serial(file_sync_probe)]
+    async fn file_fdatasync_group_commit_wait_budget_batches_late_follower() {
+        use std::sync::mpsc;
+
+        let _group_commit = set_file_fdatasync_group_commit_for_test(true);
+        set_file_fdatasync_group_commit_wait_for_test(Duration::from_millis(50));
```
overtrue: [P2] This 50 ms override bypasses the production path that parses, defaults, and caps the new setting, and it is 50× above the supported 1000 µs ceiling. I replaced the non-test fallback with a 30-second delay and all six file-fdatasync group-commit tests still passed, so there is no mutation detector for the stated default-zero and cap contract. Please cover the real configuration path (default, nonzero, and cap) and exercise a legal budget with deterministic time control.

overtrue (changes requested):

crates/ecstore/src/disk/os.rs
houseme: Addressed in 1fd0c33e9. The production wait-budget path now goes through a shared normalization helper, and tests cover the default 0, a legal non-zero value, the 1000 µs cap, and an over-cap value. The group-commit test injects the raw microsecond budget (rather than a Duration override), uses the legal 1000 µs budget, and advances Tokio virtual time deterministically. The focused file-fdatasync group-commit tests, cargo fmt --all --check, and ecstore Clippy all pass.

overtrue (approved):
LGTM

PR #6328, opened 2026-08-21
fix(iam): preserve MinIO permanent credentials in migration

### LINKED ISSUE #6295: MinIO IAM migration: permanent users and service accounts are removed as expired after successful migration
**Describe the bug**

When migrating an existing MinIO data directory to RustFS `1.0.0-rc.2`, IAM objects are migrated successfully, but immediately afterwards RustFS treats the migrated IAM identities as expired and deletes them.

In my environment, RustFS successfully migrated 162 IAM objects. However, during the subsequent IAM initialization (`build_iam_sys`), the following identities were removed as expired:

```text
Reg: 62
Svc: 52
Sts:  1
```

Total:

```text
IAM user identity expired and was removed: 115
```

These counts exactly match the identities that existed in the original MinIO installation:

```text
62 regular users
52 service accounts
1 STS identity
```

The suspected cause is MinIO's handling of permanent credentials.

MinIO uses:

```text
1970-01-01T00:00:00Z
```

as a sentinel value meaning that the credential does **not expire**.

MinIO's expiration logic explicitly excludes this value from being treated as expired.

RustFS appears to migrate this expiration timestamp unchanged, then later interprets it as a normal timestamp in the past:

```text
MinIO:
expiration = 1970-01-01T00:00:00Z
             ↓
          no expiry

RustFS migration:
expiration is migrated unchanged
             ↓

RustFS IAM loader:
1970-01-01 < current time
             ↓
          expired
             ↓
identity and policy mapping are deleted
```

The deletion appears to occur in the expiration handling around:

```text
crates/iam/src/store/object.rs
```

near the code path that logs:

```text
IAM user identity expired and was removed
```

---

**To Reproduce**

1. Create IAM users and service accounts in MinIO using permanent credentials.
2. Verify that the MinIO IAM data uses `1970-01-01T00:00:00Z` as the expiration value for those credentials.
3. Make a copy of the existing MinIO data directory.
4. Start RustFS `1.0.0-rc.2` using that copied MinIO data.
5. Allow the MinIO IAM migration to complete.
6. Check the RustFS logs.
7. Observe that IAM migration completes successfully.
8. Immediately afterwards, observe messages such as:

```text
IAM user identity expired and was removed
```

For service accounts, the policy mapping is also removed before the identity is deleted.

In my case, the final count of removed identities was:

```text
Reg: 62
Svc: 52
Sts:  1
```

This exactly matches the original MinIO IAM identities.

---

**Expected behavior**

Permanent IAM users and service accounts migrated from MinIO should remain valid after migration.

RustFS should treat MinIO's Unix epoch sentinel:

```text
1970-01-01T00:00:00Z
```

as meaning "no expiration", rather than treating it as an expiration date in the past.

After migration, I expect the following IAM objects to remain:

```text
users:              62
service-accounts:   52
policies:           14
policydb:           18
groups:             14
sts:                 1
```

One possible fix would be to normalize the MinIO Unix epoch sentinel to RustFS's native "no expiration" representation during migration.

Alternatively, the RustFS expiration check could explicitly treat `1970-01-01T00:00:00Z` as non-expiring for MinIO compatibility.

---

**Screenshots**

Not applicable.

I can provide the relevant RustFS logs showing the successful IAM migration followed immediately by removal of the migrated identities.

---

* OS: Ubuntu Linux
* RustFS version: `1.0.0-rc.2`
* Source storage: Existing MinIO data directory
* Filesystem: XFS
* Deployment: Docker

Browser information is not applicable because the issue occurs in the RustFS server-side IAM migration/loading process.

---

Not applicable.

---

**Additional context**

The IAM migration itself appears to succeed. The problem occurs only afterwards, when RustFS initializes the IAM subsystem and checks credential expiration.

This is important because the migration can appear successful before RustFS silently removes the migrated credentials.

In my environment:

```text
162 IAM objects migrated
```

was followed by removal of:

```text
62 regular users
52 service accounts
1 STS identity
```

The numbers exactly match the corresponding identities in the original MinIO installation.

MinIO treats both a zero expiration time and its Unix epoch sentinel as non-expiring credentials. RustFS `1.0.0-rc.2` appears to interpret the Unix epoch sentinel literally as an already-expired timestamp.

This can make a MinIO drop-in migration destructive from an IAM perspective, because existing application credentials stop working immediately after RustFS starts.

I reproduced this using a copy of the MinIO data and did not manually modify `xl.meta` or IAM metadata.

GatewayJ:
## Related Issues

Fixes #6295.

## Summary of Changes

- Normalize MinIO's exact Go zero time and Unix epoch permanent-credential sentinels to RustFS's canonical no-expiration value while migrating regular users and service accounts.
- Keep STS identities and every non-sentinel past, adjacent, or future expiration unchanged.
- Add focused boundary coverage plus a real `.minio.sys` migration test that verifies the persisted target, repeated IAM loads, policy-mapping retention, valid service-account JWT handling, and every credential field.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore bucket::migration::tests -- --nocapture`
- `cargo test -p rustfs-iam --test minio_iam_migration_test`
- `cargo clippy -p rustfs-iam --test minio_iam_migration_test -- -D warnings`
- `make embedded-secrets-check`
- `make pre-pr`

## Impact

MinIO permanent regular users and service accounts remain valid after fresh migration or after a previously deleted target is migrated again. There are no API or configuration changes, and STS expiration remains fail-closed.

## Additional Notes

- Correctness: exact sentinel, path, STS, adjacent-time, past-time, future-time, and credential-preservation attacks passed.
- Simplicity: the production change is one local constant and one in-place migration condition with no new API or abstraction.
- Security: only regular users and service accounts crossing the MinIO migration boundary receive the exact sentinel conversion; STS and real expiry values are unchanged.
- Concurrency/durability: no lock, await, I/O, or write-order behavior changed.
- Compatibility: the migration writes the existing canonical `null` representation and leaves the `.minio.sys` source unchanged.
- Performance: each migrated identity adds only bounded path and timestamp comparisons, with no extra I/O or allocation.
- Test coverage: explicit MinIO-shaped JSON now exercises source migration, persisted-target normalization, repeated regular-user and service-account loads, policy-mapping survival, both sentinels, STS exclusion, predicate precedence, exact timestamp equality, and full-field preservation without relying on the production deserializer for expected values.
- Existing target idempotency is unchanged: if an older RustFS version wrote a sentinel target and crashed before IAM cleanup, the first patched startup may delete that target and a second startup remigrates it. The issue's observed post-deletion state recovers on the first patched startup.

---

### DIFF
--- a/crates/ecstore/src/bucket/migration.rs
+++ b/crates/ecstore/src/bucket/migration.rs
@@ -41,6 +41,7 @@ const IAM_FORMAT_FILE_PATH: &str = "config/iam/format.json";
 const IAM_USERS_PREFIX: &str = "config/iam/users/";
 const IAM_SERVICE_ACCOUNTS_PREFIX: &str = "config/iam/service-accounts/";
 const IAM_STS_PREFIX: &str = "config/iam/sts/";
+const MINIO_GO_ZERO_TIME: OffsetDateTime = time::macros::datetime!(0001-01-01 00:00 UTC);
 const IAM_GROUPS_PREFIX: &str = "config/iam/groups/";
 const IAM_POLICIES_PREFIX: &str = "config/iam/policies/";
 const IAM_POLICY_DB_PREFIX: &str = "config/iam/policydb/";
@@ -120,6 +121,15 @@ fn normalize_iam_config_blob(path: &str, data: &[u8]) -> std::result::Result<Opt
     if is_identity_path(path) {
         let mut identity: UserIdentity =
             serde_json::from_slice(data).map_err(|err| format!("parse IAM identity failed: {err}"))?;
+        if (path.starts_with(IAM_USERS_PREFIX) || path.starts_with(IAM_SERVICE_ACCOUNTS_PREFIX))
+            && identity
+                .credentials
+                .expiration
+                .as_ref()
+                .is_some_and(|expiration| *expiration == MINIO_GO_ZERO_TIME || *expiration == OffsetDateTime::UNIX_EPOCH)
+        {
+            identity.credentials.expiration = None;
+        }
         if identity.update_at.is_none() {
             identity.update_at = Some(OffsetDateTime::now_utc());
         }
@@ -441,7 +451,10 @@ mod tests {
     use crate::bucket::replication::{
         BucketReplicationResyncStatus, ReplicationMigrationBridge, ResyncStatusType, TargetReplicationResyncStatus,
     };
+    use rustfs_policy::auth::UserIdentity;
     use std::collections::HashMap;
+    use time::OffsetDateTime;
+    use time::format_description::well_known::Rfc3339;
 
     #[test]
     fn test_normalize_policy_mapping_legacy_timestamp_and_fields() {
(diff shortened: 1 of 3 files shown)


GatewayJ:
@codex please review

overtrue (approved):
LGTM

PR #6336, opened 2026-08-21
fix(admin): advertise IAM admin capabilities in runtime capabilities

reatang:
## Related Issues

Fixes rustfs/backlog#1900 (server side).

## Summary of Changes

`rc admin policy detach` fails with `admin.iam.policy-detach was not advertised by the server` even though the detach REST endpoint (`POST /rustfs/admin/v3/idp/builtin/policy/detach`) is implemented and works. Root cause: the server has no dynamic advertisement of IAM admin capabilities — the rc client's capability report only gains `admin.iam.*` entries from a client-side contract pinned to the exact version `1.0.0-beta.10`, so every newer server loses the whole IAM capability family.

This PR adds an `advertised` list to the `/rustfs/admin/v4/runtime/capabilities` response:

- New `AdvertisedAdminCapability { name, status }` entries for the IAM admin capability family: `admin.iam.policy-attach`, `admin.iam.policy-detach`, `admin.iam.policy-entities`, `admin.iam.access-keys-bulk`, `admin.iam.access-keys-bulk.ldap`, `admin.iam.access-keys-bulk.openid`.
- Statuses are derived from the public admin route inventory via the existing `admin_route_capability` helper (same mechanism as the site-replication summary entries), so the advertisement tracks the actually registered routes instead of a hardcoded claim.
- The capability names are a cross-repo wire contract with the rc client (`IAM_POLICY_DETACH_CAPABILITY` etc. in rustfs/cli); the field-level comment pins that entries may be added but never renamed or removed.

A companion rustfs/cli PR consumes this field (probe results > server advertisement > pinned per-version fallback) and additionally extends the client's pinned IAM contract to the `1.0.0-rc.*` family so already-deployed rc.3 servers work without this server change.

## Verification

- `cargo test -p rustfs --lib admin::` — pass (includes new `runtime_capabilities_response_advertises_iam_capabilities` contract test asserting all six entries are supported, names are unique, and the serialized wire shape `{name, status: {state, reason?}}` matches what rc parses)
- `cargo clippy -p rustfs --all-targets -- -D warnings` — pass
- `make pre-commit` — pass
- Manual end-to-end: started a patched server locally, ran the issue's reproduction with a patched rc build (`alias set` → `admin user add` → `policy attach readonly` → `policy detach readonly`): detach succeeds and `rc admin capabilities` lists all six `admin.iam.*` entries with route-inventory reasons, proving the dynamic advertisement path (pinned entries carry no reason). Also verified the raw endpoint output via signed curl.

Tier rationale: localized behavior change confined to one response struct in the `rustfs` binary crate (no downstream crate consumes these types); targeted tests build every changed target, so the full `make pre-pr` gate is not required.

Adversarial review (all seven roles): correctness (route/inventory mapping, degenerate version strings), simplicity, security (endpoint authz unchanged — still gated on `ServerInfoAdminAction`; advertisement exposes only static route-registration facts), concurrency/durability (N/A — no shared state), compatibility (additive JSON field; both known consumers — rc's serde snapshot and the e2e `serde_json::Value` assertions — tolerate unknown fields; old rc clients fall back to their pinned contract), performance (per-request construction of a six-entry Vec on an admin endpoint), test-coverage (reverting the field or renaming any entry fails the new contract test). One finding surfaced and fixed on the client side (advertisement merge ordering vs client probes); no server-side findings.

## Impact

- API: additive field `advertised` in the authenticated `/rustfs/admin/v4/runtime/capabilities` response. No existing field changes; no auth behavior changes.
- User-facing: after the companion rc release, `rc admin policy detach` works against servers with this change via dynamic advertisement (and against existing rc.3 deployments via the client's extended pinned contract).
- Known residual: already-released rc ≤ 0.1.31 does not parse the new field and keeps its beta.10-only pin; those clients need the rc upgrade.

## Additional Notes

The attach/detach gate asymmetry in rc (attach sends without a capability check, detach fails closed) is documented client behavior and intentionally left unchanged; with correct advertisement it no longer causes user-visible failures.


reatang:
CI status note for reviewers — the four red feature-matrix jobs (`Test and Lint (rio-v2|sftp|swift)`, `ILM Integration (serial)`) are an infrastructure timeout, not a failure introduced by this diff:

- In both the original run and the `--failed` rerun, each job's log ends with ~50 minutes of silence after `Compiling rustfs-s3select-query` (crate 1049 of the build, i.e. right before the large `rustfs` crate compiles/links), then `The operation was canceled` at the job timeout and orphan `rustc`/`rust-lld` processes being terminated. No test ran and no compile error was emitted.
- The same jobs on `main` currently show the same behavior: run 32564075645 had all Test and Lint variants still running after 2h before being superseded, and runs 32569631227 / 32555342619 were likewise cancelled with these jobs unfinished.
- Everything that completed for this PR passed: Quick Checks, the default `Test and Lint` (40m), Build RustFS Debug Binary, End-to-End Tests, io_uring Integration, Typos, CLA.

The diff touches only `rustfs/src/admin/handlers/system.rs` (additive runtime-capabilities response field + its contract test) and cannot affect the sftp/swift/rio-v2 feature builds' compile behavior. Happy to rerun once the runner/cache situation on `main` recovers.

PR #6340, opened 2026-08-21
fix(heal): make resume checkpoints crash consistent

cxymds:
## Related Issues
Fixes rustfs/backlog#1927.

## Summary of Changes
- Replace truncating resume-checkpoint writes with conditional, durability-gated publication through the existing local backend CAS path.
- Preserve and quarantine malformed, task-ID-mismatched, and future-schema snapshots so recovery selection cannot retry them automatically.
- Reject stale and post-cleanup writers instead of allowing them to overwrite newer progress or recreate deleted checkpoints.

## Verification
- `cargo fmt --all --check`
- `cargo test -p rustfs-heal heal::resume::tests`
- `cargo test -p rustfs-heal --lib`
- `cargo test -p rustfs-ecstore compare_and_update_file`
- `git diff --check`
- `make pre-pr`

## Impact
Resume checkpoints now fail closed when a backend cannot provide conditional publication. Corrupt or future-schema snapshots are retained as recovery evidence and require explicit operator cleanup or replacement before resumption.

## Additional Notes
- Correctness adversary: attacked truncation, empty/future/task-ID snapshots, stale writers, and cleanup ordering; no remaining break found after conditional CAS and selector coverage.
- Simplicity adversary: removed the redundant generation field and retained only the save lock plus raw-byte CAS state needed for the stale-writer invariant.
- Security reviewer: attacked task paths, marker handling, and error paths; no auth, injection, or disclosure issue found.
- Concurrency/durability reviewer: attacked concurrent writers, cancellation, and crash publication ordering; the existing temp/sync/rename CAS path preserves the required ordering.
- Compatibility reviewer: attacked schema upgrade/rollback and unsupported backends; future snapshots are version-agnostic quarantined, and unsupported CAS implementations return an explicit failure.
- Performance reviewer: removed the extra healthy-path read/parse; current bytes are only inspected after a CAS mismatch.
- Test-coverage skeptic: added regressions for truncated/future snapshots, stale managers, cleanup resurrection, and blocked selector states.
- Cross-target Windows compilation was attempted but the local host lacks the Windows SDK headers required by `aws-lc-sys`; non-Unix conditional publication is explicitly fail-closed rather than using a racy fallback.


crates/heal/src/heal/resume/utils.rs
```diff
@@ -67,6 +67,7 @@ impl ResumeUtils {
                 // Extract task ID from filename: {task_id}_ahm_resume_state.json
                 if let Some(task_id) = entry.strip_suffix(&format!("_{RESUME_STATE_FILE}"))
                     && validate_resume_task_id(task_id).is_ok()
+                    && CheckpointManager::is_resumable(disk, task_id).await
```
overtrue: Please fail closed instead of filtering this task out. An exact-head production-path test shows that a blocked or corrupt checkpoint disappears from the candidate list, so `get_or_create_task_id` generates a new UUID and starts a fresh heal instead of surfacing `RecoveryRequired`. That bypasses the operator-cleanup contract this marker is meant to enforce.

crates/heal/src/heal/resume/checkpoint.rs
```diff
@@ -148,11 +241,22 @@ impl CheckpointManager {
...
-        let mut checkpoint: ResumeCheckpoint =
-            serde_json::from_slice(&checkpoint_data).map_err(|e| Error::TaskExecutionFailed {
-                message: format!("Failed to deserialize checkpoint: {e}"),
-            })?;
+        Self::load_from_data(disk, task_id, checkpoint_data).await
+    }
+
+    async fn load_from_data(disk: DiskStore, task_id: &str, checkpoint_data: Vec<u8>) -> Result<Self> {
+        validate_resume_task_id(task_id)?;
+        let mut checkpoint: ResumeCheckpoint = match serde_json::from_slice(&checkpoint_data) {
```
overtrue, line 262: Please authenticate the persisted checkpoint before trusting it. A same-length edit from `victim-a` to `victim-b` remains valid JSON and passes the task/schema checks here; an exact-head resume test then treats `victim-b` as already processed and skips the object that still needs healing. A checksum or equivalent integrity envelope is needed at this boundary.

overtrue (changes requested):

crates/heal/src/heal/resume/utils.rs
cxymds: Addressed both review findings in 0cd2ae20e. Corrupt or blocked checkpoints now fail closed from resumable-task discovery instead of being filtered into a new task, and checkpoints are persisted with a SHA-256 sidecar that rejects same-length progress tampering and isolates the task. Added focused production-path tests; `cargo test -p rustfs-heal --lib heal::resume::tests` passes (55/55).

crates/heal/src/heal/resume/checkpoint.rs
cxymds: Addressed both review findings in 0cd2ae20e. Corrupt or blocked checkpoints now fail closed from resumable-task discovery instead of being filtered into a new task, and checkpoints are persisted with a SHA-256 sidecar that rejects same-length progress tampering and isolates the task. Added focused production-path tests; `cargo test -p rustfs-heal --lib heal::resume::tests` passes (55/55).

crates/heal/src/heal/resume/checkpoint.rs
```diff
@@ -307,21 +422,139 @@ impl CheckpointManager {
...
+                EcstoreConditionalFileUpdate::Updated => {}
+                EcstoreConditionalFileUpdate::Missing | EcstoreConditionalFileUpdate::Mismatch => {
+                    return Err(Error::TaskExecutionFailed {
+                        message: "Checkpoint changed while saving; refusing to overwrite newer progress".to_string(),
+                    });
+                }
+            }
+        }
+
+        let digest_path = Self::digest_path(&checkpoint.task_id);
```
overtrue: This digest write is not crash-consistent with the checkpoint CAS. If generation N+1 commits before this write fails or the process exits, restart hashes the valid checkpoint against generation N’s stale or partial digest, marks it blocked, and refuses resume. Please publish both as one recoverable generation and add a fault-injection restart test for this window.

overtrue (changes requested):

crates/heal/src/heal/resume/checkpoint.rs
cxymds: Fixed in e3a362989. New checkpoints carry the integrity digest inside the same JSON payload protected by the checkpoint CAS, so payload and authentication record are published as one recoverable generation. Legacy sidecars remain readable for old snapshots, while new writes no longer have a crash window between checkpoint and digest writes. Added a restart regression covering a missing legacy sidecar; all 56 heal resume tests pass.

overtrue (approved):
LGTM

overtrue (approved):
LGTM

overtrue (approved):
LGTM

overtrue (approved):
LGTM

cxymds:
Fixed the failing Test and Lint (sftp) check and pushed 4d92835d7 to the PR branch. The failure was caused by two current-main Rust lint errors: removed an unnecessary mut in crates/ecstore/src/store/object.rs and replaced unnecessary sort_by with sort_by_key in crates/ecstore/src/store/rebalance/support.rs. Also merged current main without conflicts. Verification passed: cargo fmt --all --check; cargo check -p rustfs-heal --lib; cargo check -p rustfs-ecstore --lib.

overtrue (approved):
LGTM

overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #6342, opened 2026-08-21
fix(heal): fence format repair during pool transitions

cxymds:
## Related Issues
Fixes rustfs/backlog#1925.

## Summary of Changes
`handle_heal_format` now takes a per-pool metadata fence before writing formats. It acquires the existing namespace locks in pool.bin then rebalance.bin order, validates the persisted pool topology and endpoint identity, rejects missing, malformed, stale, decommissioning, or active-rebalance snapshots with `SlowDown`, and treats completed decommissioned sources as a non-fatal `NoHealRequired` skip. The lock lease is checked before and after each format operation so a lost fence fails closed. Recoverable format errors, including top-level and nested `SlowDown`, now reach the heal task manager without being wrapped as terminal task failures. Shared ECStore test fixtures persist the validated pool topology required by this fence.

This is intentionally a bounded runtime fence for the infrastructure available on the current main branch. Pool-generation/CAS fields and startup membership-generation primitives are not present yet; generation-bound commit, dispatch unification, and startup anti-resurrection coverage remain follow-up work tied to the infrastructure tracked by the related backlog work.

Adversarial review verdicts: correctness attacked missing and empty metadata, identity changes, stop-transition states, completed-source handling, and mid-operation lease loss with no unresolved break; simplicity found reuse of existing metadata loaders, namespace locks, and error types without a parallel state manager; security found no new untrusted input or secret exposure and malformed metadata fails closed; concurrency/durability verified pool.bin to rebalance.bin lock order, lease retention through writes, and fail-closed lease loss; compatibility verified unchanged heal result/error aggregation and replacement target-scoped behavior; performance verified per-pool rather than all-pool lock scope and bounded metadata reads around existing async IO; test coverage verified regression checks for barriers, identity, completed skips, later-pool progress, and top-level recoverable propagation.

## Verification
- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore store::heal::tests::heal_format_pool_rebalance_barriers_and_identity_are_fail_closed --lib`
- `cargo test -p rustfs-ecstore store::heal::tests::handle_heal_format_continues_after_a_pool_error --lib`
- `cargo test -p rustfs-heal heal::task::tests::erasure_set_format_slowdown_is_propagated --lib`
- Focused heal integration tests: `test_heal_format_basic`, `test_heal_format_with_data`, `test_heal_resume_across_page_boundary_e2e`, and `test_heal_storage_api_direct`
- `cargo clippy -p rustfs-heal --lib --tests -- -D warnings`
- `make pre-pr` (14,071 tests passed, 137 skipped; all doctests passed)

## Impact
Format-heal requests now retry instead of writing while pool membership, decommission, or rebalance metadata is unavailable or in transition. Completed retired sources are reported as no work. No public API or on-disk format is changed, and replacement format healing remains target-scoped and unchanged. Test-only ECStore fixtures now model the durable topology that production bootstrap supplies.

## Additional Notes
N/A

---


crates/ecstore/src/store/heal.rs
```diff
@@ -69,11 +202,18 @@ impl ECStore {
             r.set_count += result.set_count;
             r.before.drives.append(&mut result.before.drives);
             r.after.drives.append(&mut result.after.drives);
+
+            // Sets::heal_format cannot observe this guard before each disk write;
```
overtrue: Please prevent format writes after this lease is lost. In distributed mode, a target disk can stall inside `save_format_file` while refresh loses quorum and a successor acquires the expired pool/rebalance locks and persists a new transition. The stale call then resumes and renames `format.json`; this check returns `SlowDown` only after the unauthorized write. `Sets::heal_format` needs a pre-write loss check or an equivalent generation fence at the write boundary.

overtrue (changes requested):

crates/ecstore/src/store/heal.rs
cxymds: Added a pre-write generation/lease fence in `Sets::heal_format_with_fence`: every `format.json` write now checks whether the pool or rebalance lease was lost and aborts with `SlowDown` before writing. The caller also keeps the final fail-closed check before reporting success. Verified with `cargo check -p rustfs-ecstore --lib` and `cargo test -p rustfs-ecstore --lib store::heal::tests::handle_heal_format -- --nocapture`.

overtrue (approved):
LGTM

cxymds:
Resolved the main-branch conflict in f7cd90ad9 and pushed the merge result. The existing heal guard tests and the main-branch multi-pool heal tests pass; please re-review the updated head.

overtrue (approved):
LGTM

PR #6351, opened 2026-08-21
fix(heal): supervise scheduler task panics

cxymds:
## Related Issues

Fixes rustfs/backlog#1928

## Summary of Changes

- Wrap the full heal scheduler worker and retry child lifecycle with unwind supervision.
- On a panic, clean active, retry, queue, alias, MRF, and replacement-anchor state idempotently and retain a redacted `Failed` terminal status.
- Preserve normal terminal results, MRF repaired notices, and existing manual-cancel semantics; prevent cancelled retry work from being resurrected as failed.
- Add regression coverage for worker, retry-child, cleanup, and cancellation-versus-panic paths.

## Verification

- `cargo test -p rustfs-heal scheduler_panic_releases_active_slot --lib -- --nocapture`
- `cargo test -p rustfs-heal retry_child_panic_finishes_parent_once --lib -- --nocapture`
- `cargo test -p rustfs-heal cleanup_panic_is_supervised --lib -- --nocapture`
- `cargo test -p rustfs-heal heal::manager::tests --lib -- --nocapture`
- `cargo clippy -p rustfs-heal --lib --tests -- -D warnings`
- `cargo fmt --all --check`
- `make pre-pr` (14,072 nextest passed; doctests passed)

## Impact

Heal tasks that panic no longer retain scheduler capacity or zombie aliases. They become queryable, redacted failed tasks; public request formats and configuration are unchanged.

## Additional Notes

High-risk adversarial review verdicts: correctness — no break found; simplicity — no smaller safe replacement found; security — panic payload is not added to structured logs or task status; concurrency/durability — cancellation and panic ownership are linearly resolved without new lock-order inversions; compatibility — existing manual cancellation behavior and MRF delivery are preserved; performance — added work is confined to worker lifecycle and panic cleanup; test coverage — worker, retry, cleanup, cancellation, and readmission regressions are covered.


crates/heal/src/heal/manager/scheduler.rs
```diff
@@ -13,6 +13,341 @@
...
+        }
+        let removed = removed_task.is_some();
+        publish_active_heal_count(&active);
+        (removed, active.len())
+    })
+    .catch_unwind()
+    .await
+    .unwrap_or((false, 0));
+
+    let _ = AssertUnwindSafe(async {
```
overtrue: Please make this cleanup atomic with duplicate admission. An exact-head focused test pauses after the old active entry is removed while its retry still exists; a concurrent same-target submit returns `Merged`, then this cleanup removes the retry and archives the old task as `Failed`, so the new request is never queued or executed. The test deterministically gets `Merged` instead of `Accepted`. Cleanup must remove only the task generation it owns or serialize the full ownership handoff with admission.

overtrue (changes requested):

crates/heal/src/heal/manager/scheduler.rs
cxymds: Fixed in 4e647ca22. Panic cleanup now acquires queue -> retrying in the same order as admission and checks both ownership locations atomically; if a retry has already been retained or admitted, cleanup preserves it instead of removing/canceling the newer retry. The focused scheduler panic test passes.

cxymds: After merging the latest main, the scheduler panic-supervision path covered by this comment was removed and replaced by the new scheduler flow. I kept main as the conflict winner rather than reapplying the old cleanup patch into incompatible code; `cargo fmt --all --check` and `cargo check -p rustfs-heal --lib` pass on a9fd0f3c9. Please re-review the updated diff.

Issue #1916, opened 2026-02-23
.Values.ingress.tls.secretName Without efect


mkrueger92:
https://github.com/rustfs/rustfs/blame/3cdd2b313b9274290d31a06c441969c267f3fd86/helm/rustfs/templates/ingress.yaml#L34

In commit [e307816](https://github.com/rustfs/rustfs/commit/e30781654d26006cda575fa0746775eb3d29b8d3) this line was changed and now the configuration .Values.ingress.tls.secretName has no effect anymore. This prevents that users can use existing secrets.

It would be nice to be able to use existing secrets for tls certificates.

majinghe:
@mkrueger92 yes, `values.ingress.tls.secret` will no longer be used, now secret tls name is `{{ include "rustfs.fullname" $ }}-tls`. Thx for your PR #1918 ,after PR is merged, i will close this issue.

PR #6356, opened 2026-08-22
perf(ecstore): throttle decommission checkpoints

cxymds:
## Related Issues

- Related: #1908.
- This PR does not address the cross-node stale `pool.bin` snapshot problem tracked by #1906, nor the mixed-version and peer-reload follow-ups tracked by #1916 and #1917.

## Summary of Changes

- Keep decommission stage and current-object updates in memory instead of writing the full pool metadata after every stage.
- Add a local single-flight checkpoint path that rechecks the 30-second/1,000-item thresholds under the save gate, writes a snapshot with an exact progress watermark, and advances the live baseline only after the snapshot succeeds for the same active run.
- Back off failed non-terminal checkpoints for one second without marking progress as persisted; terminal, cancel, and bucket-complete paths retain their existing strong saves.
- Release local metadata locks before peer pool-metadata reloads. Runtime-only checkpoint fields are skipped by serde, so the `pool.bin` format is unchanged.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-ecstore --lib pools_tests` (188 passed)
- `git diff --check`
- `make pre-pr` (all pre-PR checks passed)

## Impact

Decommission no longer performs a full `pool.bin` save for each migrate/cleanup stage. Progress metadata is persisted at most once per 30 seconds or 1,000 counted items during an active run, while terminal transitions remain strongly persisted. No S3 API, on-disk schema, or object-copy/cleanup ordering changes are introduced.

## Additional Notes

The checkpoint counters and last-update timestamp are intentionally best-effort between successful checkpoints; object safety remains protected by the existing copy-before-cleanup and source-equivalence checks. The Terra adversarial review covered correctness, simplicity, concurrency/durability, compatibility, security, performance, and test coverage. Cross-node conditional/generation writes and real pool.bin fault-injection coverage remain follow-up work under the related issues above.

---


overtrue (approved):
LGTM

PR #6357, opened 2026-08-22
ci: detect incomplete and stale scheduled validations

overtrue:
## Related Issues

- rustfs/backlog#1897
- rustfs/backlog#2006

## Summary of Changes

- Report failed, cancelled, timed-out, stale, skipped, and otherwise incomplete scheduled validations through the shared deduplicating issue action.
- Add a daily freshness monitor for 12 critical scheduled workflows so a schedule that is never created or has gone stale cannot look healthy.
- Move critical schedules away from minute zero and keep their UTC slots distinct to reduce top-of-hour delay and drop risk.
- Keep the monitored workflow set and age limits in one manifest, with static wiring guards and deterministic boundary tests.

## Verification

- `python3 scripts/check_scheduled_validation_freshness.py --self-test`
- `python3 scripts/check_test_wiring.py --self-test`
- `black --check scripts/check_scheduled_validation_freshness.py`
- `actionlint` on all changed workflow files
- Read-only live freshness probe against `rustfs/rustfs`; it correctly identified five existing stale or missing scheduled validations.

## Impact

CI-only. Existing scheduled validations run at different minutes within the same UTC hour. The freshness monitor may open an infrastructure issue immediately for existing stale or never-observed schedules; it does not change RustFS runtime behavior.

## Additional Notes

Adversarial review found no blocker: exact age boundaries, missing runs, API failures, unique cron slots, least-privilege job permissions, and token persistence were checked. Rollback is limited to the freshness workflow, manifest, cron-minute changes, and shared action extension.

---

### DIFF
--- a/scripts/check_test_wiring.py
+++ b/scripts/check_test_wiring.py
@@ -274,11 +274,55 @@ def yaml_block(lines: list[str], key: str, indent: int) -> list[str] | None:
     return lines[start:end]
 
 
-def workflow_job_block(lines: list[str], job_name: str) -> str | None:
-    block = yaml_block(lines, job_name, 2)
-    if block is None:
+def workflow_step_block(job_lines: list[str], action: str) -> tuple[int, list[str]] | None:
+    uses_index = next(
+        (
+            index
+            for index, line in enumerate(job_lines)
+            if line.split("#", 1)[0].strip() in {f"uses: {action}", f"- uses: {action}"}
+        ),
+        None,
+    )
+    if uses_index is None:
         return None
-    return "\n".join(line.split("#", 1)[0] for line in block)
+    start = next(
+        (
+            index
+            for index in range(uses_index, -1, -1)
+            if job_lines[index].lstrip().startswith("- ")
+        ),
+        uses_index,
+    )
+    indent = len(job_lines[start]) - len(job_lines[start].lstrip())
+    end = next(
+        (
+            index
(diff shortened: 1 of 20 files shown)


scripts/check_test_wiring.py
```diff
@@ -245,6 +258,69 @@ def check_profile_definitions(root: Path) -> list[str]:
...
+            start = lines.index("  alert-on-failure:") + 1
+        except ValueError:
+            errors.append(f"{relative}: missing alert-on-failure job")
+            continue
+        end = next(
+            (index for index in range(start, len(lines)) if re.fullmatch(r"  [A-Za-z0-9_-]+:", lines[index])),
+            len(lines),
+        )
+        job = "\n".join(line.split("#", 1)[0] for line in lines[start:end])
+        required = (
```
overtrue, line 411: This guard does not verify the issue-writing wiring it claims to protect. Changing an alert job from `issues: write` to `issues: read` leaves both this check and actionlint green, but the local action can no longer create or update an issue; because the watchdog excludes `failure`, scheduled failures can become silent again. Please require the action invocation, token input, and `issues: write`, with mutation coverage.

overtrue: Fixed in 97b3a36f7. The guard now requires `issues: write`, the shared action invocation, and `github-token`, with an independent mutation for each failure mode.

houseme (approved):

PR #6363, opened 2026-08-22
fix(ecstore): fence deletes against decommission commits

overtrue:
## Related Issues

Closes rustfs/backlog#1910.

## Summary of Changes

- Fence decommission PUT, multipart, delete, and source-cleanup commits against namespace-lock loss.
- Preserve versioned and suspended delete-marker semantics while routing batch deletes away from decommission sources.
- Retain source-set locks through cleanup preflight and preserve per-pool batch delete errors.
- Add focused race, lock-domain, delete-marker, and error-aggregation regression coverage.

## Verification

- `cargo fmt --all --check`
- `make clippy-check`
- `cargo nextest run -p rustfs-ecstore -E '<12 focused decommission fence and delete-marker tests>'` — 12 passed
- Correctness and test-coverage adversary — APPROVE, no blocking findings
- Simplicity adversary — APPROVE, no blocking findings

## Impact

Decommission now fails closed when its mutation fence is lost and no longer removes valid delete markers or hides source-pool delete errors. No public API or persisted-format changes.

## Additional Notes

Draft while the full workspace test phase completes in CI. Local `make pre-pr` passed formatting, repository guard scripts, script tests, and Clippy; its full workspace test phase was stopped at the disk-safety threshold after the focused tests had passed.

---

### DIFF
--- a/crates/ecstore/src/store/object.rs
+++ b/crates/ecstore/src/store/object.rs
@@ -353,6 +353,8 @@ impl fmt::Display for ObjectLockDiagMode {
 
 pub(crate) struct ObjectLockDiagGuard {
     guard: rustfs_lock::NamespaceLockGuard,
+    #[cfg(test)]
+    test_namespace_lock_fence: Option<NamespaceLockFence>,
     enabled: bool,
     op: &'static str,
     bucket: Option<String>,
@@ -374,6 +376,8 @@ impl ObjectLockDiagGuard {
     ) -> Self {
         Self {
             guard,
+            #[cfg(test)]
+            test_namespace_lock_fence: None,
             enabled,
             op,
             bucket,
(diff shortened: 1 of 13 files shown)


crates/ecstore/src/store/object.rs
```diff
@@ -2817,32 +3273,104 @@ impl ECStore {
...
+            }
+
+            let pool_opts = opts.clone();
+            futures.push(async move {
+                #[cfg(test)]
+                let pool_object_names = pool_objects
+                    .iter()
+                    .map(|object| object.object_name.clone())
+                    .collect::<Vec<_>>();
+                let result = pool.delete_objects(bucket, pool_objects, pool_opts).await;
```
overtrue, line 3336: This drops the pool's DeleteAccounting for successful body deletes. The S3 DeleteObjects path needs it to decrement bucket usage, so an unversioned compressed batch delete leaves usage at 1000 instead of returning to 0. The existing production test compressed_delete_requests_restore_usage_baseline fails on this exact head. Please use delete_objects_with_accounting and propagate each pool result.

Issue #1904, opened 2026-02-21
JSON S3 Select fails with "DataFusion error: Arrow error: Json error: Not valid JSON: EOF while parsing an object"


myrust-go:
## Description

RustFS S3 Select implementation has a critical bug when processing JSON DOCUMENT format. The operation fails with the following error:

DataFusion error: Arrow error: Json error: Not valid JSON: EOF while parsing an object at line 1 column 1

This makes JSON-based S3 Select queries unusable, severely limiting RustFS compatibility with AWS S3.

## Steps to Reproduce
1. Set up RustFS (tested on https://play.rustfs.com)
2. Create a bucket and upload a JSON document file:
 ```
  ```json
   {
     "employees": [
       {"name": "Alice", "age": 30, "salary": 75000},
       {"name": "Bob", "age": 25, "salary": 65000}
     ]
   }

```
3. Execute S3 Select query:

```
response = s3_client.select_object_content(
    Bucket="test-bucket",
    Key="employees.json",
    Expression="SELECT e.name FROM s3object[*].employees[*] e WHERE e.salary > 70000",
    ExpressionType='SQL',
    InputSerialization={'JSON': {'Type': 'DOCUMENT'}},
    OutputSerialization={'JSON': {}}
)
```

## Expected Behavior

The query should return filtered JSON results:
```
{"name": "Alice"}
```

## Actual Behavior

The operation fails with:
```
DataFusion error: Arrow error: Json error: Not valid JSON: EOF while parsing an object at line 1 column 1
```

## Additional Information
- CSV S3 Select works correctly - only JSON format is affected
- JSON LINES format also fails with similar errors
- Tested on: https://play.rustfs.com/ (public demo instance)
- Credentials: rustfsadmin / rustfsadmin
- RustFS Version: Latest (as of 2026-02-21)

## Root Cause Analysis

Based on code analysis, the issue appears to be in the `s3s` library integration. RustFS uses:
- `datafusion = "52.1.0"`
- `s3s = { version = "0.13.0-alpha.3", git = "[https://github.com/s3s-project/s3s.git"](https://github.com/s3s-project/s3s.git%22) }`
The problem likely stems from:
Incomplete JSON parsing implementation in the s3s library
Buffer handling issues during streaming JSON processing
Missing JSON validation before passing data to DataFusion

## Impact
- High severity: Breaks JSON-based analytics workflows
- Affects Iceberg compatibility: Iceberg often uses JSON metadata
[...]

PR #6367, opened 2026-08-22
fix(ecstore): fail closed on unresolved decommission entries

cxymds:
## Related Issues

- Related to #1904 (decommission listing can silently skip unresolved metadata).
- Part of the decommission/rebalance remediation tracked by #1901.

## Summary of Changes

- Convert a partial listing whose `MetaCacheEntries::resolve` returns `None` into a typed fail-closed error instead of silently dropping the entry.
- Record the first unresolved-listing or entry failure in the shared error slot, cancel the active listing, and propagate the failure after listing workers drain.
- Apply the same failure propagation to the post-decommission verification sweep so unresolved metadata cannot be reported as a clean completion.
- Preserve meaningful listing failures over cancellation-only errors and ignore late error writes after cancellation.
- Include bucket, prefix, pool index, set index, candidate count, and disk-error count in the unresolved metadata error context.
- Avoid copying an object name on the successful resolution path.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-ecstore --lib pools_tests --no-fail-fast` (188 passed)
- `make pre-pr` (14,084 tests passed, 137 skipped; workspace doctests passed)
- Terra high-risk adversarial review: correctness, simplicity, security, concurrency/durability, compatibility, performance, and test-coverage roles completed with no blocking findings.

## Impact

This is a safety-focused runtime change with no S3 API, xl.meta, on-disk, or on-wire format changes. When listing metadata cannot be resolved, decommission now fails closed and retains source data instead of completing silently; operators must repair the degraded metadata and retry or clear the failed operation according to the existing decommission workflow.

## Additional Notes

This PR intentionally covers the immediate data-safety boundary only. The persistent unresolved-entry ledger, automatic retry/heal workflow, and independent final-sweep design requested by the broader #1904 plan remain follow-up work. End-to-end fault injection through the concrete `list_path_raw` partial callback is also a follow-up coverage improvement; the current regression tests cover the shared error propagation and cancellation contracts directly.

---

Thank you for contributing to RustFS!

### DIFF
--- a/crates/ecstore/src/core/pools.rs
+++ b/crates/ecstore/src/core/pools.rs
@@ -814,19 +814,20 @@ fn resolve_decommission_partial_listing_entry(
     set_index: usize,
 ) -> Result<MetaCacheEntry> {
     let candidate_count = entries.as_ref().iter().flatten().count();
-    let candidate = entries.as_ref().iter().flatten().map(|entry| entry.name.clone()).next();
+    if let Some(entry) = entries.resolve(resolver) {
+        return Ok(entry);
+    }
 
-    entries.resolve(resolver).ok_or_else(|| {
-        decommission_unresolved_listing_error(
-            bucket,
-            prefix,
-            candidate.as_deref(),
-            candidate_count,
-            disk_error_count,
-            pool_index,
-            set_index,
-        )
-    })
+    let candidate = entries.as_ref().iter().flatten().map(|entry| entry.name.as_str()).next();
+    Err(decommission_unresolved_listing_error(
+        bucket,
+        prefix,
+        candidate,
+        candidate_count,
+        disk_error_count,
+        pool_index,
+        set_index,
+    ))
 }
 
 async fn record_decommission_entry_error(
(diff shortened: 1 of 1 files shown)


crates/ecstore/src/core/pools.rs
```diff
@@ -5053,27 +5133,62 @@ impl SetDisks {
...
+                                event = EVENT_DECOMMISSION_BUCKET,
+                                component = LOG_COMPONENT_ECSTORE,
+                                subsystem = LOG_SUBSYSTEM_POOLS,
+                                bucket = %bucket,
+                                prefix = %prefix,
+                                state = "unresolved_entry",
+                                error = %err,
+                                "Decommission listing failed closed on unresolved metadata"
+                            );
+                            record_decommission_entry_error(&unresolved_error, &unresolved_rx, err).await;
```
overtrue, line 5179: This wiring is not covered by the new tests: deleting this call restores the silent resolve(None) path while every added test still exercises only the resolver, recorder, or precedence helpers. Please add a focused test that drives a real list_path_raw partial result through list_objects_to_decommission and asserts both migration and the final sweep fail closed; cancellation itself returns Ok(()), so helper-only coverage cannot catch a regression here.

overtrue (changes requested):

overtrue (approved):
LGTM

PR #6368, opened 2026-08-22
fix(rpc): bound internode concurrency under multipart load

### LINKED ISSUE #6350: 1.0.0-rc.3 regression: multipart uploads fail with 503 SlowDown above ~64 in-flight parts — rename_part reports write-quorum failure while quorum is met
Multipart uploads that succeed on `1.0.0-rc.2` fail on `1.0.0-rc.3` once a modest number of parts are in flight, and they fail *wholesale* — a single-digit percentage of failed part-PUTs takes every upload with it, so the observable result is that nothing uploads.

The threshold is low enough to be ordinary traffic rather than a stress test: it breaks from about 64 concurrent part-PUTs, which is **8 simultaneous 200 MiB uploads at the AWS SDK's default 8 MiB part size and default part concurrency**. Anything driving parallel multipart — the SDKs' transfer managers, backup products, `s3 cp`/`rclone` at default settings — reaches that on a small cluster. Plain `PutObject` is unaffected at four times that concurrency (below), so the scope is the multipart path specifically. No data is lost or corrupted: every integrity check after every run came back clean. This is a write-availability regression, in the release immediately after the one it works on.

**Environment.** `1.0.0-rc.3` (`1aae6803739a`), stock `linux-x86_64-musl` release asset, 3 nodes × 4 drives, EC:4 over 12, Alpine on KVM, 4 vCPU / 8 GB per node, 16-core host, no oversubscription.

**Same host, same hour, same client, same fixture — only the release differs.** Each arm is a freshly built cluster on the same 6720-object / 14.5 GiB fixture, at `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT=30`.

| uploads × concurrent parts, object/part size | 1.0.0-rc.2 | 1.0.0-rc.3 |
|---|---|---|
| 8 × 16, 200 MiB / 8 MiB | **8/8, 0 errors** | 0/8 |
| 16 × 16, 200 MiB / 8 MiB | **16/16, 0 errors** | 0/16, 125 × 503 |
| 16 × 16, 400 MiB / 8 MiB | **16/16, 0 errors** | 0/16, 558 × 503 |
| 24 × 16, 400 MiB / 8 MiB | **24/24, 0 errors** | 0/24, 1068 × 503 |

Every rc.2 arm is clean; every rc.3 arm fails.

Clients get `503 SlowDown` plus a minority of `500 InternalError`. Six independent rc.3 runs reproduce it, one on a freshly restored cluster verified idle first (`healing: false`, 12/12 online), so it is not residual damage. Driving one client per node instead of a single endpoint gives 0/72 either way, so it is not entry-point concentration. Serial part uploads are unaffected. Integrity after every run: 6720 objects OK, 0 missing, 0 corrupt.

**Where the threshold is, and what it tracks.** Below ~32 concurrent part-PUTs everything is clean; it breaks from ~64 upward on these 4-vCPU nodes:

| uploads × part-workers | concurrent part-PUTs | result |
|---|---|---|
| 1 × 2, 1 × 4, 2 × 4, 4 × 8 | 2 – 32 | **clean, 0 errors** |
| 8 × 8 | 64 | 0/8 uploads, 13 × 503 |
| 8 × 12 | 96 | 0/8 uploads, 84 errors |
| 8 × 16 | 128 | 2/8 uploads, 16 errors |
| 16 × 8 | 128 | 0/16 uploads, 144 errors |

**Plain `PutObject` is clean at the same and higher concurrency**, which bounds the scope and argues against generic write saturation:

| plain `PutObject` | concurrent | result |
|---|---|---|
| 128 × 8 MiB | 64 | **128/128, 0 errors** |
| 256 × 8 MiB | 128 | **256/256, 0 errors** |
| 512 × 8 MiB | 256 | **512/512, 0 errors** |

896 objects and ~7 GiB through the same endpoint, same client settings, same `retries=0`, with no errors at four times the concurrency where multipart fails.

Two things fall out of the multipart table. The *amplification*: at 64 concurrent only ~13 of ~200 part-PUTs fail (~6%), yet **no upload completes**, because one failed part aborts its whole upload. And, more diagnostically, the last two rows have **identical** total part concurrency but differ ~9× in errors — the variable is the number of concurrent *uploads* (16 vs 8). So the failure tracks concurrent uploads, not raw part parallelism, which points at per-upload state rather than I/O saturation. Exact counts vary run to run; the clean/broken boundary does not.

**Server side.** At `warn` plus `rustfs_ecstore::set_disk=debug`, one run logged this four times (parts 20, 22, 24):

```
Set disk multipart write quorum unavailable
op=upload_part  stage=rename_part
required=8  achieved=11  failed=1  total=12   offline_disks=1
retryable_failures=0  dominant_error=nil_dominated
returned_error="erasure write quorum"
```

three of them identical except `achieved=10 failed=2 offline_disks=2`.

**`achieved` is 10–11 against `required` 8 in every instance** — quorum met with room to spare, `dominant_error=nil_dominated`, `retryable_failures=0` — yet the call returns `erasure write quorum`, which `rustfs/src/error.rs` maps to `S3ErrorCode::SlowDown`. Hence the 503, and hence why its advice misleads: slowing down is not the remedy.

`rename_part` (`crates/ecstore/src/set_disk/core/io_primitives.rs`, unchanged at main tip `bc07cfd11`) has two exits that log this. The `reduce_write_quorum_errs` one cannot have produced it, since the summary shows more successes than quorum requires and no dominant error. The other is `recover_part_transaction` returning `committed == false`, which then sets `DiskError::ErasureWriteQuorum` unconditionally — a quorum error for a part-transaction *commit* failure. That is by elimination, not instrumentation: the `RUSTFS_ISSUE3031_DIAG_ENABLE` hook in that same function emitted nothing in any run, which may be worth a look by itself.

Two separable problems, then, the second independent of the first:

1. Above the threshold the part-transaction commit does not settle. I have not isolated why; the machinery looks unchanged from rc.2 by symbol count, but `io_primitives.rs` (+1764 lines) and `disk/os.rs` (+1506) changed a lot between the tags. That it scales with concurrent uploads, and that `offline_disks` is 1–2 in the summaries above while all 12 drives are online and serving, both look like threads worth pulling.
2. A commit failure is reported as `ErasureWriteQuorum` → `SlowDown`, misdirecting both the client and anyone reading the code. #5839 and #3031 were earlier quorum-accounting problems on this same function, so distinguishing "quorum short" from "commit did not settle" may be worth making explicit.

**Already ruled out:** the two early-stop defaults that flipped in rc.3 (#6140, #6156) — setting both back to `false` changes nothing; FD exhaustion — peak 1229 of 4096; and `FILE_FDATASYNC_GROUP_COMMIT` / `PUT_RENAME_EARLY_ACK`, both default-off here.

**One observability note:** in the run that produced 218 client-visible 503s, the node terminating those requests logged a single WARN — and at the default `info` level there is no trace of the failure at all.

Happy to help narrow it — I have both release images and a rig that reproduces in under a minute, so I can test a patch and report before/after on the same cluster, or bisect the 172-commit range (the `rc.3-preview` tags have no published binaries, so that needs builds).

overtrue:
## Related Issues

Fixes #6350

## Summary of Changes

Limit each cached internode gRPC channel to 64 active requests. Excess RPCs now wait in the existing channel buffer instead of exhausting h2's per-connection small-DATA-frame budget and tearing down unrelated lock and disk RPCs during highly concurrent multipart uploads.

The failure first appears with the dependency refresh at `84bd76a3`, which moved h2 from 0.4.15 to 0.4.16. The quorum path itself is unchanged; the reported write-quorum error is downstream of the shared RPC connection closing under load.

## Verification

- `cargo fmt --all --check`
- `cargo zigbuild -p rustfs-protos --release --target aarch64-unknown-linux-musl --locked`
- `cargo zigbuild -p rustfs --release --target aarch64-unknown-linux-musl --locked`
- 3-node × 4-disk cluster, 8 concurrent 200 MiB multipart uploads, 16 part workers: rc.2 `8/8`, rc.3 `0/8`, pre-fix main `0/8`, patched main `8/8` twice

## Impact

Internode RPCs above 64 active requests per channel are queued. There are no S3 API, configuration, authentication, or wire-format changes.

## Additional Notes

- Correctness: exercised the original 128-part-worker pressure case twice; all uploads completed and no quorum errors surfaced.
- Simplicity: one shared channel limit covers control and optional bulk channels without changing quorum or multipart code.
- Security: authentication, request signing, deserialization, and secrets are untouched.
- Concurrency/durability: the Tower limit queues at channel readiness; it does not drop requests or alter commit ordering.
- Compatibility: the gRPC protocol and cached-channel keys are unchanged.
- Performance: the repeated patched run averaged 25.8 seconds per upload versus 27.9 seconds on rc.2 in the same setup.
- Test coverage: removing the limit reproduces the pre-fix `0/8`; the patched build completed `8/8` in both fresh-cluster runs.

---

### DIFF
--- a/crates/protos/src/lib.rs
+++ b/crates/protos/src/lib.rs
@@ -2106,6 +2106,9 @@ pub enum ChannelClass {
     Bulk,
 }
 
+// Keep multiplexed unary RPCs below h2's per-connection small-frame budget.
+const INTERNODE_RPC_CONCURRENCY_LIMIT: usize = 64;
+
 /// Whether control/bulk channel isolation is enabled (env-gated, default off for safe rollout).
 fn channel_isolation_enabled() -> bool {
     rustfs_utils::get_env_bool(
@@ -2188,6 +2191,7 @@ async fn build_channel(dial_addr: &str, cache_key: &str) -> Result<Channel, Box<
     let mut connector = Endpoint::from_shared(dial_addr.to_string())?
         // Fast connection timeout for dead peer detection
         .connect_timeout(connect_timeout)
+        .concurrency_limit(INTERNODE_RPC_CONCURRENCY_LIMIT)
         // TCP-level keepalive - OS will probe connection
         .tcp_keepalive(Some(tcp_keepalive))
         // Disable Nagle so latency-sensitive control-plane RPCs (locks/health) are not batched


houseme (approved):

lddlww:
hi，does this feature released？

PR #6374, opened 2026-08-22
fix(ecstore): reject equal-time latest identity conflicts before index fallback

overtrue:
## Related Issues

- Fixes rustfs/backlog#1918.
- Part of rustfs/backlog#1901.

## Summary of Changes

- Compare production-visible object identity before using the pool index to resolve equal-time candidates.
- Fail closed with `ErasureReadQuorum` when equal-time replicas differ.
- Normalize RustFS and MinIO internal metadata aliases while preserving case-sensitive dynamic ARN and resource identities.

## Verification

- `cargo fmt --all --check`
- `git diff --check HEAD^ HEAD`
- Correctness, security, compatibility, and test-coverage review: APPROVE.
- Simplicity, performance, concurrency/durability, and test-quality review: APPROVE.
- Focused resolver tests are covered by the pull request CI; no local build was started after the final metadata-case fix to avoid unsafe artifact growth.

## Impact

- Equal-time replica divergence is no longer hidden by deterministic pool ordering.
- Equivalent internal metadata aliases remain compatible; dynamic target identifiers remain case-sensitive.
- No API, configuration, deployment, or documentation changes.

## Additional Notes

- Pool-specific erasure geometry and the transient reader handle remain excluded from identity comparison because they are not shared replica identity.

---

### DIFF
--- a/crates/ecstore/src/store/rebalance/support.rs
+++ b/crates/ecstore/src/store/rebalance/support.rs
@@ -12,10 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::cmp::Ordering;
+use std::collections::HashMap;
 
 use crate::error::{Error, Result, StorageError, is_err_object_not_found, is_err_version_not_found};
 use crate::object_api::{ObjectInfo, ObjectOptions};
+use rustfs_utils::http::metadata_compat::strip_internal_prefix;
 use rustfs_utils::path::decode_dir_object;
 use time::OffsetDateTime;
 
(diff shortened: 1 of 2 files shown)


crates/ecstore/src/store/rebalance/support.rs
```diff
@@ -137,37 +141,158 @@ pub(super) fn rebalance_disk_set_lookup_error(pool_idx: usize, set_idx: usize, p
...
+        && left.name == right.name
+        && left.storage_class == right.storage_class
+        && left.mod_time == right.mod_time
+        && left.size == right.size
+        && left.actual_size == right.actual_size
+        && left.is_dir == right.is_dir
+        && same_user_defined_identity(left, right)
+        && left.user_tags == right.user_tags
+        && left.version_id == right.version_id
+        && left.data_dir == right.data_dir
```
overtrue, line 236: A data-movement target preserves the source mod_time/version/etag but generates a new data_dir and data-moved metadata. Before source cleanup, normal multi-pool GET/HEAD sees both equal-time copies here and returns ErasureReadQuorum. A two-pool production-path test on this exact head reproduces it immediately after migrate_object. Please normalize metadata that migration intentionally rewrites before deciding the copies conflict.

overtrue: Fixed in #6447. Equal-time source and committed migration targets now use the directional movement identity check, while unmarked data_dir conflicts still fail closed.

PR #6375, opened 2026-08-22
fix(replication): deny non-owner replication config edits under site replication

reatang:
## Related Issues

Fixes rustfs/backlog#1948 (B5-rc T1, audit item A1/P2-17; spec rustfs/backlog#1946, parent rustfs/backlog#1675).

## Summary of Changes

Under site replication, a user granted only bucket-scoped `s3:PutReplicationConfiguration` could overwrite or delete the bucket's replication config verbatim — wiping the operator-managed `site-repl-*` rules (and their `ReplicationService` bucket targets on DELETE) — and the result was broadcast to every peer, silencing replication net-wide for up to a reconciler period (600s). A bucket-scoped permission escalated to a cluster-wide blast radius.

- **Gate (interface layer)**: `PutBucketReplication` / `DeleteBucketReplication` now deny non-owner requests with MinIO-parity `XMinioReplicationDenyEdit` (HTTP 400) while site replication is enabled. The gate runs after policy authorization and only on the external S3 path; the reconciler and peer bucket-meta ingestion never route through it. A site-replication state read failure fails closed. The spec placed this in the access layer; the layer-dependency ratchet forbids new infra→interface imports, so it lives at the S3 handler entry in `ecfs.rs` — same request-path semantics.
- **Defense in depth (usecase)**: PUT now merges the incoming config with the stored `site-repl-*` rules (reusing the same merge as peer ingestion — user rules replaced, local site rules preserved, incoming `site-repl-*` impostors dropped); DELETE keeps `site-repl-*` rules and never garbage-collects a bucket target a surviving site-replication rule still references. Buckets without site-replication rules keep the exact previous overwrite/delete semantics. Read-merge-write stays under `lock_bucket_targets_metadata` with incarnation guards intact.
- Moved `is_site_replication_rule` / `merge_incoming_replication_config` / `replication_target_arn_deployment_id` verbatim from the admin site-replication handler down to `rustfs-replication` so the app layer can reuse them without new layering violations; removed the now-dead `replication_targets_without_config_targets` wrapper.

## Verification

- `cargo test -p rustfs --lib bucket_usecase::tests` (89 passed, incl. 5 new merge/split tests)
- `cargo test -p rustfs --lib replication_config_edit_gate` (end-to-end through the S3 trait methods: non-owner PUT/DELETE → `XMinioReplicationDenyEdit` 400; owner and SR-off pass through to the usecase)
- `cargo test -p rustfs --lib site_replication::tests` (226 passed), `cargo test -p rustfs-replication`
- `cargo fmt --all`; `make pre-commit` (fmt + unsafe/arch/layer/logging guards + quick check)
- Adversarial review (high-risk tier, all 7 roles): correctness — attacked role-based configs, disabled SR rules, ARN trim mismatch, rollback-after-target-write-failure, no break; security — fail-closed state read, gate unreachable from peer/reconciler paths, test override is `cfg(test)`-only; concurrency — read-merge-write under the existing lock with incarnation guards; test-coverage — gate wiring is revert-detected via trait-method tests. Residual: the usecase-internal merge/split call sites lack a full-store revert test (no such harness in this crate's unit tests).

## Impact

Behavior changes (S3-visible, rolling-upgrade safe — no on-disk or on-wire format change):

1. Non-owner `PutBucketReplication` / `DeleteBucketReplication` on a site-replication deployment now returns 400 `XMinioReplicationDenyEdit` (was 200 + net-wide config overwrite). MinIO parity.
2. Owner `DeleteBucketReplication` on a site-replication-managed bucket now removes only operator-authored rules; `GetBucketReplication` afterwards returns the surviving `site-repl-*` rules instead of `NoSuchReplicationConfiguration`, and site-replication bucket targets are no longer garbage-collected by the delete.
3. The PUT broadcast payload now carries the merged config (including local `site-repl-*` rules); peers already filter incoming `site-repl-*` rules on ingestion — the bucket-meta sync path has always shipped them on the wire, so mixed-version peers within the supported rc line are unaffected.

4. **Operator rule priorities and roles are preserved only under the derived-rule contract (review round 2, c6799da20).** The S3 edit path probes every remote peer via `peer/edit-capabilities` (`derived-rule-contract`); if every peer supports it, operator priorities and an operator `Role` target survive the merge and the broadcast carries `derivedRuleContract: true`. If any peer predates the contract, the edit is merged under the legacy contract (prefix classification, 1..n renumbering) so the whole cluster stays consistent — operator rules are listed in priority order first, so the winning overlapping rule does not flip. Rolling-upgrade notes: during the mixed period edits are canonicalized cluster-wide (priority values lost, order kept); configs canonicalized then are **not** renumbered back after the last site upgrades — re-submit the config to restore the values; adding an older-version site after priorities were preserved is not gated and leaves that bucket drifted until the next edit.

## Additional Notes

`site-repl-*` rule preservation deliberately renumbers priorities and clears a site-replication ARN in `role`, matching the existing peer-ingestion merge so the reconciler's no-op check stays byte-stable.


crates/replication/src/config.rs
```diff
@@ -265,6 +265,78 @@ pub fn active_replication_rule_destination_arns(config: &ReplicationConfiguratio
...
+        return Some(parts[4].to_string());
+    }
+
+    None
+}
+
+/// Whether `rule` is a site-replication rule (`site-repl-*` id) owned by the
+/// local site's reconciler rather than authored by an operator.
+pub fn is_site_replication_rule(rule: &ReplicationRule) -> bool {
+    rule.id.as_deref().is_some_and(|id| id.starts_with("site-repl-"))
```
overtrue: This prefix check also classifies a user-authored rule as site-owned. Rule IDs do not reserve `site-repl-*`, so on a site-replication-disabled bucket an owner can create `site-repl-user`; DeleteBucketReplication then keeps that rule and target while returning success. Please distinguish reconciler-owned rules from arbitrary IDs and add a PUT-to-DELETE regression.

crates/replication/src/config.rs
```diff
@@ -265,6 +265,78 @@ pub fn active_replication_rule_destination_arns(config: &ReplicationConfiguratio
...
+            .flat_map(|config| config.rules)
+            .filter(is_site_replication_rule),
+    );
+
+    if rules.is_empty() {
+        return None;
+    }
+
+    for (index, rule) in rules.iter_mut().enumerate() {
+        rule.priority = Some(i32::try_from(index + 1).unwrap_or(i32::MAX));
```
overtrue: Reassigning priorities changes the submitted replication policy. On a site-replication bucket, an owner PUT with overlapping same-target rules ordered priority 5 then priority 1 is rewritten to 1 then 2, so the lower-priority delete-marker rule wins. Please preserve operator priorities and add a merge-to-replication-decision regression.

overtrue (changes requested):

crates/replication/src/config.rs
reatang: Agreed — the prefix alone is not ownership. A rule is now reconciler-owned only when it matches what the reconciler derives: id `site-repl-<deployment-id>` for a *current remote site-replication peer* and a destination ARN naming that same deployment id (`site_replication_rule_deployment_id` + `is_reconciler_owned_site_replication_rule`). The S3 put/delete path reads the remote peer set (`site_replication_remote_peer_deployment_ids`, empty when SR is disabled) and keeps exactly those rules; everything else, including an owner's `site-repl-user`, is operator state. Incoming rules claiming a current peer's id are dropped so the reconciler rule's id stays unique. The peer ingestion merge and the reconciler keep their prefix predicate unchanged.

Regression: `put_then_delete_replication_without_site_replication_treats_site_repl_id_as_user_rule` (PUT verbatim, DELETE removes the config and garbage-collects the target) plus `delete_replication_split_keeps_only_reconciler_derived_rules`. Commit a3733c1a1.

crates/replication/src/config.rs
reatang: Fixed. Operator priorities are now kept verbatim; only the derived site rules move, to the lowest priorities no operator rule uses (`assign_site_replication_rule_priorities`). The same helper is used by the S3 edit merge, the peer ingestion merge, the reconciler pass and the peer-removal prune — the latter two renumbered 1..n as well, so leaving them would have rewritten the operator's priorities within one reconciler period.

The helper is a pure, idempotent function of the rule list, so the reconciler's `rules == existing_rules` no-op check still holds after a merged write (tested), and the historical on-disk layout (operator rules 1..k, site rules k+1..n) produces identical bytes, so nothing is rewritten on upgrade. Regression `merge_keeps_operator_priorities_and_replication_decision` feeds the merged config to `filter_target_replication_decisions` and asserts the winning rule matches the submitted config (the priority-5 delete-marker rule wins). Commit a42d81b26.

overtrue (approved):
LGTM

crates/replication/src/config.rs
```diff
@@ -265,6 +265,160 @@ pub fn active_replication_rule_destination_arns(config: &ReplicationConfiguratio
...
+/// satisfy (no bucket target backs it) and replication silently stops. Only
+/// operator-authored rules travel: the site-replication peer ingestion path
+/// and the S3 put/delete-bucket-replication path both keep the local site's
+/// `site-repl-*` rules through this merge. `incoming == None` models a
+/// delete of the operator-authored rules.
+pub fn merge_incoming_replication_config(
+    incoming: Option<ReplicationConfiguration>,
+    local: Option<ReplicationConfiguration>,
+) -> Option<ReplicationConfiguration> {
+    merge_replication_config_keeping_site_rules(incoming, local, is_site_replication_rule)
```
overtrue: This still treats every incoming `site-repl-*` ID as reconciler-owned. An owner-authored `site-repl-user` rule is preserved locally by the new S3 merge, then dropped by this peer-ingestion merge, so sites persist different operator configs. Please classify incoming rules by the same derived ID/ARN contract and add a PUT-to-peer regression.

crates/replication/src/config.rs
```diff
@@ -265,6 +265,160 @@ pub fn active_replication_rule_destination_arns(config: &ReplicationConfiguratio
...
+    if rules.is_empty() {
+        return None;
+    }
+
+    assign_site_replication_rule_priorities(&mut rules, &is_site_rule);
+
+    // A site-replication ARN in `role` is the sender's, and the reconciler's
+    // per-peer target lookup reads it — carrying it over would pin the
+    // receiver's targets to the sender's identity.
+    let role = match replication_target_arn_deployment_id(&incoming_role) {
```
overtrue: This clears any valid target ARN carried in `Role`, including an owner-submitted role target such as `arn:minio:replication::operator-dep:bucket`. The merged config then selects rule destination ARNs instead of the validated role target. Please preserve operator roles and add a merge-to-target-selection regression.

crates/replication/src/config.rs
```diff
@@ -265,6 +265,160 @@ pub fn active_replication_rule_destination_arns(config: &ReplicationConfiguratio
...
+
+/// Give the site rules in `rules` the lowest priorities no operator rule uses,
+/// in rule order, leaving every operator rule's priority untouched. Operator
+/// priorities decide which rule wins per target, so they are part of the
+/// submitted policy; site rules are derived state and only need to be unique
+/// (`validate_replication_config_structure` rejects duplicates). The result
+/// is a pure function of the rule list, so the site-replication reconciler,
+/// the peer ingestion merge and the S3 edit merge all converge on the same
+/// bytes and the reconciler's no-op check holds.
+pub fn assign_site_replication_rule_priorities(rules: &mut [ReplicationRule], is_site_rule: impl Fn(&ReplicationRule) -> bool) {
```
overtrue, line 469: This is not rolling-upgrade safe. A new peer broadcasts operator priorities `5,1`, but an old peer production merge rewrites them to `1,2`, which can select the opposite overlapping rule; upgrading later cannot recover the original priorities. Please add a mixed-version compatibility path or gate before preserving priorities.

overtrue (changes requested):
Please address the inline blockers.

reatang:
Quick Checks failed on the layer guard after the review fix (`bucket_usecase` imported `site_replication_remote_peer_deployment_ids` from the admin handlers — app→interface). Fixed in 74171bd67: the S3 handlers in `ecfs.rs` (interface) now read the peer set and pass it into `execute_put/delete_bucket_replication` as a parameter, so the usecase stays a pure function of its inputs and the baseline is untouched. Fail-closed semantics on a state-read error are unchanged, just one layer up.

crates/replication/src/config.rs
reatang: Done in 4e0b9d934. The peer ingestion merge now classifies by the same derived id/ARN contract as the S3 merge (`is_reconciler_owned_site_replication_rule`), using the receiver's full site set (its own id included, since the sender's rule towards the receiver names it). The reconciler, the peer-removal prune and the target-online probe switched from the id prefix to the derived shape as well — otherwise the reconciler would have dropped `site-repl-user` 600s later on both sites. Regression `peer_ingestion_keeps_owner_site_repl_user_rule_and_sites_agree` merges a PUT on site A, ingests the broadcast on site B and asserts both operator rule sets are identical.

crates/replication/src/config.rs
reatang: Done in 44de73b66. `Role` is cleared only when it is a site-replication ARN naming a current site (`is_site_replication_role`); an owner role such as `arn:minio:replication::operator-dep:bucket` is kept by both merges and the reconciler's repair pass. Regression `merge_keeps_operator_role_target_for_target_selection` feeds the merged config to `replication_target_arns` / `filter_target_arns` and asserts the owner role target is selected, while a role naming a peer is still cleared.

crates/replication/src/config.rs
reatang: Done in c6799da20, as a gate. The S3 edit path probes every remote peer through the existing `peer/edit-capabilities` endpoint (new capability `derived-rule-contract`; older peers answer `success:false` or 404) and preserves operator priorities only when every peer supports it. Otherwise the edit merges under an explicit `Legacy` contract that reproduces the pre-contract merge byte for byte (prefix classification, 1..n renumbering), with one addition: operator rules are listed in priority order first, so the renumbering keeps their relative order and the winning overlapping rule does not flip. Every bucket-meta item a new site sends carries `derivedRuleContract: true` (bootstrap/retry snapshots included); a receiver merges an unmarked payload the legacy way.

Rolling-upgrade behavior (also in the commit message; will be added to the PR body): during the mixed period edits are canonicalized cluster-wide (values lost, order kept); configs canonicalized then are not renumbered back after the upgrade — re-submitting the config restores the values. Not gated: adding an older-version site after priorities were preserved (that bucket drifts until the next edit). Regressions cover the "one old peer" vs "all new" probe sets and the legacy/derived merge outputs.

PR #6376, opened 2026-08-22
fix(admin): expose per-target disableProxy through remote target admin API

reatang:
## Related Issues

Fixes rustfs/backlog#1950 (B5-rc T3, audit item A2/P2-10; spec rustfs/backlog#1946, parent rustfs/backlog#1675). Inconsistency disclosed in #6170.

## Summary of Changes

#6172 (P1-5) wired the read-proxy consumer for `disable_proxy`: `get_proxy_targets` skips targets whose `TargetClient` carries the flag, and the field is persisted/loaded via bucket-targets.json. But the admin API still rejected it: `REMOTE_TARGET_UNSUPPORTED_FIELDS` listed `disableProxy`, and the update path rejected the `proxy` op. The per-target read-proxy opt-out therefore worked in code but did not exist on the API — the only way to enable it was importing a MinIO-written bucket-targets.json.

- Move `disableProxy` from `REMOTE_TARGET_UNSUPPORTED_FIELDS` to `REMOTE_TARGET_WRITABLE_FIELDS`; `set-remote-target` (create) now accepts it.
- Add `TargetUpdateOp::Proxy`: `set-remote-target?update=true&proxy=true` overlays only the proxy group onto the stored target (MinIO `TargetUpdateType` parity, same per-group overlay mechanism as #5715); a proxy-only update body may omit credentials like the other non-creds groups.
- Bump `REMOTE_TARGET_CAPABILITY_CONTRACT_VERSION` 1 → 2; the runtime-capabilities endpoint now reports `disableProxy` as supported under `remote_targets`.
- `edge` / `edgeSyncBeforeExpiry` (fields and update ops) remain rejected — nothing implements them.
- New regression test pins that a published `TargetClient` carries `disable_proxy`, the field the proxy-target selector consults.

## Verification

- TDD: compile-level red for `TargetUpdateOp::Proxy`; assertion-level red on the contract-version pin before the bump (55 passed / 1 failed — the pin), then green.
- `cargo nextest run -p rustfs --lib -E 'test(/admin::handlers::(replication|system)::tests/)'` — 56/56
- `cargo nextest run -p rustfs-ecstore` proxy/target tests — 5/5 (incl. new `update_all_targets_publishes_disable_proxy_on_target_client`)
- `cargo test -p rustfs-replication` — all green
- `cargo fmt --all`; `make pre-commit`
- Note: under same-process `cargo test`, two pre-existing topology tests (`runtime_capabilities_response_reports_missing_topology_before_storage_init`, `server_info_response_carries_pool_topology`) share global state and can interfere; reproduced on the unmodified baseline, not under nextest process isolation. Unrelated to this change and left alone (rc scope discipline).

## Impact

Admin API: `disableProxy` becomes writable on create and via the new `proxy` update op. Capability contract version 1 → 2.

Mixed-version / rolling-upgrade behavior (the remote-target capability contract is version-locked on the client side; declared policy is `fail_closed_when_capability_unknown_or_unsupported`):

- **Old node (contract v1) receiving the new field/op**: rejected with `InvalidRequest` ("not supported by this RustFS version") — fail-closed, no silent drop. In a mixed cluster a client can see the toggle succeed or fail depending on which node serves the admin call; clients honoring the capability contract will not offer the toggle against a v1 node.
- **New node reading old metadata**: `disableProxy` has always been part of the persisted bucket-targets format (defaults to false); no migration.
- **Flag set on a new node, read by an old node**: nodes at ≥ #6172 honor `disable_proxy` in the read-proxy selector; nodes older than #6172 ignore the flag and keep proxying reads to that target until upgraded (same degradation shape as the P1-7 disclosure).

## Additional Notes

N/A

### DIFF
--- a/crates/ecstore/src/bucket/bucket_target_sys.rs
+++ b/crates/ecstore/src/bucket/bucket_target_sys.rs
@@ -3425,6 +3425,44 @@ mod tests {
         assert!(mutexes.contains_key("second"));
     }
 
+    #[tokio::test]
+    async fn update_all_targets_publishes_disable_proxy_on_target_client() {
+        // The read-proxy selector (replication_proxy::get_proxy_targets) skips
+        // targets whose TargetClient carries disable_proxy — the persisted
+        // per-target opt-out must survive client publication.
+        let sys = BucketTargetSys::default();
+        let target = |arn: &str, disable_proxy: bool| BucketTarget {
+            arn: arn.to_string(),
+            endpoint: "192.168.1.10:9000".to_string(),
+            target_bucket: "target-bucket".to_string(),
+            region: "us-east-1".to_string(),
+            disable_proxy,
+            credentials: Some(Credentials {
+                access_key: "access".to_string(),
+                secret_key: "secret".to_string(),
+                session_token: None,
+                expiration: None,
+            }),
+            ..Default::default()
+        };
+        let targets = BucketTargets {
+            targets: vec![target("arn:proxied", false), target("arn:opted-out", true)],
+        };
+
+        sys.update_all_targets("bucket", Some(&targets)).await;
+
+        let proxied = sys
+            .get_remote_target_client("bucket", "arn:proxied")
+            .await
+            .expect("client should be published");
+        assert!(!proxied.disable_proxy);
+        let opted_out = sys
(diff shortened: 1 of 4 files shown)


overtrue (approved):
LGTM

PR #6377, opened 2026-08-22
fix(admin): use madmin key names in list-remote-targets response

reatang:
## Related Issues

Fixes rustfs/backlog#1951 (B5-rc T4, audit item B2/P2-11; spec rustfs/backlog#1946, parent rustfs/backlog#1675).

## Summary of Changes

The admin `list-remote-targets` response serialized five fields under the persisted snake_case keys — `bandwidth_limit`, `storage_class`, `reset_id`, `deployment_id`, and `credentials.session_token` — while madmin/mc decode the response under their own JSON tags (`bandwidth`, `storageclass`, `resetID`, `deploymentID`, `sessionToken`). The mismatched fields silently decode to zero values, so `mc replicate ls` shows empty bandwidth-limit and reset-id columns. Symmetrically, `BucketTarget` deserialization only aliased `bandwidth`, so a MinIO-written `bucket-targets.json` silently dropped the other four fields to defaults on migration.

- Response side: rename the five keys to madmin's tags inside `remote_target_admin_json` only, following the existing Go-duration re-encode precedent in that function. The inbound `RemoteTargetRequest` DTO already accepts these names, so mc's list-then-update round-trip stays closed.
- Read side: add `#[serde(alias = ...)]` for `storageclass`, `resetID`, `deploymentID`, and `sessionToken` so MinIO-written metadata populates the matching fields.

**Persisted and wire formats are unchanged.** `BucketTarget` is simultaneously the persisted `bucket-targets.json` format (written via `serde_json::to_vec(&BucketTargets)`, read back in `metadata.rs`, and embedded verbatim in the cluster-shared `BucketMetadata` msgpack) and the msgpack struct-map form (`BucketTargets::marshal_msg`), so its serde `rename`s must not change. This PR touches neither: serde aliases affect deserialization only, and the madmin naming is applied in the admin response layer after serialization. A new pinning test locks the snake_case keys in both the JSON and msgpack encodings as a regression fence.

## Verification

- TDD (red then green):
  - `list_remote_targets_response_uses_madmin_key_names` — pins the five madmin keys in the admin response and asserts the snake_case keys are gone (red: `bandwidth` was Null).
  - `minio_written_bucket_targets_json_populates_madmin_named_fields` — a madmin-shaped `bucket-targets.json` sample populates all five fields (red: `storage_class` defaulted to empty).
  - `bucket_target_persisted_wire_keys_stay_snake_case` — persisted JSON and msgpack keys stay snake_case (green before and after).
- `cargo test -p rustfs-ecstore --lib bucket_target::tests` (17 passed)
- `cargo test -p rustfs --lib admin::handlers::replication::tests` (40 passed, incl. existing duration/round-trip pins)
- `cargo fmt --all`; `make pre-commit`
- Adversarial review: attacked credentials=None / redacted token=null / key collisions / override ordering; redaction runs before the admin JSON step so `sessionToken` output is always null; no unresolved findings.

## Impact

Admin API response: `list-remote-targets` now emits madmin-compatible key names for the five fields (mc shows bandwidth limit and reset id). On-disk and inter-node formats unchanged. Migration from MinIO-written `bucket-targets.json` no longer loses those fields.

## Additional Notes

N/A


rustfs/src/admin/handlers/replication.rs
```diff
@@ -391,6 +399,11 @@ fn remote_target_admin_json(target: &BucketTarget) -> Result<serde_json::Value,
         "avg": go_duration_nanos(target.latency.avg),
         "max": go_duration_nanos(target.latency.max),
     });
+    rename_key(&mut value, "bandwidth_limit", "bandwidth");
```
overtrue: madmin-go v3.0.109 decodes this field from `bandwidthlimit`, not `bandwidth`. The live ListRemoteTarget path therefore returns a zero bandwidth limit to mc for every nonzero target limit. Please emit the exact pinned tag and cover it with a madmin-shaped decode regression.

overtrue (changes requested):

rustfs/src/admin/handlers/replication.rs
reatang: Good catch, thanks. Verified against both the madmin-go v3.0.109 source (`bucket-targets.go`) and the tags embedded in the mc binary: `BucketTarget.BandwidthLimit` is `json:"bandwidthlimit,omitempty"`; the other four tags (`storageclass`, `resetID`, `deploymentID`, `credentials.sessionToken`) were already correct. Fixed in cddc003c1:

- `remote_target_admin_json` now emits `bandwidthlimit`.
- The ecstore `BucketTarget` reader gains a `bandwidthlimit` alias (the pre-existing `bandwidth` alias was wrong for the same reason; kept as legacy). Persisted snake_case keys are untouched.
- The inbound `RemoteTargetRequest` already accepted `bandwidthlimit`, so set-remote-target was unaffected.
- Added `list_remote_targets_response_decodes_through_madmin_tags`, which decodes the response through a mirror of the v3.0.109 json tags and asserts the nonzero bandwidth limit; the key-name pin test now asserts `bandwidthlimit` and rejects `bandwidth`. Both failed before the fix and pass now.

overtrue (approved):
LGTM

PR #6378, opened 2026-08-22
fix(admin): bound site replication lifecycle lock and parallelize add preflight

reatang:
## Related Issues

Fixes rustfs/backlog#1952 (B5-rc T5, audit item C2; spec rustfs/backlog#1946, parent rustfs/backlog#1675). Also covers the peer-timeout-constant pinning part of rustfs/backlog#1889 C1.

## Summary of Changes

The site replication add flow probed every peer site serially while holding the process-wide lifecycle lock, and every other lifecycle entry point (remove, peer join admission, peer remove, resync, service-account rotation) acquired that lock with no timeout. With k unreachable sites, each probe ate the full 10s peer request timeout (up to 2 requests per site), so the lock was held for tens of seconds to minutes while concurrent add/remove/endpoint-refresh requests hung indefinitely and background reconcile/retry-drain rounds (which `try_acquire`) were starved for the whole window.

- Probe all sites in the add preflight concurrently via `join_all`, matching the file's other peer fan-outs; k unreachable sites now cost roughly one peer timeout instead of k. Errors are still reported in request order with unchanged messages.
- Bound the lifecycle lock acquire at 30s (3× the peer request timeout — long enough for a healthy concurrent operation's full peer round-trip, short enough to turn a holder wedged on unreachable peers into a retryable 503 for waiters); on timeout, handlers return `ServiceUnavailable`. All 6 production acquire sites are covered; the success path is unchanged; background reconcilers keep using the non-blocking `try_acquire`. Lock ordering is untouched.
- Keep the 10s request / 3s connect peer timeout constants unchanged and pin them with a regression test.

## Verification

- `test_add_preflight_probes_sites_concurrently` — barrier-gated local server; only concurrent probes can finish inside half a peer request timeout (red under the serial loop: `Elapsed`).
- `test_lifecycle_guard_acquire_times_out_with_retryable_503` — paused-clock test: second acquire gets 503 instead of hanging, and succeeds after release (red under the unbounded acquire).
- `test_peer_timeout_constants_bound_unreachable_peer_probes` — pins the 10s/3s constants and `LOCK_TIMEOUT >= REQUEST_TIMEOUT`.
- Adjacent lifecycle tests (`test_add_bootstrap_scope_only_allows_expected_bucket_setup_until_guard_drops`, `test_add_lifecycle_allows_callback_before_remove_writer`) still green.
- `cargo fmt --all`; targeted `cargo test -p rustfs --lib`; `make pre-commit`
- Adversarial review (high-risk: locking/replication): correctness — attacked error ordering / borrows / multi-local-site matching; concurrency — lock order unchanged, tokio Mutex timeout dequeues cleanly; compatibility — 503 only appears where the request previously hung forever; no unresolved findings.

## Impact

Behavior changes (disclosed):

- **New 503 path**: admin lifecycle endpoints (replicate add/remove/resync/rotate, peer join/remove) can now return 503 `ServiceUnavailable` ("another site replication lifecycle operation is in progress; retry later") when the lock is held longer than 30s — previously these requests hung indefinitely. The error is retryable; re-running once the in-flight operation finishes succeeds.
- Preflight probes sites after a failing one (read-only GETs); previously probing stopped at the first failure. First-error-in-request-order reporting is unchanged.

No on-disk or on-wire format change.

## Additional Notes

The lock and the serial preflight live in the admin handler layer (`rustfs/src/admin/handlers/site_replication.rs`), not in ecstore as the original audit coordinates suggested; re-verified before the change.


overtrue (approved):
LGTM

PR #6379, opened 2026-08-22
fix(replication): apply receiver-side LWW to inbound metadata categories

reatang:
## Related Issues

Fixes rustfs/backlog#1953 (B5-rc T6, audit item A4/P1-6 — receiver-side LWW missing; spec rustfs/backlog#1946, parent rustfs/backlog#1675). Completes the receiver half of #6129 (P1-6).

## Summary of Changes

Metadata-only replication reuses the whole-object PUT/multipart transports, so in active-active topologies an inbound authorized replication write carried the source's tags, object-lock retention, and legal hold verbatim and unconditionally overwrote a category the destination had modified more recently: tag A at t1 on site A and tag B at t2>t1 on site B ended as A=t2 / B=t1 forever, with both sides reporting COMPLETED and no self-heal. Per the owner decision on rustfs/backlog#1953 this lands entirely on the receiver (route 2): the existing whole-object transports are kept unchanged — no metadata-only operation, no capability probing, no contract bump; CopyObject semantics (route 1) remain a post-stable evolution.

The receiver now judges each category independently under the object write lock, only for authorized replication requests carrying that category's source timestamp (`x-{rustfs,minio}-source-replication-{tagging,retention,legalhold}-timestamp`, shipped by the P1-6 sender): a stored timestamp strictly newer than the inbound one keeps the local values and timestamp; otherwise the inbound category wins and its internal timestamp key is pinned to the source-authored time instead of the receiver-clock value stamped by the object-lock eval-metadata path. Absent stored timestamps (pre-P1-6 data) or absent inbound timestamps keep today's overwrite behavior, and a local win never fails the write (a rejected write would loop through MRF re-delivering the stale value).

- The `put_object` hook reuses the existing commit-lock WORM-gate read (no extra disk fanout).
- `complete_multipart_upload` adds one gated read under its held lock, and the sender's complete-multipart options now forward the three timestamps so the multipart transport (used when the source object is multipart) gets the same receiver behavior.
- Local wins restore the timestamp via `insert_str` so a MinIO-written single-key version still yields both compatibility keys.
- A skipped multipart comparison on a destination-read failure is surfaced at `warn` (`replication_lww_read_unavailable`); the first-delivery not-found case stays quiet.

- **Sender-side fixes surfaced by the dual-node e2e (review round 2, 344ad517c)**: `PutObjectOptions::header()` now carries `x-amz-tagging` built from `user_tags` (MinIO `PutObjectOptions.Header` parity) — metadata replication previously delivered tagless replicas; and `metadata_target_is_eligible` no longer requires a per-target COMPLETED status on a REPLICA version (it defers to the rule's ReplicaModifications decision, MinIO `mustReplicate` parity), so replica-side tag/retention/legal-hold edits in active-active actually replicate instead of being silently dropped.

**Documented tradeoffs and asymmetries:** (1) equal timestamps let the inbound side win (MinIO parity, keeps replays idempotent) — two sites writing different values at the exact same instant can still swap states rather than converge on one; (2) on a destination-read failure the two transports differ: `put_object` rides the pre-existing WORM gate, which fails the write on non-not-found read errors (fail closed), while `complete_multipart_upload` keeps its fail-open semantics and applies the inbound metadata without comparison (now logged at `warn`); (3) non-RustFS replication sources (e.g. minio-go senders) may not send the category timestamp headers on CompleteMultipartUpload, in which case multipart objects keep today's overwrite behavior for those senders.

## Verification

- `cargo test -p rustfs-ecstore --lib replication_lww` (8 hermetic SetDisks behavior tests; 5 red before the fix — e.g. "site=remote overwrote site=local", "stale ON resurrected a released hold" — green after; 2 compatibility guards green before and after)
- `cargo test -p rustfs-ecstore --lib -- complete_multipart_upload_stale_replication_tags_keep_local replication_complete_multipart_options_sets_actual_size`
- `cargo build --bin rustfs && cargo test -p e2e_test receiver_lww_keeps_newer_tags_across_delivery_orders` (live-server convergence across delivery orders: newer timestamp wins regardless of arrival order, stale delivery still succeeds at object level)
- `cargo build --bin rustfs && cargo test -p e2e_test test_site_replication_tagging_lww_converges_active_active_real_dual_node` (2b7c569b8; real dual-node site replication: reciprocal edits with COMPLETED/REPLICA on both sides, concurrent edits with the stale delivery released last, MRF replay after restarting the peer; fails deterministically with the receiver merge disabled)
- `cargo fmt --all`, `cargo check -p rustfs`, `make pre-commit` (incl. logging guardrails)
- Adversarial validation (high risk, all seven roles; correctness + concurrency by an independent reviewer against the final diff): lock coverage on both hooks, lock-loss fencing ordering, serialization with local tag/metadata writers, `parts_metadatas` rewrite alignment, dual-key compatibility — no blocker. Findings fixed: dual-internal-key restoration on local wins; revert-detecting coverage for the timestamp-pinning branch; operator visibility of the skipped multipart comparison.

## Impact

S3-visible only for authorized replication writes: an inbound replication write carrying an older per-category timestamp no longer overwrites a newer local tags/retention/legal-hold state; the write still succeeds and all other metadata applies per the inbound request (MinIO replication LWW semantics). No wire-format or contract changes — the existing P1-6 headers are additionally sent on CompleteMultipartUpload, which already rides an authorized replication request shape; old receivers ignore them, old senders keep today's behavior. MinIO interop preserved via dual internal metadata keys. One new `warn` event (`replication_lww_read_unavailable`) on multipart LWW read failures.

## Additional Notes

Pre-existing adjacent issues observed and deliberately left out of this PR (tracked on rustfs/backlog#1953, fix coming as a stacked PR): the explicit-version WORM gate rejects inbound overwrites of versions under an active hold/retention (replication retry, no divergence); a locally cleared retention stored as empty-string mode trips that gate's strict parse; buckets without a replication config do not stamp category timestamps on local tag writes. Route 1 (CopyObject-semantics transport) remains the post-stable direction.


overtrue (approved):
LGTM

crates/e2e_test/src/replication_lww_receiver_test.rs
```diff
@@ -0,0 +1,150 @@
...
+        .expect("object tagging should be readable");
+    tagging
+        .tag_set()
+        .iter()
+        .find(|tag| tag.key() == tag_key)
+        .map(|tag| tag.value().to_string())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn receiver_lww_keeps_newer_tags_across_delivery_orders() -> TestResult {
```
overtrue, line 100: This still exercises one receiver with injected replication headers. The linked acceptance criterion requires reciprocal two-site replication, concurrent tag edits, convergence to the newer timestamp on both sites, and both replication statuses. Please cover that path with the existing two-site harness.

overtrue (changes requested):

overtrue (dismissed):
LGTM

crates/e2e_test/src/replication_lww_receiver_test.rs
overtrue, line 100: The backlog acceptance requires reciprocal two-site concurrent tagging, convergence on both sites, correct replication statuses, and MRF replay. This single-server test injects authorized replication PUTs directly, so sender, worker, status, and replay regressions can still pass. Please extend the existing dual-site harness to exercise those paths and assert both final states and statuses.

overtrue (changes requested):

crates/e2e_test/src/replication_lww_receiver_test.rs
reatang: Thanks for insisting on the dual-site path — it was the right call. I extended the existing real dual-node site-replication harness (`replication_extension_test`) with `test_site_replication_tagging_lww_converges_active_active_real_dual_node` (2b7c569b8), kept the single-server receiver test as the fast unit-level check, and the new test immediately surfaced two pre-existing sender-side gaps that the injected-PUT test could never see (fixed in 344ad517c):

1. `bucket_target_sys::PutObjectOptions::header()` never emitted `x-amz-tagging` from `user_tags`, so metadata replication delivered a tagless replica (only the timestamp header travelled). Fixed with MinIO `PutObjectOptions.Header` parity + unit test.
2. `metadata_target_is_eligible` (#5696) required a per-target COMPLETED status that a REPLICA version never carries, so every replica-side tag/retention/legal-hold edit in active-active was silently dropped. A REPLICA version now defers to the rule's ReplicaModifications decision (MinIO `mustReplicate` parity); source-side PENDING/FAILED stays gated. Unit test added.

The e2e now covers: reciprocal edits in real order (author COMPLETED / receiver REPLICA on both sides); concurrent edits with the *stale* delivery arriving last (a new content-keyed proxy hold parks each site's delivery at the peer, B's edit is newer, A's stale edit is released first — B must keep its newer tag through a quiet window while the delivery still succeeds, then both sites settle on B's value with no PENDING/FAILED); and MRF replay (B stopped, A edits, B restarted in place, replay converges forward, object stays a single version). With the receiver-side merge disabled it fails deterministically (3/3) at the rollback assertion; with it enabled it passes consistently (~25–37s). Retention/legal-hold are not yet exercised end-to-end — that needs an object-lock-enabled bucket; proposing it as a follow-up.

overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #6382, opened 2026-08-22
fix(heal): correct progress accounting

cxymds:
## Related Issues

Closes rustfs/backlog#1933.

## Summary of Changes

- Replace the legacy additive progress fallback with an explicit object-sweep ledger for scanned, healed, failed, and skipped outcomes.
- Keep stage progress separate from object progress and expose additive indeterminate, running, completed, baseline, generation, and counter-fence fields while preserving existing wire fields.
- Persist object counts, bytes, baseline generation, and overflow/unknown state through resume state and the authoritative checkpoint; migrate older schemas to an explicitly unknown baseline.
- Require ledger convergence and successful cleanup before publishing terminal 100%, and make counter overflow non-fatal while preserving an unknown progress state.
- Bind usage baselines to the heal scope and add regression coverage for retries, restart restoration, old schemas, torn checkpoint writes, terminal cleanup ordering, and overflow.

## Verification

- `cargo fmt --all --check`
- `cargo check -p rustfs-heal`
- `cargo test -p rustfs-heal --lib` (326 passed)
- `cargo test -p rustfs-heal --tests`
- `git diff --check`
- `make pre-pr` (14,090 nextest tests passed, 137 skipped; all doctests passed)
- Seven-role adversarial validation completed with no unresolved findings.

## Impact

Heal status no longer double-counts scanned and healed objects, invents a percentage without a trusted baseline, or reports 100% before the terminal ledger and cleanup converge. Existing serialized fields and checkpoint schema remain compatible through additive defaults.

## Additional Notes

Checkpoint writes publish the unknown-counter fence before newer counters so recovery cannot observe an apparently newer but unfenced summary. Build artifacts were removed with `cargo clean` after verification. PR status will not be monitored per the task workflow.


crates/heal/src/heal/erasure_healer.rs
```diff
@@ -1026,12 +1172,31 @@ impl ErasureSetHealer {
...
+                    );
+                    if telemetry_unknown {
+                        progress.mark_unknown();
+                    }
+                    progress.counter_unknown
+                };
+                if telemetry_unknown || progress_unknown {
+                    checkpoint_manager.mark_counter_unknown().await?;
+                }
+                checkpoint_manager
```
overtrue: The checkpoint ledger is not crash-atomic. Each dedup identity and this aggregate update advance the shared throttle separately, so an identity-triggered flush can survive without its counters/bytes. At the page boundary, the durable resume cursor also advances before `complete_page` flushes this ledger. Either crash window skips already-recorded work with stale telemetry. Please commit page identities and outcomes together before exposing the next cursor, with threshold and page-boundary restart regressions.

overtrue (changes requested):

crates/heal/src/heal/erasure_healer.rs
cxymds: Fixed in 69ae9ef67. Object identity and aggregate progress now persist through one checkpoint mutation. At each page boundary, the checkpoint position and ledger are durably saved before publishing the resume cursor, and identities are pruned only after the cursor save succeeds. Added a disk-reload regression test covering the checkpoint ledger and page-boundary ordering.

crates/heal/src/heal/erasure_healer.rs
```diff
@@ -1023,15 +1174,42 @@ impl ErasureSetHealer {
...
+                };
+                checkpoint_manager
+                    .record_object_outcome(
+                        key,
+                        checkpoint_outcome,
+                        *successful_objects,
+                        *failed_objects,
+                        *skipped_objects,
+                        bytes_processed,
+                        0,
```
overtrue: These zeros overwrite special-skip totals already recorded earlier in the page. In the existing mixed normal/new-version production test, a durable checkpoint assertion gets 0 instead of 1; after the cursor is saved and identities are pruned, a crash before the bucket summary restores the undercount permanently. Please pass the current skipped-new and ILM totals into this atomic snapshot.

crates/heal/src/heal/task/heal_bucket.rs
```diff
@@ -384,7 +387,10 @@ impl HealTask {
                     }
 
                     let mut progress = self.progress.write().await;
-                    progress.update_progress(scanned, healed, failed, bytes);
+                    progress.update_object_progress(scanned, healed, failed, skipped, bytes);
```
overtrue, line 397: This publishes a scanned object before its queued retry has a terminal outcome, so progress marks counter_unknown permanently and the later successful retry cannot recover it. The existing retry production-path test fails after adding an assertion that counter_unknown stays false. Please exclude pending retries from the terminal ledger or model them separately.

overtrue (changes requested):

cxymds:
Addressed both review points in 79a837147. Normal erasure-object outcomes now carry the current skipped-new-version and ILM-expired totals into the atomic checkpoint snapshot instead of overwriting them with zero. Bucket heal retries are no longer counted or published as terminal progress until a retry reaches a terminal outcome; the existing retry test passes with scanned=2 and healed=2. Verification: cargo fmt --all --check; cargo check -p rustfs-heal --lib; cargo test -p rustfs-heal --lib heal::task::tests::test_recursive_bucket_heal_retries_only_retryable_objects -- --nocapture; cargo test -p rustfs-heal --lib heal::erasure_healer::resume_loop_tests -- --nocapture.

overtrue (approved):
LGTM

overtrue (approved):
LGTM

cxymds:
Fixed the failing Test and Lint (swift) check and pushed 276d8e156. The failure was caused by two Rust lint errors from the current main integration: removed an unnecessary mut in crates/ecstore/src/store/object.rs and replaced unnecessary sort_by with sort_by_key in crates/ecstore/src/store/rebalance/support.rs. Also merged current main without conflicts. Verification passed: cargo fmt --all --check; cargo check -p rustfs-heal --lib; cargo check -p rustfs-ecstore --lib.

overtrue (approved):
LGTM

cxymds:
Fixed the CI Clippy failure by grouping the object outcome and aggregate progress fields into `CheckpointObjectOutcomeRecord`, preserving the atomic checkpoint update semantics without suppressing the lint. Verified with `cargo fmt --all --check`, `cargo check -p rustfs-heal --lib`, and `cargo clippy -p rustfs-heal --lib -- -D warnings`.

overtrue (approved):
LGTM

crates/heal/src/heal/storage.rs
```diff
@@ -785,11 +790,30 @@ impl HealStorageAPI for ECStoreHealStorage {
...
+                };
+                baseline.bytes = match baseline.bytes.checked_add(usage.size) {
+                    Some(total) => total,
+                    None => return Ok(None),
+                };
             }
         }
 
+        let identity = info.snapshot_identity();
+        let mut hasher = DefaultHasher::new();
```
overtrue: This value is persisted and compared after restart, but `DefaultHasher` is explicitly not stable across Rust toolchains. A binary/toolchain upgrade can therefore turn the same usage snapshot and bucket scope into a different generation; the resume path then marks valid counters unknown permanently. Please use a canonical, toolchain-stable digest and pin a fixed-vector restart test. The repository already documents this invariant in `site_replication_identity.rs`.

crates/heal/src/heal/progress.rs
```diff
@@ -162,6 +382,101 @@ impl HealProgress {
...
+    for progress in progresses {
+        found = true;
+        let object_sweep = matches!(progress.kind, HealProgressKind::ObjectSweep);
+        has_object_sweep |= object_sweep;
+        all_ledgers_complete &= progress.ledger_complete;
+        if object_sweep {
+            all_object_baselines_known &= progress.baseline_known;
+            match baseline_generation {
+                None => baseline_generation = Some(progress.baseline_generation),
+                Some(generation) => baseline_generation_consistent &= generation == progress.baseline_generation,
```
overtrue, line 425: This equality check conflicts with generation construction: every normal erasure-set task hashes `set_disk_id`, pool, and set into its generation, while the manager can run four different sets concurrently. Two healthy tasks using the same validated usage snapshot therefore get different generations, so this aggregate sets `baselineKnown=false` and reports 0/indeterminate. Keep task/resume identity separate from the shared snapshot generation, or aggregate per task, and add a two-set regression.

crates/heal/src/heal/erasure_healer.rs
```diff
@@ -775,20 +881,48 @@ impl ErasureSetHealer {
...
+                            *processed_objects,
+                            *successful_objects,
+                            *failed_objects,
+                            *skipped_objects,
+                            bytes_processed,
+                        );
+                        if !counter_ok {
+                            progress.mark_unknown();
+                        }
+                        progress.clone()
```
overtrue: This clones the entire `HealProgress`, including the freshly allocated `current_object` string, for every terminal version only to copy scalar checkpoint fields. A large heal now adds one extra string allocation/copy per version. Build `CheckpointObjectOutcomeRecord` while the guard is held and return that record plus the unknown flag; apply the same change to the other two clone sites.

overtrue (changes requested):
Two correctness blockers remain in the persisted and aggregated baseline-generation contract. There is also an avoidable per-object allocation in the heal loop.

crates/heal/src/heal/storage.rs
cxymds: Fixed in 9837cc7c7. Persisted heal baseline generation now uses an explicit length-delimited FNV-1a digest over canonical snapshot fields and sorted scope, which is stable across Rust toolchains. Added a fixed-vector test for the digest.

crates/heal/src/heal/progress.rs
cxymds: Fixed in 9837cc7c7. Per-set progress now retains the shared usage-snapshot generation instead of mixing set identity into it; resume/task identity remains separate. Added an aggregate regression proving two healthy sets from one snapshot remain baseline-known.

crates/heal/src/heal/erasure_healer.rs
cxymds: Fixed in 9837cc7c7. The three terminal outcome paths now build only a CheckpointObjectOutcomeRecord while holding the progress lock, then persist the record without cloning the full HealProgress/current_object string.

overtrue (changes requested):
The multi-set regression is still not pinned at its producer. The new test in crates/heal/src/heal/progress.rs:869 only aggregates hand-built values that already share generation 7, so it would still pass if crates/heal/src/heal/task/heal_bucket.rs:447 salted generations with set identity again. Please add a producer-level regression that applies the same snapshot to two distinct set IDs and asserts equal baseline_generation; the now-unused set_disk_id parameter can then be removed from this boundary.

overtrue (changes requested):
Re-reviewed at 7e7a44212. The heal/proto/admin diff is unchanged from 9837cc7c7, so the outstanding producer-level multi-set baseline regression is still missing. apply_erasure_set_usage_baseline still accepts an unused set ID, while the current multi-set test only aggregates hand-built equal generations. Please apply one snapshot to two distinct set IDs, assert equal baseline_generation, and remove the unused parameter.

overtrue (approved):
LGTM

PR #6384, opened 2026-08-22
fix(scanner): discover sub-quorum heal candidates

cxymds:
## Related Issues
Fixes rustfs/backlog#1934.

## Summary of Changes
The scanner now performs bounded raw xl.meta discovery for abandoned-child objects, preserving valid sub-quorum object and delete-marker version candidates that normal quorum resolution can omit. Discovery validates object identity, metadata shape, checksums, version UUIDs, and candidate names; deduplicates replicas; separates versioned, delete-marker, and unversioned candidates; records bounded overflow; and performs no writes or deletes. Heal admission and normal read/write quorum behavior remain unchanged. Versionless retry reconstruction is explicitly non-destructive, and existing pending-heal reason data records sub-quorum discovery without changing the durable schema.

## Verification
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-filemeta discover_heal_candidates -- --nocapture` (8 passed)
- `cargo test -p rustfs-scanner scanner_folder::tests -- --nocapture` (87 passed)
- `cargo clippy -p rustfs-filemeta -p rustfs-scanner --all-targets -- -D warnings`
- Architecture, layer-dependency, unsafe-code, and logging guard scripts
- `make pre-pr` (all fast checks, 14,110 nextest tests passed with 137 skipped, and workspace doctests passed)

## Impact
Scanner heal discovery covers divergent sub-quorum histories while retaining bounded candidate and overflow memory. Malformed, nil, or unsafe metadata is not turned into a destructive versioned heal request; unversioned retries remain non-destructive. Normal quorum resolution and S3-visible behavior are unchanged.

## Additional Notes
The final seven-role adversarial review attacked correctness, simplicity, security, concurrency/durability, compatibility, performance, and test coverage; no concrete break or missing regression check was found.

---


crates/filemeta/src/metacache.rs
```diff
@@ -546,6 +771,33 @@ impl MetaCacheEntries {
     }
 }
 
+fn valid_heal_candidate_name(bucket: &str, entry: &MetaCacheEntry) -> bool {
+    if bucket.is_empty()
+        || entry.name.is_empty()
+        || entry.is_dir()
+        || entry.name.contains('\\')
```
overtrue: This rejects a supported object key on Linux/macOS. Live PUT validation accepts `object\name`, and the local path layer preserves the backslash as filename data there. If that object becomes sub-quorum, discovery drops it and queues no heal. Please make this validation platform-aware and keep a production-contract regression for a supported backslash key.

crates/filemeta/src/metacache.rs
```diff
@@ -370,6 +417,184 @@ impl MetaCacheEntries {
...
+                    discovery.truncated = true;
+                    if discovery.truncated_objects.len() < MAX_META_CACHE_HEAL_TRUNCATED_OBJECTS
+                        && !discovery.truncated_objects.iter().any(|object| object == &candidate.object)
+                    {
+                        discovery.truncated_objects.push(candidate.object.clone());
+                    }
+                    // The remaining versions in this raw entry cannot add a
+                    // bounded candidate; avoid parsing a very long history
+                    // after the safe continuation has been recorded.
+                    break;
```
overtrue: This continuation cannot recover dropped historical versions. After 1,024 candidates only the object name is kept; scanner sends `version_id=None`, which reaches `read_all_fileinfo` with an empty version and selects only current/latest. A stable 1,025th historical version is therefore omitted every scan. Please retain a bounded resumable version cursor or equivalent exact-version enumeration, with a production-path regression.

overtrue (changes requested):

cxymds:
Addressed both review comments and pushed aa3f16fe0. Candidate-name validation now rejects backslashes only on Windows, while Linux/macOS preserve them as object-key data. Overflow candidates beyond the main cap are retained with exact version IDs and queued as versioned heal requests; they are no longer downgraded to version_id=None/latest-version inspection. Added/updated production-contract coverage. Verification passed: cargo fmt --all --check; cargo test -p rustfs-filemeta --lib metacache::tests::discover_heal_candidates_rejects_nil_and_malformed_metadata_and_is_bounded -- --nocapture; cargo check -p rustfs-scanner --lib.

crates/scanner/src/scanner_folder.rs
```diff
@@ -1904,65 +1906,112 @@ impl FolderScanner {
...
+                                if seen_truncated_objects.len() >= MAX_META_CACHE_HEAL_TRUNCATED_OBJECTS
+                                    && !seen_truncated_objects.contains(&candidate.object)
+                                {
+                                    continue;
+                                }
+                                seen_truncated_objects.insert(candidate.object.clone());
+                                if !seen_heal_candidates.insert(identity) {
+                                    continue;
+                                }
+                                let request = build_object_heal_request(
```
overtrue, line 1995: A historical null/unversioned candidate can also land in truncated_candidates after 1,024 newer UUID versions. This unconditional builder turns it into a destructive version_id=None request, which the heal path interprets as latest rather than the discovered null version. Please preserve the unversioned non-destructive branch here and add this overflow ordering regression.

crates/filemeta/src/metacache.rs
```diff
@@ -370,6 +420,185 @@ impl MetaCacheEntries {
...
+                    // bounded object-level continuation so the scanner cannot
+                    // silently lose every version of a busy object.
+                    discovery.truncated = true;
+                    if discovery.truncated_objects.len() < MAX_META_CACHE_HEAL_TRUNCATED_OBJECTS
+                        && !discovery.truncated_objects.iter().any(|object| object == &candidate.object)
+                    {
+                        discovery.truncated_objects.push(candidate.object.clone());
+                    }
+                    if discovery.truncated_objects.iter().any(|object| object == &candidate.object) {
+                        discovery.truncated_candidates.push(candidate);
```
overtrue, line 588: truncated_candidates has no candidate or byte cap and is outside the cross-replica seen map. A valid 10,000-version object already adds 8,976 overflow entries, and multiple eligible objects/disks multiply both the vector and scanner fanout. Please enforce a hard global bound or resumable cursor and test the maximum supported version history.

crates/scanner/src/scanner_folder.rs
```diff
@@ -1904,65 +1906,112 @@ impl FolderScanner {
...
-                                self.send_required_scanner_heal_request(
-                                    PendingScannerHealKind::Object,
-                                    bucket.clone(),
-                                    Some(entry.name.clone()),
-                                    version_id.clone(),
+                                if !seen_heal_candidates.insert(identity) {
+                                    continue;
+                                }
+                                let request = if candidate.is_unversioned() {
+                                    build_non_destructive_object_heal_request(
```
overtrue, line 1942: The promised non-destructive fallback is not enforced end to end. This sets remove_corrupted=false, but ECStore’s unrecoverable file-heal branches call delete_if_dangling without checking HealOpts.remove, so an old single-survivor null version can still be deleted. Please gate both dangling-delete branches on the flag and cover the storage path with a deletion spy.

overtrue (changes requested):
Three reachable heal-discovery blockers remain.

PR #6385, opened 2026-08-22
test(kms): stabilize Vault failover validation

overtrue:
## Related Issues

rustfs/backlog#1897

## Summary of Changes

- align the post-failover progress timeout with the KMS circuit's bounded 30-second recovery interval
- allow only backend and attempt-timeout errors during the injected leader outage while still failing unexpected KMS errors or wrong plaintext
- keep KV2 and Transit worker failures separate and preserve the underlying error in assertion output
- make manual Nightly GNU dispatches test the selected ref instead of always checking out `main`

## Verification

- `bash scripts/test/vault_ha_kms_live.sh` passed 10 consecutive real three-voter Vault Raft runs before the final oracle adjustment, then passed again with the final diff
- `actionlint .github/workflows/nightly-gnu.yml`
- `cargo fmt --all --check`
- `git diff --check`
- exact-head CI run https://github.com/rustfs/rustfs/actions/runs/32572514679 passed the HA lane in 31.58 seconds after killing the active leader, exercising the circuit recovery window that previously failed

## Impact

No production behavior changes. The ignored live Vault HA test now validates the existing fail-fast circuit contract: retryable backend failures are allowed only during the injected outage, recovery must complete within a fixed bound, unexpected errors and wrong plaintext still fail immediately, and the circuit and in-flight gauges must return to zero.

## Additional Notes

The first manual run exposed that all three Nightly GNU jobs forced `ref: main`. After fixing checkout, the exact-head run https://github.com/rustfs/rustfs/actions/runs/32572216393 showed the real failure: five connection failures opened the intentional 30-second KMS circuit, but the test treated the first failed decrypt operation as a permanent HA failure. The final exact-head run validates recovery instead of contradicting that circuit contract.

---

### DIFF
--- a/crates/kms/tests/vault_ha_failover_live.rs
+++ b/crates/kms/tests/vault_ha_failover_live.rs
@@ -16,13 +16,13 @@
 //!
 //! `scripts/test/vault_ha_kms_live.sh` owns the official Vault containers and
 //! kills the active node while this test continuously decrypts through a
-//! surviving standby. KV2 and Transit requests must remain successful, use a
-//! bounded number of attempts, and leave the circuit and in-flight gauges at
-//! zero after a new leader is elected.
+//! surviving standby. KV2 and Transit must recover after the bounded circuit
+//! interval, use a bounded number of attempts, and leave the circuit and
+//! in-flight gauges at zero after a new leader is elected.
 
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


crates/kms/tests/vault_ha_failover_live.rs
```diff
@@ -326,18 +371,39 @@ async fn exercise_failover(snapshotter: &Snapshotter) {
...
-    wait_for_count(&transit_completed, transit_after_election, "post-failover Transit decrypts").await;
+    wait_for_count(
+        &kv2_completed,
+        &kv2_failure,
+        kv2_after_election,
+        "post-failover KV2 decrypts",
+        POST_FAILOVER_PROGRESS_TIMEOUT,
+    )
+    .await;
+    wait_for_count(
```
overtrue, line 382: These two 35-second waits run serially, so KV2 at 34 seconds and Transit at 68 seconds still pass even though the PR promises one fixed recovery bound. A paused-time exact-head test reproduces this: both waits pass, but total elapsed exceeds 35 seconds. Please wait for both backends under one shared deadline and monitor both failure slots concurrently.

PR #6389, opened 2026-08-22
ci(perf): fix nightly regression baseline

overtrue:
## Related Issues

Part of rustfs/backlog#1897, rustfs/backlog#1603, and rustfs/backlog#1152. Follows rustfs/backlog#935.

## Summary of Changes

- Re-enable the Performance A/B workflow after correcting its nightly baseline semantics.
- Compare scheduled candidates with the last successful scheduled head instead of comparing main with itself.
- Keep a failed regression anchored to the last known-good head; with no history, the first successful run seeds itself without a second release build.
- Cache each successful nightly candidate for the next comparison and remove the release build previously started on every main push.
- Fail closed when the stored baseline is not on the candidate history, and guard the contract against workflow drift.
- Give the cold-cache path enough time to publish its artifact, summary, and next baseline after both release builds and the A/B run.

## Verification

- `actionlint .github/workflows/performance-ab.yml`
- `shellcheck scripts/security/check_performance_ab_workflow.sh`
- `bash scripts/security/check_performance_ab_workflow.sh`
- `make script-tests`
- `cargo fmt --all --check`
- `git diff --check`
- Exact-head manual run: https://github.com/rustfs/rustfs/actions/runs/32574518482 (A/B gate passed; the 120-minute job cap cancelled evidence upload and produced the timeout fix in this PR)

## Impact

The nightly performance job becomes active again and can detect aggregate regressions since its last successful run. A normal nightly restores one known-good binary and builds one candidate; main pushes no longer start separate release builds. Manual dispatch continues to compare the selected ref with current main.

## Additional Notes

There is no prior successful scheduled run. The first complete scheduled execution measures the current binary against itself and seeds that head only if the complete rig succeeds; every later successful run becomes the next baseline. Failed or cancelled runs do not move it.

---

### DIFF
--- a/.github/workflows/performance-ab.yml
+++ b/.github/workflows/performance-ab.yml
@@ -53,8 +53,8 @@ jobs:
     runs-on: sm-standard-2
     # A normal nightly restores the last successful binary and builds only the
     # candidate; daily access keeps that cache warm. A cache miss may build both
-    # but remains capped and reports a timeout through the scheduled alert.
-    timeout-minutes: 120
+    # and needs room for the A/B run plus artifact and cache publication.
+    timeout-minutes: 180
     steps:
       - name: Checkout repository
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
--- a/.github/workflows/performance-ab.yml
+++ b/.github/workflows/performance-ab.yml
@@ -22,13 +22,6 @@
 # correctness cost (e.g. the #4221 fsync durability fix) is recorded, not
 # blocked (rustfs/backlog#935 correction 1).
 
-# DISABLED. This workflow is switched off in the repository's Actions settings
-# (state: disabled_manually) and does not run on any trigger, including its cron
-# and workflow_dispatch. That state lives in GitHub's UI and is invisible when
-# reading this file, which has already misled at least one audit — hence this
-# banner. Re-enabling is a UI action; anyone doing so should first check that the
-# workflow still matches the current CI layout. See rustfs/backlog#1603.
-#
 name: Performance A/B
 
 on:
@@ -52,6 +45,7 @@ on:
     branches: [main]
 
 permissions:
+  actions: read
   contents: read
 
 env:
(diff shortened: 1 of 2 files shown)


.github/workflows/performance-ab.yml
```diff
@@ -46,96 +39,28 @@ on:
...
-    # With perf-3's cached baseline binary the common (cache-hit) nightly is
-    # measurement-only and finishes well under 50min. This ceiling stays
-    # generous only to absorb the same-commit cache-miss self-heal (~65min
-    # single build with the post-#4806 LTO profile + measurement). A timeout
-    # surfaces via the alert-on-failure job (it fires on cancelled/timed-out,
-    # not just failure). perf-6 recalibrates the budget once the noise study
-    # lands.
+    # A normal nightly restores the last successful binary and builds only the
+    # candidate; daily access keeps that cache warm. A cache miss may build both
+    # but remains capped and reports a timeout through the scheduled alert.
```
overtrue: The exact-head cold-cache/different-candidate manual run 32574518482 completed both builds and the A/B gate, then hit the two-hour job limit during artifact upload. Summary, candidate cache, and Enforce gate never ran, and no artifact was saved. This supported path needs enough end-to-end budget, with result and enforcement recorded before long post-processing.

.github/workflows/performance-ab.yml
```diff
@@ -46,96 +39,28 @@ on:
         required: false
         default: false
         type: boolean
-  push:
-    # Every main commit pre-builds and caches its release binary (perf-3) so the
-    # nightly A/B restores a ready baseline instead of paying the double build.
-    branches: [main]
-
 permissions:
+  actions: read
```
overtrue, line 43: Reactivating this workflow still lets workflow_dispatch execute the selected branch on the self-hosted sm-standard-2 runner. Exact-head run 32574518482 checked out and ran this branch, so a write collaborator can execute branch-controlled scripts or build.rs there. Please restrict non-main dispatches or use an isolated runner for them.

.github/workflows/performance-ab.yml
```diff
@@ -46,96 +39,28 @@ on:
...
     runs-on: sm-standard-2
-    # With perf-3's cached baseline binary the common (cache-hit) nightly is
-    # measurement-only and finishes well under 50min. This ceiling stays
-    # generous only to absorb the same-commit cache-miss self-heal (~65min
-    # single build with the post-#4806 LTO profile + measurement). A timeout
-    # surfaces via the alert-on-failure job (it fires on cancelled/timed-out,
-    # not just failure). perf-6 recalibrates the budget once the noise study
-    # lands.
+    # A normal nightly restores the last successful binary and builds only the
+    # candidate; daily access keeps that cache warm. A cache miss may build both
```
overtrue, line 55: Exact-head manual run 32574518482 missed the baseline cache, spent about 41 minutes on two release builds and 72 minutes in A/B, then hit the 120-minute cap during artifact upload before summary, candidate caching, or gate enforcement. Please make the uncached dispatch path fit the job budget or avoid the double build.

.github/workflows/performance-ab.yml
overtrue: This runner class is ephemeral: the repository checks that each ARC pod serves one job, and PR workflows already execute branch code on the same isolated pool. This path has no persistent runner, write token, or credential persistence, so I am not treating it as a blocker.

PR #6391, opened 2026-08-22
fix(ecstore): fence bucket heal against decommissioned pools

overtrue:
## Related Issues

- Fixes rustfs/backlog#1912.
- Part of the decommission/rebalance roadmap tracked by rustfs/backlog#1901.

## Summary of Changes

- `handle_heal_bucket` now applies the same suspended-pool fence as `handle_heal_object`: a pool-scoped bucket heal on a pool with an in-progress decommission returns `SlowDown`, and a completed decommission returns a terminal `InvalidArgument("heal", "pool", ...)` instead of fanning out.
- Local bucket-heal volume mutation (`heal_bucket_local_on_disks`) re-reads the node's live pool metadata at execution time and skips every disk belonging to a suspended pool, so stale assignments cannot recreate or delete volumes on a source pool after dispatch; fenced drives report a `skipped-decommission-suspended` drive state with an explanatory result detail.
- Unscoped heals continue across active pools and surface skipped pools instead of failing the whole operation; object-level heal fencing is unchanged.

## Testing

- New deterministic regressions: scoped bucket heal blocked for in-progress/completed decommissions, local volume fencing skips suspended pools while still healing active pools, and `remove` keeps suspended-pool volumes intact (verified each fails when the fence hunks are reverted).
- Focused: `cargo test -p rustfs-ecstore --lib heal_bucket` → 12 passed; broader suspended/decommission heal suites green after merging latest main.

### DIFF
--- a/crates/ecstore/src/cluster/rpc/peer_s3_client.rs
+++ b/crates/ecstore/src/cluster/rpc/peer_s3_client.rs
@@ -46,7 +46,7 @@ use std::sync::{
     Mutex as StdMutex,
     atomic::{AtomicBool, Ordering},
 };
-use std::{collections::HashMap, fmt::Debug, sync::Arc, time::Duration};
+use std::{collections::BTreeSet, collections::HashMap, fmt::Debug, sync::Arc, time::Duration};
 #[cfg(test)]
 use tokio::sync::Notify;
 use tokio::{net::TcpStream, sync::RwLock, time};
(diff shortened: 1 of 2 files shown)


overtrue:
Superseded by #6416 (same backlog#1912 fix, rebased with the newer heal fencing work). Closing to avoid duplicate merges.

PR #6393, opened 2026-08-22
fix(ecstore): migrate tier free versions during decommission

overtrue:
## Related Issues

Fixes rustfs/backlog#1923 and relates to rustfs/backlog#1901.

## Summary of Changes

- migrate tier free versions as exact decommission inventory entries instead of leaving them on the source pool
- preserve remote-delete identity and validate every target disk before writes and idempotent replay
- retain source state on conflicts, capacity failures, invalid metadata, lost fences, and failed write quorum
- keep cleanup-only delete-marker decisions independent from free-version records
- preserve current durable ILM, SourceChanged retry, metadata-first, MPU drain, and queue semantics
- cover migration, replay, conflicts, cleanup, final sweep, restart, and lifecycle-consumed races

## Verification

- `rustfmt --edition 2024 --check` on all changed Rust files
- `git diff --check`
- layer dependency, architecture migration, unsafe allowance, logging, doc path, and no-planning-doc guards
- two independent exact-head adversarial reviews: PASS on `6064ea9fc4cff874460ccccfe676a714821b513d`
- no local Cargo build was started because disk headroom is constrained; fresh CI is the execution gate

## Impact

Tier free versions now retain their remote-delete obligation through decommission and cannot overwrite a conflicting target version. Source cleanup remains fail-closed until target state and the exact source inventory are valid. No public S3 API or persisted wire-format version changes are introduced.

### DIFF
--- a/crates/ecstore/src/core/pools.rs
+++ b/crates/ecstore/src/core/pools.rs
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 use crate::bucket::replication::replication_state_from_filemeta;
+#[cfg(test)]
+use crate::bucket::utils::is_meta_bucketname;
 use crate::bucket::versioning_sys::BucketVersioningSys;
 use crate::bucket::{
     lifecycle::{
@@ -967,6 +969,7 @@ fn with_decommission_entry_context<E: Display>(stage: &str, bucket: &str, object
     Error::other(format!("decommission entry {stage} failed for bucket {bucket} object {object}: {err}"))
 }
 
+#[cfg(test)]
 fn load_decommission_entry_versions(entry: &MetaCacheEntry, bucket: &str, stage: &str) -> Result<FileInfoVersions> {
     entry
         .file_info_versions(bucket)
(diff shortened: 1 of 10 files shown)


crates/filemeta/src/filemeta.rs
```diff
@@ -90,6 +90,22 @@ fn legacy_data_key_for_version(version_id: Option<Uuid>) -> Option<String> {
...
+/// [`MetaObject::delete_version`] when a version whose remote transition
+/// completed is removed from xl.meta; it carries the remote tier identity for
+/// an idempotent remote delete and is never a user-visible version
+/// (`num_versions` excludes it). While the record exists it is consumed by the
+/// lifecycle free-version recovery scan and the usage scanner, which re-enqueue
+/// the pending remote delete, and by heal metadata walks. On S3 and lifecycle
+/// delete paths the same obligation is also carried by a committed tier-journal
+/// entry; deletes without such an entry (for example a removed version whose
+/// transition state decodes as unknown) rely on this record alone until the
+/// worker removes it after a successful remote delete. Decommission does not
```
overtrue: Journal-less Unknown records cannot reach a successful remote delete: validate_transition_remote_version rejects Unknown, and free_version_delete_rejects_unknown_version_state_before_backend_io asserts zero backend calls. Please document these records as retained but non-actionable, and state that decommission discards the only durable evidence instead of implying the worker eventually removes them.

crates/ecstore/src/core/pools.rs
```diff
@@ -6762,6 +6796,27 @@ mod pools_tests {
         assert!(!should_cleanup_decommission_source_entry(2, 2, 1));
     }
 
+    #[test]
+    fn decommission_free_version_accounting_reports_skipped_records() {
```
overtrue: This test only exercises the counting helper. Removing the production decommission_entry event leaves it green, so it does not protect the claimed final-sweep accounting. Please cover a real free-version decommission path and assert the recorded disposition.

crates/filemeta/src/filemeta.rs
overtrue: Updated the behavior and docs: decommission now migrates the Unknown record unchanged, and lifecycle retains it when remote identity validation cannot issue a delete. The durable evidence is no longer discarded.

crates/ecstore/src/core/pools.rs
overtrue: Removed the counting-only test. The regressions now run the real decommission entry path for migration, conflicts, source cleanup, final sweep, heal visibility, and the lifecycle-consumed race.

overtrue (commented):
Correctness and compatibility review for 76aad3772bc5: APPROVE. No blockers.

overtrue (commented):
Simplicity and test-coverage review for 76aad3772bc5: APPROVE. No blockers.

overtrue (commented):
Correctness, concurrency, durability, and compatibility review for a7052bc76fd0: APPROVE. No blockers.

overtrue (commented):
Simplicity, test coverage, and performance review for a7052bc76fd0: APPROVE. No blockers.

overtrue:
Conflict note: merging latest main now collides in pools.rs sweep/entry regions (free-version migration fences vs the newer ILM receipt + bounded-listing flow). Earlier I resolved one round of this and fixed the migration fixtures (decode-faithful free-version records + bucket creation) — those fixes are already on this branch. The remaining collision needs the same designed-composition approach as #6400 (both the free-version disposition accounting and the durable-ILM receipt ordering must survive). Deferring to branch owner given active pushes; ping me to co-resolve.

PR #6394, opened 2026-08-22
fix(scanner): unify unknown metadata size accounting

cxymds:
## Related Issues

Fixes rustfs/backlog#1935.

## Summary of Changes

Scanner metadata-size handling now resolves each object version once into an explicit Known, Unknown, or Corrupt result. Unknown compressed-size sentinels continue heal, replication, version/delete-marker, and physical-tier accounting without contaminating logical totals; corrupt metadata fails closed for destructive lifecycle actions. Lifecycle size predicates are evaluated only for the selected rule, while time/version-only actions retain their existing semantics. Unknown and corrupt records are written to a bounded, deduplicated reconciliation ledger with canonical bounded identities, retry attempts, TTL/cap pruning, restart replay, and scope-based clearing. Legacy cache snapshots keep their original wire shape when the ledger is empty.

## Verification

- `cargo test -p rustfs-scanner --lib --no-default-features -- --test-threads=1` (502 passed)
- Focused malformed-size, size-resolution, reconciliation-replay, bounded-identity, accounting, and queued-unknown-version tests
- `cargo clippy -p rustfs-scanner -p rustfs-data-usage --all-targets --no-default-features -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`
- Architecture, logging, unsafe-code, migration-rule, and doc-path guards
- `make pre-pr` (14,150 tests passed, 137 skipped; all doctests passed)
- Adversarial validation: correctness, simplicity, security, concurrency/durability, compatibility, performance, and test-coverage roles found no actionable issue after the queued unknown-version count fix.

## Impact

Malformed or legacy unknown size metadata no longer causes ILM and non-ILM accounting to diverge. Size-dependent lifecycle actions defer safely until reconciliation; time/version-only actions remain available where metadata is valid. Reconciliation state is bounded and persisted in the scanner usage cache, with no new public API or required configuration. Physical tier statistics remain distinct from logical usage.

## Additional Notes

Analysis used Sol, implementation used Luna, and final adversarial review used 5.4 as requested. No new tracing sites were added. The ledger is intentionally bounded and records unresolved work as partial scanner state until a later scan resolves it.

---


crates/scanner/src/scanner_folder/item_actions.rs
```diff
@@ -32,6 +33,259 @@ pub(super) enum GetSizeFailureAction {
...
+        .find(|rule| rule.id.as_deref().unwrap_or_default() == rule_id)
+        .and_then(|rule| rule.filter.as_ref())
+        .is_some_and(filter_has_size)
+}
+
+fn lifecycle_event_allowed(resolution: &SizeResolution, event: &Event, lifecycle: &BucketLifecycleConfiguration) -> bool {
+    match resolution {
+        // Corrupt metadata cannot safely authorize a destructive action, even
+        // when the evaluator happened to produce a time-only event.
+        SizeResolution::Corrupt { .. } => false,
```
overtrue: This blocks time/version-only expiration whenever only the size metadata is corrupt. rustfs/backlog#1935 explicitly requires those actions to keep their existing semantics; the new test at line 1574 currently pins the opposite result. Please defer only actions that require size and cover this with the real evaluator plus an expiry spy.

crates/scanner/src/scanner_folder/item_actions.rs
```diff
@@ -32,6 +33,259 @@ pub(super) enum GetSizeFailureAction {
...
+            || filter.object_size_less_than.is_some()
+            || filter
+                .and
+                .as_ref()
+                .is_some_and(|and| and.object_size_greater_than.is_some() || and.object_size_less_than.is_some())
+    };
+    lifecycle
+        .rules
+        .iter()
+        .find(|rule| rule.id.as_deref().unwrap_or_default() == rule_id)
```
overtrue: An empty event rule ID is not a stable rule identity: intrinsic restore-expiry events use it, while persisted lifecycle input can also contain an id-less size-filter rule. That makes an unknown-size restored object skip cleanup because this lookup binds it to the unrelated rule. Please avoid that match and add a production-path restore-expiry regression test.

overtrue (changes requested):
Two reachable lifecycle regressions need fixing.

crates/scanner/src/scanner_folder/item_actions.rs
cxymds: Fixed in 2086ade96. Corrupt or unknown logical size now defers only lifecycle actions whose selected rule has a size predicate; time/version-only expiration, including restore-expiry events, keeps its existing behavior. The focused scanner regression test now covers both branches.

crates/scanner/src/scanner_folder/item_actions.rs
cxymds: Fixed in 2086ade96. Empty intrinsic event rule IDs no longer match persisted id-less lifecycle rules, so an id-less size-filter rule cannot accidentally suppress restore-expiry cleanup. The regression test covers this exact lookup case.

crates/scanner/src/scanner_folder/item_actions.rs
```diff
@@ -32,6 +33,261 @@ pub(super) enum GetSizeFailureAction {
...
+
+    SizeResolution::Known { logical, physical }
+}
+
+fn resolve_sizes(object_infos: &[ObjectInfo]) -> Vec<SizeResolution> {
+    object_infos.iter().map(resolve_size).collect()
+}
+
+fn lifecycle_rule_has_size_filter(lifecycle: &BucketLifecycleConfiguration, rule_id: &str) -> bool {
+    if rule_id.is_empty() {
```
overtrue: An explicit `<ID/>` survives parsing and validation as an empty rule ID. A size-filter rule can then be selected using the physical object size, but this early return treats its Unknown/Corrupt action as size-independent and allows deletion or transition. Please normalize/reject empty IDs or carry unambiguous event provenance, with a pinned-XML production-path test.

crates/scanner/src/scanner_folder/item_actions.rs
```diff
@@ -319,34 +575,45 @@ impl ScannerItem {
...
+                    // A valid compressed legacy sentinel has no logical size,
+                    // but heal and replication still need to run. The
+                    // physical size is only an input to those operations; it
+                    // is not folded into the logical total below.
+                    SizeResolution::Unknown { physical, .. } => {
+                        self.heal_actions(oi, *physical, size_summary).await;
+                        size_summary.actions_accounting_unknown(oi);
                         continue;
                     }
+                    SizeResolution::Corrupt { .. } => continue,
```
overtrue: This keeps the original accounting split for Corrupt metadata. With no lifecycle the version is skipped here, but adding an unrelated future rule produces NoneAction and records the same version through actions_accounting_unknown. Please make both production branches use the same accounting semantics and test apply_actions with and without a nonmatching lifecycle.

crates/scanner/src/scanner_folder.rs
```diff
@@ -748,6 +800,55 @@ impl FolderScanner {
...
+        // Keep an unresolved identity in place while refreshing its object
+        // scope. This lets repeated observations increment `attempts`; only
+        // debts absent from the current pass are considered resolved.
+        let current_keys = summary
+            .size_reconciliation
+            .iter()
+            .map(|entry| entry.key.clone())
+            .collect::<HashSet<_>>();
+        for info in [&mut self.new_cache.info, &mut self.update_cache.info] {
+            prune_size_reconciliation(info, now);
```
overtrue: This runs two full ledger prunes plus scope retention for both cache copies on every scanned object. A live 10,000-entry ledger means about 100,000 entry visits per otherwise healthy object, so a large bucket becomes O(objects × ledger) despite the storage cap. Please batch pruning/updates per scan boundary and add a work-bound regression.

overtrue (changes requested):
Three exact-head regressions still need fixing.

cxymds:
Investigated the new `Test and Lint (sftp)` failure. The log showed two baseline Clippy errors in the GitHub merge ref, not a lifecycle regression: an unnecessary `mut` in `crates/ecstore/src/store/object.rs` and an `unnecessary_sort_by` in `crates/ecstore/src/store/rebalance/support.rs`. Synced latest `origin/main`, fixed both in 854b6dfb0, and verified with `cargo fmt --all --check`, `cargo check -p rustfs-ecstore --lib`, and `cargo clippy -p rustfs-ecstore --lib -- -D warnings`.

crates/scanner/src/scanner_folder/item_actions.rs
cxymds: Fixed in 6dbff1c03. Lifecycle validation now rejects parsed `Some("")` rule IDs. For legacy in-memory/persisted rules that still have no ID, an empty event ID is treated as size-dependent when the matching id-less rule has a size filter, so intrinsic restore expiry fails closed instead of authorizing deletion or transition. Added `core::tests::validate_rejects_empty_rule_id` and scanner coverage for the id-less size-filter lookup.

crates/scanner/src/scanner_folder/item_actions.rs
cxymds: Fixed in 6dbff1c03. The no-lifecycle branch now records `actions_accounting_unknown` for corrupt metadata just like the lifecycle/NoneAction branch. This keeps physical/tier accounting and unknown logical totals consistent whether lifecycle configuration is absent or present; `malformed_size_has_same_ilm_accounting` passes.

crates/scanner/src/scanner_folder.rs
cxymds: Fixed in 6dbff1c03. Per-object updates remain incremental, but scope retention and both full ledger prunes are now accumulated and executed once at the scan-folder boundary. The batch uses sets for current identities and scopes, so it performs one bounded pass per cache rather than scanning the ledger for every object.

PR #6395, opened 2026-08-22
feat(ecstore): coalesce GET ReadVersion RPCs

houseme:
## Related Issues
Related to rustfs/backlog#1879.

## Summary of Changes
Add a default-off GET metadata ReadVersion coalescing path that batches per-remote-disk ReadVersion calls after the service is ready, while keeping startup, ready-before, internal metadata, local disk, and non-GET read_all_fileinfo paths on the existing unary flow.

Add the BatchReadVersion client/server foundation with msgpack/json compatibility payloads, mixed-version auto fallback on Unimplemented, independent grpc_batch_read_version metrics, and server-side bounded per-item read parallelism defaulting to the conservative p=4 candidate from the issue evidence.

Harden coalesced response mapping so missing, duplicate, out-of-order, per-item error, and identity-mismatched responses fail closed instead of being silently assigned to the wrong request slot.

## Verification
- `cargo fmt --all --check` passed on `531ebf0a2`.
- `cargo test -p rustfs-ecstore batch_read_version` passed on `531ebf0a2`.
- `cargo test -p rustfs-ecstore metadata_read_version_coalescer` passed before the final rebase; the covered code was unchanged by the final rebase.
- `cargo test -p rustfs-ecstore batch_metadata_rpc_mode` passed before the final rebase; the covered code was unchanged by the final rebase.
- `cargo test -p rustfs-io-metrics internode_metrics` passed before the final rebase; the covered code was unchanged by the final rebase.
- `make pre-pr` passed on the immediately previous base after running outside the sandbox so localhost listener tests could bind: clippy passed, 14153 nextest tests passed, and doctests passed.
- `PATH=/Users/zhi/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin:$PATH make pre-pr` on `531ebf0a2` passed guard/script/clippy phases and ran 9532/14170 nextest tests before failing in two rustfs-iam OIDC localhost issuer fixture tests unrelated to this RPC/coalescing diff: `oidc_discovery_accepts_extra_root_ca_for_https_provider` and `oidc_discovery_refreshes_extra_root_ca_when_generation_changes`.

## Impact
Default behavior is unchanged unless `RUSTFS_GET_METADATA_READ_VERSION_COALESCE` is set to `auto` or `on`. In `auto`, old peers that do not implement BatchReadVersion fall back to unary ReadVersion. The server-side batch item parallelism defaults to 4 and can be overridden with `RUSTFS_BATCH_READ_VERSION_SERVER_PARALLELISM`, clamped to the batch item limit.

## Additional Notes
This intentionally leaves h2 window, connection pooling, and payload gate settings untouched so attribution stays focused on reducing per-object internode ReadVersion RPC count. The backlog issue contains the testing1 repeat A/B evidence for the conservative p=4 candidate.

Adversarial review summary: correctness found and fixed response identity/duplicate-index fail-closed handling; simplicity removed temporary lifecycle attribution; security checked RPC response identity and decode failures; concurrency/durability checked coalescer lock/timeout/cancellation boundaries; compatibility checked default-off and Unimplemented fallback; performance checked bounded p=4 and local/non-GET bypass; test coverage mapped behavior claims to focused tests and the pre-PR gate evidence above.

---

### DIFF
--- a/crates/ecstore/src/set_disk/core/io_primitives.rs
+++ b/crates/ecstore/src/set_disk/core/io_primitives.rs
@@ -317,11 +317,18 @@ fn map_batch_read_version_responses(
             *slot = if response.success {
                 Ok(response.file_info)
             } else {
-                Err(DiskError::other(response.error))
+                Err(batch_read_version_response_error(response.error_code, response.error))
             };
         }
     }
     results
+}
+
+fn batch_read_version_response_error(error_code: u32, error: String) -> DiskError {
+    match DiskError::from_u32(error_code) {
+        Some(DiskError::Io(_)) | None => DiskError::other(error),
+        Some(error) => error,
+    }
 }
 
 pub(in crate::set_disk) fn bounded_metadata_fanout_order(
@@ -6858,6 +6865,7 @@ mod tests {
                 success: false,
                 file_info: FileInfo::default(),
                 error: "disk read failed".to_string(),
+                error_code: 0,
             },
             BatchReadVersionResp {
                 index: 0,
@@ -6866,6 +6874,7 @@ mod tests {
                 success: true,
                 file_info: ok_file_info,
                 error: String::new(),
+                error_code: 0,
             },
         ];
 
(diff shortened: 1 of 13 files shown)


crates/ecstore/src/set_disk/core/io_primitives.rs
```diff
@@ -94,6 +95,235 @@ fn metadata_distribution_key(bucket: &str, object: &str) -> String {
...
+            continue;
+        }
+        seen[response.index] = true;
+        if response.path != expected.path || response.version_id != expected.version_id {
+            *slot = Err(DiskError::other("coalesced read_version response identity mismatch"));
+        } else {
+            *slot = if response.success {
+                Ok(response.file_info)
+            } else {
+                Err(DiskError::other(response.error))
```
overtrue: Please preserve the typed per-item disk error here. On the enabled, service-ready distributed GET path, the real batch producer turns FileNotFound and FileVersionNotFound into strings, and this rebuilds them as Io(Other). A focused producer-to-mapper test then loses the not-found classification, so remote-disk quorum can return an internal/quorum error instead of NoSuchKey or NoSuchVersion.

overtrue (changes requested):

houseme:
Follow-up for https://github.com/rustfs/rustfs/pull/6395#discussion_r3836507393: fixed in c9e80947b. BatchReadVersionResp now carries a serde-defaulted error_code next to the compatibility error string; the server producer sets err.to_u32() per failed item and the mapper restores non-Io DiskError variants from that code before falling back to DiskError::other for legacy/unknown codes. Added focused coverage for preserving FileNotFound/FileVersionNotFound in the mapper and for JSON/msgpack response payload compatibility. Verification: cargo fmt --all --check; git diff --check; cargo test -p rustfs-ecstore batch_read_version; cargo test -p rustfs batch_read_version_response.

PR #6396, opened 2026-08-22
fix(scanner): fence unknown tier accounting

cxymds:
## Related Issues
Fixes #1936

## Summary of Changes
Freeze scanner tier accounting against one registry snapshot per cycle, fold retired or unknown tiers into a fixed `UNKNOWN_TIER` bucket, and bound diagnostics by count, bytes, TTL, hashing, and redaction. Preserve the no-config empty-map shape for STANDARD/RRS, keep logical and physical accounting independently conserved, and fail closed for lifecycle actions that require a recognized tier. Add generation fencing and authenticated capability negotiation so cached and remote results from different registries cannot mix.

## Verification
- `cargo fmt --all --check` ✅
- `git diff --check` ✅
- `cargo test -p rustfs-data-usage --lib -- --nocapture` (52 passed) ✅
- `cargo test -p rustfs-scanner --lib scanner_io::publish_gate_tests -- --nocapture --test-threads=1` (44 passed) ✅
- `cargo test -p rustfs-scanner --lib -- --nocapture --test-threads=1` (528 passed) ✅
- Capability and generation-focused ecstore/scanner tests (8 passed) ✅
- `make pre-pr` (14,205 passed, 137 skipped, all doctests passed) ✅

## Impact
Unknown tier keys no longer disappear from scanner accounting or create unbounded map and metric cardinality. Persisted cache data remains map-encoded and backward-compatible; unknown entries are accounted even when bounded diagnostics overflow. Remote peers without authenticated generation support use the existing coordinator path.

## Additional Notes
Adversarial validation covered correctness, simplicity, security, concurrency and durability, compatibility, performance, and test-coverage roles; no findings remained. A real cross-version HTTP integration test was not added because the negotiated legacy-proof branches are covered by the authenticated transport helpers and focused unit tests.


cxymds:
Fixed the failed Test and Lint (sftp) check and pushed b0fe2b277. The failure was caused by two current-main Rust lint errors: removed an unnecessary mut in crates/ecstore/src/store/object.rs and replaced unnecessary sort_by with sort_by_key in crates/ecstore/src/store/rebalance/support.rs. Also merged current main without conflicts. Verification passed: cargo fmt --all --check; cargo check -p rustfs-scanner --lib; cargo check -p rustfs-ecstore --lib.

overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #6399, opened 2026-08-22
fix(admin): four site-replication interop correctness fixes (B5-rc T2)

reatang:
## Related Issues

Fixes rustfs/backlog#1949 (B5-rc T2; spec rustfs/backlog#1946; parent rustfs/backlog#1675).

## Summary of Changes

Four small site-replication interop correctness fixes in the admin layer, one commit each:

- **A3 - send `versioningEnabled=true` on outbound make-with-versioning bucket ops.** The outbound query only carried `operation`/`createdAt`/`lockEnabled`. MinIO's own create-bucket hook sends `versioningEnabled=true` on this op, so both outbound builders (bootstrap plan and create-bucket broadcast hook) now route through one shared `make_with_versioning_bucket_op_path` builder that always appends the flag. This aligns the outbound wire contract with MinIO; RustFS's inbound handler force-enables versioning either way, so RustFS-to-RustFS behavior is unchanged.
- **A5 - propagate purge-deleted-bucket errors.** The purge-deleted-bucket peer op no longer swallows storage errors and returns 200. Bucket-not-found is tolerated (idempotent purge); every other error is propagated to the caller with its real status.
- **B8 - derive the fallback site deployment ID with UUIDv5.** `deployment_id_for_endpoint` used `DefaultHasher`, whose algorithm is not stable across Rust releases. The fallback now derives `Uuid::new_v5(NAMESPACE_URL, canonical endpoint)`, so the fallback ID format becomes a 36-character UUID, and a structured `warn!` is logged when a peer metainfo response arrives with an empty `deploymentID`. Already persisted fallback IDs are non-empty and are never re-derived, so existing state is unaffected.
- **C5 - stream the site-replication devnull body without the 1 MB cap.** The `/site-replication/devnull` endpoint used to buffer the body with a 1 MiB limit and reject larger netperf payloads with 400. It now drains the body as a stream with no upper bound; the endpoint still sits behind admin authentication.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs --lib site_replication` - 244 passed, 0 failed
- `make pre-commit` - passed

```
Verified against real MinIO RELEASE.2025-09-07 (single-drive, no docker) with mc RELEASE.2025-08-13:
- C5: signed POST to /minio/admin/v3/site-replication/devnull with 2 MiB and 8 MiB bodies returns 204 on this branch;
  main (adb90fc6e) returns 400 "length limit exceeded" for the same requests (0.5 MiB passes on both).
- A3: with RustFS->MinIO site replication paired, buckets created before pairing (bootstrap) and after pairing
  (broadcast) are versioning-enabled on MinIO; `mc admin trace` shows the outbound bucket-ops query carrying
  `versioningEnabled=true`. Note: this MinIO release force-enables versioning in PeerBucketMakeWithVersioningHandler
  regardless of the flag, so the change aligns the wire contract rather than fixing observable behavior there.
- Object replication smoke RustFS->MinIO and MinIO->RustFS: COMPLETED.
- Pairing itself required four out-of-scope local shims (null Roles map, ServerInfo deploymentID fallback,
  OpenID.Region only when enabled, join peers keyed by real deployment ID); MinIO-initiated site netperf still
  gets 403 from RustFS devnull because MinIO sends it unsigned. Tracked separately.
```

## Impact

- Outbound admin peer bucket ops carry one additional query parameter (`versioningEnabled=true`). Older RustFS peers ignore it; MinIO peers already expect it.
- Purge-deleted-bucket now returns the real error status instead of 200 when storage fails (not-found still succeeds).
- Devnull requests larger than 1 MiB now return 204 instead of 400; the endpoint remains admin-authenticated.
- Fallback deployment IDs (only generated when a peer reports an empty `deploymentID`) change format from a hashed string to a 36-character UUID; persisted IDs are not re-derived.
- The `rustfs` crate enables the `uuid/v5` feature, which pulls in `sha1_smol` (Cargo.lock +1 entry).

## Additional Notes

Out-of-scope findings from the real-MinIO run, recorded for separate triage (not fixed in this PR):

1. MinIO idp-settings responses carry `Roles: null`, which RustFS fails to deserialize during pairing.
2. Before MinIO enables site replication its ServerInfo `deploymentID` is empty, and the RustFS add preflight hard-rejects that case upstream of the B8 fallback, so the fallback is not reachable on that real path.
3. RustFS reports `OpenID.Region` even when OpenID is disabled, which triggers an IDP mismatch on the MinIO side.
4. Join peers are keyed by the fallback deployment ID rather than the real one.
5. MinIO-initiated site netperf hits RustFS devnull with unsigned requests and gets 403.

These are documented and awaiting an owner decision on how to handle native RustFS<->MinIO pairing.

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -55,6 +55,7 @@ use crate::storage::storage_api::{
 use base64::Engine;
 use base64::engine::general_purpose::STANDARD as BASE64_STANDARD;
 use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use futures::StreamExt;
 use hmac::{Hmac, Mac};
 use http::header::{CONTENT_TYPE, HOST};
 use http::{HeaderMap, HeaderValue, Uri};
@@ -10219,13 +10220,25 @@ impl Operation for SiteReplicationStatusHandler {
     }
 }
 
+/// `POST /v3/site-replication/devnull` — peer link-check upload drain.
+/// MinIO streams multi-megabyte probe bodies here during site netperf link
+/// checks and expects an unbounded discard (its handler copies to io.Discard);
+/// buffering through the 1MB admin body cap turned any larger probe into a
+/// 400 and a false link failure. Stream and discard instead — no size cap.
+async fn drain_site_replication_devnull(mut input: Body) -> S3Result<()> {
+    while let Some(chunk) = input.next().await {
+        chunk.map_err(|e| s3_error!(InvalidRequest, "failed to read devnull stream: {}", e))?;
+    }
+    Ok(())
+}
+
 pub struct SiteReplicationDevNullHandler {}
 
 #[async_trait::async_trait]
 impl Operation for SiteReplicationDevNullHandler {
     async fn call(&self, req: S3Request<Body>, _params: Params<'_, '_>) -> S3Result<S3Response<(StatusCode, Body)>> {
         validate_site_replication_admin_request(&req, AdminAction::SiteReplicationOperationAction).await?;
-        let _ = read_plain_admin_body(req.input).await?;
+        drain_site_replication_devnull(req.input).await?;
         Ok(empty_response(StatusCode::NO_CONTENT))
     }
 }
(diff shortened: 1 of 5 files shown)


overtrue (approved):
LGTM

rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -2096,6 +2097,18 @@ async fn remote_add_preflight_info(site: &PeerSite) -> S3Result<SiteReplicationA
             format!("invalid site replication metainfo from `{}`: {e}", site.endpoint),
         )
     })?;
+    if info.deployment_id.is_empty() {
```
overtrue, line 2100: `remote_add_preflight_info` leaves this ID empty, but the add handler immediately calls `validate_add_preflight_topology`, which rejects it before `normalize_peer_info` can apply the UUIDv5 fallback. MinIO metainfo can return an empty ID for a new site, so this still blocks the B8 join path. Please derive the endpoint fallback before topology validation and cover that path.

overtrue (changes requested):

overtrue (approved):
LGTM

rustfs/src/admin/handlers/site_replication.rs
overtrue: The linked T2 scope requires stable UUIDv5 fallback derivation and warning coverage; native MinIO add-preflight reachability is explicitly left out of scope. I’m dropping this thread.

PR #6406, opened 2026-08-22
fix(ecstore): resolve lint regressions

overtrue:
## Related Issues

N/A

## Summary of Changes

- Remove an unused mutable binding from batch-delete accounting.
- Use `sort_by_key` with `Reverse` for the existing descending candidate order.

These two mechanical fixes restore the workspace Clippy gate without changing behavior.

## Verification

- `cargo fmt --all --check`

## Impact

No runtime, API, storage-format, or compatibility impact.

## Additional Notes

The branch was refreshed onto the current `main` head before verification.


overtrue:
The lint fixes are now on main. After syncing the branch, only whitespace changes remain, so this PR is superseded.

PR #6413, opened 2026-08-22
fix(object-lock): unblock authorized replication writes on locked versions and tolerate cleared lock metadata

reatang:
> Rebased onto `main` after #6379 merged (no conflicts); the PR now carries only the four object-lock commits. Review round 2 (bucket-default retention) is addressed in the last commit.

## Related Issues

rustfs/backlog#1953 — adjacent defects registered in https://github.com/rustfs/backlog/issues/1953#issuecomment-5379241883 (owner decision: fix under the same ticket, stacked on the LWW PR). Parent rustfs/backlog#1675.

## Summary of Changes

Three pre-existing defects found while verifying the receiver-side LWW fix (#6379):

1. **Replication WORM gate (fixed, `d1e543190` + `3f3ae22bd`)** — the explicit-version PUT gate at the set layer and the app-layer pre-check (`validate_existing_object_lock_for_write`) rejected every inbound authorized replication write whenever the destination version was under a legal hold, COMPLIANCE or GOVERNANCE retention. A source-side hold release or retention change could never land; replication looped through MRF forever. Both gates now pass an authorized replication write (`ReplicateObjectAction` → `opts.replication_request`) **only when it carries the source timestamp of every category that currently locks the version** (`replication_write_may_pass_worm_gate`), so the receiver-side LWW merge judges each of them: a category locked more recently on this site is kept, otherwise the source's newer state wins. A write without that timestamp stays WORM-rejected (a metadata replace would lift the lock unjudged). The set layer evaluates the lock gate first so malformed persisted lock metadata still fails closed. Non-replication overwrites are unchanged.
2. **Cleared lock metadata read as invalid (fixed, `9828a867c`)** — a local PutObjectRetention / PutObjectLegalHold "clear" persists empty strings (MinIO on-disk shape); the read side treated them as corrupt and then blocked explicit-version PUTs and version deletes. Empty values are now read as "key absent"; non-empty malformed values still fail closed.
3. **Local tags on a bucket without replication config carry no timestamp** — **not fixed** (see Additional Notes).

## Verification

- `cargo test -p rustfs --lib -- validate_existing_object_lock` — 7 passed (new replication allow/reject cases; the allow case was red before the app-layer change).
- `cargo test -p rustfs-ecstore --lib -- replication_lww_tests objectlock_sys::tests put_object_tmp_cleanup_tests::explicit_version` — 52 passed (new: `inbound_newer_legal_hold_release_updates_locked_version`, `inbound_stale_legal_hold_release_keeps_newer_local_hold`, `inbound_without_legal_hold_timestamp_stays_rejected_on_held_version`, `replication_write_on_malformed_lock_metadata_still_fails_closed`, `non_replication_overwrite_of_locked_version_is_still_rejected`, `replication_write_passes_worm_gate_only_with_every_locking_category_timestamp`, `deletion_treats_cleared_empty_lock_metadata_as_unlocked`, `explicit_version_overwrite_and_delete_succeed_after_local_lock_clear`).
- `make pre-commit` — passed. `cargo clippy -p rustfs-ecstore --lib --tests -- -D warnings` — clean.
- Adversarial review (high risk, all seven roles): correctness — found and fixed the unjudged-category silent unlock (`3f3ae22bd`: source never held a lock, replica holds a local hold, source changes only tags → the unconditional bypass would have replaced metadata and dropped the hold); security — bypass reachable only after `ReplicateObjectAction`, timestamps parsed only on authorized requests; concurrency — reuses the existing commit-lock read, no new locks/IO, gate-before-bypass ordering tested; compatibility — MinIO senders emit the required timestamp headers, pre-P1-6 senders keep today's rejection, on-disk/wire formats unchanged; performance — helper runs only when the gate already found a lock.

## Impact

- Inbound authorized replication of hold releases / retention changes onto locked replicas now converges instead of looping through MRF.
- Compliance boundary: a lock on a replica can only be lifted by a newer, source-authored decision delivered through an authorized replication write; non-replication writes and writes without the category's source timestamp remain rejected. This is stricter than MinIO's `checkPutObjectLockAllowed`, which skips the existing-version check for any replica write.
- Principals holding `ReplicateObjectAction` can apply the source's lock decisions to replicas (same trust MinIO grants, plus the LWW timestamp constraint).
- Objects whose retention/legal hold was cleared locally are no longer wedged for later explicit-version PUTs and deletes.

## Additional Notes

- **Basis for (1)**: the "replica skips the existing-version lock check" rule is taken from memory of MinIO's `checkPutObjectLockAllowed` (`!replica` guard); there is no equivalent code in this repo to cite. In-repo design consistency supports it: `has_replication_retention_update` and the LWW merge in #6379 only make sense if the gate lets inbound lock changes through.
- **(3) not fixed, by design**: MinIO's `PutObjectTaggingHandler` also stamps the tagging timestamp only under `dsc.ReplicateAny()`; RustFS mirrors that. A bucket without replication config never participates in LWW, so no timestamp is needed.
- **Out of scope, found while fixing (2)**: `put_object_retention` does not validate that mode and retain-until-date are set as a pair; a request with only one of them persists incomplete retention that the read side rejects as "incomplete". Needs its own issue.
- **Out of scope, unchanged behavior**: the set-layer explicit-version *copy* gate still rejects replica writes; a MinIO source doing metadata-only replication via CopyObject onto a locked replica keeps today's rejection (RustFS senders use PUT). The complete-multipart path has no set-layer WORM gate; its app-layer pre-check is not under the commit lock (pre-existing TOCTOU structure; categories with inbound timestamps remain protected by the in-lock LWW).
- After (2), a cleared retention on a bucket with default retention falls into the gate's existing bucket-default branch (computed from mod_time) instead of erroring; this is the pre-existing rule for absent keys.


crates/ecstore/src/bucket/object_lock/objectlock_sys.rs
```diff
@@ -136,12 +136,41 @@ pub fn add_years(dt: OffsetDateTime, years: i32) -> OffsetDateTime {
...
+    }
+    if has_legal_hold(user_defined) && opts.replication_legalhold_timestamp.is_none() {
+        return false;
+    }
+    let ret = objectlock::get_object_retention_meta(user_defined);
+    let retention_locked = ret
+        .mode
+        .as_ref()
+        .is_some_and(|mode| is_retention_active(mode.as_str(), ret.retain_until_date.as_ref()));
+    !(retention_locked && opts.replication_retention_timestamp.is_none())
```
overtrue: This returns true when the version has no explicit retention keys, even if the bucket default currently puts it under COMPLIANCE or GOVERNANCE retention. An authorized same-version replication PUT carrying only a tagging timestamp then passes the commit-time WORM gate and overwrites that protected version without a source retention decision. Please include the authoritative bucket-default lock state in this decision and add the focused default-retention replication regression.

overtrue (changes requested):

crates/ecstore/src/bucket/object_lock/objectlock_sys.rs
reatang: Fixed in 02a2750fc — thanks, this was a real gap (same finding as the Copilot thread). `replication_write_may_pass_worm_gate` only derived the locking categories from the explicit lock keys, while the commit-time gate also locks a version through the bucket default retention (mode + days/years from the version's mod_time) when no explicit retention keys are present.

The helper now takes the same `ObjectLockConfigState` + `ObjectInfo` the gate judges, and both share one evaluation (`legal_hold_locks` / `active_retention`: explicit keys first, bucket default otherwise), so every locking category — default retention included — must carry its source timestamp before the write may pass to LWW. Malformed lock metadata or a non-authoritative bucket state is an error (fail-closed), never a pass; the gate still runs first. Both the set-layer commit gate and the app-layer pre-check (`validate_existing_object_lock_for_write`, now given the bucket lock state) apply the rule.

Regressions added: a hermetic set-layer test (bucket default COMPLIANCE/GOVERNANCE, version without explicit retention keys, tagging-only replication PUT is rejected and the version untouched; with the retention source timestamp the write passes and LWW applies), an app-layer pre-check test, and helper truth-table cases for default retention, expired default, cleared keys, delete markers and the fail-closed paths. Verified the new tests go red against the previous key-only derivation.

reatang:
Rebased onto `main` after #6379 merged — the three LWW commits are gone from this PR, only the four object-lock commits remain (84826583c → 9c0c35334), no conflicts. Note: `main` at f694a0000 currently fails `make pre-commit` on its own (s3s footprint 212 files vs baseline 211, and a clippy `too_many_arguments` in `crates/heal`); neither is touched by this PR, so those lanes may stay red until main is fixed.

overtrue (approved):

PR #6360, opened 2026-08-22
perf(ecstore): bound decommission entry workers

cxymds:
## Related Issues

- Related: #1907 and #1908.
- The broader decommission/rebalance roadmap is tracked in #1901.
- Incorporates the decommission checkpoint work from #6356; this branch is now based on the latest `main`.

## Summary of Changes

- Replace callback-bound decommission listing with a bounded listing-to-queue-to-worker pipeline so listing can continue while object entries are processed.
- Add cancellation-aware queue backpressure, retry-attempt draining, worker join/error propagation, and a shared run-level entry budget to prevent bucket and pool-index fan-out from multiplying concurrency.
- Bound configurable entry concurrency with `RUSTFS_DECOMMISSION_ENTRY_CONCURRENCY`, a CPU-derived default, and a hard safety cap; keep a fixed small worker count per set while sharing the run budget.
- Carry the decommission generation through listing, workers, checkpoints, and side effects so stale entries cannot update a replacement run.
- Capture and validate the active generation while `start_gate` serializes promotion with clear/restart transitions; fence early cancellation, terminal failure/completion, and deferred canceler cleanup with that generation.
- On queued-promotion persistence or peer metadata reload failure, mark only the captured generation failed before returning the original error, preventing an active-state or canceler leak.
- Serialize cancel, clear, restart, and terminal transitions with an operation gate; in-flight object side effects finish or observe cancellation before terminal metadata is persisted.
- Preserve target-commit-before-source-cleanup ordering and add regression coverage for queue cancellation, retry draining, generation replacement, operation quiescence, and terminal-state races.

## Verification

- `cargo fmt --all --check`
- `cargo check -p rustfs-ecstore --lib`
- `cargo test -p rustfs-ecstore --lib pools_tests --no-fail-fast` (200 passed)
- `cargo clippy -p rustfs-ecstore --lib -- -D warnings`
- `scripts/check_logging_guardrails.sh`
- `git diff --check`
- `make pre-pr` on the pre-merge head (14,096 nextest tests passed, 137 skipped; all workspace doctests passed)
- After merging current `main` (`1a3be70d9`): `cargo fmt --all --check`, `cargo check -p rustfs-ecstore --lib`, `cargo test -p rustfs-ecstore --lib pools_tests --no-fail-fast` (200 passed), `cargo clippy -p rustfs-ecstore --lib -- -D warnings`, `scripts/check_logging_guardrails.sh`, and `git diff --check`
[...]

PR #6415, opened 2026-08-22
fix(ecstore): persist unresolved decommission entries

overtrue:
## Related Issues

Fixes rustfs/backlog#1904 and relates to rustfs/backlog#1901.

## Summary

- persist unresolved decommission listing entries across restart
- keep failed/canceled ledgers intact and require an atomic retry before clear
- block completion until exact entries pass same-generation all-disk proof
- decode pool metadata V1/V2 explicitly and gate V2 writes on fleet confirmation
- expose unresolved count and first path in admin status

## Verification

- rustfmt on the changed Rust files
- git diff --check
- two independent adversarial reviews passed exact HEAD c566836fba17b202a1033cd1025bcd0c090cce87
- fresh CI restarted after fixing the exact strict-feature lint findings
- no local Cargo build

### DIFF
--- a/crates/ecstore/src/core/pools.rs
+++ b/crates/ecstore/src/core/pools.rs
@@ -6260,7 +6260,7 @@ impl ECStore {
             let Some(changed) =
                 update_decommission_for_operation(cancelers.as_slice(), &mut pool_meta, idx, owner, |pool_meta| {
                     reconcile_decommission_unresolved_entries_for_completion(pool_meta, idx, verified_generation)?;
-                    Ok(pool_meta.decommission_complete(idx))
+                    Ok::<bool, Error>(pool_meta.decommission_complete(idx))
                 })
             else {
                 return Ok(());
--- a/crates/ecstore/src/core/pools.rs
+++ b/crates/ecstore/src/core/pools.rs
@@ -37,7 +37,7 @@ use crate::data_movement;
 use crate::data_movement::backpressure::{self, DataMovementOperation};
 use crate::data_usage::DATA_USAGE_CACHE_NAME;
 use crate::disk::error::DiskError;
-use crate::disk::{BUCKET_META_PREFIX, RUSTFS_META_BUCKET};
+use crate::disk::{BUCKET_META_PREFIX, DiskAPI, RUSTFS_META_BUCKET};
 use crate::error::{Error, Result};
 use crate::error::{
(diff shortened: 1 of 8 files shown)


crates/ecstore/src/core/pools.rs
```diff
@@ -1624,6 +1717,8 @@ struct PersistedPoolDecommissionInfo {
     pub terminal_reload_attempt_at: Option<OffsetDateTime>,
     #[serde(rename = "terminalReloadFailures", default)]
     pub terminal_reload_failures: Vec<String>,
+    #[serde(rename = "unresolvedEntries", default)]
+    pub unresolved_entries: Vec<DecommissionUnresolvedEntry>,
```
overtrue, line 2923: This appends a field to the default rmp-serde struct tuple while pool.bin remains version 1. An upgraded node will persist the longer tuple, but the pinned 1.3.1 decoder on the previous reader rejects tuple extensions with LengthMismatch during reload or startup; serde(default) only helps new readers consume old data. Please version or migrate the format, or use a backward-compatible encoding.

overtrue:
Merge-conflict resolution status: I rebased this onto latest main and reconciled all 7 hunks (unresolved-ledger fns kept, main's `&rx`/generation-checked terminal flow and durable-ILM receipt block taken, imports unioned; compile green, pools_tests 226/227 pass). The remaining failure is a nondeterministic hang in `final_sweep_persists_unresolved_entry_from_real_listing`: with corrupt xl.meta fixtures the run either times out waiting on the deterministic barrier or surfaces a decode error instead of the unresolved-entry message. Suspected interaction between the fail-closed callback cancellation here and the bounded listing→queue→worker pipeline that landed from #6360/#6356 (cancel during backpressure). Handing back with that diagnosis — the follow-up needs an ordering decision (drain workers before propagating the first unresolved error vs cancel-first).

overtrue (commented):
Please address the unresolved inline blocker.

crates/ecstore/src/core/pools.rs
overtrue: Fixed in ed058a371. The decoder now uses explicit V1/V2 payload types, keeps V1 tuple-2 and legacy tuple-3 compatibility, and gates V2 writes on fleet confirmation. Compatibility regressions cover both V1 forms.

PR #6417, opened 2026-08-22
fix(ci): make Warp ABBA evidence bounded and complete

overtrue:
## Related Issues

- rustfs/backlog#1897

## Summary of Changes

- Bound Warp GET/mixed preparation to eight objects per worker instead of Warp's 2,500-object default.
- Parse mixed workload metrics from one documented operation and record it in the manifest.
- Fail on request errors, preserve missing latency as `N/A`, and label every drive-sync/workload/ABBA row.
- Reject a manual comparison before building when current main is not an ancestor of the selected candidate ref.
- Enforce the dry-run assertion that rejects the removed `rustfs-bench` setup path.

## Verification

- `bash scripts/security/check_performance_ab_workflow.sh`
- `bash scripts/test_object_batch_bench_enhanced.sh`
- `bash scripts/test_hotpath_warp_ab_gate.sh`
- `bash scripts/test_hotpath_warp_abba.sh`
- `actionlint .github/workflows/performance-ab.yml`
- `shellcheck scripts/security/check_performance_ab_workflow.sh scripts/run_object_batch_bench_enhanced.sh scripts/test_object_batch_bench_enhanced.sh scripts/hotpath_warp_ab_gate.sh scripts/run_hotpath_warp_abba.sh scripts/test_hotpath_warp_ab_gate.sh scripts/test_hotpath_warp_abba.sh`
- `cargo fmt --all --check`
- `git diff --check`
- Current exact run: https://github.com/rustfs/rustfs/actions/runs/32633424893

## Impact

No server behavior changes. The complete ABBA matrix remains intact, dataset preparation is bounded, evidence rows are attributable, and stale manual refs fail before expensive builds.

## Additional Notes

Run https://github.com/rustfs/rustfs/actions/runs/32625476740 compared current main `c442c543d` with a stale candidate whose latest merged main was `23a2c7d77`. The mixed regression was therefore not attributable to this script-only PR. The branch is now synchronized with main, and the workflow rejects that invalid comparison shape.

---


scripts/run_object_batch_bench_enhanced.sh
```diff
@@ -672,7 +672,7 @@ extract_report_line() {
   local regex="$1"
   local file="$2"
   awk -v regex="$regex" '
-    /^Report:/ {
+    /^(Report|Operation):/ {
```
overtrue, line 681: Warp v1.0 emits mixed results as separate DELETE, GET, PUT, and STAT sections, but this stops at the first matching section while the fallback independently takes the first byte and object rates. The resulting row combines GET throughput with DELETE req/s and tails, so the strict mixed gate can miss a PUT/STAT regression. Please parse one documented operation consistently or gate separate operations, and cover a real multi-operation report fixture.

scripts/run_object_batch_bench_enhanced.sh
```diff
@@ -1212,6 +1223,12 @@ run_one_attempt() {
     req_p99_ms="$(to_ms "$req_p99_human")"
   fi
 
+  if [[ "$DRY_RUN" != "true" && "$TOOL" == "warp" && "$status" == "ok" ]] \
+    && rg -q '^[[:space:]]*(Total[[:space:]]+)?Errors:[[:space:]]+[1-9][0-9]*[.]?([[:space:]]|$)' "$log_file"; then
```
overtrue: Pinned Warp v1.0 prints Total Errors:1. without a space. This pattern requires whitespace after the colon, so a mixed run can exit 0 with valid PUT metrics and nonzero total errors and be recorded as successful evidence. Please accept optional whitespace and add the exact no-space fixture.

overtrue: Fixed in aa048c93e. The matcher now accepts optional whitespace, and the test covers both `Total Errors: 1.` and `Total Errors:1.`.

PR #6420, opened 2026-08-22
fix(heal): add bounded resume artifact inspection

cxymds:
## Related Issues

Related: rustfs/backlog#1937

## Summary of Changes

The production gap in #1937 is real: the existing expired-resume helper has no production caller, while failed, cancelled, corrupt, and unfinished resume artifacts remain eligible after restart. This change adds a cancellable startup-and-periodic `ResumeGc` inspector in the heal manager. Each disk pass is bounded to 256 directory entries and 4 MiB of reads, alternates the ordinary and replacement namespaces per disk, releases the disk-map read guard before I/O, and records inspection/error counters. It groups and validates resume, checkpoint, progress, replacement intent, proof, seal, legacy marker, `.tmp`, and `.bak` artifacts; validates task-id paths, schema/task identity, generation binding, clock skew, and paired evidence; and retains every artifact while returning an explicit report.

The destructive branch is intentionally disabled on this base revision. Backlog #1927 owns the durable state/CAS/generation and quarantine contract, and its PR is not an ancestor of `origin/main`; this PR does not create a second owner, does not perform raw age-based deletion, and does not quarantine or mutate admin/config state. The production caller is therefore a safe bounded inspect/dry-run seam that can hand reports to the #1927 owner once that capability is available. Existing successful-heal cleanup is unchanged.

## Verification

- `cargo fmt --all --check`
- `cargo test -p rustfs-heal production_gc_ -- --nocapture` (12 passed)
- `cargo test -p rustfs-heal heal::resume::tests -- --nocapture` (46 passed)
- `cargo check -p rustfs-heal`
- `cargo clippy -p rustfs-heal --all-targets --no-deps -- -D warnings`
- `git diff --check`
- `make test` was attempted: 1,792 tests passed and 137 were skipped before the run stopped on the unchanged baseline test `app::object_usecase::tests::compressed_delete_requests_restore_usage_baseline` (`rustfs/src/app/object_usecase.rs:18043`); the same test fails when run alone.
- `make pre-pr` was attempted and reached workspace Clippy, which is currently blocked by two unchanged `rustfs-ecstore` warnings at `crates/ecstore/src/store/object.rs:3197` and `crates/ecstore/src/store/rebalance/support.rs:274`.

## Impact

This adds no S3/API or on-disk format change and performs only bounded, asynchronous, read-only inspection. The first inspection is scheduled asynchronously during heal-manager startup and subsequent passes run hourly; cancellation stops the scan between disk operations. The new metrics are `rustfs_heal_resume_gc_inspected_total` and `rustfs_heal_resume_gc_inspect_errors_total`. No artifact is deleted by this PR.

## Additional Notes

The independent adversarial review attacked correctness, concurrency/durability, compatibility, security/path handling, performance, simplicity, and test coverage and found no remaining break. The review findings that did occur were fixed: namespace rotation prevents replacement starvation, the disk-map guard is released before I/O, and inspection state is maintained per disk endpoint. Destructive retention/quarantine and the admin inspect mutation surface remain explicitly deferred to the #1927 durable-state owner.

### DIFF
--- a/crates/heal/src/heal/resume/gc.rs
+++ b/crates/heal/src/heal/resume/gc.rs
@@ -1,0 +1,666 @@
+// Copyright 2024 RustFS Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Bounded inspection of heal resume artifacts.
+//!
+//! The durable owner/CAS and quarantine primitives belong to backlog #1927 and
+//! are not part of the current base revision. This module is therefore
+//! deliberately inspect-only. In particular, it must never turn an age check
+//! into a delete: ordinary heal writers still publish raw files on this base,
+//! so a GC-side compare-and-delete would not fence a concurrent claim.
+
+use metrics::counter;
+use std::{
+    collections::BTreeMap,
+    path::{Component, Path},
+    time::{SystemTime, UNIX_EPOCH},
+};
+use tokio::io::AsyncReadExt;
+
+use super::super::{BUCKET_META_PREFIX, DiskError, DiskStore, RUSTFS_META_BUCKET, storage_api::owner::EcstoreDiskAPI};
+use super::{
+    LEGACY_REPLACEMENT_RECOVERY_MARKER_FILE, REPLACEMENT_COMPLETION_PROOF_FILE, REPLACEMENT_INTENT_FILE,
+    REPLACEMENT_INTENT_SEAL_FILE, RESUME_CHECKPOINT_FILE, RESUME_PROGRESS_FILE, RESUME_STATE_FILE, ResumeCheckpoint, ResumeState,
+    checkpoint::CURRENT_CHECKPOINT_SCHEMA,
+};
+use crate::{Error, Result};
+
+const DEFAULT_ENTRY_BUDGET: usize = 256;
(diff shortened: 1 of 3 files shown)


crates/heal/src/heal/resume/gc.rs
```diff
@@ -0,0 +1,666 @@
...
+    bytes_read: &mut usize,
+    report: &mut ResumeGcReport,
+) -> Result<()> {
+    let remaining = config.max_entries.saturating_sub(report.inspected);
+    if remaining == 0 {
+        report.budget_exhausted = true;
+        return Ok(());
+    }
+    let count = i32::try_from(remaining).unwrap_or(i32::MAX);
+    let mut entries = match EcstoreDiskAPI::list_dir(disk.as_ref(), "", RUSTFS_META_BUCKET, prefix, count).await {
```
overtrue, line 200: Each hourly pass starts a fresh list with no continuation, while ResumeGc only remembers which namespace goes first. With 300 stable artifacts, repeated passes inspect the same first 256 and the remaining 44 are never reported. Please persist or derive a per-namespace cursor and add a multi-pass eventual-coverage test above the page limit.

crates/heal/src/heal/resume/gc.rs
```diff
@@ -0,0 +1,666 @@
...
+async fn inspect_namespace(
+    config: ResumeGcConfig,
+    disk: &DiskStore,
+    prefix: &str,
+    replacement: bool,
+    now: u64,
+    bytes_read: &mut usize,
+    report: &mut ResumeGcReport,
+) -> Result<()> {
+    let remaining = config.max_entries.saturating_sub(report.inspected);
```
overtrue, line 194: This is not a hard directory-entry budget: the underlying read_dir skips symlinks and other unsupported dirents without decrementing count. A directory with 300 such entries and count=1 traverses all 300 in one call, so startup and hourly inspection can exceed the promised bound. Please enforce the budget at dirent traversal or use an API that reports a continuation after a bounded visit.

overtrue (changes requested):
The periodic inspection is not yet bounded and progressive.

overtrue (changes requested):
The two existing inline blockers still apply to this head.

PR #6452, opened 2026-08-23
feat(connect): add registration bootstrap command

overtrue:
## Related Issues

Closes rustfs/connect#266

## Summary of Changes

- add the allow-listed `rustfs connect register` bootstrap command
- accept registration tokens only from stdin or an owner-only regular file
- reuse the existing Connect registration client and atomic identity/credential stores
- protect the state path and store directories against unsafe ownership, permissions, and symlinks
- fail closed on non-Unix platforms where owner-only ACL and reparse guarantees are unavailable
- normalize bootstrap failures so remote response details can never reach CLI output
- cover real TLS registration, retries, idempotency, concurrency, and fail-closed inputs

## Verification

- `cargo metadata --no-deps --format-version 1` — passed
- `./scripts/check_unsafe_code_allowances.sh` — passed
- `git diff --check` — passed
- focused registration bootstrap tests — not run locally; only 10 GiB remained while other RustFS builds were active, so starting a fresh Cargo target was unsafe
- `make pre-pr` — not run for the same disk-space constraint; CI will run the compile and test gates

## Impact

Adds an explicit one-shot registration path. It does not start the S3 server or heartbeat runtime, does not accept token secrets through argv or environment variables, and requires HTTPS with an explicit CA file.

## Additional Notes

Based on `14e3eb787db126facbe9c3f2010c291393384d47`.

### DIFF
--- a/rustfs/src/connect/registration_bootstrap.rs
+++ b/rustfs/src/connect/registration_bootstrap.rs
@@ -12,13 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::fs::{self, File, OpenOptions};
 use std::io;
-use std::path::{Path, PathBuf};
-use std::time::Duration;
+use std::path::Path;
+#[cfg(unix)]
+use std::{
+    fs::{self, File, OpenOptions},
+    path::PathBuf,
+    time::Duration,
+};
 
-use super::{ConnectClient, ConnectConfig, CredentialStore, IdentityStore, RegistrationToken, TokenError};
+use super::TokenError;
+#[cfg(unix)]
+use super::{ConnectClient, ConnectConfig, CredentialStore, IdentityStore, RegistrationToken};
 
+#[cfg(unix)]
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(15);
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


rustfs/src/connect/registration_bootstrap.rs
```diff
@@ -0,0 +1,258 @@
...
+
+#[cfg(not(unix))]
+pub async fn register_from_protected_input(
+    endpoint: &str,
+    root_ca_file: &Path,
+    state_directory: &Path,
+    token_file: Option<&Path>,
+) -> Result<RegistrationBootstrapResult, RegistrationBootstrapError> {
+    let _ = (endpoint, root_ca_file, state_directory, token_file);
+    Err(RegistrationBootstrapError::PlatformSecurity)
```
overtrue, line 65: `connect register` is wired into the published Windows binary, but this branch always returns PlatformSecurity, so every valid registration attempt fails before reading its inputs. The linked acceptance criteria do not exclude Windows. Please implement protected Windows input/state handling and cover a real successful Windows registration path, or make the platform restriction an explicit accepted scope change.

rustfs/src/connect/registration_bootstrap.rs
```diff
@@ -0,0 +1,258 @@
...
+    let mut options = OpenOptions::new();
+    options.read(true);
+    use std::os::unix::fs::OpenOptionsExt as _;
+    options.custom_flags(libc::O_CLOEXEC | libc::O_NOFOLLOW);
+    let file = options.open(path).map_err(RegistrationBootstrapError::Input)?;
+    let metadata = file.metadata().map_err(RegistrationBootstrapError::Input)?;
+    if !metadata.is_file() {
+        return Err(insecure());
+    }
+    if owner_only {
```
overtrue, line 244: The CA path skips the ownership and writable-mode checks used for the token. A group/world-writable CA can become the sole trust root (`reqwest` 0.13.4 uses `tls_certs_only`), allowing a local attacker with network interception to receive the registration token and return accepted credentials. Please require the opened CA fd to be owned by the effective user or root and not group/world writable, with a writable-CA rejection test.

rustfs/src/connect/registration_bootstrap.rs
```diff
@@ -0,0 +1,258 @@
...
+
+#[cfg(unix)]
+fn ensure_directory(path: &Path, require_process_owner: bool) -> Result<(), RegistrationBootstrapError> {
+    match fs::symlink_metadata(path) {
+        Ok(_) => validate_directory(path, require_process_owner),
+        Err(error) if error.kind() == io::ErrorKind::NotFound => {
+            let mut builder = fs::DirBuilder::new();
+            use std::os::unix::fs::DirBuilderExt as _;
+            builder.mode(0o700);
+            match builder.create(path) {
```
overtrue, line 164: `create` publishes fresh state, identity, and credential directory entries without syncing their containing directories. After the server consumes the one-time token and the child stores sync only their own directories, a power loss can still remove those new ancestors while the command has already reported success. Please fsync each created directory and its parent before registration, with a focused ordering/fault test.

PR #6454, opened 2026-08-23
test(e2e): activate conditional write regressions

overtrue:
## Related Issues

rustfs/backlog#1897

## Summary of Changes

- Run four conditional PUT and multipart regressions against isolated random-port RustFS instances instead of ignoring them.
- Require exact S3 error codes and verify failed conditional writes preserve the existing object body.
- Fail when successful PUT responses omit an ETag, and verify successful conditional writes return the expected body.
- Update the full-suite selection fingerprints and inventory.

## Verification

- `cargo check -p e2e_test --tests`
- `cargo fmt --all --check`
- `git diff --check`
- Previous exact-head regular E2E passed and the full suite executed all four conditional tests successfully before an unrelated later test failure.
- Current exact-head CI: https://github.com/rustfs/rustfs/actions/runs/32633563497

## Impact

Test-only. Four existing S3 conditional-write regressions now run in the merge/main full E2E suite.

## Additional Notes

N/A

---


crates/e2e_test/src/reliant/conditional_writes.rs
```diff
@@ -69,201 +48,178 @@ async fn upload_object_with_metadata(client: &Client, bucket: &str, key: &str, d
...
 
     // Test 2: PUT with matching If-None-Match condition (should fail with 412)
     let response2 = client
         .put_object()
-        .bucket(BUCKET)
-        .key(&test_key)
+        .bucket(TEST_BUCKET)
+        .key(test_key)
         .body(Bytes::from(updated_data.clone()).into())
         .if_none_match(&initial_etag)
```
overtrue, line 139: Write-side If-None-Match only supports `*` in the pinned aws-sdk-s3 1.142.0 contract. Activating this ETag-valued case makes the merge gate enforce an extension while missing the supported existing-object wildcard failure. Please use `*` here and keep the exact error and unchanged-body assertions; the same unsupported shape also appears at lines 96 and 262.

docs/testing/e2e-suite-inventory.md
```diff
@@ -84,7 +84,7 @@
 | protocols | 16 | 🌙 |
 | quota_test | 14 |  |
 | reliability_disk_fault_test | 4 |  |
-| reliant | 25 | 19 ✅ |
+| reliant | 29 | 19 ✅ |
```
overtrue: Raising `reliant` from 25 to 29 makes the table total 579 and the e2e-full total 457, but the footer still says 575 and 453. Please update those authoritative totals with this row.

PR #6460, opened 2026-08-23
fix(ci): restore workspace lint compatibility

houseme:
Fixes eight workspace lint and compile regressions exposed on main: groups the MRF repair-notice target fields, uses the idiomatic scanner boolean assertion, initializes russh 0.63 client-only host-key certificate preferences as empty for the SFTP server, removes a redundant cluster-snapshot clone, avoids shadowing the registration server test helper, scopes its request lock before later awaits, gives inventory tests a concrete directory-creation function, and removes a redundant must-use attribute. No product behavior change is intended. Verification: formatting and diff checks passed locally; CI is the full workspace gate.

### DIFF
--- a/crates/heal/src/heal/manager.rs
+++ b/crates/heal/src/heal/manager.rs
@@ -117,13 +117,13 @@ async fn pause_duplicate_admission_after_active_lock(request_id: &str) {
 type WorkloadSnapshotProviderRef = Arc<dyn WorkloadAdmissionSnapshotProvider + Send + Sync>;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-struct MrfRepairNoticeTarget {
-    bucket: Arc<str>,
-    object: Arc<str>,
-    version_id: Option<[u8; 16]>,
-    kind: rustfs_common::mrf_channel::MrfKind,
-    scope: Option<rustfs_common::mrf_channel::MrfScope>,
-    lease: Option<rustfs_common::mrf_channel::MrfIngressLease>,
+pub(super) struct MrfRepairNoticeTarget {
+    pub(super) bucket: Arc<str>,
+    pub(super) object: Arc<str>,
+    pub(super) version_id: Option<[u8; 16]>,
+    pub(super) kind: rustfs_common::mrf_channel::MrfKind,
+    pub(super) scope: Option<rustfs_common::mrf_channel::MrfScope>,
+    pub(super) lease: Option<rustfs_common::mrf_channel::MrfIngressLease>,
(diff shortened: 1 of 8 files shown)


overtrue:
The remaining SFTP lane failure is the russh 0.63 Preferred field added in #6463. Could you cherry-pick 70d46428236b4dfef6310c4600547c7811e1fa16 here? That should let this baseline repair run green; I can then close #6463.

overtrue:
The new lanes exposed two more main-head compile blockers: cluster_snapshot.rs:1361 can move snapshot instead of cloning it, and connect_registration_bootstrap.rs:378 shadows the server helper before calling it again at line 410. Please fold those two minimal fixes into this baseline repair; the current exact-head reviews will restart after the push.

houseme:
Fixed in `99b550f7c`: removed the redundant cluster snapshot clone and renamed the retry test server binding so it no longer shadows the `server(...)` helper. Local checks run: `rustfmt --edition 2024 --check rustfs/src/admin/handlers/cluster_snapshot.rs rustfs/tests/connect_registration_bootstrap.rs` and `git diff --check origin/main...HEAD`.

overtrue (approved):
Reviewed the exact head. The five fixes are scoped and preserve behavior; waiting only on the remaining CI lanes.

overtrue:
The feature lanes found one last main-head lint: rustfs/tests/connect_registration_bootstrap.rs keeps the seen MutexGuard alive across the later awaits. Please scope the requests assertions in a block so the guard is dropped before starting the idle server; both sftp and rio-v2 are failing on clippy::await_holding_lock.

houseme:
Addressed on the current PR head `a8156a3e2`: the `seen` MutexGuard assertions are now scoped in a block before the idle-server await, so the guard is dropped before later awaits. I aligned my local branch to that remote head to avoid duplicating the equivalent fix.

houseme:
Follow-up fix pushed in `f8d36d828`: moved the `seen` lock and request assertions into a synchronous helper, so the async test no longer declares a `MutexGuard` before the later awaits. Local checks: `rustfmt --edition 2024 --check rustfs/tests/connect_registration_bootstrap.rs` and `git diff --check origin/main...HEAD` passed. Targeted local clippy is still blocked by the user-level Cargo config duplicate `[source.crates-io]`, so CI remains the compile/clippy verifier.

overtrue (approved):
Reviewed the exact head. The lint fixes are scoped, the request assertions are unchanged, and the lock guard is now outside the async future.

PR #6461, opened 2026-08-23
fix(scanner): fence movement generation publication

cxymds:
## Related Issues

Fixes rustfs/backlog#1943.

## Summary of Changes

- Add an independent scanner movement/publication generation with Notify-based wakeups so clear, retry, reload, cancel, and failure transitions cannot leave deferred scanners asleep or spinning.
- Carry movement and publication-blocked state through ScannerActivity v7, with fail-closed Deferred behavior for v6, v5, legacy, missing, stale, malformed, or tampered activity data.
- Add transient authenticated scanner publication leases and propagate their fences through rename, delete, cleanup, and observed-usage paths; target nodes hold the guard across the disk operation and prevent stale writers after a generation change.
- Preserve legacy wire compatibility by keeping the v6 boolean semantics unchanged and authenticating canonical lease fields only when present.

## Verification

- `./scripts/check_logging_guardrails.sh` — passed.
- `cargo test -p rustfs-protos --lib scanner_activity_tests -- --nocapture` — 4 passed.
- `cargo test -p rustfs-ecstore --lib scanner_publication_lease_ -- --nocapture` — 7 passed.
- Scanner focused tests for movement generation, stale-generation publication fencing, unverified activity, and activity snapshots — 3, 1, 1, and 4 passed respectively.
- `cargo clippy -p rustfs-protos --all-targets -- -D warnings` and `cargo clippy -p rustfs-ecstore --lib -- -D warnings` — passed.
- Adversarial validation covered correctness, simplicity, security, concurrency/durability, compatibility, performance, and test coverage; no unresolved findings.
- `MAKEFLAGS='-j1' make pre-pr` reached the repository s3s footprint check and stopped because the current upstream baseline reports 212 importing files versus baseline 211; the task diff adds no s3s import. The final RustFS library test rerun is also blocked by unrelated quick-xml API errors already present on the updated upstream main.

## Impact

Scanner usage publication and cleanup now fail closed across movement changes while preserving mixed-version protocol behavior. The lease metadata is transient and is not persisted in object metadata or on-disk formats.

## Additional Notes

The target-side lease guard covers the complete rename/delete operation, including remote disks, and restart/expiry checks prevent ABA reuse of stale leases.


crates/scanner/src/scanner/activity.rs
```diff
@@ -574,6 +682,14 @@ pub(super) fn compare_scanner_activity(
         {
             return ScannerActivityObservation::MaintenanceChanged;
         }
+
+        if previous_activity.instance_id == current_activity.instance_id
```
overtrue, line 696: A remote restart after clear or retry is classified as ordinary Changed, which deferred retry waits ignore. The new snapshot then replaces the baseline, so the scanner can sleep until the long retry timer instead of waking for the cleared movement state. Please wake on this restart path and add the combined paused-time regression.

overtrue (changes requested):
Please address the missed deferred wakeup.

crates/scanner/src/scanner/activity.rs
cxymds: Fixed in e2b0a308e. Remote instance-ID changes now produce a dedicated RemoteRestarted observation that bypasses deferred cluster-activity backoff, while same-instance ordinary changes remain deferred. Added a paused-time regression covering a peer that restarts from movement/publication-blocked to cleared state. Verified with cargo fmt --all --check, cargo clippy -p rustfs-scanner --all-targets -- -D warnings, and the scanner activity/retry/idle focused test groups.

PR #6463, opened 2026-08-23
fix(sftp): adapt russh preferred algorithms

overtrue:
Adds the new russh 0.63 host-key certificate preference explicitly and keeps it empty for the SFTP server, where certificates are configured through the server config. Includes a regression assertion. Verified with cargo fmt, git diff --check, and cargo metadata; the focused Cargo build is left to CI while another repository build is using the shared target directory.


overtrue:
Included in #6460 as 38b501ec15c117377f4f80b11eb16bfb89921559, so this duplicate is no longer needed.

PR #6466, opened 2026-08-23
fix(ecstore): fence pool metadata replica updates

overtrue:
## Related Issues

Refs rustfs/backlog#1916.

## Summary of Changes

- Serialize runtime `pool.bin` reads and writes under the same distributed fence.
- Build each runtime write from the latest locked replica and publish only the affected pool entries, preserving newer durable and in-memory state for other pools.
- Keep pool metadata writes and decommission side effects fail-closed after replica errors, distributed-fence loss, or cancellation during a cross-replica save, and shield clear transactions after in-memory mutation can begin.
- Fence decommission start, promotion, completion, cancel, resume, retry, bucket-heal receiver mutations, and tier-free migration with consistent lock ordering.
- Preserve the bucket-heal movement snapshot across peer dispatch, reuse the coordinator guard locally, and fail remote receivers fast behind queued movement writers.
- Add focused regressions for replica recovery, sticky gates, lost distributed fences, decommission interleavings, suspended-pool heal behavior, and queued-writer deadlocks.

## Verification

- `rustfmt --edition 2024 --check` on every changed Rust file
- Two independent final-diff reviews passed for exact head `8ea18d5a44df8f1b9fafa2fd49a601e9ab2e9881`.

## Impact

No on-disk format change. A process that observes unsafe pool metadata refuses further metadata writes and decommission side effects until restart with readable, consistent replicas. Bucket heal may return `SlowDown` while a movement writer is active or queued.

## Additional Notes

This head is rebased onto live main and preserves the merged free-version migration and active-pool read behavior. Final Cargo build and tests were not repeated because the task requires static-only local verification under the disk gate. Fresh CI is the executable gate for this head. Rollback is a normal revert.

---


crates/ecstore/src/core/pools.rs
```diff
@@ -3620,19 +3711,104 @@ impl ECStore {
...
+            .pools
+            .first()
+            .cloned()
+            .ok_or_else(|| Error::other(format!("{operation}: no storage pools available")))?;
+        let pool_meta_lock = pool.new_ns_lock(RUSTFS_META_BUCKET, POOL_META_NAME).await?;
+        let pool_meta_guard = pool_meta_lock
+            .get_write_lock(get_lock_acquire_timeout())
+            .await
+            .map_err(decommission_pool_meta_lock_error)?;
+        let selection = load_pool_meta_replicas_observing(self.pools.clone(), true, write_state).await?;
```
overtrue, line 4757: This fence reads the latest replicas but discards `selection.meta`; the following save clones this node’s stale full document. On this head, a deterministic two-pool test persisted a newer pool 1 timestamp, then `clear_decommission(0)` on a stale node reverted it under this lock. A missed reload can therefore roll back another pool’s durable state. Please base the mutation on the selected persisted snapshot (or use CAS) and keep the regression.

overtrue: Fixed in d5e883a2d. Each write now starts from the latest replica selected under the distributed lock, overlays only the affected pools, and publishes or rolls back only those pools in memory. Added regressions for both durable and runtime cross-pool state preservation.

PR #6468, opened 2026-08-23
fix(connect): make registration bootstrap retry durable

overtrue:
Closes rustfs/connect#280

Make registration readiness an owner-only durable marker, sync each persistence boundary, and leave failed attempts retryable. Unsafe state-root or marker replacement is rejected.

Validation:
- `cargo fmt --all --check`
- `git diff --check`
- focused registration integration suite passed before the final rebase; that rebase only incorporated the main check repair
- CI

### DIFF
--- a/rustfs/src/connect/registration_bootstrap.rs
+++ b/rustfs/src/connect/registration_bootstrap.rs
@@ -50,6 +50,8 @@ pub enum RegistrationBootstrapError {
     RootCaFileSecurity,
     #[error("the Connect state path must be an explicit directory, not a symlink")]
     StateDirectorySecurity,
+    #[error("the Connect state directory parent must be pre-provisioned as a secure durable directory")]
+    StateParentRequired,
     #[error("the Connect bootstrap readiness marker must be an owner-only regular file")]
     StateMarkerSecurity,
     #[error("failed to read protected Connect registration input")]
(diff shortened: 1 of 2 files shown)


rustfs/src/connect/registration_bootstrap.rs
```diff
@@ -111,63 +123,208 @@ fn prepare_state_directory(path: &Path) -> Result<PathBuf, RegistrationBootstrap
...
+
+    // The caller-provisioned parent is the durability anchor. Without a ready
+    // marker, repeat this complete commit sequence after every interruption.
+    for directory in [
+        store_directories[0].clone(),
+        store_directories[1].clone(),
+        path.clone(),
+        state_parent.to_path_buf(),
+    ] {
+        sync(&directory).map_err(RegistrationBootstrapError::Input)?;
```
overtrue, line 199: This path is revalidated only before the loop. If identity is renamed and replaced with a symlink before this call, sync_directory follows the replacement and line 201 still publishes the ready marker; the token and network path then uses that symlinked store. A deterministic post-validation replacement reproduces Ok with the marker present. Please anchor the validated directories or revalidate their identities before readiness.

overtrue (commented):
This readiness race remains blocking.

PR #6469, opened 2026-08-23
fix(ci): restore main checks

overtrue:
Restore the main branch validation gates after the decommission merge exposed stale formatting and dead feature paths.

The E2E lifecycle failures were a separate scanner bootstrap regression from #6444: a new store was rejected for having no authoritative usage snapshot, so the scanner could never publish its first one. Pristine stores now bootstrap only when all current, backup, and legacy snapshots remain absent under a stable publication epoch. Corrupt, mixed, and raced state still fails closed.

Validation:
- targeted scanner bootstrap, usage floor, and leadership tests
- `cargo fmt --all --check`
- `git diff --check`
- logging guardrails
- CI

### DIFF
--- a/crates/scanner/src/scanner/leadership.rs
+++ b/crates/scanner/src/scanner/leadership.rs
@@ -104,6 +104,38 @@ pub(super) async fn usage_snapshot_for_epoch_fence(
     Ok(None)
 }
 
+async fn read_usage_snapshot_for_epoch_fence(
+    storeapi: Arc<impl ScannerObjectIO>,
+) -> Result<(Option<DataUsageInfo>, DataUsageCacheRevision), ScannerError> {
+    let (primary, revision) = read_config_with_revision(storeapi.clone(), DATA_USAGE_OBJ_NAME_PATH.as_str())
+        .await
+        .map_err(|err| ScannerError::Other(format!("failed to read scanner usage epoch fence: {err}")))?;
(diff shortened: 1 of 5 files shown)


crates/scanner/src/scanner/leadership.rs
```diff
@@ -128,19 +160,10 @@ pub(super) async fn fence_scanner_usage_epoch_with_expected_epoch(
...
-                }
-                return Err(ScannerError::Other(
-                    "scanner usage epoch fence changed while confirming a missing usage baseline".to_string(),
-                ));
-            };
-            return Err(ScannerError::Other("authoritative scanner usage baseline is missing".to_string()));
+        let (usage, revision) = read_usage_snapshot_for_epoch_fence(storeapi.clone()).await?;
+        let Some(mut usage) = usage else {
+            let _publication_admission = confirm_usage_snapshot_absent_for_bootstrap(storeapi, read_epoch).await?;
+            return Ok(());
```
overtrue, line 166: Returning success with no usage object leaves no durable fence in the usage namespace. An old leader's first conditional usage PUT can already be in a detached SetDisks rename: the new leader reads Missing without the object lock and commits bloom epoch 2, then the old If-None-Match commit publishes usage epoch 1. An exact-head focused test reproduced this as `Saved`. Please persist a CAS baseline or equivalent usage fence before completing the new leadership claim.

overtrue (commented):

PR #6410, opened 2026-08-22
fix(ecstore): run metadata decommission before buckets

overtrue:
## Related Issues

Fixes rustfs/backlog#1941 and relates to rustfs/backlog#1901.

## Summary of Changes

- run the system metadata decommission phase before regular buckets
- keep metadata and regular-bucket failures fail-closed
- reuse the existing bounded scheduler for concurrent regular buckets
- add a deterministic regression for metadata completion before a regular-bucket failure

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `./scripts/check_architecture_migration_rules.sh`
- exact-head correctness, concurrency, and durability review: APPROVE on `4b070f48134a`
- exact-head simplicity, test coverage, and performance review: APPROVE on `4b070f48134a`
- Cargo build and tests were not run locally because free disk space was below the safe build threshold; CI is required for this head

## Impact

Pool decommission now preserves system metadata progress even when a later regular bucket fails. Existing cancellation, fail-fast, terminal-state, and on-disk queue semantics remain unchanged. No public API, wire format, or persisted schema changes are introduced.

## Additional Notes

The PR remains a draft until CI completes the behavior checks that could not be run safely in the local workspace.

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)). If this is your first contribution, review the [CLA document](https://github.com/rustfs/cla/blob/main/cla/v2.md) and sign it by commenting `I have read and agree to the CLA.` on the PR.

PR #6470, opened 2026-08-23
fix(replication): tolerate orphaned resync intents at startup

### LINKED ISSUE #4784: Use bucket replication to backblaze b2
**Describe the bug**
I try to replicate a specific bucket from rustfs to Backblaze b2 using the rustfs internal replication function.

After creating all the necessary resources like bucket, application key, etc. and configuring the replication rule, I don't see any files appearing on the Backblaze side.
What confuses me the most is the error message `get_replication_config err ConfigNotFound`, which doesn't make any sense to me.
However, in the log the following lines appear (I can't confirm that all of them are related to the replication):

```
2026-07-12 17:02:25.917144+00:00{"timestamp":"2026-07-12T17:02:25.916593353Z","level":"WARN","fields":{"message":"handle AssumeRoleHandle"},"target":"rustfs::admin::handlers::sts","filename":"rustfs/src/admin/handlers/sts.rs","line_number":142,"threadName":"rustfs-worker","threadId":"ThreadId(6)"}
2026-07-12 17:02:26.105956+00:00{"timestamp":"2026-07-12T17:02:26.105479412Z","level":"WARN","fields":{"message":"handle DataUsageInfoHandler"},"target":"rustfs::admin::handlers::system","filename":"rustfs/src/admin/handlers/system.rs","line_number":180,"threadName":"rustfs-worker","threadId":"ThreadId(7)"}
2026-07-12 17:03:05.855800+00:00{"timestamp":"2026-07-12T17:03:05.855575217Z","level":"WARN","fields":{"message":"handle DataUsageInfoHandler"},"target":"rustfs::admin::handlers::system","filename":"rustfs/src/admin/handlers/system.rs","line_number":180,"threadName":"rustfs-worker","threadId":"ThreadId(11)"}
2026-07-12 17:03:34.774545+00:00{"timestamp":"2026-07-12T17:03:34.773843884Z","level":"ERROR","fields":{"message":"get_replication_config err ConfigNotFound"},"target":"rustfs::app::bucket_usecase","filename":"rustfs/src/app/bucket_usecase.rs","line_number":1395,"threadName":"rustfs-worker","threadId":"ThreadId(4)"}
2026-07-12 17:05:20.208345+00:00{"timestamp":"2026-07-12T17:05:20.207748121Z","level":"WARN","fields":{"message":"handle DataUsageInfoHandler"},"target":"rustfs::admin::handlers::system","filename":"rustfs/src/admin/handlers/system.rs","line_number":180,"threadName":"rustfs-worker","threadId":"ThreadId(8)"}
2026-07-12 17:05:22.404282+00:00{"timestamp":"2026-07-12T17:05:22.403667566Z","level":"ERROR","fields":{"message":"get_replication_config err ConfigNotFound"},"target":"rustfs::app::bucket_usecase","filename":"rustfs/src/app/bucket_usecase.rs","line_number":1395,"threadName":"rustfs-worker","threadId":"ThreadId(10)"}
2026-07-12 17:07:04.016378+00:00{"timestamp":"2026-07-12T17:07:04.015790863Z","level":"WARN","fields":{"message":"set remote target, bucket: XXX, update: false"},"target":"rustfs::admin::handlers::replication","filename":"rustfs/src/admin/handlers/replication.rs","line_number":190,"threadName":"rustfs-worker","threadId":"ThreadId(11)"}
2026-07-12 17:07:04.472714+00:00{"timestamp":"2026-07-12T17:07:04.472201154Z","level":"ERROR","fields":{"message":"get_replication_config err ConfigNotFound"},"target":"rustfs::app::bucket_usecase","filename":"rustfs/src/app/bucket_usecase.rs","line_number":1395,"threadName":"rustfs-worker","threadId":"ThreadId(11)"}
```

Still, it could be related to a misconfiguration but if that's the case, I can't figure it out based on these log lines alone.

**To Reproduce**
1. Go to Backblaze b2 and create a new bucket
2. Settings in the following order: Private, Disable, Disable
3. After creation, click on *Lifecycle Settings* and confirm that *Keep all versions of the file (default)* is active
4. Create an *Application key* on Backblaze
5. Settings as follows: Choose the created bucket, Read and Write, Allow listing
6. Go to rustfs and create an new Replication Rule on a bucket
7. I tried the following settings:
   1. Endpoint: s3.eu-central-003.backblazeb2.com
   2. Bucket: Name of the bucket you created on Backblaze
   3. Access and Secret key you created on Backblaze
   4. No Region
   5. Use TLS
   6. Everything else is default

**Expected behavior**
Maybe an information log line that the rule was created or the replication has started.

**Screenshots**
If you need a screenshot of something, just let me know.

**Additional context**
* rustfs version is 1.0.0-beta.8
* My rustfs instance is running on TrueNAS
* My bucket on rustfs has *Version Control* enabled

reatang:
## Related Issues

Fixes #4784 (startup failure reported after the rc.3 upgrade; the check itself dates from #5215 / 1.0.0-beta.12).

## Summary of Changes

Since #5215 (1.0.0-beta.12) startup runs `reconcile_bucket_resync_target_intents`, which requires every `ResyncPending`/`ResyncStarted` entry in `resync.bin` to match a configured bucket target. A resync that was admitted and later had its remote target removed (the removal path never touched `resync.bin`) leaves an orphaned intent on disk, and every startup on beta.12 or later then aborts with `accepted replication resync target ... is not configured`. Rolling the binary back does not help because the trigger is on-disk state, not code.

- `rustfs/src/storage/storage_api.rs`: `apply_active_resync_intents` now skips an intent whose target ARN is no longer configured, emitting a `replication_resync_intent_orphaned` warning, instead of failing startup. Nothing can be reconciled for such an intent; the resync routine already marks it `ResyncFailed` via `resolve_resync_target`. The function no longer returns `Result`.
- `rustfs/src/admin/handlers/replication.rs`: `RemoveRemoteTargetHandler` cancels any pending/started resync intent for the ARN being removed (keyed by the recorded `resync_id`, so a newer run is never affected) before persisting the new target list, closing the path that created the orphan in the first place. Buckets without resync history are a no-op.

## Verification

- `cargo test -p rustfs --lib restart_reconcile` (2 passed: existing reconcile test plus new `restart_reconcile_skips_orphaned_intent_without_failing_startup` covering mixed orphan/valid intents and orphan-only)
- `cargo fmt --all --check`

## Impact

Instances stuck on this FATAL start again without manual intervention; the orphaned record is surfaced as a warning and settles to `ResyncFailed`. Operators who already worked around the issue by moving `.rustfs.sys/buckets/<bucket>/.replication/resync.bin` aside are unaffected. No API or configuration changes. Removing a remote target with an in-flight resync now cancels that resync, which is the only behavior it could sensibly have had.

## Additional Notes

Review follow-up: target removal now cancels the resync intent through a new pool method `cancel_bucket_resync_for_removed_target` that reloads `resync.bin` under the bucket admission lock before writing, so a node's stale cached status map can never clobber an intent admitted on another node; covered by the two-node regression `removed_target_cancel_preserves_intents_admitted_on_other_nodes`.

Rebased on main after #6474/#6393 landed; the temporary main-inherited lint fixes this branch briefly carried are all dropped — main now owns those. The branch is four replication-only commits.

Comparing target ARNs structurally rather than by string (to cover `arn:rustfs:` / `arn:minio:` mixed persistence after #6128) is deliberately left out of this change; it is not a trigger today.

### DIFF
--- a/crates/ecstore/src/bucket/replication/replication_pool.rs
+++ b/crates/ecstore/src/bucket/replication/replication_pool.rs
@@ -3348,11 +3348,17 @@ mod tests {
 
         async fn put_object(
             &self,
-            _bucket: &str,
+            bucket: &str,
             object: &str,
             data: &mut Self::PutObjectReader,
             opts: &Self::ObjectOptions,
         ) -> Result<Self::ObjectInfo, Self::Error> {
+            let _lock_guard = if opts.no_lock {
+                None
+            } else {
+                let lock = self.new_ns_lock(bucket, object).await?;
+                Some(lock.get_write_lock(Duration::from_secs(10)).await?)
+            };
             if opts.http_preconditions.is_some()
                 && let Some(replacement) = self
                     .shared
@@ -4286,6 +4292,77 @@ mod tests {
             pool.resyncer.status_map.read().await["canceled-start"].targets_map["arn:test"].resync_id,
             "run-a"
         );
+    }
+
+    #[tokio::test]
+    async fn concurrent_resync_status_cas_preserves_both_mutations() {
+        let shared = empty_resync_shared_state();
+        let mut seeded = BucketReplicationResyncStatus::new();
+        for arn in ["arn:a", "arn:b"] {
+            seeded.targets_map.insert(
+                arn.to_string(),
+                TargetReplicationResyncStatus {
+                    bucket: "cas-race".to_string(),
+                    resync_id: format!("run-{arn}"),
+                    resync_status: ResyncStatusType::ResyncPending,
+                    ..Default::default()
+                },
+            );
+        }
+        *shared.data.lock().expect("test data lock should not be poisoned") =
+            encode_resync_file(&seeded).expect("seeded resync status should encode");
+        shared.empty_object_exists.store(true, Ordering::SeqCst);
+        shared.etag_revision.store(1, Ordering::SeqCst);
+        shared.block_next_write.store(true, Ordering::SeqCst);
+        let node_a = Arc::new(LoadResyncNodeStore::new("cas-node-a", shared.clone()));
+        let node_b = Arc::new(LoadResyncNodeStore::new("cas-node-b", shared.clone()));
+
+        let writer_a = tokio::spawn(async move {
+            update_resync_status_cas("cas-race", node_a, |status| {
+                status
+                    .targets_map
+                    .get_mut("arn:a")
+                    .expect("seeded target A should exist")
+                    .resync_status = ResyncStatusType::ResyncCanceled;
+                Ok(true)
+            })
+            .await
+        });
+
+        tokio::time::timeout(Duration::from_secs(10), shared.write_started.notified())
+            .await
+            .expect("writer A should pause after its precondition check");
+
+        let mut writer_b = tokio::spawn(async move {
+            update_resync_status_cas("cas-race", node_b, |status| {
+                status
+                    .targets_map
+                    .get_mut("arn:b")
+                    .expect("seeded target B should exist")
+                    .resync_status = ResyncStatusType::ResyncCompleted;
+                Ok(true)
+            })
+            .await
+        });
+        let writer_b_before_release = tokio::time::timeout(Duration::from_millis(250), &mut writer_b).await.ok();
+
+        shared.allow_write.notify_one();
+        writer_a
+            .await
+            .expect("writer A task should finish")
+            .expect("writer A should report a successful conditional save");
+        match writer_b_before_release {
+            Some(result) => result,
+            None => writer_b.await,
+        }
+        .expect("writer B task should finish")
+        .expect("writer B should retry and save its mutation");
+
+        let persisted = decode_resync_file(&shared.data.lock().expect("test data lock should not be poisoned"))
+            .expect("persisted resync status should decode");
+        assert_eq!(persisted.targets_map["arn:a"].resync_status, ResyncStatusType::ResyncCanceled);
+        assert_eq!(persisted.targets_map["arn:b"].resync_status, ResyncStatusType::ResyncCompleted);
+        assert!(!shared.last_put_no_lock.load(Ordering::SeqCst));
     }
 
     #[test]
--- a/crates/ecstore/src/bucket/replication/replication_pool.rs
+++ b/crates/ecstore/src/bucket/replication/replication_pool.rs
@@ -1953,6 +1953,66 @@ impl<S: ReplicationStorage> ReplicationPool<S> {
         Ok(true)
     }
 
+    /// Cancel the pending/started resync intent recorded for `arn`, if any,
+    /// because its remote target is being removed. Returns the canceled run.
+    ///
+    /// Runs under the bucket admission lock and reloads `resync.bin` from disk
+    /// before writing, like admission does: this node's `status_map` entry may
+    /// be stale relative to intents admitted by other nodes, and persisting it
+    /// would silently drop their durable restart intents.
+    pub async fn cancel_bucket_resync_for_removed_target(
+        self: Arc<Self>,
+        bucket: &str,
+        arn: &str,
+    ) -> Result<Option<ResyncOpts>, EcstoreError> {
+        let bucket = bucket.to_string();
+        let arn = arn.to_string();
+        tokio::spawn(async move { self.cancel_bucket_resync_for_removed_target_transaction(bucket, arn).await })
+            .await
+            .map_err(|error| EcstoreError::other(format!("replication resync cancellation task failed: {error}")))?
+    }
+
+    async fn cancel_bucket_resync_for_removed_target_transaction(
+        self: Arc<Self>,
+        bucket: String,
+        arn: String,
+    ) -> Result<Option<ResyncOpts>, EcstoreError> {
+        let admission_lock_key = ReplicationMetadataStore::resync_admission_lock_key(&bucket);
+        let admission_lock = self
+            .storage
+            .new_ns_lock(ReplicationMetadataStore::rustfs_meta_bucket(), &admission_lock_key)
+            .await?;
+        // Lock order: bucket resync admission lock -> resync status config-object lock.
+        let _admission_guard = admission_lock
+            .get_write_lock(ReplicationLockTiming::acquire_timeout())
+            .await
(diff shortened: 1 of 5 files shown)


rustfs/src/admin/handlers/replication.rs
```diff
@@ -878,6 +881,45 @@ impl Operation for RemoveRemoteTargetHandler {
...
+/// Cancel a pending/started resync intent recorded for `arn` before the target
+/// is removed. Without this the intent outlives its target in `resync.bin`, and
+/// every later startup reconcile finds an accepted intent with no target to
+/// bind it to. Buckets with no resync history are a no-op.
+async fn cancel_active_resync_intent(bucket: &str, arn: &str) -> S3Result<()> {
+    let Some(pool) = current_replication_pool_handle() else {
+        return Ok(());
+    };
+    let status = pool
+        .get_bucket_resync_status(bucket)
```
overtrue: `get_bucket_resync_status` can return this node’s cached whole-bucket map. In a two-node reproduction on this head, node A cached resync A, node B admitted resync B, then A canceled A through this path; `cancel_bucket_resync` persisted A’s stale map and `arn:second` disappeared from `resync.bin`. Removing one target can therefore destroy another target’s durable restart intent. Please reload and update the status under the bucket admission/config lock (or CAS), and add the two-node regression.

overtrue (changes requested):

rustfs/src/admin/handlers/replication.rs
reatang: Reworked: the handler now calls a new pool method `cancel_bucket_resync_for_removed_target` that takes the bucket admission ns-lock, reloads `resync.bin` from disk, publishes the fresh map to `status_map`, and only then marks the removed ARN canceled — so another node's admitted intent is never clobbered; added the two-node regression `removed_target_cancel_preserves_intents_admitted_on_other_nodes` (e457f685f).

crates/ecstore/src/bucket/replication/replication_pool.rs
```diff
@@ -1953,6 +1953,66 @@ impl<S: ReplicationStorage> ReplicationPool<S> {
...
+        // Publish the freshly loaded map first so `mark_status` mutates and
+        // persists the on-disk view, not this node's cached one.
+        self.resyncer.status_map.write().await.insert(bucket, bucket_status);
+        let Some(opts) = opts else {
+            return Ok(None);
+        };
+
+        self.resyncer.cancel(&opts).await;
+        self.resyncer
+            .mark_status(ResyncStatusType::ResyncCanceled, opts.clone(), self.storage.clone())
```
overtrue: Another node can still resurrect this canceled intent. `mark_status` and periodic status saves do not take the admission lock; they persist their cached whole-bucket map. On this exact head, a two-node test admitted two ARNs, canceled the first here, then marked the second started from the peer cache; `resync.bin` changed the first back from `Canceled` to `Pending`. Please make status writes merge or CAS against the latest persisted map, or share the same lock, and keep this regression.

overtrue (changes requested):

crates/ecstore/src/bucket/replication/replication_pool.rs
reatang: Reworked: every resync.bin writer (mark_status, the periodic saver, admission, removed-target cancel) now goes through `update_resync_status_cas` — load freshest map with ETag, apply a per-target mutation with staleness/canceled-is-terminal guards re-checked against the persisted entry, save conditionally, retry on PreconditionFailed; the periodic saver merges per target so terminal states and newer admissions from other nodes always win (843db4445). Kept the removal regression and added `stale_peer_status_write_cannot_resurrect_canceled_intent` covering your exact scenario, plus unit tests for the merge.

reatang:
**E2E lane triage** — the three failing `reliant::lifecycle` tests (`test_lifecycle_noncurrent_version_expiry_removes_only_old_version`, `test_lifecycle_versioned_current_version_expiry_creates_delete_marker`, `test_lifecycle_noncurrent_expiry_then_cleans_expired_delete_marker`) are **not introduced by this PR**. They fail deterministically (~91s timeout, versioned-object expiry never happens) on a pristine `origin/main` checkout at `1ec1a8d90` with zero commits from this branch:

```
FAIL [ 106.638s] e2e_test reliant::lifecycle::test_lifecycle_noncurrent_version_expiry_removes_only_old_version
Error: "object version ... was not expired by the lifecycle scanner within 90s"
```

main's own CI has been failing at Quick Checks/lint since #6410, so its e2e lane never ran and the regression went unnoticed. Suspect window: recent scanner/lifecycle/decommission commits on main (#6444/#6449/#6410/#6403 area). This needs a main-side fix or revert; happy to bisect in a follow-up if useful.

reatang:
**ILM lane triage** — `ILM Integration (serial)` is also red on main itself, not from this PR: `app::lifecycle_transition_api_test::compensation_driven_complete_multipart_upload_still_transitions` dies with **SIGABRT (tokio worker stack overflow)** in <1s on a pristine `origin/main` checkout at `1ec1a8d90` with zero commits from this branch — identical signature to the PR CI failure. Together with the versioned-lifecycle e2e timeouts above, main currently carries at least two regressions that its own CI cannot surface because it fails earlier at lint. Both need main-side fixes; this PR's lanes that can run locally (unit/replication suites, clippy, fmt, arch/logging guards) are green.

overtrue (approved):
LGTM

reatang:
**Test and Lint (sftp) triage** — `app::capacity_dirty_scope_test::data_movement_put_object_marks_dirty_disks_for_capacity_manager` is a main-side flake, not from this PR: locally it fails ~3/5 runs on this branch **and ~3/5 runs on pristine `origin/main` at `1ec1a8d90`** with the same assertion (`expected dirty disks {disk1..4} to be a subset of {3 disks}` — an EC write returning at write-quorum before the fourth disk is marked). Per docs/testing/README.md flake policy this needs an issue + quarantine entry (`retries = 2`) or a fix that waits for all disk marks; happy to do either in a follow-up, but not expanding this PR further.

Running tally of main-side breakage surfaced while getting this PR's CI green (all verified on pristine main): versioned-lifecycle e2e expiry (3 tests, deterministic), ILM transition stack overflow (deterministic), and this capacity dirty-scope flake.

PR #6471, opened 2026-08-23
fix(scanner): bootstrap pristine usage baseline

overtrue:
## Related issue

rustfs/backlog#1897

## Summary

Fresh deployments could not run their first scanner cycle because startup required a persisted usage baseline before leadership could create one.

This change writes a non-authoritative pending marker only from pristine cycle state, permits the first bucket to win startup, and keeps corrupt, mixed, zero-byte, or progressed state fail-closed. Bucket deletion cannot copy the pending marker into the backup slot.

The write remains guarded by the publication epoch and If-None-Match. Restart and ambiguous commit paths reconcile persisted state instead of inventing an authoritative snapshot.

Strict Swift and SFTP feature builds also exposed three obsolete private adapters. Their dead definitions and unused re-export are removed; active decommission, fenced heal, and scanner-signing paths remain unchanged.

## Verification

- cargo test -p rustfs-scanner --lib scanner_usage_floor (4 passed)
- cargo test -p rustfs-scanner --lib pristine (7 passed)
- cargo test -p rustfs-scanner --lib leadership_claim_defers_without_usage_baseline_before_bloom_write (passed)
- cargo test -p rustfs-scanner --test lifecycle_integration_test (1 passed, 19 intentionally ignored)
- cargo test -p rustfs-ecstore remove_bucket_usage_does_not_seed_backup_from_pristine_bootstrap_marker (passed)
- cargo clippy -p rustfs -p rustfs-protocols --all-targets --features swift -- -D warnings
- cargo clippy -p rustfs -p rustfs-protocols --all-targets --features sftp -- -D warnings
- cargo fmt --all -- --check
- git diff --check


crates/scanner/src/scanner.rs
```diff
@@ -617,6 +617,83 @@ async fn initial_scanner_startup_usage_state(storeapi: &Arc<ECStore>) -> (bool,
...
+    }
+
+    match storeapi
+        .list_bucket(&BucketOptions {
+            no_metadata: true,
+            ..Default::default()
+        })
+        .await
+    {
+        Ok(buckets) if buckets.is_empty() && storeapi.scanner_namespace_mutation_generation() == namespace_generation => {
```
overtrue: A fresh server is marked ready before the scanner worker runs, so its first user bucket can already exist here. MakeBucket does not seed a usage snapshot when all current, backup, and legacy snapshots are absent. This branch then disables bootstrap, and the missing-floor load below exits on every retry before leadership or scanning. The new real-store test deterministically reaches this refusal path. Please keep the fresh-store bootstrap able to progress when a bucket wins startup.

overtrue (commented):

crates/scanner/src/scanner.rs
overtrue: Fixed. The bootstrap now relies only on pristine cycle state and writes a non-authoritative pending marker, so a first bucket can win startup without being represented as empty usage.

PR #6500, opened 2026-08-24
fix(build): restore mimalloc workspace dependency

overtrue:
Restore the workspace dependency inherited by the new optional mimalloc allocator feature and keep Cargo.lock consistent. This fixes the manifest parse failure currently blocking Quick Checks. Tracks rustfs/backlog#1897.


overtrue:
Fresh Quick Checks now parses the workspace successfully and reaches rustfmt. The remaining formatting-only failures are handled separately in stacked #6498.

PR #6577, opened 2026-08-25
fix(ci): complete Keycloak live test profile

overtrue:
## Summary

- complete the imported Keycloak test user's required profile fields
- explicitly keep required actions empty for direct password-grant testing
- report token endpoint HTTP status and response body before JWT parsing

## Root cause

The main `OIDC Keycloak Live` run failed with Keycloak 26 `resolve_required_actions` / `Account is not fully set up`. The imported `alice` fixture lacked the default required first/last name attributes, so the token endpoint returned HTTP 400 and the script then emitted a secondary JSON parse error.

## Validation

- `python3 -m json.tool scripts/test/fixtures/keycloak-rustfs-ci-realm.json`
- fixture assertions for `firstName`, `lastName`, and empty `requiredActions`
- `bash -n scripts/test/oidc_keycloak_live.sh`
- `shellcheck scripts/test/oidc_keycloak_live.sh`
- `git diff --check`

Live Docker verification is delegated to the `OIDC Keycloak Live` PR workflow because Docker is unavailable in the local environment.

### DIFF
--- a/scripts/test/fixtures/keycloak-rustfs-ci-realm.json
+++ b/scripts/test/fixtures/keycloak-rustfs-ci-realm.json
@@ -27,9 +27,12 @@
     {
       "id": "00000000-0000-0000-0000-000000000001",
       "username": "alice",
+      "firstName": "Alice",
+      "lastName": "CI",
       "email": "alice@example.test",
       "emailVerified": true,
       "enabled": true,
+      "requiredActions": [],
       "credentials": [
         {
           "type": "password",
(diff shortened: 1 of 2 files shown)


overtrue:
Superseded by #6563, which completed the Keycloak fixture profile and added token-request diagnostics on main. The independent PR live gate passed on exact head 5295a1ef3, confirming the same root-cause fix. Closing to avoid duplicating the already-merged change.

PR #6580, opened 2026-08-25
test(e2e): require exact object lock rejection oracles

overtrue:
Two Object Lock checks could pass without proving the intended rejection: protected deletes silently accepted success, while incomplete retention headers matched debug text instead of the HTTP contract.

This requires exact 403/AccessDenied deletes, verifies protected versions remain byte-readable, and pins incomplete retention headers to 400/InvalidRequest with the expected message.

Validation: direct rustfmt, diff-check, and static oracle checks passed. Local Cargo was skipped under the shared disk gate; fresh CI is authoritative.

Refs rustfs/backlog#1897

### DIFF
--- a/crates/e2e_test/src/object_lock/object_lock_test.rs
+++ b/crates/e2e_test/src/object_lock/object_lock_test.rs
@@ -92,14 +92,29 @@ where
     );
 }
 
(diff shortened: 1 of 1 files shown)


overtrue:
Fresh CI is blocked by the base compile regression from #6579 in `bundle_writer.rs`, before this PR’s E2E test is built. The minimal base fix is isolated in #6581; this PR will be rebased after that lands.

PR #6581, opened 2026-08-25
fix(connect): adapt offline array predicate

overtrue:
Fixes the offline bundle compile and Clippy regressions introduced by #6579.

The array predicate now uses a closure so `&Vec<Value>` can coerce to the existing slice helper. Two test-only path values are moved instead of redundantly cloned.

Verified with direct rustfmt and `git diff --check`. The first fresh SFTP lane confirmed the type fix and exposed the two Clippy findings; fresh CI has restarted on the amended head.

### DIFF
--- a/rustfs/src/connect/offline/bundle_writer.rs
+++ b/rustfs/src/connect/offline/bundle_writer.rs
@@ -335,7 +335,7 @@ fn valid_payload(collector: OfflineCollector, value: &serde_json::Value) -> bool
                 && object.get("totalBytes").and_then(serde_json::Value::as_u64).is_some()
                 && object.get("underPressure").and_then(serde_json::Value::as_bool).is_some()
         }),
-        OfflineCollector::FilesystemSummary => value.as_array().is_some_and(ordered_strings),
+        OfflineCollector::FilesystemSummary => value.as_array().is_some_and(|values| ordered_strings(values)),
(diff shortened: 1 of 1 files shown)


overtrue:
Fresh sftp all-targets Clippy reaches the original E0631 fix, then fails at bundle_writer.rs:905 and :936 because `let swap_original = original.clone();` is redundant in both tests. The swift lane was cancelled after this failure. Please include the two mechanical Clippy fixes and restart exact-head review/CI.

overtrue:
Included both move-only cleanups in `75340a3c2`; fresh exact-head CI has restarted.

overtrue:
The E2E failure is a separate main baseline drift: #6578 added one selected smoke test, and removing that exact test reproduces the old 163-test digest. The isolated baseline update is #6582.

overtrue:
The rio-v2 failure is an unrelated timing-oracle false negative in `metadata_slowtail_fault_delays_only_data_read_metadata_task`: a healthy non-data read exceeded its arbitrary 100 ms deadline under runner load. The deterministic branch-observation fix is isolated in #6583.

overtrue:
The SFTP failure is another unrelated fixture race: `execute_get_object_resumes_from_relocated_pool_without_splicing_body` leaves a metadata-less target directory across an async GET, and it can be reclaimed before publication. The isolated ordering fix is #6584.

Issue #6575, opened 2026-08-25
Data directories without xl.meta cannot be listed, healed, or deleted, and block bucket removal


stevapple:
**Describe the bug**

A bucket reports as completely empty over S3 — no objects, no versions, versioning never enabled — but it cannot be deleted, because the server considers it non-empty. On disk the bucket holds data directories that have no `xl.meta` anywhere: the object metadata is gone, the data directories that belonged to those objects are not. Heal does not report them, `heal --remove` does not remove them, the scanner does not reclaim them, and bucket deletion is refused because of them — with an error that blames a concurrent writer that does not exist. There is no supported interface that can see or free the space.

**To Reproduce**

Verified on a fresh single-node 1.0.0-rc.3 with one volume; outputs below are verbatim. The manual metadata deletion in step 2 stands in for whatever removed the metadata in production (context below) — the point is that once the state exists, no supported interface can see or clear it.

1. Start a single-node rustfs 1.0.0-rc.3 with one volume. Create a bucket and upload a few objects; on disk each object directory holds `xl.meta` plus one `<uuid>/part.1` data directory.
2. Stop the server. Delete every `xl.meta` under the bucket on the filesystem, leaving the data directories and their part files in place. Start the server.
3. The bucket is now empty over S3:

```
$ rc ls repro/orphan-test --recursive
(no output)

$ rc ls repro/orphan-test --recursive --versions
(no output)
```

4. Deleting it fails, and the reason given does not match what is happening — nothing is writing to this bucket:

```
$ rc rb repro/orphan-test
Failed stage: bucket_delete
✗ Bucket 'orphan-test' is not empty after verification; a concurrent writer may have added new state
```

5. Heal with `--remove` sees nothing — no findings, and the part files remain on disk. `--remove` is documented as removing dangling objects and parts, but it appears not to consider data directories with no metadata to be dangling, presumably because the enumeration starts from `xl.meta`:

```
$ rc admin heal start repro -b orphan-test --remove
✓ Heal operation started successfully.

$ rc admin heal status repro
Heal Status: Idle
  No active heal operation.
```

6. Delete the object directories directly on the filesystem — the bucket now deletes immediately, so the data directories with no `xl.meta` are demonstrably what `bucket_delete` refuses on:

```
$ rc rb repro/orphan-test
Completed stages: bucket_delete
✓ Bucket 'orphan-test' removed successfully.
```

**Expected behavior**

[...]

PR #6591, opened 2026-08-25
fix(ecstore): handle metadata-less bucket residue

houseme:
## Related Issues
Fixes rustfs/backlog#2019
Refs rustfs/rustfs#6575

## Summary of Changes
- Detect metadata-less on-disk residue during ordinary non-force DeleteBucket and return the existing S3-visible BucketNotEmpty contract before physical bucket deletion.
- Route scanner-discovered missing-xl.meta objects through a non-destructive heal request so scanner activity cannot remove uncommitted data-dir residue.
- Add explicit heal --remove handling for unversioned metadata-less data directories, using the existing data-dir delete primitive and fail-closed shape checks for offline disks, unknown entries, nested directories, and pre-commit delete markers.
- Keep MinIO-compatible API behavior by mapping the new internal diagnostic error back to BucketNotEmpty at the S3 boundary.

## Verification
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-ecstore metadata_less --lib`
- `cargo test -p rustfs-ecstore reclaim_orphan_data_dirs_aborts_when_meta_missing --lib`
- `cargo test -p rustfs-ecstore reclaim_orphan_data_dirs_recovers_committed_delete_marker_without_meta --lib`
- `cargo test -p rustfs-scanner test_build_non_destructive_object_heal_request_disables_removal --lib`
- `cargo test -p rustfs-scanner test_scan_folder_missing_xl_meta_stops_erasure_data_dir_descent --lib`
- `cargo test -p rustfs test_api_error_from_storage_error_mappings --lib`
- `cargo test -p rustfs-ecstore heal_rename_outcome_matrix_reports_partial_and_retries_failed_targets --lib`
- `cargo nextest run -p rustfs-ecstore set_disk::ops::heal::heal_result_report_tests::heal_rename_outcome_matrix_reports_partial_and_retries_failed_targets`
- `PATH=/Users/zhi/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin:$PATH make pre-pr` reached 7,746/14,749 tests before fail-fast stopped on `rustfs-ecstore set_disk::ops::heal::heal_result_report_tests::heal_rename_outcome_matrix_reports_partial_and_retries_failed_targets`; the failure happened in test setup before heal execution (`target shard should be removed before heal: No such file or directory`). The same test passed immediately afterward with both cargo test and nextest exact runs.

## Impact
- Ordinary DeleteBucket now treats raw metadata-less files as non-empty instead of attempting physical removal, preserving data until an explicit operator cleanup path is used.
- Explicit heal --remove can reclaim only narrow, validated unversioned metadata-less data directories containing positive `part.N` files and no unknown/pre-commit shapes.
- Scanner behavior becomes safer: metadata-missing objects remain diagnostic/heal candidates but are not auto-removed by scanner queueing.
- S3 clients continue to receive BucketNotEmpty for the affected DeleteBucket path.

## Additional Notes
- The implementation intentionally does not add a time-based grace window because the current DiskStore API does not expose a cheap stat-only data-dir probe; adding one would be a separate performance/safety follow-up.
- Compared with MinIO, this keeps the API-compatible empty-bucket contract while adding RustFS-specific fail-closed handling for the metadata-less residue shape reported in rustfs/rustfs#6575.

### DIFF
--- a/crates/ecstore/src/error/mod.rs
+++ b/crates/ecstore/src/error/mod.rs
@@ -119,6 +119,8 @@ pub enum StorageError {
     BucketExists(String),
     #[error("Bucket not empty: {0}")]
     BucketNotEmpty(String),
+    #[error("Bucket not empty: {bucket} ({details})")]
+    BucketNotEmptyWithDetails { bucket: String, details: String },
     #[error("Bucket name invalid: {0}")]
     BucketNameInvalid(String),
 
@@ -493,6 +495,10 @@ impl Clone for StorageError {
             StorageError::MethodNotAllowed => StorageError::MethodNotAllowed,
             StorageError::BucketNotFound(a) => StorageError::BucketNotFound(a.clone()),
             StorageError::BucketNotEmpty(a) => StorageError::BucketNotEmpty(a.clone()),
+            StorageError::BucketNotEmptyWithDetails { bucket, details } => StorageError::BucketNotEmptyWithDetails {
+                bucket: bucket.clone(),
+                details: details.clone(),
+            },
             StorageError::BucketNameInvalid(a) => StorageError::BucketNameInvalid(a.clone()),
             StorageError::ObjectNameInvalid(a, b) => StorageError::ObjectNameInvalid(a.clone(), b.clone()),
             StorageError::BucketExists(a) => StorageError::BucketExists(a.clone()),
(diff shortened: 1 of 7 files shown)


houseme:
Follow-up commit c606ff7ad fixes the Linux CI compile error in `rustfs/src/connect/offline/bundle_writer.rs`.

The issue was `Option::is_some_and` receiving a function item with signature `fn(&[serde_json::Value]) -> bool`, while `serde_json::Value::as_array()` yields `Option<&Vec<serde_json::Value>>`. Wrapping the call in a closure lets the argument coerce to a slice at the call site.

Additional verification:
- `cargo fmt --all --check`
- `git diff --check`

Linux cross-check was attempted with `cargo check -p rustfs --lib --target x86_64-unknown-linux-gnu`, but the local macOS environment lacks `x86_64-linux-gnu-gcc`, so it cannot complete dependency build locally. The source-level type mismatch is addressed by the closure.

houseme:
Follow-up commit 98aa4422b fixes the `clippy::redundant_clone` errors in `rustfs/src/connect/offline/bundle_writer.rs`.

The two temporary path values are now moved into their swap closures after `swapped_output` is derived, while the `moved` path remains cloned because it is still used by assertions after the closure.

Additional verification:
- `cargo fmt --all --check`
- `git diff --check`

PR #6596, opened 2026-08-25
feat(admin): self-service account management and TOTP two-factor authentication

sineld:
## Related Issues

N/A — no tracking issue exists yet. Happy to open one and link it if maintainers prefer.

## Summary of Changes

Adds a self-service account surface and TOTP two-factor authentication for RustFS identities, with the second factor enforced on the interactive login path.

**Endpoints**

| Route | Gate |
| --- | --- |
| `GET /v3/account/info` | possession of the credential |
| `POST /v3/account/password` | credential + knowledge of the current secret |
| `GET /v3/account/mfa` | possession of the credential |
| `POST /v3/account/mfa/enroll` | credential; credential kind must be able to manage a factor |
| `POST /v3/account/mfa/activate` | credential + a code from the pending secret |
| `POST /v3/account/mfa/disable` | credential + a valid code **and** the account password |
| `POST /v3/account/mfa/recovery-codes` | credential + a valid code |
| `GET /v3/mfa/challenge` | possession of the credential |
| `GET /v3/user/mfa` | `admin:GetUser` |
| `DELETE /v3/user/mfa` | `admin:EnableUser` (break-glass) |
| `PUT /v3/set-user-secret-key` | `admin:CreateUser` |

`AssumeRole` gains STS's own `SerialNumber`/`TokenCode` parameters, so an SDK or script presents a factor without a RustFS-specific protocol.

**What this deliberately does not do**, because the alternatives are worse:

- **A request signed directly with a long-term access key stays ungated.** Gating it would break every script and CLI the moment a human enabled 2FA on their own account, and would add no protection — whoever holds the secret key already has full access without presenting a code. This is the division AWS draws. Making 2FA meaningful for direct API access needs an `aws:MultiFactorAuthPresent` policy condition, which I've left as a follow-up.
- **Root credentials are not made mutable.** The root secret key signs every STS session token, derives the internode RPC secret, and decrypts MinIO-migrated IAM blobs. Rotating it at runtime would invalidate every session cluster-wide and break node-to-node auth. `account/info` reports `credentials_source: "env"` and `mutable.password: false`, and the docs recommend a `consoleAdmin` IAM user for day-to-day administration. Root *can* enroll a second factor — that split is deliberate and tested.
- **Username changes are not supported for any identity.** The access key is the primary key for policy mappings, group membership, service-account parents and bucket-policy principals; a rename orphans service accounts and silently breaks bucket-policy ARNs. `mutable.username` is `false` everywhere.
- **Secret keys are not hashed.** SigV4 requires the server to know the secret, so this is not possible in any S3 implementation. At-rest encryption plus a length floor is what protects them; there is deliberately no maximum length.

**Two decisions worth a reviewer's attention:**

1. **Enrollment is refused without `RUSTFS_IAM_MASTER_KEY`** (`501` with the remedy in the message). A TOTP secret is credential-equivalent; one written in plaintext could be lifted off a disk, which is worse than no second factor because the user believes they have one. IAM identities tolerate a missing master key for backward compatibility — a new feature has no such history to honour. If you'd rather this degrade to plaintext for parity, say so and I'll change it.
2. **Two `EventName` variants, not one per operation.** `mask()` gives every variant its own bit in a `u64` and the budget is nearly spent — **63 of 64 used after this PR**. Per-operation detail lives in `api.name` and the `iamOperation` tag. Splitting further needs `mask()` widened first; flagging it because the next feature to add an event will hit the wall.

Storage is one encrypted object per identity under `.rustfs.sys/config/mfa/`, a sibling of `config/iam/` so the IAM cache loader's startup walk doesn't sweep it up. Writes are optimistic `If-Match` with bounded retries; records are deliberately uncached, because a cache would need cluster-wide invalidation to keep the replay mark and lockout counter honest. Login challenges are stateless HMAC tokens so a cluster without session affinity can complete the exchange.

Full rationale, limitations and the rate-limit/replay table: [`docs/operations/two-factor-auth.md`](docs/operations/two-factor-auth.md).

## Merge order

**This PR lands first.** The console and CLI PRs call endpoints that do not exist
until it merges.

```
1.  rustfs/rustfs#6596     this PR                ← must land first
2a. rustfs/console#208     console UI             ← after 1
2b. rustfs/cli#358         rc admin commands      ← after 1
```

2a and 2b are independent of each other. Nothing here depends on them, so this can
merge on its own schedule; they are kept as drafts until it does.

## Verification

```
cargo fmt --all -- --check                                          # clean
cargo clippy -p rustfs -p rustfs-iam -p rustfs-madmin \
             -p rustfs-s3-types -p e2e_test --all-targets           # clean
cargo nextest run -p rustfs -p rustfs-iam -p rustfs-madmin \
                  -p rustfs-s3-types                                # 4416 passed, 0 failed
cargo nextest run -p e2e_test -E 'test(admin_mfa_test)'             # 4 passed
```

Note for anyone reproducing locally: plain `cargo test -p rustfs --lib` reports ~69 failures because many tests initialize process-global `OnceLock` state and collide as threads in one process. `cargo nextest run` (process-per-test, as `.config/nextest.toml` and `ARCHITECTURE.md` prescribe) is green.

New coverage: 85 unit tests in `crates/iam/src/mfa/` (including the RFC 6238 Appendix B vectors), 11 in `caller_identity`, and 4 e2e tests.

The e2e suite earned its place during development — it caught enrollment being refused for root because its *password* is immutable, which would have left the default deployment (an administrator signing into the console as root) unable to protect the one login the feature exists for.

## Impact

- **Additive.** An identity with no enrollment takes no new code path in `AssumeRole`; existing deployments are unchanged.
- **New dependencies:** `qrcode` (server-side QR so neither client needs an encoder) and `data-encoding` (already in the lock graph transitively). No new TOTP crate — RFC 6238 is implemented over the workspace's existing `hmac`/`sha1`.
- **New env var (optional but required for 2FA):** `RUSTFS_IAM_MASTER_KEY` must be set for enrollment. It already exists for IAM at-rest encryption.
- **Console/CLI:** unlocks the corresponding PRs in `rustfs/console` and `rustfs/cli`, which cannot work until this merges.
- **`idp_compat.rs`** now delegates its byte-identical `json_response` to a shared helper; no behaviour change.

## Additional Notes

Deliberately left as follow-ups rather than smuggled in here:

- `aws:MultiFactorAuthPresent` condition key, fed from the `x-rustfs-mfa-verified` session claim. This is what would make 2FA meaningful for direct API access.
- Mutable root credentials, which needs the three couplings above untangled first — notably a dedicated STS signing key (GHSA-m77q-r63m-pj89).
- A rename/migration operation for access keys, if that's wanted at all.

GHSA-m77q-r63m-pj89 is unaffected: a holder of the root secret can still forge a session token, `mfa` claim included. Stated in the docs so nobody reads this PR as closing it.

**CLA:** not signed by me. This project requires the contributor to comment the CLA acceptance line on the PR themselves; that is a legal assertion I am not in a position to make on the author's behalf.


sineld:

mertefecerit:
+1

trueWD:
+1

PR #6597, opened 2026-08-25
perf(sse): drop the second KMS decrypt from encrypted GET responses

reatang:
## Related Issues
Part of the SSE/KMS P0 plan (rustfs/backlog#2023, item P0-2; audit context in rustfs/backlog#2022).

## Summary of Changes
An SSE-encrypted GET performed two KMS `Decrypt` calls per request: the object layer's encryption resolver unwraps the data-key envelope to build the decrypted stream, and the S3 layer then called `sse_decryption` again purely to derive response headers — discarding the returned key bytes.

This PR replaces that second call with `classify_sse_read_response` (rustfs/src/storage/sse.rs), which reproduces the replaced call's behavior from stored metadata alone, item by item: SSE-C request-key validation with the same errors and precedence, per-key `kms:Decrypt` authorization ahead of every other failure mode (so a denied caller still learns nothing about the key), and the same KMS audit-summary fields (`sseType`/`kmsKeyId`/`kmsKeyVersion`/`kmsOutcome`/`kmsErrorClass`) on the request's audit entry. The recorded success outcome stays honest because a failed unwrap aborts the read inside the object layer before response classification is ever reached.

Net effect: SSE-KMS GET goes from 2 backend `Decrypt` round trips to 1, halving KMS load on read-heavy encrypted workloads, with response headers, error semantics, and audit tags unchanged.

## Verification
- `cargo test -p rustfs --lib storage::sse` — 114 passed, including 4 new tests: header parity against the unwrap-based path, audit-tag parity for allowed and denied principals, SSE-C validation parity, and a proof that classification succeeds with the DEK provider torn down (zero KMS round trips).
- `cargo test -p e2e_test --lib kms::bucket_default_encryption_test` (5 passed), `kms::encryption_metadata_test` (3), `kms::copy_object_self_copy_sse_test` (3), `kms::kms_authorization_negative_matrix_test` (2, covers the moved authorization point), `kms::configured_roundtrip_test::test_configured_local_kms_admin_and_versioned_cleanup` (1).
- `make pre-commit` equivalent: all steps green; the `test-wiring-check` step was run as `uv run --python 3.12 python ./scripts/check_test_wiring.py` because the shell's default Python 3.9 lacks `tomllib` (environment issue, not introduced here).
- `cargo clippy -p rustfs --lib --tests` — clean.
- Not verified here: `kms::configured_roundtrip_test::test_configured_vault_kms_admin_and_versioned_cleanup` fails on this machine because no `vault` binary is on PATH; it fails identically on `main` (environmental).

## Impact
No API or wire-format change. Response headers for SSE-C/SSE-S3/SSE-KMS GET are byte-identical by construction and covered by parity tests. One behavioral narrowing: transient KMS failures occurring between stream construction and response setup can no longer fail the request at the header-derivation step, because that step no longer contacts KMS — strictly fewer failure modes. `DecryptionMaterial.algorithm` is now `#[allow(unused)]` (still constructed for multipart session recovery callers).

## Additional Notes
`sse_decryption` remains in place for its legitimate key-recovering callers (multipart session recovery, the object-layer resolver). Follow-up P0 items from rustfs/backlog#2023 land separately.


reatang:

PR #6598, opened 2026-08-25
perf(ecstore): enable encrypted range part-seek by default

reatang:
## Related Issues
Part of the SSE/KMS P0 plan (rustfs/backlog#2023, item P0-3 Phase A); the underlying read-amplification defect is rustfs/backlog#1316.

## Summary of Changes
Range GETs on encrypted objects in the default build read the whole ciphertext from offset 0 and discard the decrypted prefix, because the part-boundary seek shipped behind `RUSTFS_ENCRYPTED_RANGE_SEEK` defaulted to `false`. This PR flips `DEFAULT_RUSTFS_ENCRYPTED_RANGE_SEEK` to `true`.

Safety rests on the marker chain shipped with the seek, where every misfit direction falls back to the conservative full read: MPUs created without a candidate layout marker never become seek-eligible; CompleteMultipartUpload promotes the candidate to the quorum marker only after revalidating it against the object's `data_dir` under the uploadId write lock; and reads seek only when the quorum marker matches the current `data_dir`. Objects that are single-part, compressed, or predate the marker keep the full-read path unchanged. `RUSTFS_ENCRYPTED_RANGE_SEEK=false` remains as the kill switch.

The stale default-off regression test is rewritten as `test_legacy_range_seek_defaults_enabled`: the unset-env default must produce the same part-aligned plan as the explicit opt-in, seek past the leading parts, and not span the whole ciphertext.

## Verification
- `cargo test -p rustfs-ecstore --lib object_api::readers` — 53 passed (covers seek-eligibility gates, marker validation, kill switch, byte-exact range matrix, tamper fail-hard).
- `cargo test -p rustfs-ecstore --lib set_disk::ops::multipart` — 71 passed (candidate/quorum marker write and promotion paths).
- `cargo nextest run -p rustfs-ecstore` — two local full runs each hit one unrelated flaky failure under load (`tier_free_version_recovery_continues_after_deleted_marker_bucket`, then `heal_writer_failures_emit_one_aggregate_warning_per_object`); both pass in isolation and neither touches encryption or range planning. CI's clean runners are the authoritative full-suite gate.
- `cargo test -p e2e_test --lib kms::bucket_default_encryption_test` and `kms::encryption_metadata_test` with a freshly built `target/debug/rustfs` — multipart SSE objects served end-to-end under the new default.

## Impact
Deployment-visible behavior change: eligible encrypted multipart range reads now seek to the covering part boundary instead of reading the full object, which shows up as `rustfs_io_get_encrypted_range_read_amplification` dropping toward 1 on the `part_seek` path. No wire-format or metadata change — markers were already written whenever the flag was on, and remain quorum-validated. Operators who need the old behavior set `RUSTFS_ENCRYPTED_RANGE_SEEK=false`.

Remaining from backlog#1316 and tracked in rustfs/backlog#2023: single-part objects structurally cannot seek in the v1 frame format (variable-length frames); that lands with the fixed-frame v2 format PR. The 4-node regression and kill-switch drill called for by backlog#1316 before/around this default change are an operational step for the release owner.

## Additional Notes
The `RUSTFS_COMPAT_TODO(backlog-1316)` comment is replaced with the rationale for why default-on is safe under the marker chain.


reatang:

PR #6599, opened 2026-08-25
feat(kms): seal persisted config secrets with RUSTFS_KMS_CONFIG_SECRET

reatang:
## Related Issues
Part of the SSE/KMS P0 plan (rustfs/backlog#2023, item P0-1; audit context in rustfs/backlog#2022).

## Summary of Changes
The dynamic-configuration flow persisted `KmsConfig` to cluster storage as raw JSON, so inline authentication material — the Vault token, an AppRole `secret_id`, the Local backend master key — landed in `config/kms_config.json` in cleartext, readable by anyone with access to the backing store.

This PR adds `rustfs_kms::config_secret`: when the per-node environment variable `RUSTFS_KMS_CONFIG_SECRET` is set, those field values are individually sealed in place before persistence — Argon2id (same parameters as the Local key store) derives an AES-256-GCM key from the operator secret and a per-value random salt, and the field's logical label (aligned with `KMS_CONFIG_REDACTION_RULES`) is bound as AEAD associated data so sealed values cannot be swapped between fields. Sealed values carry the versioned prefix `RUSTFS-KMS-ENC[v1]:`.

Backward compatibility is warn-only by design (owner decision): with the secret unset, saves keep the plaintext format byte-for-byte semantics and emit a `kms_config_secret_unset` warning naming the exposed fields; plaintext values load forever, and a node that gains the secret reseals on its next save. Sealed values fail closed — a missing or wrong secret, or a tampered payload, is a load error that leaves KMS unconfigured, never a silent fall-back to plaintext. The sealing secret must be an independent trust root: reusing the Local master key or Static secret is refused, mirroring the backup-KEK rule.

## Verification
- `cargo test -p rustfs-kms --lib config_secret` — 9 tests: seal/open round trip, warn-only no-op without a secret, fail-closed on missing/wrong secret, AAD prevents cross-field ciphertext swaps, plaintext documents pass through and are reported, the legacy `Vault` enum alias still seals, empty inline AppRole `secret_id` (file-based auth) is not counted as an exposure, and trust-root independence.
- `cargo test -p rustfs --lib admin::handlers::kms_dynamic` — handler-level round trip through `seal_persisted_kms_config` / `open_persisted_kms_config` / `decode_persisted_kms_config`, including the refusal to reuse a backend secret; pre-existing beta.5 migration tests stay green.
- `cargo clippy -p rustfs-kms -p rustfs --lib --tests` — clean.
- `make pre-commit` equivalent green (`test-wiring-check` run via `uv run --python 3.12` because the shell's default Python 3.9 lacks `tomllib`).

## Impact
Opt-in security hardening; no behavior change until `RUSTFS_KMS_CONFIG_SECRET` is set. Operational notes: the secret must be identical on every node that loads the persisted config — a node without it fails closed with an error naming the variable; documentation guidance remains to prefer file-based Vault auth (`secret_id_file`, `TokenFile`, Kubernetes `jwt_path`), which never persists inline secrets at all. Static backend configs remain non-persistable, unchanged. Residual risk stated plainly: deployments that keep inline secrets and never set the variable stay exposed exactly as before, with a warning at every save and load.

## Additional Notes
Sealing is field-level, so the persisted document keeps its shape and every non-secret field stays readable by older loaders; a rolling upgrade should complete before the secret is configured, since old nodes cannot unseal. Cluster status surfacing of the per-node seal state (`config_secret_state`) is deliberately left to a follow-up to keep the wire-shape snapshot churn separate.


reatang:

PR #6600, opened 2026-08-25
feat(rio): authenticated fixed-frame v2 encryption layout behind a write switch

reatang:
## Related Issues
Part of the SSE/KMS P0 plan (rustfs/backlog#2023, item P0-4; audit context in rustfs/backlog#2022). Groundwork for the single-part range seek (P0-3 Phase B).

## Summary of Changes
The default-build encryption stream format (legacy rio v1) authenticates only each frame's ciphertext: the 8-byte frame header (length + plaintext CRC32) and the end marker sit outside the AEAD, frames carry no position binding, and nothing marks the last frame — so header rewrites, frame reordering and truncation of trailing frames are not cryptographically detected.

This PR adds a v2 frame layout to the same format family, byte-layout compatible and dispatched per frame by the type byte:
- **Authenticated headers and positions**: the 8-byte header plus the frame's index are bound as AEAD associated data (`type 0x01`), so header tampering, reordering and cross-position splicing fail authentication.
- **Authenticated stream end**: the final frame carries its own type byte (`0x02`) inside the AAD; a clean EOF or an end marker before the final frame is now an error, and every stream — including the empty one — ends in an authenticated final frame. A v2 multipart stream additionally fails if it ends before all listed part segments were read.
- **Fixed-length frames**: the writer accumulates a full 8 KiB block before emitting, so every non-final frame costs exactly 8218 ciphertext bytes and single-part objects gain a closed-form plaintext-to-ciphertext offset mapping (consumed by the follow-up seek PR).

Key hierarchy, nonce derivation, DEK envelopes and object metadata are unchanged. v1 objects remain readable forever (the reader keeps the v1 path, including its historical nonce fallbacks, which v2 frames deliberately reject). Unknown frame types are now a hard error instead of being parsed as v1 data.

**Write side is off by default** (`RUSTFS_ENCRYPTION_FRAME_V2`, default `false`): mixed-version clusters cannot read v2 frames, and encrypted ciphertext travels verbatim through ILM transition, decommission and SSE-C replication passthrough. This release ships read support everywhere; the write default flips in a following release once every node (and every RustFS warm/replication target receiving raw ciphertext) can read v2.

## Verification
- `cargo test -p rustfs-rio --lib encrypt_reader` — 21 passed: 15 pre-existing v1 tests unchanged, plus 5 new v2 suites covering round trips at every block-boundary length (0, 1, ±1 around 8 KiB, multiples), fixed-frame byte-exact stream shapes, a tamper matrix (header CRC flip, final-flag rewrite, frame swap, truncation, forged early end marker, unknown type, forged zero-length frame — each fails; the untampered control still decrypts), mixed-version-segment rejection, multipart round trip plus missing-tail-part detection, and the block-aligned-part regression test.
- `cargo nextest run -p rustfs-ecstore` with the write-switch plumbing in place — two local full runs each hit one unrelated flaky failure under load (tier free-version recovery, then a heal warning-aggregation test); both pass in isolation and neither touches encryption. CI's clean runners are the authoritative full-suite gate. The switch itself is exercised via the `RUSTFS_ENCRYPTION_FRAME_V2=true` e2e below.
- e2e with `RUSTFS_ENCRYPTION_FRAME_V2=true` against a freshly built binary: `kms::bucket_default_encryption_test` (5, includes a 5 MiB + 1 MiB multipart whose block-aligned first part regressed an earlier draft — an authenticated empty final frame surfaced as a zero-byte read, i.e. EOF, dropping the second part; fixed, with a dedicated rio regression test), `kms::encryption_metadata_test` (3), `kms::copy_object_self_copy_sse_test` (3). Control run with the switch off: 5 passed (v1 behavior unchanged).

## Impact
No change until the switch is enabled. Once enabled, newly written encrypted objects use v2 frames: readable only by nodes at this release or later. Rollout contract (documented on the switch): complete the cluster upgrade — including RustFS warm tier targets and SSE-C replication targets that receive raw ciphertext — before enabling. Rollback: disabling the switch stops new v2 writes; existing v2 objects remain readable by any v2-capable node.

## Additional Notes
The follow-up PR stamps the `fixed-8k-v2` layout marker and adds the closed-form single-part range seek; this PR intentionally contains only the format, its authentication semantics and the switch, so the crypto change reviews in isolation.


reatang:

PR #6601, opened 2026-08-25
feat(ecstore): closed-form range seek for single-part v2 encrypted objects

reatang:
## Related Issues
Part of the SSE/KMS P0 plan (rustfs/backlog#2023, item P0-3 Phase B; read-amplification defect rustfs/backlog#1316). Stacked on the v2 frame format PR.

## Summary of Changes
Single-part encrypted objects in the legacy format structurally could not serve range reads without decrypting from byte 0: v1 frames are emitted per upstream read, so their lengths vary and no closed-form plaintext-to-ciphertext mapping exists. The v2 format fixed the frame length (8218 ciphertext bytes per 8 KiB plaintext frame), making the mapping closed-form.

This PR consumes that:
- Single-part PUTs that encrypt locally under the v2 write switch stamp the internal marker `encrypted-frame-layout-fixed8k-v1`, **valued with the object's `data_dir` token** (the same self-invalidation scheme as the multipart layout markers): ciphertext passthrough and data movement never stamp it, and any path that re-homes metadata onto a new object identity mints a new `data_dir`, disqualifying a carried-over marker automatically.
- The encrypted read plan gains a v2 branch: an uncompressed single-part object with the marker seeks to `frame_index * 8218`, decrypts from that frame (the frame index rides the plan's sequence-number slot into the decrypt reader, whose AAD binds absolute frame indices), skips `offset % 8192` plaintext bytes and serves exactly the requested range. The read amplification metric gains the `frame_seek` path label.
- A lying marker fails closed: seek math landing outside the object, or frames that fail v2 authentication at the seek target, surface as read errors — never as plaintext from the wrong offset.
- No strip-list or replication transport-table changes are needed: the `data_dir` binding makes a re-homed marker self-disqualifying. SSE-C passthrough replicas therefore fall back to the conservative full read (honest, slower) rather than trusting a marker minted for another object identity.

Compressed objects keep the conservative full read (their offsets address the decompressed stream; combining the S2 index with frame seek is follow-up work). Multipart objects keep the part-boundary seek from the earlier default-flip PR; sub-part v2 seek inside parts is likewise follow-up.

## Verification
- `cargo test -p rustfs-rio --lib encrypt_reader` — 22 passed, including the new mid-stream decrypt-at-frame test: a window starting at frame 2 decrypts with the matching absolute index, and the same window with a wrong index fails authentication.
- `cargo nextest run -p rustfs-ecstore -E 'test(/object_api::readers/)'` — 57 passed, including 4 new: closed-form offset math at every boundary (head, crossing, aligned, tail, out-of-bounds), a byte-exact range matrix through real v2 ciphertext with plan-offset assertions, marker mismatch/absence falling back to the conservative full read, and a lying marker over short-frame v1 ciphertext failing closed at read time.
- `cargo nextest run -p rustfs-ecstore` (targeted) — `put_object_stamps_the_v2_frame_layout_marker_only_when_enabled` proves the marker is stamped with the object's `data_dir` only under the write switch and never on plaintext objects.
- New e2e `kms::encrypted_range_get_test` (registered in the kms module): SSE-S3 single-part PUT then a ranged-GET matrix (head, boundary-crossing, aligned, mid-frame, tail, suffix, full body) compared byte-for-byte — run against freshly built binaries both with `RUSTFS_ENCRYPTION_FRAME_V2=true` (frame-seek path) and unset (v1 conservative path); `kms::bucket_default_encryption_test` re-run green under the switch.
- Module-filtered full runs of `set_disk::ops::multipart` flake randomly on this loaded machine with and without this diff (each failing test passes in isolation; a stashed baseline reproduces it identically) — CI's clean runners are the authoritative gate.

## Impact
With the v2 switch on, single-part encrypted range reads stop reading the whole object — `rustfs_io_get_encrypted_range_read_amplification` drops toward 1 on the new `frame_seek` path. Objects written before the switch (v1 frames, no marker) keep the full-read path unchanged. No new configuration: seekability follows the v2 write switch.

## Additional Notes
The marker is advisory for planning only; correctness never rests on it — decryption authenticates every frame against its absolute index regardless of how the plan positioned the read.

## Adversarial Review
Adversarial validation ran on the final stacked diff with the correctness, security, compatibility and concurrency/streaming lenses. One CONFIRMED finding, fixed in this PR: a range reaching into the last plaintext-bearing frame planned a storage window that excluded the authenticated final frame, so drain-to-plaintext-end reads of block-aligned objects failed with a spurious "truncated before its final frame" error; the window now extends to the ciphertext end whenever the range touches the last plaintext frame, with unit (offset math + aligned-tail matrix) and e2e (aligned-object ranges) regressions. Null verdicts, with the attacks named: v1/v2 type-byte relabeling and final-flag rewrites fail AEAD authentication (type byte and index are AAD); cross-position and cross-part frame splices fail (index in AAD, part number in the nonce); forged empty streams are caught by app-layer length enforcement exactly as with v1; client-side marker forgery is impossible because every put strips incoming markers and the value must equal the server-minted `data_dir`; mixed-version MPUs (a v1 part next to a v2 part) decode per segment; the rio-v2 feature build is untouched.

### DIFF
--- a/crates/e2e_test/src/kms/encrypted_range_get_test.rs
+++ b/crates/e2e_test/src/kms/encrypted_range_get_test.rs
@@ -1,0 +1,140 @@
+// Copyright 2024 RustFS Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Ranged GETs over encrypted single-part objects.
+//!
+//! Byte-exactness must hold on every frame layout the server can write:
+//! legacy v1 (variable frames, conservative full read) and, when
+//! `RUSTFS_ENCRYPTION_FRAME_V2=true` reaches the server under test, the
+//! fixed-frame v2 layout whose marker enables the closed-form frame seek.
+//! The matrix crosses frame boundaries, starts mid-frame, and ends inside
+//! the final short frame, so a mispositioned seek cannot pass.
+
+use super::common::LocalKMSTestEnvironment;
+use crate::common::{TEST_BUCKET, init_logging};
+use aws_sdk_s3::primitives::ByteStream;
+use aws_sdk_s3::types::ServerSideEncryption;
+use tracing::info;
+
+const FRAME_PLAINTEXT: usize = 8 * 1024;
+
+#[tokio::test]
+async fn sse_s3_single_part_ranged_gets_are_byte_exact() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    init_logging();
+    info!("Testing ranged GETs over an SSE-S3 single-part object");
+
+    let mut kms_env = LocalKMSTestEnvironment::new().await?;
+    let _default_key_id = kms_env.start_rustfs_for_local_kms().await?;
+    kms_env.wait_for_kms_ready().await?;
+
+    let s3_client = kms_env.base_env.create_s3_client();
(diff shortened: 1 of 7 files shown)


reatang:

PR #6671, opened 2026-08-26
fix(tests): cfg-gate the connect inventory fixture uid for non-Linux builds

overtrue:
## Related Issues

N/A (follow-up to the CI fixture repairs around #6617/#6620; found while running the full local gate for #6670).

## Summary of Changes

`INVENTORY_UID` in `rustfs/tests/connect_registration.rs` is only referenced by the inventory heartbeat tests, which are `#[cfg(target_os = "linux")]` like the `InventorySnapshot` imports above it. On macOS `cargo clippy --all-targets -- -D warnings` (and any `-D warnings` build of that test target) therefore fails with a dead_code error on the constant. This gates the constant with the same `#[cfg(target_os = "linux")]` the inventory import already carries.

## Verification

- `cargo check -p rustfs --test connect_registration` on macOS — the dead_code warning is gone.
- Linux builds are unaffected: the constant and all its uses remain compiled together.

## Impact

Test-only, non-Linux developer-experience fix. N/A for runtime behavior.

## Additional Notes

N/A.


overtrue:
Superseded: main already fixed this via #6641 (08d16d067) while this was in flight.

PR #6725, opened 2026-08-27
test(ecstore): accept quorum-persisted fencing epochs

### LINKED ISSUE #6711: flaky test: multipart fencing epoch read-back assumes all-disk persistence
## Symptom

`rustfs-ecstore set_disk::ops::multipart::tests::object_transaction_fencing_persists_epoch_on_multipart_commit` failed non-deterministically on the `Test and Lint (rio-v2)` leg of #6704 (run 33029849943, 2026-08-27):

```
panicked at crates/ecstore/src/set_disk/ops/multipart.rs:3702:39:
disk 1 should persist object metadata: file not found
```

The PR is `.config/nextest.toml`-only and cannot affect this test; the same test passed on the rio-v2 legs of #6696 and #6700 the same day.

## Suspected cause

The `object_transaction_epochs` helper (multipart.rs:3696) reads `xl.meta` back from **every** disk and panics if any single disk has not persisted — but a multipart commit is quorum-based, so a lagging disk can legally still be missing its metadata when the helper reads it. Same failure class as rustfs#6703 (fixture assumes all-disk materialization where the product only guarantees quorum). The rio-v2 lane's different IO scheduling plausibly widens the timing window, but the assumption is wrong under rio-v1 too.

One open question worth checking while fixing: whether the rio-v2 path makes the post-commit visibility window materially wider than v1 — if so this is also useful signal for the rio-v2 lane (backlog#1316), which is why the quarantine keeps the junit `flaky` marker visible rather than hiding the failure.

## Affected suite

`package(rustfs-ecstore)` lib tests; observed on the rio-v2 leg, reachable from every `Test and Lint` leg.

## Quarantine

Added in PR #6704 (`.config/nextest.toml`, ci profile, `retries = 2`) linking this issue, per docs/testing/README.md flake policy. 30-day fix-or-delete window applies. A robust fix should read back only quorum-many disks (or poll each disk with a bounded readiness wait) before asserting epochs.

overtrue:
## Related issue

Fixes #6711

## Background

The multipart fencing test read `xl.meta` from every disk immediately after a quorum-based commit. A legal lagging disk could therefore fail the fixture even though the committed object and epoch were durable on write quorum.

## Fix

The test helper now ignores only missing object metadata, still fails on other disk errors, and requires write-quorum epoch coverage. The regression deterministically removes one disk metadata file before checking that the remaining quorum carries one non-nil epoch.

## Validation

- `make pre-commit`
- `cargo clippy --workspace -- -D warnings`
- `cargo test -p rustfs-ecstore set_disk::ops::multipart::tests::object_transaction_fencing` (3 passed)
- `cargo test --workspace` reached 606 passed but failed 15 unrelated e2e cases, including unavailable Vault KMS, WebDAV 500s, Snowball path-limit assertions, and cluster startup fixtures

### DIFF
--- a/crates/ecstore/src/set_disk/ops/multipart.rs
+++ b/crates/ecstore/src/set_disk/ops/multipart.rs
@@ -3693,19 +3693,30 @@ mod tests {
         }
     }
 
-    async fn object_transaction_epochs(disks: &[DiskStore], bucket: &str, object: &str) -> Vec<Option<Uuid>> {
+    async fn object_transaction_epochs(
+        disks: &[DiskStore],
+        bucket: &str,
+        object: &str,
+        write_quorum: usize,
+    ) -> Vec<Option<Uuid>> {
         let mut epochs = Vec::with_capacity(disks.len());
         for (disk_index, disk) in disks.iter().enumerate() {
-            let file_info = disk
-                .read_version("", bucket, object, "", &ReadOptions::default())
-                .await
-                .unwrap_or_else(|err| panic!("disk {disk_index} should persist object metadata: {err}"));
+            let file_info = match disk.read_version("", bucket, object, "", &ReadOptions::default()).await {
+                Ok(file_info) => file_info,
+                Err(DiskError::FileNotFound | DiskError::FileVersionNotFound) => continue,
+                Err(err) => panic!("disk {disk_index} should read object metadata: {err}"),
+            };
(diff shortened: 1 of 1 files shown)


overtrue:
Superseded by #6723, which fixes #6711 with a deterministic rename-tail regression while preserving the stronger all-disk epoch assertion.

PR #6733, opened 2026-08-27
test(heal): add node-outage heal E2E script and workflow

majinghe:
## What

Node-outage heal test for the **3x4 cluster** (3 nodes × 4 disks, one `RUSTFS_VOLUMES` expression on every node: `http://rustfs-node{1...3}:9000/data/rustfs{1...4}/mnmd`), following the Obsidian "RustFS Heal 测试步骤" note and reusing the pool-expansion test's admin-API approach (SigV4, no `rc`).

- `scripts/test/rustfs_heal_test.sh` — environment setup (purge + recreate dirs), install & simultaneous 3-node start, warp write with a mid-write node outage, restart, cluster heal via `POST /rustfs/admin/v3/heal/`, monitoring via the heal task (`clientToken`), result analysis and cleanup.
- `.github/workflows/rustfs-heal-test.yml` — workflow_dispatch on the `smoke-testing` runner, nightly deb by default, thresholds as inputs, log artifacts, reset before/after.
- `scripts/test/rustfs_heal_test.md` — README.

## Validation (live, nightly deb, test cluster)

```
used before heal: vm000=40GB vm001=40GB vm002=16GB   (vm002 stopped mid-write)
heal summary=finished
per-node disk usage: vm000=40GB vm001=40GB vm002=40GB
heal test PASSED: cluster heal complete, 0 failed, vm002 reached 40GB
```

## Notes / caveats

- The script writes `RUSTFS_HEAL_TASK_TIMEOUT_SECS` (default 6h) into the node config: the server default (5 min) is far too short for healing tens of GiB — a cluster heal task times out at 5 minutes otherwise.
- The cluster-level `background-heal/status` aggregator returns 501 in this single-pool topology (no notification system), so the script monitors the started heal task via `POST /rustfs/admin/v3/heal/?clientToken=<token>` instead; `summary` must reach a terminal success and the outage node's disk must reach the target.


cxymds:
## Proposed modifications

I reviewed PR #6733 against backlog #2035 (heal E2E) and #2031 (pool expansion), together with the current main behavior. The rebuild path can complete, but the test still has paths that may report zero progress or pass without proving that the intended failure scenario occurred. I suggest the following scope before merge.

### Must-fix in this PR (#6733)

1. **Make the E2E assertions fail closed.** Validate the expected 3-node × 4-disk, single-pool topology before writing data. For the foreground Heal scenario, disable the background scanner (or record and exclude background repairs), require the outage threshold and inactive service to be observed, and require the write target to be reached. A missing outage or an incomplete write should fail the test, not emit a warning. After restart, replace `|| true` with a bounded readiness retry and fail if the cluster does not become active.

2. **Fix Heal progress parsing and semantics.** The Heal API serializes progress fields in camelCase (`objectsFailed`, `objectsHealed`, `objectsScanned`, `progressPercentage`), while the script currently reads snake_case names in both monitoring and final analysis. Read the canonical names (optionally retaining a clearly documented legacy fallback), preserve the raw response, and distinguish `null`/missing progress from numeric zero instead of coercing it to zero. If an active task still returns `progress: null` with the scanner disabled, retain that evidence and open a server-side follow-up rather than masking it in the test.

3. **Make the workflows deterministic and non-overlapping.** Fix the actionlint SC2054 findings by quoting comma-separated Warp arguments (for example, `"1,2"`). Use one concurrency group for the Heal and pool-expansion workflows because both mutate the same `vm000`/`vm001`/`vm002` environment, and keep `cancel-in-progress: false` so a cancellation cannot leave a node down. For `workflow_run`, require an exact successful conclusion, pin checkout/package selection to the triggering SHA or an explicitly identified artifact, and normalize boolean inputs for both dispatch modes.

4. **Align documentation and retain bounded operational safeguards.** The script header/docs currently refer to a 3-pool configuration and background-heal status, while this test uses one pool and polls the Heal task endpoint (`POST /rustfs/admin/v3/heal/?clientToken=...`). Update the wording and the known-issues note accordingly. Keep the 21,600-second test override documented for the 40 GiB workload (the product default remains five minutes) and retain bounded retries for transient Heal-start coordination failures, while failing fast on authentication or invalid-request errors and logging the final response.

### Scope boundary

The core #2031 cold-start/rebalance fix is already present on main via PR #6720, so #6733 should add smoke verification and avoid duplicating that runtime change. I would keep the default timeout and notification fallback behavior out of this test/CI PR unless the scope is intentionally expanded.

### Follow-up if #2035 is reproduced

If a raw active-task response still has no progress after the test-side fixes, handle that in a separate runtime change: preserve the last progress snapshot for active, retrying, and completed tasks; add manager/channel/admin integration coverage; and make the POST background-heal status endpoint return an explicit degraded/local snapshot when notification state is unavailable. The Heal start path should remain fail-closed.

### Validation / acceptance

- `bash -n scripts/test/rustfs_heal_test.sh`
- `shellcheck -S error scripts/test/rustfs_heal_test.sh`
- `actionlint -config-file .github/actionlint.yaml .github/workflows/rustfs-heal-test.yml .github/workflows/rustfs-pool-expand-test.yml`
- `./scripts/test/rustfs_heal_test.sh --all --dry-run`
- A live 3-node × 4-disk run proves the outage, write threshold, restart/readiness, Heal-start retry, observable progress (or an explicitly captured runtime defect), terminal success, zero failures, and artifact/cleanup behavior.
- A 2→3 pool run proves all nodes become active, rebalance starts and completes, and no `init retry budget exhausted` failure occurs.

**Recommended merge shape:** keep #6733 test/CI focused, and track any runtime progress/status fallback as a separate follow-up. If this scope is acceptable, it can be implemented in that order.

majinghe:
Thanks for the review — addressed in the latest commit (934f0de80):

1. **Fail closed**: step 3 now verifies the outage node's service is actually inactive after `systemctl stop` and fails if the write target (`warp_stop_gb`) or the outage threshold is not reached; step 4 uses a bounded readiness loop for the admin API to report an active pool after restart (no more `|| true`).
2. **Progress parsing**: confirmed the API serializes camelCase (`objectsScanned/objectsHealed/objectsFailed/progressPercentage`). The script now reads those with a snake_case fallback and distinguishes `null`/absent progress from numeric zero — null while running is logged as evidence (backlog #2035) instead of being coerced. Added a `heal_progress_field` helper.
3. **Determinism**: `RUSTFS_HEAL_AUTO_HEAL_ENABLE=false` is written into the node config so the background scanner cannot mask the outage; heal and pool workflows share one concurrency group; `workflow_run` now requires an exact `success` conclusion; checkout is pinned to the triggering SHA; comma-separated step args are quoted (SC2054).
4. **Docs**: header and README updated (single 3x4 pool, task-endpoint monitoring, 21,600s timeout override and auto-heal note in Known issues). Heal start fails fast on 400/403 and retries only transient server errors.

loverustfs (approved):

PR #6748, opened 2026-08-27
test(heal): relative disk target and fail fast on terminal-but-short

majinghe:
## Problem

The absolute 40 GiB heal target was calibrated to the environment's previous default where the **background scanner (auto-heal) is enabled**. That scanner is now disabled (`RUSTFS_HEAL_AUTO_HEAL_ENABLE=false`, per review, so the outage stays observable and the explicit heal is the only repair mechanism). With only the explicit heal, the recovered node lands at ~36 GiB for 40 GiB surviving nodes — so the fixed 40 GiB criterion never passes even though the heal completes cleanly (0 failures).

A second issue: when the task reached a terminal success but the disk target was not met, the monitor kept polling until the timeout instead of failing fast.

## Changes

- **Relative disk target (default)**: the outage node must reach at least **90% of the least-used surviving node** after heal. `heal_target_gb` becomes an optional absolute floor (default `0` = relative only), so the criterion scales with the write thresholds.
- **Fail fast**: if the heal task reaches a terminal success (`finished`/`completed`/...) but the disk target is not met, the script fails immediately with the numbers and the final status JSON (no more polling until timeout).
- Drop the misleading "progress absent" warning on the final task response: the server clears progress after completion; mid-run progress is reported correctly (validated: `scanned/healed` climb, e.g. 27 → 633, in lockstep with the outage node's disk usage).

## Validation (live, nightly package)

```
heal: summary=finished ... vm002_used=40GB (target 36GB)
per-node disk usage: vm000=40GB vm001=40GB vm002=40GB
heal test PASSED: cluster heal complete, 0 failed, vm002 reached 40GB (target 36GB)
```

### DIFF
--- a/.github/workflows/rustfs-heal-test.yml
+++ b/.github/workflows/rustfs-heal-test.yml
@@ -16,9 +16,9 @@ on:
         required: false
         default: '40'
       heal_target_gb:
-        description: 'Outage node must reach N GiB after heal to pass'
+        description: 'Absolute disk floor (GiB) for heal success; 0 = relative (>=90% of the least-used surviving node)'
         required: false
-        default: '40'
+        default: '0'
       cleanup_before:
         description: 'Reset the nodes before the test (DESTROYS existing data/config)'
         type: boolean
--- a/.github/workflows/rustfs-heal-test.yml
+++ b/.github/workflows/rustfs-heal-test.yml
@@ -15,10 +15,6 @@ on:
         description: 'Stop warp when surviving nodes reach N GiB'
         required: false
         default: '40'
-      heal_target_gb:
-        description: 'Absolute disk floor (GiB) for heal success; 0 = relative (>=90% of the least-used surviving node)'
-        required: false
-        default: '0'
       cleanup_before:
         description: 'Reset the nodes before the test (DESTROYS existing data/config)'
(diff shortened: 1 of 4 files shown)


majinghe:
Final approach (commit ac9fd1bec supersedes the relative-target gate in the earlier commit):

**Heal success = the server's own verdict + an end-to-end data read-back:**

1. The heal task reaches a terminal success (`finished`/`completed`) with `objectsFailed == 0` — the server's per-object EC scan/repair result (`items[]` per object, progress counters).
2. S3 read-back: list the test bucket and `GET` a sample of objects, requiring HTTP 200 for every read — proves the data is still reconstructable after the repair.

**Why the disk-usage GB gate was dropped:** with erasure coding each node holds *different* shards, so a fixed per-node GB number (40 GiB, or 90% of survivors) is not a code-level invariant — it depends on the EC layout, not on heal correctness (the same heal legitimately landed at 36–40 GiB across runs). Disk usage is now printed as observability (with a warning if the outage node gained no usage), not a pass/fail gate. The `heal_target_gb` input and relative-target logic were removed.

The object read-back uses a discard mode (status-code only) so binary bodies are never captured.

houseme (approved):

PR #6790, opened 2026-08-28
fix(site-replication): send the reverse-reachability probe as POST

reatang:
## Related Issues

Fixes rustfs/backlog#2072. Also addresses the error-detail truncation part of rustfs/backlog#2073 (F5).

## Summary of Changes

The reverse-reachability probe that runs on the joining site during `site-replication add`/`join` sent `PUT` to the devnull endpoint, while the admin router only registers `POST /rustfs/admin/v3/site-replication/devnull`. Every probe therefore failed with 501, and every add/join response carried a false `initialSyncErrorMessage: "... is not reachable from this site"`, drowning out the real network-asymmetry warnings the probe exists to surface. Verified on a real two-VM deployment: the message appears on 100% of add calls while replication is fully healthy.

- Add a `POST` constructor to `PeerAdminRequest` (it only had `PUT`/`GET`) and send the probe through a named `devnull_probe_request` helper.
- A new route-registration test builds the actual probe request and asserts its method/path pair is a registered admin route, so a probe/router method drift fails CI instead of shipping as permanent warning noise.
- `summarize_peer_error_detail` now keeps both the head and the tail of oversized details. Nested peer errors append the decisive status/code at the tail, and the previous prefix-only cut dropped exactly that part (the same truncation hid the 501 root cause during diagnosis).

## Verification

- `cargo test -p rustfs --lib site_replication` (255 passed)
- `cargo test -p rustfs --lib route_registration_test` (8 passed, including the new probe-route assertion)
- `cargo test -p rustfs --lib test_status_peer_error_summarizes` (updated for head+tail truncation)
- `cargo clippy -p rustfs --lib --tests` (clean)
- `make pre-commit` equivalent: fmt-check through quick-check all pass (test-wiring-check run via uv because the local system python lacks `tomllib`; the script itself passes)

## Impact

Operator-facing only: add/join responses stop reporting healthy peers as unreachable, and truncated peer errors retain their final status/code. No wire-format or API changes; the devnull endpoint itself is unchanged.

## Additional Notes

Found during full site-replication testing on real hardware (report and root-cause analysis in rustfs/backlog#2073).

### DIFF
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -4283,7 +4283,7 @@ async fn probe_reverse_peer_reachability(state: &SiteReplicationState, local_pee
                 continue;
             }
         };
-        if let Err(err) = PeerAdminRequest::put(&connection, SITE_REPLICATION_DEVNULL_PATH, &state.service_account_access_key)
+        if let Err(err) = devnull_probe_request(&connection, &state.service_account_access_key)
             .send(&secret_key, &serde_json::json!({}))
             .await
         {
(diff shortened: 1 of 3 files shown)


reatang:

PR #6791, opened 2026-08-28
fix(iam): stop stamping the wall clock on policy-less group reads

reatang:
## Related Issues

Fixes rustfs/backlog#2069. Unblocks the convergence path tracked in rustfs/backlog#2071.

## Summary of Changes

`build_group_desc` stamped a fresh `OffsetDateTime::now_utc()` on every read of a group that has no mapped policy, instead of reporting the group's stored `updatedAt` (which `GroupInfo` already carries). The value feeds content-addressed consumers: the site-replication repair plan hashes each IAM item into its task ids, and the task ids into the preflight token. One policy-less group therefore made every `repair` dry-run produce a different preflight token, so `execute` permanently failed with `412 preflight is stale` — leaving no working path to settle escalated retry entries, which the retry queue explicitly defers to repair.

Verified on a real two-VM deployment: with a policy-less group present, three consecutive dry-runs returned three different tokens and execute always failed; after removing the group, tokens were stable, execute succeeded across all four families, and the long-stuck escalated deletion entry was settled.

- Report the stored `GroupInfo.update_at` for policy-less groups; the mapped-policy branch is unchanged.
- Unit tests pin `build_group_desc` determinism for both branches.
- A repair-token determinism test over a policy-less group pins the end-to-end invariant: equal state must produce equal preflight tokens.

An audit of the other `build_sr_info` inputs (bucket metadata timestamps, users, policies, policy mappings) found no further wall-clock fallbacks on the read path.

## Verification

- `cargo test -p rustfs-iam --lib` (338 passed, including the two new `build_group_desc` tests and the pre-existing disabled-group test)
- `cargo test -p rustfs --lib test_site_replication_repair_preflight` (2 passed, including the new determinism test)
- `cargo clippy -p rustfs-iam --lib --tests` and `cargo clippy -p rustfs --lib --tests` (clean)
- fmt-check through quick-check all pass (test-wiring-check run via uv because the local system python lacks `tomllib`)

## Impact

`GroupDesc.updatedAt` for policy-less groups now reports the stored group timestamp (or is omitted when the legacy record has none) instead of the read time. This is the accurate value for admin `group` queries and makes site-replication repair usable again. No API shape changes.

## Additional Notes

Found during full site-replication testing on real hardware; the live bisection that localized the flapping task id to the policy-less group is documented in rustfs/backlog#2069.

### DIFF
--- a/crates/iam/src/manager.rs
+++ b/crates/iam/src/manager.rs
@@ -2320,9 +2320,13 @@ fn filter_policies_from_docs(policy_docs: &CacheEntity<PolicyDoc>, policy_name: 
 }
 
 fn build_group_desc(name: &str, group_info: GroupInfo, mapped_policy: Option<MappedPolicy>) -> GroupDesc {
+    // A group without a mapped policy must report the group's own stored
+    // timestamp, not the wall clock: this value feeds content-addressed
+    // consumers (the site-replication repair plan hashes it), and a fresh
+    // clock on every read makes equal states hash differently.
     let (policy, updated_at) = mapped_policy
         .map(|policy| (policy.policies, Some(policy.update_at)))
-        .unwrap_or_else(|| (String::new(), Some(OffsetDateTime::now_utc())));
+        .unwrap_or_else(|| (String::new(), group_info.update_at));
 
     GroupDesc {
         name: name.to_string(),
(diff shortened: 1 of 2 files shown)


reatang:

Issue #5963, opened 2026-08-11
Site replication silently stops replicating when peer states diverge — every client command still reports success


mrabbah:
**Affected versions:** `1.0.0-beta.11` and `1.0.0-rc.1` (both reproduced)

## Summary

When the two sites' persisted site-replication states diverge, the target site
rejects every peer operation with
`400 InvalidRequest — "site replication is not enabled"`, but **nothing surfaces
on the source side**: `mb` returns success, `admin replicate info <source>` still
lists both sites, and the console still shows *Enabled / 2 sites*. Replication is
dead while the whole control plane reports a healthy cluster.

## Reproduction (two containers, single host)

```bash
# site1 -> :30190, site2 -> :30290, same access/secret key
rc admin replicate add site1 site2          # works, replication verified

docker compose -f site1/... stop            # make site1 unreachable
rc admin replicate remove site2 --all       # returns 0: "removed successfully"
docker compose -f site1/... start

rc admin replicate info site1               # still shows a 2-site cluster
rc admin replicate info site2               # "Site replication is not configured"

rc mb site1/demo                            # "✓ created successfully"
rc ls site2                                 # demo is absent, forever
```

Target log:

```
PUT /minio/admin/v3/site-replication/peer/bucket-ops?bucket=demo&operation=make-with-versioning
  -> 400  S3Error(InvalidRequest, "site replication is not enabled")
```

`rc admin replicate status site1 --json` reports `HasBucket: false` for the peer
deployment, but no client command exits non-zero.

## Code pointers (`rustfs/src/admin/handlers/site_replication.rs`, tag 1.0.0-beta.11)

- L. 605 — `fn enabled(&self) -> bool { self.peers.len() > 1 }`; the target's
  local view alone decides whether peer ops are accepted.
- L. ~7018 (`SRPeerBucketOpsHandler`) —
  `if state.pending_remove.is_some() || (!state.enabled() && !bootstrap_peer_bucket_operation_allowed(...))`
  → the error above. Same guard at L. 8601 in `1.0.0-rc.1`.
- L. ~6899 (`SRPeerJoinHandler`) — when `join_req.updated_at <= state.updated_at`
  (or is absent), the join is a **silent no-op that returns HTTP 200**, so the
  initiator reports `configured successfully` although the peer applied nothing.
  Makes `add` sensitive to residual state and to clock skew between hosts.
- L. ~6887 `SRPeerJoinHandler` never resets `state.pending_remove`; only
[...]

PR #6793, opened 2026-08-28
fix(site-replication): rotate-svc-acct no longer wedges replication

reatang:
## Related Issues

Fixes rustfs/backlog#2068.

## Summary of Changes

A single `POST /rustfs/admin/v3/site-replication/rotate-svc-acct` call could permanently kill both replication directions, reproduced 100% (twice) on a real two-VM deployment: the rotating site installs the new site-replicator secret locally, but the join push to the peers could only sign with that new secret — so every peer rejected it with `SignatureDoesNotMatch`, the rotation sat in `pendingOperation` forever (>18 minutes observed, across two reconciler cycles), outbound pushes failed against peers still holding the old secret, inbound peer requests failed against the local new secret, and `remove`/`peer edit` were rejected while the rotation was pending.

Three defects lined up:

1. A fresh `PendingRotation` seeds `secret_candidates` only from the legacy plaintext state secret — empty on modern deployments.
2. The pre-rotation secret was recorded into persisted state, but the push read a stale clone of the pending record taken before that write.
3. The "current" secret was read after the local overwrite, so it was already the new one — the candidate list collapsed to `[new]`.
4. The reconcile tick bailed out entirely on `pending_rotation.is_some()`, so nothing ever retried (the resume the tick has for `pending_remove` since rustfs/rustfs#5963 had no rotation counterpart).

Fix, mirroring the pending-remove structure:

- Capture the still-installed secret before the local overwrite and sign the join push with `[previous, new]`, assembled by a pure `rotation_secret_candidates` helper (unit-tested for the fresh, resume, and unreadable-secret orders).
- Extract the push loop into `drive_pending_rotation`, shared by the handler and a new `resume_pending_rotation` called from the reconcile tick — a rotation that still wedges (peer down mid-rotation) now heals within one tick instead of never.
- New dual-node e2e test: one rotate call returns `success`, `pendingOperation` clears on both sites, and replication then flows in both directions. Red/green verified: the test fails in 15s against the unfixed handler and passes in ~14s with the fix.
- `scripts/test/site_replication_smoke.py` gains a `rotate` subcommand for the local lab (verified end to end locally).

Deliberately not done: rolling back the local secret when the push fails — a rollback racing the resume would reintroduce a split-brain window, and with the resume in place the worst case is bounded by one reconcile tick.

## Verification

- `cargo nextest run -p e2e_test -E 'test(test_site_replication_rotate_svc_acct_completes_and_replication_survives_real_dual_node)'` — PASS (14s); the same test FAILS against the unfixed handler (red/green check)
- `cargo test -p rustfs --lib site_replication` (254 passed, including the new `rotation_secret_candidates` unit test)
- `cargo clippy -p rustfs --lib --tests` and `cargo clippy -p e2e_test` (clean)
- `make fmt-check logging-guardrails-check unsafe-code-check architecture-migration-check` (pass); test-wiring check passes with the refreshed `e2e-repl-nightly` selection digest (57 tests)
- `scripts/test/site_replication_smoke.py up && rotate && clean` on alternate ports — rotation settles and bidirectional smoke passes

## Impact

Operational fix only: rotation completes in one call on healthy clusters and self-heals within one reconcile tick otherwise. The rotate response, wire paths, and peer join format are unchanged. Existing wedged deployments recover automatically after upgrading (the resume drives the persisted pending rotation forward).

## Additional Notes

Root cause, real-hardware timeline, and the interim operator workaround (calling rotate a second time) are documented in rustfs/backlog#2068; the full test campaign is in rustfs/backlog#2073.


reatang:

PR #6821, opened 2026-08-28
fix(kms): restore persisted configuration after restart

### LINKED ISSUE #6574: KMS configured via the admin API is silently lost on restart: startup load runs before storage is initialized
**Describe the bug**

KMS configured through the admin API works and is persisted to cluster storage, but it is not restored when the server restarts. After a restart the service reports `state: not-configured` with no error, while the persisted configuration is still on disk and intact. The startup load runs before the object store is available, takes an early return, and nothing retries it.

The effect is that KMS silently turns itself off on every restart, and there is no supported way to load the persisted configuration back other than re-submitting the whole thing — including secrets that the persisted copy deliberately does not expose.

**To Reproduce**

1. Configure KMS via `POST /rustfs/admin/v3/kms/configure` on a multi-node cluster. It takes effect immediately and is persisted (`config/kms_config.json`, present on disk with all erasure shards, unchanged since the day of configuration).
2. Restart the nodes.
3. KMS now reports not-configured, with no error surfaced:

```
$ rc admin kms status <alias> --json
{
  "data": {
    "operation": "status",
    "state": "not-configured",
    "backend": null,
    "healthy": null,
    "error_message": null,
    "config": null
  }
}
```

4. `kms start` does not help, because start operates on the in-memory service manager rather than reading the persisted configuration:

```
$ rc admin kms start <alias>
✗ Failed to start KMS: KMS start request was rejected
```

5. The startup log explains it. This line appears once per boot, on every boot:

```
{"level":"WARN","message":"admin kms dynamic state","component":"admin","subsystem":"kms",
 "event":"kms_config_load_skipped","reason":"storage_uninitialized","result":"config_load_skipped",
 "target":"rustfs::admin::handlers::kms_dynamic","filename":"rustfs/src/admin/handlers/kms_dynamic.rs"}
```

**What the code suggests**

`rustfs/src/init.rs` calls `admin::handlers::kms_dynamic::load_kms_config()` during startup to restore the persisted configuration. That function begins by resolving an object store handle and returns `None` if it cannot:

```rust
let Some(store) = current_object_store_handle_for_context(context.as_deref()) else {
    warn!(… event = "kms_config_load_skipped", reason = "storage_uninitialized",
          result = "config_load_skipped", …);
    return None;
};
```

On our cluster that branch is taken every time, so the persisted configuration is never read and KMS is left unconfigured. The ordering appears to be the whole bug, and it looks deterministic rather than a race: on current `main` (b15928220), `init_startup_runtime_services` (`rustfs/src/startup_services.rs`) calls `init_kms_system(config)` as its very first step, while the app context / object-store handle that `load_kms_config()` resolves is published later in the same function (by `ensure_startup_after_iam` inside `init_iam_runtime`, per the comment on the audit init just below it). The configuration is fine, the loader is fine — it simply runs too early, every time.

Two things make it harder to notice and harder to recover from than it needs to be:

1. The failure is not reflected anywhere an operator would look. `kms status` returns `not-configured` with `error_message: null`, which is indistinguishable from never having configured KMS at all. Only the startup log distinguishes them, and only if the operator knows to look for that event.
2. There is no way to ask a running server to re-read the persisted configuration. `reload_persisted_kms_config()` exists and does exactly the right thing, but it is only reachable through the peer reload RPC that a configure/reconfigure request broadcasts (`rustfs/src/storage/rpc/node_service.rs`). Recovering therefore means re-submitting the full configuration, including secrets, which the persisted copy deliberately does not expose. An operator who no longer has the original JSON cannot restore a configuration that is sitting on their own disks.

**Expected behavior**

1. The persisted KMS configuration should be loaded after storage is initialized, so a restart restores the configured state.
2. If the load is skipped or fails, `kms status` should say so rather than reporting `not-configured` with a null error, since the two states call for very different operator responses.
3. There should be a supported way to re-read the persisted configuration on a running server — an admin endpoint mapping to the existing `reload_persisted_kms_config()` would be enough.

**Environment**

- 4 nodes, erasure coded, Linux (bare metal, systemd)
- Version: 1.0.0-rc.3; the early-return ordering is still present on current `main` by inspection
- KMS configured through the admin API (not command-line flags)

**Additional context**

Impact: any restart silently disables KMS cluster-wide. Buckets with SSE-KMS default encryption then reject writes, and the failure surfaces to clients as a request error rather than as anything pointing at KMS. Nothing in the status output indicates that a configuration exists but was not loaded.

reatang:
## Related Issues

Fixes #6574.

Architecture follow-up: https://github.com/rustfs/backlog/issues/2064.

Unrelated test flakes observed during local preflight: https://github.com/rustfs/backlog/issues/2081 and https://github.com/rustfs/backlog/issues/2082.

## Summary of Changes

- Load the persisted dynamic KMS configuration from the explicitly initialized `ECStore` instead of depending on an `AppContext` that is published later in startup.
- Distinguish an absent persisted configuration from storage, unseal, and decode failures, and expose initialization failures through KMS status without replacing an already usable service.
- Add authenticated `POST /rustfs/admin/v3/kms/reload` service control, reuse the persisted-config reload path, broadcast reloads to peers, and cover the route policy, audit mapping, registration matrix, and contract snapshot.
- Add an end-to-end regression test that configures local KMS through the admin API, restarts RustFS with the same data, verifies the restored status, and completes an SSE-KMS object round trip.

## Verification

- `./scripts/check_logging_guardrails.sh` — passed.
- `cargo test -p rustfs-kms initialization_error --lib -- --nocapture` — passed, 2 tests.
- `cargo test -p rustfs persisted_config_loader --lib -- --nocapture` — passed, 2 tests.
- `cargo test -p rustfs kms_admin_route_contract_is_stable --lib` — passed.
- `cargo test -p rustfs kms_reload_route_is_registered --lib` — passed.
- `cargo test -p rustfs test_register_routes_cover_representative_admin_paths --lib` — passed.
- `cargo test -p e2e_test test_admin_configured_local_kms_is_restored_after_restart -- --nocapture --test-threads=1` with localhost proxy bypass — passed, including restart and an SSE-KMS object round trip.
- `make pre-pr` — formatting, architecture, logging, security guards, offline enrollment E2E, Clippy, and script tests passed; workspace nextest was interrupted by the unrelated timeout tracked in rustfs/backlog#2081, whose exact nextest rerun passed.
- `make test` — script tests passed; workspace nextest was interrupted by the unrelated timeout tracked in rustfs/backlog#2082, whose exact nextest rerun passed.

## Impact

Admin-configured KMS now survives server restarts, and operators can request a reload of the persisted configuration without resubmitting redacted secrets. Failed persisted-config loads are distinguishable from a never-configured service. The persisted format and existing configuration endpoints remain unchanged; the reload endpoint and status error reporting are additive.

## Additional Notes

This PR is opened as a draft because the required local workspace nextest gate did not reach completion after two unrelated, non-deterministic timeout failures. Both failures passed in exact isolation and have dedicated backlog issues with the first-failure evidence.

High-risk review verdicts: correctness is covered by explicit-store startup wiring and the restart E2E; security requires `kms:ServiceControl` and does not expose config payloads or secrets; concurrency and durability retain lifecycle serialization and the existing best-effort peer convergence model; compatibility keeps the persisted format and ambient loader API unchanged; performance adds one bounded config read during startup and no hot-path work.

Rollback is a revert of this commit. No data migration or persisted-format rollback is required.


reatang:
Pushed a28ea8846: adds the missing `POST /kms/reload` row to the admin contract matrix (response reuses the configure snapshot shape, noted in the snapshot-coverage section), and adds a runbook procedure distinguishing a `load_failed` startup from a never-configured server, with the reload-based recovery path. Closes the two doc findings from the review; the multi-node cold-start retry note stays with the backlog#2064 follow-up.

PR #6828, opened 2026-08-28
fix(version): do not bump version when HEAD equals latest tag

cuishuang:
## Related Issues

N/A

## Summary of Changes

Resolve `HEAD` and the latest tag to their commit IDs before checking ancestry.

When both references point to the same commit, treat `HEAD` as not newer and preserve the current tag instead of incrementing its version.

Add a regression test using an annotated Git tag that covers both an equal `HEAD` and a strict descendant commit.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs --lib version::tests::`

## Impact

Builds performed exactly at the latest tag now report that tag instead of incorrectly incrementing its version. Builds from commits after the tag retain the existing increment behavior.

No public API, configuration, or dependency changes.

## Additional Notes

N/A

### DIFF
--- a/rustfs/src/version.rs
+++ b/rustfs/src/version.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use shadow_rs::shadow;
+use std::path::Path;
 use std::process::Command;
 
 shadow!(build);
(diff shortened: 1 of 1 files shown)


cuishuang:

PR #6868, opened 2026-08-29
perf(ecstore): reuse local fd metadata snapshots

houseme:
## Related Issues

- rustfs/backlog#2060

## Summary of Changes

Reuse the validated part-file length alongside cached LocalDisk descriptors. StdBackend cache hits now skip a redundant metadata syscall, and the io_uring cache-hit path reuses the same immutable length snapshot while retaining zero-length and out-of-bounds checks. Generation fencing, TTL eviction, and heal/delete/rename invalidation remain unchanged, so a snapshot is never reused across a replaced inode.

## Verification

- `git diff --check`
- `cargo test -p rustfs-ecstore std_backend_truncate_append_stream_and_full_read_restore_bytes --lib`

## Impact

Linux LocalDisk buffered and io_uring read cache hits avoid repeated file metadata syscalls. Mutation, heal, delete, rename, zero-length, and out-of-bounds semantics remain fail-closed. No public API, wire-format, or on-disk format changes are introduced.

## Additional Notes

This is an independent P2 follow-up to the PUT eager-threshold and metric atomic-load PRs. Linux target cross-check was attempted but the local environment lacks `x86_64-linux-gnu-gcc` required by `aws-lc-sys`; the package-scoped native checks above passed. Rollback is a single revert of this commit.

---

### DIFF
--- a/crates/ecstore/src/disk/local.rs
+++ b/crates/ecstore/src/disk/local.rs
@@ -3148,9 +3148,10 @@ impl LocalIoBackend for StdBackend {
                 direct_read_copy_fault_delta: MmapPageFaultDelta,
                 blocking_task_duration: StdDuration,
                 used_direct_io: bool,
-                /// The descriptor opened by THIS call (None on a cache hit), handed
-                /// back so the async caller can index it in the fd cache.
-                opened_fd: Option<Arc<std::fs::File>>,
+                /// The descriptor and size snapshot opened by THIS call (None on a
+                /// cache hit), handed back so the async caller can index it in the
+                /// fd cache.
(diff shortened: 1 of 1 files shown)


houseme:
Fixed the compile failure reported by CI: `FdCache::insert_if_fresh` now inserts the renamed `entry` parameter instead of the stale `file` identifier. Revalidated with `cargo fmt --all --check`, `cargo check -p rustfs-ecstore --lib`, the focused LocalDisk regression test, and `cargo clippy -p rustfs-ecstore --lib --tests -- -D warnings`. Commit: `8952faca1`.

Issue #2060, opened 2026-03-03
Cannot modify the deafult password


exing567:
**Describe the bug**
Cannot change the admin password.

**Expected behavior**

**Screenshots**
<img width="1911" height="998" alt="Image" src="https://github.com/user-attachments/assets/1e00248b-bbf5-4a18-84a9-460b1d5e6eb6" />

**Device**
 - Device: ThinkPa
 - OS: Win 11
 - Browser: Edge
 - Version: N/A

**Additional context**
I use Linux Quick Install to install the RustFS on a VPS.

overtrue:
Thanks for reporting this.

This behavior is expected and **by design**.
The system account credentials are defined at service startup (for example via `RUSTFS_ACCESS_KEY` / `RUSTFS_SECRET_KEY`) and are not allowed to be modified through the Console/API.

If you want to change that account’s password, please update the startup credentials and restart RustFS.

reatang:
@exing567 
Has this answered your question? If you have any further questions, please submit a new issue. I will close this issue now. Finally, thank you for trying RustFS.

--- 

是否已经解释了您提的问题，如果还有其他问题，可提交新的issue，这个issue我就关闭了。
最后感谢您尝试使用RustFS。

PR #6869, opened 2026-08-29
perf(storage): keep 1MiB PUTs on eager path

houseme:
## Related Issues
Related: rustfs/backlog#2087

## Summary of Changes
Raise the default ordinary small-eager PUT boundary from 512 KiB to 1 MiB.
Known-length, unencrypted, uncompressed 1 MiB PUTs now use the eager ingestion path that materializes the request before EC storage processing; the existing `RUSTFS_SMALL_EAGER_PUT_MAX_SIZE_BYTES` override remains available for rollback and A/B testing.

The change is intentionally limited to the PUT path selector. EC encoding, bitrot, quorum, rename, conditional-write handling, encryption, compression, and extract requests retain their existing gates.

## Verification
- `CARGO_TARGET_DIR=<isolated-target> cargo test -p rustfs should_use_small_eager_put_path --lib` (7 passed)
- `cargo fmt --all --check`
- `git diff --check`
- Remote testing on Warp v1.6.1, testing `node1...4:29000`, 1 MiB PUT, `--disable-multipart`: c8/15s baseline 214.32 MiB/s, candidate 240.99 MiB/s; B2 274.24 MiB/s; A2 269.75 MiB/s. The short ABBA pair is noisy and does not prove a fixed percentage gain.
- A c64/60s attempt produced repeated service-unavailable responses before a valid report and was excluded from throughput conclusions.

## Impact
Ordinary 1 MiB PUTs allocate up to one request-sized eager buffer per in-flight request. The default memory footprint therefore increases with concurrency; operators can set `RUSTFS_SMALL_EAGER_PUT_MAX_SIZE_BYTES=524288` to restore the prior boundary. No API or on-disk format changes.

## Additional Notes
The runtime evidence supports the intended path change but is not a claim that every topology will reach the historical 459 MiB/s target. Further c64/5m validation should be repeated after the testing cluster's service-unavailable condition is resolved.

---

### DIFF
--- a/rustfs/src/app/object/put.rs
+++ b/rustfs/src/app/object/put.rs
@@ -25,15 +25,14 @@ const ENV_ZERO_COPY_EAGER_PUT_MAX_SIZE_BYTES: &str = "RUSTFS_ZERO_COPY_EAGER_PUT
 
 /// Maximum body size materialized by the ordinary eager PUT path.
 ///
-/// Bodies above this boundary stay streaming so a 1 MiB request does not
-/// reserve a full request-sized buffer while the EC writer is consuming it.
-/// The environment override keeps the boundary reversible for workload A/B
-/// tests and for deployments whose measured workload favors eager ingestion.
+/// Bodies above this boundary stay streaming while the EC writer consumes them.
+/// The environment override keeps the boundary reversible for workload A/B tests
+/// and for deployments whose measured workload favors streaming ingestion.
 const ENV_SMALL_EAGER_PUT_MAX_SIZE_BYTES: &str = "RUSTFS_SMALL_EAGER_PUT_MAX_SIZE_BYTES";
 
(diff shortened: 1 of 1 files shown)


houseme:
Adversarial review: no correctness, persistence, or compatibility blocker was found. Performance delivery remains gated: the current c64 run returned repeated service-unavailable responses, and the completed c8 measurements were noisy (B2/A2 +1.7%). Before merge, repeat a valid c64/5m comparison with readiness, error classification, RSS/BytesPool, and P99 capture; do not treat the historical 459 MiB/s target as restored yet.

houseme:
Superseded by main commit a22fa7461 (PR #6863), which keeps the 512 KiB baseline and adapts the eager ceiling to active PUT concurrency. The fixed 1 MiB default in this PR is not sufficient to explain or guarantee the 459 MiB/s target; backlog #2060 records the required c64/5m validation and commit-stage instrumentation.

PR #6892, opened 2026-08-30
perf(ecstore): consolidate non-inline read planning

houseme:
## Related Issues

Related to rustfs/backlog#1309 and rustfs/backlog#1308.

## Summary of Changes

Refactor the accumulated non-inline GET read-plan patches into one incremental quorum planner. Candidate shard coverage is tracked by an allocation-free mapped bitmask, with one reconstruction reserve retained before metadata early-stop. Reserve scheduling is activated only after a candidate is classified as eligible; healthy inline and gate-off paths keep their existing fanout behavior. A short pending-data hedge delay races a genuinely slow data shard without turning every healthy request into a full fanout. The published environment variable remains compatible, while internal naming and comments now describe the actual non-inline quorum behavior.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-ecstore non_inline_data_read_early_stop --lib -- --nocapture` — 5 passed.
- `cargo test -p rustfs-ecstore metadata_slowtail_fault --lib -- --nocapture` — 6 passed, including initial-slow and deferred-tail hedge cases.
- `make pre-pr` guards, builds, fuzz, wiring, and offline-enrollment stages passed; nextest stopped on two existing sandbox TLS bind failures. The two named TLS tests passed when rerun individually with network permissions.
- Two fresh independent correctness/concurrency and performance/simplicity reviews of the final head — PASS.

## Impact

With `RUSTFS_GET_METADATA_TWO_PHASE_READ_PLAN_ENABLE` unset or disabled, the existing path is preserved and no shard-accounting work is performed. When enabled, only plain, single-part, non-inline candidates may stop after a mapped quorum plus one reserve; unsafe, transformed, remote, inline, multipart, conflicting, or malformed candidates retain full-wait/fallback behavior. No wire or on-disk format changes are introduced.

## Additional Notes

This revision removes the prior per-request erasure construction, shared Reed-Solomon cache access, temporary `Vec<bool>` allocation, and repeated metadata scans. It uses a 100ms delayed hedge for any sole pending shard that could stall the reserve, skips offline spare slots, and cancels/drains the remaining tasks on early success. The gate remains disabled by default pending broader rollout evidence.

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](https://github.com/rustfs/rustfs/blob/main/CODE_OF_CONDUCT.md)). If this is your first contribution, review the [CLA document](https://github.com/rustfs/cla/blob/main/cla/v2.md) and sign it by commenting `I have read and agree to the CLA.` on the PR.

### DIFF
--- a/crates/ecstore/src/set_disk/core/io_primitives.rs
+++ b/crates/ecstore/src/set_disk/core/io_primitives.rs
@@ -3159,7 +3159,6 @@ impl SetDisks {
                     _ = tokio::time::sleep_until(deadline) => {
                         if bounded_fanout
                             && !force_full_wait
-                            && !defer_pending_inline_data_shard
                             && join_set.len() == 1
                             && non_inline_candidate_eligible == Some(true)
                             && !accumulator.candidate_has_read_reserve()
--- a/crates/ecstore/src/set_disk/core/io_primitives.rs
+++ b/crates/ecstore/src/set_disk/core/io_primitives.rs
@@ -3304,6 +3304,16 @@ impl SetDisks {
             }
 
             let pending_responses = join_set.len();
+            // Inline verification can still depend on a missing data shard;
+            // issue one immediate spare when only that shard remains. The
+            // non-inline path keeps its delayed hedge below to avoid healthy
+            // reads paying speculative I/O before the candidate is classified.
+            let should_hedge_single_pending_inline_read = read_data
+                && !force_full_wait
+                && !defer_pending_inline_data_shard
+                && pending_responses == 1
+                && non_inline_candidate_eligible != Some(true)
+                && accumulator.can_still_reach_early_stop_with_pending(pending_responses);
             // A non-inline plan must retain one extra matching shard as a
             // reconstruction reserve. Schedule that reserve only after the
             // candidate is known to be eligible, so inline GETs do not pay an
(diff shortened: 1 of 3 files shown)


houseme:
Follow-up fix pushed as `92b063baf3`.

The 10-second barrier failures came from delaying the single-pending hedge until a candidate was classified as non-inline. Inline or not-yet-classified candidates can still require a spare to complete pending data-shard verification. The fix keeps the 100 ms delayed hedge for confirmed non-inline reserve reads, while immediately scheduling one spare only for the pending inline/unknown case that can still reach early-stop quorum.

Validation on the new head: the six previously failing bounded metadata/inline early-stop tests pass; six slow-tail/non-inline/quorum/inline path tests pass; `cargo fmt --all --check`, `git diff --check`, and `cargo clippy -p rustfs-ecstore --all-targets -- -D warnings` pass. No test timeout or assertion was relaxed.

houseme:
Follow-up cleanup pushed as `59687da734`: removed a redundant `defer_pending_inline_data_shard` check from the delayed-hedge timer branch (the variable is freshly initialized before that select branch and cannot be set there). This keeps the state machine unchanged while making the hot path explicit.

After the cleanup, 12 focused early-stop/slow-tail/non-inline/inline tests pass, together with `cargo fmt --all --check`, `git diff --check`, and the ECStore all-target Clippy gate.

PR #6903, opened 2026-08-30
fix(ecstore): fail closed on unverifiable data quorum

houseme:
## Related Issues

Fixes rustfs/backlog#1308.

## Summary of Changes

Reject verification-based erasure decoding when a data shard is missing and only the exact data-shard count remains available. This closes a fail-open path where `decode_data_with_reconstruction_verification` could fall back to unverified decoding and return bytes without an error. The data-shards-only gate remains disabled by default; this change hardens the opt-in path without changing healthy or gate-off behavior.

## Verification

- `cargo test -p rustfs-ecstore ...` focused verification and gate-on 2+2 regression: passed.
- Lockstep and demand-bound suites: 6/6 and 7/7 passed.
- Erasure decoder suite: 31/31 passed.
- `cargo fmt --all --check`
- `git diff --check`

## Impact

Opt-in data-shards-only reads now fail closed instead of returning unverified data when verification cannot be performed. No default configuration is enabled and no wire or on-disk format changes are introduced.

## Additional Notes

The 4-node Linux fault-injection and rollout decision remain tracked by rustfs/backlog#1325; this PR does not claim production rollout readiness.

---


houseme:
Follow-up fix after merging the latest `main` (`37b23a16d`):

The first implementation rejected exact data-shard quorum in `Erasure::decode_data_with_reconstruction_verification` globally. That correctly protected the opt-in data-shards-only lockstep path, but incorrectly changed default legacy/prefetch/decode-reader and transition rollback behavior. Commit `5bf1ac7853` scopes the surplus-source requirement to the lockstep policy; default reconstruction semantics remain unchanged.

Post-fix validation: the seven affected decode/lockstep regressions pass, the test-util decommission scenario passes, the transition rollback scenario passes, `cargo fmt --all --check` and `git diff --check` pass, and `cargo clippy -p rustfs-ecstore --all-targets -- -D warnings` passes.

houseme:
Follow-up fix pushed after the reported rio-v2 failure:

The failure was test-fixture setup, not the decoder change. The heal regression removed `part.1` through a data directory copied from one source disk, which is not guaranteed to be the current directory on every target disk under concurrent/rio-v2 layouts. Commit `219163e8fe` resolves the actual object directory on each target disk and removes its current `part.1`; production heal behavior and assertions are unchanged.

Validation after the fix: default heal regression passed, rio-v2 heal regression passed, `cargo fmt --all --check`, `git diff --check`, and `cargo clippy -p rustfs-ecstore --all-targets -- -D warnings` passed. The prior merged-main decode scope fix remains in `5bf1ac7853`.

PR #6927, opened 2026-08-31
fix(ecstore): recover late parity after exact quorum

houseme:
## Related Issues

Related to rustfs/backlog#1309.

## Summary of Changes

Recover omitted parity when the opt-in two-phase non-inline GET plan reaches exact decode quorum after a selected data shard fails during part open or bitrot validation. The initial materialization stays private, decode reports whether it used exact quorum, and the caller performs at most one full metadata refresh before exposing a reader. The fallback validates a compact canonical identity and uses the refreshed `FileInfo`, avoiding a deep clone on healthy requests.

The late-materialization path remains limited to whole latest plain single-part objects larger than 512 KiB that fit one EC block. Codec-streaming rollout is mutually exclusive at metadata fanout entry, so codec requests keep full metadata fanout and do not enter late refresh.

## Verification

- 4+2 single selected-shard bitrot and missing-part recovery: passed with exact body.
- 4+2 quorum-minus-one: passed with fail-closed/no body exposure.
- 2+2 late-parity bitrot/missing regressions: passed.
- Real 1 MiB codec mixed-feature GET: codec path selected, exact body, four metadata reads, no late refresh.
- Non-inline early-stop suite: 5/5 passed.
- Metadata slow-tail suite: 6/6 passed.
- Decode-focused suite: 141/141 passed.
- `cargo fmt --all --check`
- `git diff --check`

## Impact

Only the opt-in two-phase non-inline read plan gains a single fail-closed full-fanout retry after exact-quorum reconstruction. Gate-off, inline, codec-streaming, transformed, multipart, range, explicit-version, and remote-object paths preserve their existing behavior. No wire or on-disk format changes are introduced.

## Additional Notes

Two independent high-risk reviews covered correctness/concurrency and performance/compatibility/test coverage. The rollout gate remains disabled by default; testing1 multi-node evidence is still required before enabling it.

---


houseme:
Follow-up for the two failing Test and Lint lanes is now pushed in `eb66f5c918`.

The failures had two PR-attributable causes:

- The branch merge had dropped the existing 1 KiB test storage-class override from `get_object_reader_wires_mid_size_to_single_inflight`. Its 256 KiB EC2+2 fixture therefore sat exactly on the default inline boundary and correctly selected inline path 3. The explicit non-inline fixture contract is restored.
- The new late-materialization branch enlarged the `SetDisks::get_object_reader` poll frame enough to cross the debug test stack boundary. At a 1,984 KiB stack, the base passed, the unboxed PR head aborted, and the boxed late branch passed. The full refresh plus retry is now boxed only after `has_late_metadata_fanout()` is true, so gate-off and healthy paths do not allocate it.

Post-merge validation on the latest `main`:

- all three previously aborting `store::init` tests passed at `RUST_MIN_STACK=2031616`;
- the mid-size SingleInFlight wiring test passed;
- late-parity bitrot and missing-part regressions passed;
- the real codec mutual-exclusion regression passed;
- `cargo clippy -p rustfs-ecstore --lib --tests --features test-util -- -D warnings` passed;
- `cargo fmt --all --check` and `git diff --check` passed.

Independent final reviews found no correctness, cancellation/lock-lifetime, compatibility, test-coverage, or healthy-path performance issue. The new CI run for this head is pending; this update does not claim it is green yet.

Issue #6174, opened 2026-08-17
perf(ecstore): severe ListObjectsV2 regression after beta.8 on nested prefixes


Joduus:
## Summary

`ListObjectsV2` delimiter listings became dramatically slower between `1.0.0-beta.8` and `1.0.0-beta.12` on the same dataset, storage and client.

```text
beta.8:        ~0.48 s
beta.12 warm:  ~7.2 s
beta.12 cold:  ~183 s
```

The listed prefix contains 633 immediate child prefixes with additional nested directories/files below them.

Direct enumeration of the backing ZFS directory takes only ~16–24 ms, so the parent directory itself is not the bottleneck.

## Reproduction

```bash
time rclone lsf remote:bucket/prefix --max-depth 1 >/dev/null
```

Environment:
- TrueNAS / ZFS on HDDs
- rclone S3 backend
- RustFS beta.8 vs beta.12, unchanged data/storage

On beta.12, a warm ~7 s listing produced roughly **125k demand-metadata accesses** according to `arcstat`, despite returning only 633 immediate prefixes. Cold runs then amplify this into many random HDD metadata reads.

A synthetic nested-tree `rclone copy` also regressed from **<1 minute on beta.8** to **>1 hour on beta.12**. I have not isolated raw PUT latency, so this may be caused by repeated destination listing/existence checks.

## Suspected regression

This appears to line up with #4296 / commit `2585855d23fe334f5f5fa04ae55da8967b7bec2c`.

That change replaced the previous cheap handling of non-empty physical prefix directories with a call to `directory_has_visible_listing_entry()` for non-recursive listings. That helper recursively walks child directories and reads/parses `xl.meta` until it can prove a visible object exists.

Later releases renamed/hardened this path as `directory_has_listing_entry()`, but the recursive subtree probe remains.

This matches the observed behavior: a level with one immediate prefix is fast, while a level with hundreds of immediate prefixes becomes expensive because each prefix may trigger its own subtree visibility probe.

## Expected behavior

A delimiter listing of hundreds of immediate CommonPrefixes should not require recursively traversing descendant metadata for every prefix.

The correctness goal of #4296 is valid, but the implementation seems to cause severe I/O amplification. A fast path for cases where delete-marker visibility cannot matter, cached/indexed namespace visibility, or another bounded approach would avoid this regression.

I can provide `arcstat`, `iostat`, request logs, or run targeted A/B tests on specific commits if useful.

PR #6974, opened 2026-08-31
fix(ecstore): reclaim stale object prefixes

### LINKED ISSUE #6898: Stale / orphaned objects after upgrade to 1.0.0-rc.4
I am testing OpenObserve, using RustFS as the object storage.  During testing, I was trying to figure out why objects that should have been deleted are not getting deleted.  At the moment, I think I have now removed OpenObserve from the equation and both through the RustFS UI as well as CLI, been able to show that objects show up when running 1.0.0-rc.4 and are NOT removable.  However, when downgrading to 1.0.0-rc.3, the objects do not appear.  Switching between the two RustFS versions results in the same behavior.

If I try to delete the objects on rc.4, the call looks successful, but the objects simply remain.

In the screenshots attached, you'll notice that the `metrics` path doesn't even exist on rc3, but when updating to rc4, it's there with some objects.

Image

Image

overtrue:
## Related Issues

Fixes #6898.

## Summary of Changes

Reclaim stale directory prefixes after an exact empty recursive `max-keys=1` listing proves that a prefix has no live objects. Cleanup is limited to authoritative never-versioned buckets, preserves the bounded non-recursive listing fast path from #6174, and removes only empty directories or individually classified files in UUID data directories carrying a valid committed delete marker.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo +1.97.1 clippy -p rustfs-ecstore --lib --ignore-rust-version -- -D warnings`
- `cargo +1.97.1 test -p rustfs-ecstore orphan_dir_entries_must_be_single_relative_components --lib --ignore-rust-version`
- `cargo +1.97.1 test -p rustfs-ecstore orphan_cleanup_preserves_object_published_after_scan --lib --ignore-rust-version`
- `cargo +1.97.1 test -p rustfs-ecstore orphan_cleanup_keeps_commit_marker_when_residue_delete_fails --lib --ignore-rust-version`
- `cargo +1.97.1 test -p rustfs-ecstore purge_orphan_dir_object_ --lib --ignore-rust-version`
- `cargo +1.97.1 test -p rustfs-ecstore empty_directory_listing_purge_requires_complete_exact_recursive_request --lib --ignore-rust-version`
- `cargo +1.97.1 test -p rustfs-ecstore empty_recursive_listing_purges_committed_delete_residue --lib --ignore-rust-version`
- `cargo +1.97.1 test -p rustfs-ecstore test_scan_dir_nonrecursive_fast_path_preserves_probe_bound --lib --ignore-rust-version`
- `make pre-pr` completed every pre-test gate before the rebase; the test phase hit an unrelated host-capacity failure in the rebalance test.
- `cargo nextest run --all --exclude e2e_test --no-fail-fast`: 15,742 passed, 3 failed before the rebase. Both non-capacity failures passed in isolation; the rebalance test consistently requires about 12.3 TB while the host has about 1.81 TB available.
- `cargo test -p rustfs-ecstore --features test-util prepared_snapshot_transition_duplicate_and_late_get_use_committed_remote_generation --lib -- --nocapture`
- `cargo nextest run -p rustfs-ecstore --features test-util decommission_entry_rejects_subquorum_free_version_conflict_and_retains_source --no-capture`

## Impact

Users can remove stale prefixes left by interrupted delete cleanup without reintroducing the large metadata-probe regression from #6174. Versioned, ambiguous, malformed, non-authoritative, or path-escaping disk state fails closed and retains the conservative behavior.

## Additional Notes

- Correctness: cleanup requires a complete empty exact-prefix result and rescans every online disk before deletion.
- Security: malformed markers, `xl.meta`, uncommitted data, and disk/RPC entries that are not single relative path components veto cleanup.
- Concurrency and durability: only files classified during the scan are removed, committed markers are deleted last so failures remain retryable, and every directory removal is non-recursive so a concurrent object publication survives.
- Compatibility: versioned and legacy-ambiguous buckets keep conservative prefix verification, with no on-disk or wire-format change.
- Performance: top-level delimiter listings retain zero recursive visibility probes; cleanup is restricted to the `max-keys=1` request shape used by recursive removal.
- Tests: request gating, committed and malformed residue, concurrent publication, partial-delete retryability, path-component validation, nested metadata, physical probe bounds, and the end-to-end listing cleanup path are covered.
- Rollback: reverting this commit restores the previous listing behavior but also restores the stale-prefix failure.

### DIFF
--- a/crates/ecstore/src/bucket/metadata_sys.rs
+++ b/crates/ecstore/src/bucket/metadata_sys.rs
@@ -1136,6 +1136,16 @@ pub(crate) async fn has_authoritative_never_versioned_state(bucket: &str) -> Res
     bucket_meta_sys.has_authoritative_never_versioned_state(bucket).await
 }
 
+pub(crate) async fn has_authoritative_never_versioned_state_in(
+    ctx: &crate::runtime::instance::InstanceContext,
+    bucket: &str,
+) -> Result<bool> {
+    let bucket_meta_sys_lock = bucket_metadata_sys_of(ctx)?;
+    let bucket_meta_sys = bucket_meta_sys_lock.read().await.clone();
+
+    bucket_meta_sys.has_authoritative_never_versioned_state(bucket).await
+}
+
 pub async fn get_website_config(bucket: &str) -> Result<(WebsiteConfiguration, OffsetDateTime)> {
     let bucket_meta_sys_lock = get_bucket_metadata_sys()?;
     let bucket_meta_sys = bucket_meta_sys_lock.read().await;
--- a/crates/ecstore/src/disk/local.rs
+++ b/crates/ecstore/src/disk/local.rs
@@ -7092,6 +7092,9 @@ impl LocalDisk {
                                 .await?
                         {
                             meta.name.push_str(SLASH_SEPARATOR);
+                            // Conservative listings verify physical prefixes. Never-versioned
+                            // buckets use the bounded fast path and reclaim residue after an
+                            // exact recursive listing proves that prefix empty.
                             if opts.recursive
                                 || opts.incl_deleted
                                 || opts.skip_hidden_prefix_check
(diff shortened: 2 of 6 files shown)


stevapple:
@overtrue Could you take some time to resolve the conflicts, so auto-merge can go on?

PR #7023, opened 2026-09-01
ci(functional): retry chain handoffs and alert on stall

majinghe:
## Why

Follow-up to #7013 found during post-merge review: the chain handoff step (repository_dispatch to the next suite) was `continue-on-error` with a **single attempt**. If the dispatch API call failed — PF_TESTING_GH_TOKEN lacking `contents:write`, or a transient GitHub API error — the chain stalled **silently**: every job stayed green, nothing told anyone.

## What changes

Each of the 7 handoff steps now:
1. Retries the dispatch 3 times (20s/40s/60s backoff).
2. On final failure, files an alert issue in `rustfs/backlog` containing the failed handoff, the expected next event type, likely causes, and the exact one-line recovery command — then exits 1 (job stays green per the chain design, but the stall is now visible and actionable).

Verification: all YAML validated; handoff blocks in all 7 suites upgraded identically.

### DIFF
--- a/.github/workflows/rustfs-heal-test.yml
+++ b/.github/workflows/rustfs-heal-test.yml
@@ -284,21 +284,52 @@ jobs:
 
       - name: "Continue functional chain (next: Pool expansion)"
         # Only chain-triggered runs forward to the next suite; standalone
-        # workflow_dispatch runs stop after their own cleanup.
+        # workflow_dispatch runs stop after their own cleanup. A failed
+        # handoff must never pass silently: it retries, then files an alert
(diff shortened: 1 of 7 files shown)


majinghe:
Regression found: this PR's handoff scripts used ${{VAR}} (Actions expression syntax) for shell variables, which invalidated all seven touched workflow files on main — repository_dispatch was silently dropped and last night's functional chain never started (0 shared-VM suite runs since 2026-09-01 23:11 +0800). Fix in #7034.

PR #7041, opened 2026-09-02
fix(ilm): enqueue committed tier free versions

cxymds:
## Related Issues

- Part of rustfs/backlog#2097.
- Test-side acceptance bound: rustfs/auto-testing#31.

## Summary of Changes

- Retain a typed, drain-once in-memory receipt only when an ordinary single or batch delete commits a transitioned source's durable tier free-version.
- Dispatch committed receipts from the outer store wrapper after lifecycle, namespace, metadata, capacity, and tier-operation guards have completed.
- Deduplicate multi-pool physical copies by complete destructive identity while preserving the exact local free-version UUID required by the worker.
- Keep queue absence, saturation, cancellation, and process interruption fail-safe: the persisted marker remains authoritative and the existing recovery/scan path remains the fallback.
- Add production-path regressions for restored and non-restored deletes, unversioned and explicit-version batch deletes, aggregate errors, local rollback, queue failure, remote failure, and restart recovery.
- Make the CI race tests causal: scanner coverage now synchronizes tier GET, the actual delete namespace write-lock attempt, and remote DELETE; the batch rollback test keeps discovery online and injects a quorum of failures only after local metadata commits.
- Synchronize the committed E2E smoke selection digest with the additional `negative_sigv4` security regression merged in rustfs/rustfs#7052; the profile filter is unchanged.

## Verification

- `cargo test -q -p rustfs-ecstore --features test-util store::init::tests::batch_transitioned_delete_post_commit_failures_roll_back_without_free_version_receipt -- --exact` — passed.
- `cargo test -q -p rustfs-ecstore --features test-util,rio-v2 store::init::tests::batch_transitioned_delete_post_commit_failures_roll_back_without_free_version_receipt -- --exact` — passed.
- The exact batch rollback test passed 30/30 with default features and 30/30 with `rio-v2` by repeatedly invoking the corresponding final ecstore test binaries.
- A focused `cargo nextest run -p rustfs-ecstore --features test-util` selection covering causal single deletes, aggregate batch errors, batch post-commit rollback, and the existing single-delete failpoint passed 6/6.
- The deterministic scanner GET/delete race passed 30/30; all six existing consumers of the strengthened delete-lock barrier passed 6/6.
- Two independent final reviews at the exact head covered correctness, concurrency/durability, compatibility, simplicity, test coverage, and performance; both approved with no findings.
- `cargo check -p rustfs-ecstore --lib` — passed during final review.
- `cargo check -p rustfs-scanner --tests` — passed during final review.
- `./scripts/check_logging_guardrails.sh` — passed; the receipt fallback warning uses the existing lifecycle event/component/subsystem fields and contains no secret-bearing payload.
- `python3 ./scripts/check_test_wiring.py --check-profile e2e-smoke CI_LISTING` — passed with 168 selected tests and SHA-256 `db9bd8cdcb0abe43461aa6b36499b17cabd4098e5b34e300b1a0f0d0f34d9884`.
- `./scripts/check_security_smoke_count.sh check CI_LISTING` — passed with 21 selected auth-rejection tests against a floor of 18.
- `python3 ./scripts/check_test_wiring.py` — passed.
- `python3 ./scripts/check_test_wiring.py --self-test` — 11 passed.
- GitHub `Quick Checks`, `Typos`, `CLA`, and `io_uring Integration (real)` passed for head `92553ac96178ebebf117c3913830a19bbd4e51c2`; the remaining jobs are running.
- `RUST_MIN_STACK=33554432 cargo test -p rustfs-ecstore --features test-util --lib free_version -- --nocapture` — 92 passed on the production diff before the test-only CI follow-ups.
- The exact ignored ILM integration selection passed 55/55 on the first deterministic scanner follow-up; later runtime-code changes are confined to test synchronization and ecstore unit-test fault injection.

Per task constraints, `make pre-pr` was not run.

## Impact

- Ordinary transitioned-object deletes now hand committed tier free-versions directly to the expiry worker after request guards drop; durable metadata remains the source of truth and recovery fallback.
- No S3, HTTP, RPC, trait, wire, journal, or `xl.meta` format changes are introduced.
- `ObjectOptions` gains a doc-hidden additive in-memory carrier whose default is `None`; exhaustive external Rust struct literals may need to add the field or use `..Default::default()`.
- Non-transitioned single deletes add one request-scoped collector allocation. Batch deletes add one collector and one box per request, not per object; there is no additional disk read, namespace scan, RPC, blocking queue send, or remote I/O on the request path.
- The latest CI follow-ups affect only test fault injection, assertions, and the immutable E2E selection digest; default production behavior is unchanged.

## Additional Notes

- The prior default and `rio-v2` failures came from a test fixture that removed two disks before discovery. Depending on timing, it could return raw `ErasureWriteQuorum` before reaching the rollback commit path. The replacement fixture leaves every disk online through discovery and makes three disks report `Unexpected` only after their batch metadata commits, then verifies both versions remain, no receipt is published, and the remote object is retained.
- The same prior default run also timed out in the unchanged `public_upload_part_stages_before_decommission_capacity_admission` one-second probe. That file is outside this PR diff and the exact probe passed locally 10/10; it is treated as unrelated runner contention unless reproduced.
- The first run at head `3747762177ed39399182a9b11a9f75c27417166c` passed the serial ILM job but its E2E archive guard observed 168 smoke tests with the new digest while the repository still carried the 167-test digest. rustfs/rustfs#7052 added the selected security regression and had the same E2E failure before merge. The final one-line commit accepts that added test without changing the smoke filter or lowering the security floor.
- Merge base used for the frozen diff: `bc789332b6c2b10cd6fc3425af773adc94dbb5e2`.
- Head: `92553ac96178ebebf117c3913830a19bbd4e51c2`.
- Frozen binary diff SHA-256: `00b804a9d652a9ed73dcdddac9a061e964be79190e8c57470fbe5a439785c760`.
- Current `origin/main` observed before the final PR metadata refresh: `68f47b9219132559e97ac3b984cb6c76a66e7fd1`; GitHub reports the PR mergeable.
- rustfs/backlog#2097 must remain open until TIER-107/TIER-108 pass in both target topologies and the final immutable-artifact 53-case machine gate is green.


cxymds:
CI follow-up pushed in dfde5fde77501bc9f1a578db75745cda6277608e.

Attribution:

- The 5,056-byte future failures in Test and Lint and Test and Lint (rio-v2), plus the ILM Integration scanner assertion failure, were introduced by this PR.
- Neither the frozen base nor subsequent main commits #7043 and #7046 contain an independent fix.

Corrections:

- The single and batch delete wrappers now keep the inner operation, unconditional receipt drain/enqueue, and original result inside the existing single boxed async state. The 4 KiB threshold was not raised, and the test now covers both wrappers.
- The two ILM tests use a one-shot failing remote-DELETE barrier with one worker only. Startup recovery is intentionally not started, so the queued scanner fallback cannot be masked by a concurrent recovery sweep.

Verification:

- free-version regressions: 92/92
- default stack-bound regression: 1/1
- affected ILM tests: three consecutive 2/2 runs (6/6)
- cargo fmt and git diff checks: pass
- final frozen correctness review: APPROVE
- final frozen compatibility/performance/CI review: APPROVE

Frozen merge base: bc789332b6c2b10cd6fc3425af773adc94dbb5e2
Frozen diff SHA-256: ad4923b373baaefdc806ef222f24a37a9bc84996294d0a9861c447ba5acd4108

The exact rio-v2 regression passed before the #7043 dependency rebase. Post-rebase local rebuild attempts were interrupted before test execution by shared-host memory pressure, without compiler or test diagnostics; the newly triggered GitHub CI run is the isolated verification for that matrix. make pre-pr was not run, per task constraints.

PR #7049, opened 2026-09-02
fix(ecstore): stop scan_dir emitting entries past a limit hit inside a subdirectory

HateTM:
## Summary

- Recursive `ListObjectsV2` over a wide prefix can silently drop a contiguous block of keys and report `IsTruncated=false`. Repro: 7,881 keys uploaded under one prefix, recursive listing with `MaxKeys=1000` returns exactly 7,826 keys — a 55-key block inside one subdirectory is missing, while a key that sorts *after* that subdirectory is still returned as the last item of the stream. Reproduced identically on `1.0.0-rc.4` and `1.0.0-rc.5`.
- Root cause: in `LocalDisk::scan_dir` (`crates/ecstore/src/disk/local.rs`), the per-entry loop flushes pending subdirectories from `dir_stack` and recurses into them via `Box::pin(self.scan_dir(...))` before processing the current sibling entry. If that recursive call reaches `opts.limit` partway through the subdirectory, it returns `Ok(())` cleanly, but the *outer* loop had no limit re-check after the flush — so it went on to write the current entry anyway, even though that entry sorts **after** the now-truncated subdirectory.
- `gather_results` (`crates/ecstore/src/store/list_objects.rs`) builds the next page's continuation marker from the last entry written to the stream. Because that last entry was the later-sorting sibling rather than the true last key inside the subdirectory, the next page's `forward_to` cursor skips straight past the subdirectory's unscanned tail — permanently, not just deferred to the next page — while `disk_has_more = list_result.err.is_none()` still reports a clean, non-truncated listing.

## Fix

Add a limit re-check immediately after the `dir_stack` flush loop, before the current entry is written. Once the flush loop's recursion has exhausted the page limit, `scan_dir` now returns immediately instead of emitting a later-sorting entry, so the stream's true last-written key — and therefore the continuation marker — always reflects where the scan actually stopped.

## Test plan

- [x] Added `scan_dir_does_not_emit_entries_past_a_limit_hit_inside_a_subdirectory`, reproducing the reported shape (one subdirectory with more objects than the page limit, plus a sibling object sorting after it). Fails before the fix (subdirectory truncated *and* the later sibling still emitted); passes after.
- [x] `cargo test -p rustfs-ecstore --lib disk::local::test::scan_dir` — all scan_dir tests pass, including the pre-existing `scan_dir_records_whole_parent_read_dir_before_page_limit`.
- [x] `cargo test -p rustfs-ecstore --lib disk::local::test::` — 239/241 pass; the 2 failures (`test_rename_data_writes_old_metadata_backup_for_inline_overwrite`, `strict_inline_rename_retains_admission_until_commit_fsync`) are unrelated fsync/rename-concurrency tests that also fail on unmodified `main` under parallel execution and pass individually — pre-existing flakiness, not caused by this change.

CLA document:
https://github.com/rustfs/cla/blob/main/cla/v2.md


HateTM:
## Update: this PR does not fully close the reported bug — root cause is deeper than `scan_dir`

I deployed this fix (plus a follow-up patch, see below) against a real production single-disk bucket with ~1.17M objects under one prefix (the same shape as the original report: `site-archive/pir/Huawei/...`) and confirmed the core symptom — a recursive `ListObjectsV2` still silently returns `IsTruncated=false` partway through, dropping the rest of the tree.

### What this PR's fix does and does not do

The fix in this PR (checking `opts.limit` right after the `dir_stack` flush loop in `scan_dir`) is real and correct for the mechanism it targets: it stops `scan_dir` from emitting a later-sorting sibling entry once a recursive call into a pending subdirectory has already exhausted the page limit. I verified this both with a unit test (`scan_dir_does_not_emit_entries_past_a_limit_hit_inside_a_subdirectory`) and by observing production behavior change: before this fix, the last key returned by a truncated listing was a sibling entry sorting *after* the actual gap (e.g. `.../2382_Красная Горбатка/Безымянный.png`); after this fix, the last key returned is a legitimate entry from inside the exhausted subdirectory itself (e.g. `.../2382_Красная Горбатка/АП/IMG_20210702_120826.jpg`). I also added a second, follow-up patch that re-checks the limit *inside* the flush loop (not just once after it), to cover the case where more than one pending subdirectory would otherwise be popped and recursed into after the limit was already reached.

**Neither patch fixes the actual false-EOF symptom**, though. I reproduced the exact same cut (~7825/1.17M objects, `IsTruncated=false`) with both patches deployed, at a two-level-nested subdirectory boundary (`2382_Красная Горбатка/АП/`).

### Actual root cause (confirmed via debug logging in production)

I enabled `RUST_LOG=rustfs_ecstore::store::list_objects=debug,warn` and captured the exact `gather_results` outcome for the failing page:

```
"list_objects gather_results drained all candidates without hitting limit"
bucket: "site-archive", prefix: "pir/"
scanned_entries: 2201, returned_entries: 825, filtered_entries: 1376
marker: "pir/Huawei/ВК Владимир 2021/2375_Гусь ЧКС/IMG_20211125_103822.jpg"
```

`forward_to`/`per_disk_limit` are wired correctly here (`per_disk_limit = opts.limit + 4 + opts.limit/16 = 1067`, matching the log exactly, `forward_to: opts.marker.clone()` — this is not a marker-plumbing bug). The disk-side `scan_dir` producer received/emitted 2201 raw `MetaCacheEntry` messages before its channel closed. Of those, only 825 survived `gather_results`' own filtering (marker/prefix/directory-marker/delete-marker checks) into `entries`. Because `825 < opts.limit (1001)`, the check

```rust
if opts.limit > 0 && entries.len() >= opts.limit as usize { ... LimitReached ... }
```

never fires, so `gather_results` falls through to the "finish not full, return eof" branch, which sends `err: Some(Unexpected)`. Downstream, `disk_has_more = list_result.err.is_none()` then evaluates to `false` — a clean, "successful" EOF is reported to the S3 API layer even though the disk-side producer stopped because it exhausted its own `per_disk_limit` budget (1067 raw entries), not because the tree was actually exhausted.

### Why this is a protocol/contract bug, not a `scan_dir` bug

`scan_dir` returning `Ok(())` is indistinguishable to its caller between "stopped because `opts.limit`/`per_disk_limit` was reached" and "stopped because there is genuinely no more data on disk." Both cases produce `err: None` from the producer's perspective, and `gather_results` currently infers "more data available" *only* from whether its own post-filter `entries.len()` reached `opts.limit` — which is the wrong signal whenever a meaningful fraction of the raw scanned entries get filtered out downstream (as happens here: 1376 of 2201, i.e. ~62%, filtered before ever reaching `entries`).

A correct fix likely needs the disk-side producer to explicitly signal "I stopped due to hitting my own limit" (as opposed to genuine tree exhaustion) through to `gather_results`/`disk_has_more`, independent of how many of the raw entries survive `gather_results`' filtering — rather than inferring it from `entries.len() >= opts.limit`, which silently assumes scanned-vs-returned parity that doesn't hold once filtering discards a large share of what the disk-side limit was actually budgeting for.

I don't have enough context on the multi-disk quorum/callback architecture in this layer (`ListingSupplement`, `agreed`/`partial`/`finished` callbacks, the `MetaCacheEntry` wire format across `list_path_raw`) to confidently design that signal myself without more risk of another partial fix. Flagging this in detail here so whoever picks it up next has the full repro and root-cause trace without having to redo the diagnostic work.

### Repro shape for local reproduction

- Single-disk bucket, ~1.17M objects under one prefix, deeply nested directory tree (photo archive: `site/subdir/*.jpg`, several levels deep, uneven fan-out per subdirectory).
- Recursive `ListObjectsV2`, `MaxKeys=1000`, paginated via `ContinuationToken` (or equivalently `StartAfter` from the same position — both trigger the same failure identically, so this is not a continuation-token/cache-specific bug).
- Failure is deterministic at a specific tree position once a directory's fan-out and the ~5-15% typical filter ratio combine such that `per_disk_limit` gets exhausted mostly on entries that don't survive `gather_results`' filtering.

CLA document:
https://github.com/rustfs/cla/blob/main/cla/v2.md

HateTM:

CLA document:
https://github.com/rustfs/cla/blob/main/cla/v2.md

HateTM:
Re-triggering CLA Check after signing.

HateTM:

PR #7083, opened 2026-09-02
fix(replication): fence stale metadata status writeback

cxymds:
## Related Issues

Closes rustfs/backlog#2099

### Problem

Metadata mutations can be issued after the target object has committed while the source still records the initial object replication as `PENDING`. The old eligibility check rejected that target, so a successful local `PutObjectTagging` could disappear without a replication task.

Simply admitting `PENDING` is unsafe: an older object worker can later write `COMPLETED`/`FAILED` over the newer metadata mutation's `PENDING`, suppressing recovery and publishing stale completion state.

### Changes

- Treat an exact target ARN with historical `PENDING`, `FAILED`, `COMPLETED`, or legacy `COMPLETE` state as eligible for metadata replication, while preserving the replication-rule gate. Missing, empty, `REPLICA`, unknown, or conflicting states remain fail closed.
- Require RustFS/MinIO status aliases to agree and require exactly one status record for the target ARN. Duplicate records and mixed-case alias conflicts fail closed.
- Persist a source-local UUID mutation generation alongside the existing opaque raw replication timestamp for every replication-admitted source mutation. Keep this generation local; it is removed from outbound replication metadata.
- Capture a deterministic semantic fingerprint of the authorized source snapshot so a rolling-upgrade legacy writer that preserves the UUID and reuses a timestamp still cannot make a stale write-back pass after changing tags, lock metadata, user metadata, version identity, or target admission.
- Canonicalize the suspended-bucket null-version identity inside that fingerprint: queue-side `Some(nil UUID)` and storage-reread `None` represent the same persisted xl.meta and must not self-supersede.
- Validate terminal status write-back against timestamp, UUID, and fingerprint using an exact CAS after the quorum read while holding the object write lock. Legacy missing fields use exact `Option` matching; empty, invalid, and dual-alias conflicts fail closed.
- Validate every terminal attempt, including unchanged `FAILED -> FAILED`. `ValidateOnly` performs the locked read/fence/CAS checks without a metadata write or cache invalidation.
- Map a failed CAS to `Superseded`: preserve the newer `PENDING`, consume the old MRF task, and suppress the old generation's event, failure log, and statistics.
- Carry the renewable replication lock's loss signal into the terminal storage commit. A worker that loses its lease retries without writing status, publishing terminal side effects, or consuming its MRF entry; malformed generations fail in the same retryable way before any target I/O.
- Keep terminal CAS quorum failures side-effect free: they return for retry without entering dangling-object deletion, including `ValidateOnly` and lease-lost update attempts.
- Keep the replication terminal write-back condition behind a pointer-sized box in `ObjectOptions`, preventing an optional generation snapshot from inflating every by-value storage future and exhausting CI test stacks.
- Canonicalize storage-authored internal metadata updates by removing stale RustFS/MinIO aliases (including mixed-case legacy keys), writing one matching dual-key pair, and rebuilding the in-memory replication state before the xl.meta RMW.
- Strip source-object replication bookkeeping before COPY and trusted internal PUT/MPU admission. Local source writes mint their own generation/status; authorized inbound replication remains exempt.
- Keep source user metadata that resembles RustFS/MinIO replication bookkeeping in the ordinary `x-amz-meta-*` namespace. A real ODM write-back regression proves forged status, generation, and replica markers cannot become historical target admission after a later tag match.
- Preserve the existing authorized `REPLICA`-only multipart compatibility path even when an older or heterogeneous client does not repeat the source-request marker on Complete; untrusted and local completions still clear staged replica bookkeeping.
- Recompute multipart replication admission immediately before object commit, atomically replace staged generation/PENDING metadata, and reuse the same decision for scheduling. This covers both S3 multipart and the new internal/ODM multipart path.

### Compatibility and security

- No S3 API, replication configuration, public status, timestamp format, or replication wire contract changes.
- The UUID is optional internal dual-key metadata; legacy timestamp-only and fully absent generations remain readable.
- Existing single RustFS or MinIO aliases remain readable during rolling upgrades; agreeing dual aliases are accepted, while conflicting or empty aliases are rejected.
- Authorized multipart requests carrying only `X-Amz-Replication-Status: REPLICA` retain their persisted anti-cascade state; this does not broaden header authorization or trust caller-supplied metadata.
- The existing `cfg.replicate(...)` gate remains mandatory. An object that was never admitted to a target is still not created there after a later matching tag mutation (the PR #5696 safety boundary).
- Rollout note: upgrade or drain all replication workers before relying on the new fence. Older workers do not send the write-back condition.

### Performance

- No hot storage object lock is held across remote replication I/O. The existing renewable replication coordination lease remains held and is now fenced at terminal commit.
- Terminal write-back adds one locked current-object quorum validation. Unchanged terminal retries use `ValidateOnly` and avoid quorum metadata writes and cache invalidation.
- The semantic fingerprint hashes bounded in-memory object metadata and sorted target ARNs; it adds no disk or wire field.
- Internal alias canonicalization uses one bounded grouping pass and one metadata-map retain pass per local metadata mutation; it adds no remote I/O.
- The boxed condition allocates only for terminal replication publication; ordinary requests retain a pointer-sized optional field and lower stack pressure through deep storage futures.

## Verification

- `cargo test -p rustfs-replication` (184 unit + 2 compatibility tests passed)
- `cargo test -p rustfs-filemeta` (263 unit + 2 property tests passed)
- `cargo test -p rustfs-replication metadata_replication_requires_prior_target_admission -- --test-threads=1`
- `cargo test -p rustfs-replication consistent_internal_metadata_accepts_single_or_matching_aliases_and_rejects_conflicts -- --test-threads=1`
- `cargo test -p rustfs-ecstore metadata_replication_requires_both_current_rule_match_and_historical_admission -- --test-threads=1`
- `cargo test -p rustfs-ecstore invalid_generation_retries_before_remote_replication -- --test-threads=1`
- `cargo test -p rustfs-ecstore terminal_writeback_carries_replication_lock_loss_fence -- --test-threads=1`
- `cargo test -p rustfs-ecstore metadata_mutations_canonicalize_mixed_case_minio_generation_aliases -- --test-threads=1`
- `cargo test -p rustfs-ecstore metadata_mutation_generation_tests -- --test-threads=1` (7 passed)
- `cargo test -p rustfs-ecstore replication_terminal_cas_quorum_failure_never_deletes_surviving_metadata -- --test-threads=1`
- `cargo test -p rustfs-ecstore replication_status_writeback_ -- --test-threads=1` (9 passed, including suspended null-version Update and ValidateOnly)
- `cargo test -p rustfs-ecstore terminal_ -- --test-threads=1` (73 focused tests passed on the final base)
- `cargo test -p rustfs-ecstore replication_put_options_filter_and_map_metadata -- --test-threads=1`
- `cargo test -p rustfs-ecstore complete_multipart_upload_commits_replication_admission_metadata_atomically -- --test-threads=1`
- `cargo test -p rustfs --lib test_complete_multipart_opts_with_replica_status -- --test-threads=1`
- `cargo test -p rustfs --lib authorized_replica_only_multipart_complete_preserves_persisted_anti_cascade_state -- --test-threads=1`
- `cargo test -p rustfs --lib write_back_ -- --test-threads=1` (10 on-demand migration single/MPU adapter tests passed, including reserved-looking source user metadata)
- `cargo test -p rustfs --lib internal_multipart_completion_recomputes_replication_admission_atomically -- --test-threads=1`
- `cargo test -p rustfs --lib internal_multipart_roundtrip_completes_and_abort_leaves_nothing -- --test-threads=1`
- `cargo test -p rustfs --lib internal_put_ -- --test-threads=1` (8 passed)
- `cargo test -p rustfs --lib internal_multipart_ -- --test-threads=1` (2 passed)
- `cargo test -p rustfs --lib internal_put_drops_foreign_admission_before_later_matching_metadata -- --test-threads=1`
- `cargo test -p rustfs --lib internal_put_replaces_mixed_case_foreign_generation_with_local_canonical_aliases -- --test-threads=1`
- `cargo test -p rustfs --lib internal_complete_multipart_upload_rejects_unordered_parts -- --test-threads=1`
- `cargo test -p rustfs --lib local_copy_does_not_inherit_source_replication_generation -- --test-threads=1`
- `cargo test -p e2e_test test_site_replication_tagging_during_initial_pending_window_converges -- --test-threads=1`
- `cargo test -p e2e_test test_bucket_replication_acceptance_matrix_local_dual_targets -- --test-threads=1`
- `cargo test -p e2e_test test_site_replication_tagging_lww_converges_active_active_real_dual_node -- --test-threads=1`
- `target/debug/deps/rustfs_ecstore-3fe3b962ebbc2c5e --exact object_api::types::tests::replication_status_writeback_condition_remains_indirected --nocapture`
- `target/debug/deps/rustfs_ecstore-3fe3b962ebbc2c5e replication_status_writeback --nocapture` (10 passed)
- `target/debug/deps/rustfs_ecstore-3fe3b962ebbc2c5e --exact set_disk::ops::object::metadata_mutation_generation_tests::replication_terminal_cas_quorum_failure_never_deletes_surviving_metadata --nocapture`
- `RUST_MIN_STACK=2097152 target/debug/deps/rustfs-e8ca872b0542a213 --exact <test> --nocapture` (all 7 tests that overflowed on Linux passed serially)
- Same-host A/B stack threshold: the pre-fix binary aborted at 2,031,616 bytes for `data_usage_endpoint_serves_snapshot_without_live_listing`; the fixed binary passed.
- `cargo fmt --all --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo clippy -p rustfs -p rustfs-protocols --all-targets --features swift -- -D warnings`
- Full `.github/workflows/ci.yml` Quick Checks command set, including architecture migration, layer dependency, `error(format!)`, logging, extension-schema, body-cache, S3 footprint, FIPS/KMS wording, embedded-secret, and test-wiring guards
- `./scripts/check_s3s_footprint.sh`

The full focused matrix and all three real-process E2E scenarios passed before the final rebase. The PR was then rebased onto `origin/main@d04611ba3889c93f0d2bf27e504eab5d45c3fb86`; both CI clippy commands, the complete Quick Checks command set, CAS write-back tests (9), terminal-focused tests (73), atomic multipart admission, and authorized `REPLICA` multipart compatibility tests were rerun successfully. The rebase preserves #7075's ETag-aware single-PUT diagnostic and #7060's multipart checksum-type validation/response behavior.

### Independent adversarial reviews

Frozen review snapshot: base `5e58b1d3a2324adf74c9d148aaf18813067d7915`, head `751da58259842b843b96a5bd1fe59dbf514d14f7`, diff SHA-256 `44adc95bcfd7992070959ded5176c6aefa820b6293a66172dba844abfc9b59b7`.

- Correctness / concurrency / durability / compatibility: CLEAN — boxing changes only in-memory ownership; CAS comparison, lease and namespace fences, quorum handling, persisted metadata, wire behavior, and rolling-upgrade semantics are unchanged.
- Simplicity / test coverage / performance: CLEAN — the single production allocation occurs only for terminal write-back, default options shrink, and the structural guard, 10 write-back tests, quorum regression, and 7 stack-sensitive tests cover the change.


cxymds:
CI follow-up for the failed **Quick Checks** run:

- Root cause was introduced by this PR: five direct type references bypassed the ECStore replication boundary/facade rules.
- Rebasing onto current `main` also exposed a false `error(format!)` ratchet decrease: the new write-back test module preceded the two existing production call sites. The tests were moved behind the existing first test boundary; the baseline remains unchanged at 2 for `set_disk/ops/object.rs` and the guard still covers those calls.
- The boundary references now flow through `replication_filemeta_boundary`, `replication_storage_boundary`, and the bucket replication facade. These changes are import/test-layout only and do not alter runtime replication semantics.
- Rebase preserved #7075's ETag-aware single-PUT diagnostic and #7060's multipart checksum behavior.

Updated snapshot:

- Base: `d04611ba3889c93f0d2bf27e504eab5d45c3fb86`
- Head: `d6270cf4591b4371598c4a36101b873e2336736f`
- Diff SHA-256: `4d0aa7192b5abf0ddb9a011080cb73b38caca6c5cfdd3f2c227abf79ea093122`

Local verification after the rebase/fix:

- Complete `ci.yml` Quick Checks command set: passed
- Replication status write-back/CAS: 9 passed
- `terminal_` focused set: 73 passed
- Atomic multipart replication admission: passed
- Authorized `REPLICA` multipart persistence and option compatibility: passed
- `cargo fmt --all --check` and `git diff --check`: passed

Per project instruction, `make pre-pr` was not run.

cxymds:
Follow-up for the subsequent **Test and Lint** / **Test and Lint (swift)** failures:

- Both jobs reported the same PR-introduced Clippy findings: a large outcome enum, collapsible conditions, a map key/value iteration that only used values, and redundant test-only clones.
- The large enum was not boxed. Successful persistence now writes the returned `ObjectInfo` through an explicit mutable output and returns a zero-payload outcome; `PreconditionFailed`/`Superseded` and other errors leave that output unchanged. This preserves behavior and avoids adding a terminal-writeback heap allocation.
- The remaining changes are evaluation-order-preserving condition/iterator simplifications and removal of redundant test clones.

Updated snapshot:

- Base: `d04611ba3889c93f0d2bf27e504eab5d45c3fb86`
- Head: `bae840fe4a0bce3083071040a88e2d58e231b6d9`
- Diff SHA-256: `2a048f430f9f5c3b092387295bf261d13a3e4eb58adb5c891f7c9c8c87977986`

Local verification:

- `cargo clippy --all-targets -- -D warnings`: passed
- `cargo clippy -p rustfs -p rustfs-protocols --all-targets --features swift -- -D warnings`: passed
- Replication status write-back/CAS: 9 passed
- `terminal_` focused set: 73 passed
- Complete `ci.yml` Quick Checks command set: passed
- `cargo fmt --all --check` and `git diff --check`: passed

Per project instruction, `make pre-pr` was not run.

PR #7107, opened 2026-09-03
fix(scanner): preserve explicit cycle retry cadence

marshawcoco:
## Related Issues

Related to #5950.

## Summary of Changes

- preserve an explicit scanner cycle as the minimum delay after a superseded scan
- retain the existing short exponential convergence retry for the default adaptive cadence
- cover environment, persistent-config, and compatibility-config cycle sources

## Verification

- `cargo test -p rustfs-scanner superseded_retry_ --lib -- --nocapture`
- `cargo test -p rustfs-scanner superseded_retry_preserves_explicit_cycle_cadence --lib -- --nocapture`
- `cargo nextest run -p rustfs-scanner` (686 passed, 19 skipped)
- `cargo fmt --all --check`
- `git diff --check`

## Impact

Continuous writes no longer cause superseded scans to bypass an operator-configured cycle interval and repeatedly walk the namespace early. Default scanner retry behavior, configuration names, and APIs are unchanged.


houseme (commented):
Reviewed the current merge result against `main` and the latest #5950 production evidence.

No blocking findings from the scanner scheduling side. This does not look like a patch-on-patch workaround: it keeps the existing retry backoff as the canonical primitive and adds the missing policy boundary for superseded cycles when the operator selected an explicit cycle interval. That is the right place to fix the regression because dirty-usage wakeups and clean-idle backoff already respect explicit cycle policy; the remaining early full-walk path was the superseded convergence retry.

Performance-wise, the change adds only one small `Duration` calculation per scheduler loop and removes the expensive behavior reported in #5950: repeated physical namespace walks every few seconds/minutes despite `RUSTFS_SCANNER_CYCLE=21600`. The default adaptive path is still preserved for deployments that did not override the scanner cycle.

Focused local checks:

- `cargo test -p rustfs-scanner superseded_retry_ --lib -- --nocapture` passed, 7 tests.
- `git diff --check origin/main...HEAD` passed.

CI still has long jobs pending, and the PR is still draft, so I would wait for those before merging. From the code-review side, I do not see a better high-performance refactor needed for this scope.

PR #7118, opened 2026-09-04
fix(version): avoid deriving RustFS version from runtime working directory

cuishuang:
## Related Issues

N/A

## Summary of Changes

Remove runtime Git repository detection from `version::get_version()`.

The previous implementation executed Git commands without binding them to the RustFS source repository. As a result, starting RustFS from an unrelated Git working directory could cause the binary to report that repository's tag as its own version. This value is also used by the update checker as `current_version` and in the `RustFS/<version>` User-Agent.

`get_version()` now consistently derives the version from build-time metadata generated by `shadow_rs`, while preserving the existing `RUSTFS_BUILD_VERSION` override and version reference formatting.

A regression test was added to ensure the returned version matches the build metadata.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `CARGO_INCREMENTAL=0 RUSTFLAGS='-C debuginfo=0' cargo test -p rustfs --no-default-features --lib version::tests -- --nocapture`

All 9 version module tests passed.

## Impact

This changes version reporting for binaries launched from unrelated Git repositories. RustFS will no longer inherit version information from the process working directory.

Build-time version metadata, `RUSTFS_BUILD_VERSION`, CLI display output, update checks, and User-Agent formatting remain supported.

## Additional Notes

The change is limited to `rustfs/src/version.rs`. Runtime version detection no longer requires the `git` executable or a Git repository.

### DIFF
--- a/rustfs/src/app/bucket_list_through.rs
+++ b/rustfs/src/app/bucket_list_through.rs
@@ -336,7 +336,7 @@ async fn fetch_source_page(
             delimiter: params.delimiter.as_deref(),
             // S3 ignores start-after once a continuation token is present, so
             // the client's own start-after only applies to the first page.
-            start_after: token.is_none().then(|| params.start_after_for_query.as_deref()).flatten(),
+            start_after: token.is_none().then_some(params.start_after_for_query.as_deref()).flatten(),
             continuation_token: token,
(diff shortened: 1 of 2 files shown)


cuishuang:

PR #7132, opened 2026-09-04
fix(ilm): reject invalid retention counts and validate lifecycle filters

overtrue:
## Related Issues

backlog#2201

### `NewerNoncurrentVersions` had no lower bound

Validation checked `NoncurrentDays` but never the retention count, and evaluation read it through `usize::try_from(count).unwrap_or(usize::MAX)`. A negative count was accepted over HTTP and then behaved as "retain (almost) everything", so the rule silently stopped expiring versions — the one outcome a retention rule must never produce by accident.

A negative count is now rejected during validation. It is also no longer read as "retain everything" anywhere it can still arrive from older persistence or an import:

- `eval_inner` takes no action for such a rule and emits the existing `lifecycle_noncurrent_expiry_skipped` diagnostic with `reason = "corrupt_newer_noncurrent_versions"`.
- `noncurrent_versions_expiration_limit` (the batch path) yields no limit event instead of `usize::MAX`.
- `Evaluator::eval` returns a typed corruption error, so callers that can report one do.

Fail-closed is deliberate: the safe reading of a corrupt retention count is "delete nothing", and the removed `unwrap_or(usize::MAX)` had the same effect *silently*. `NoncurrentVersionTransition` was already covered by `NoncurrentVersionTransitionOps::validate`; the gap was the expiration side.

### Count-only noncurrent expiration

`NewerNoncurrentVersions` with no `NoncurrentDays` is a MinIO extension, not an AWS form. It was rejected at PUT as an actionless rule and, had it been persisted, would never have executed. It is now accepted and honoured with MinIO's semantics: the newest N noncurrent versions are kept, every older one is due as soon as it became noncurrent (zero-day, via the existing `expected_expiry_time(_, 0)` path). With both fields set, the count gates candidates and the age condition still decides when they are due — unchanged.

Zero keeps the meaning the batch limit path has always given it — no count constraint — rather than gaining a new "retain nothing" reading that only one of the two paths would honour. A zero-count rule with no age condition therefore still has no action and is rejected as such.

### `LifecycleRuleFilter` structure

The generated DTO is all-`Option`, so none of the S3 schema constraints were checked. Now validated: at most one top-level predicate, an `And` that combines at least two, no repeated tag key, tag key (1–128) and value (≤256) limits, non-negative sizes, and `ObjectSizeGreaterThan < ObjectSizeLessThan`. The legacy `Prefix` / `Filter` exclusivity check is unchanged.

An empty filter stays valid. The issue asks for "exactly one" top-level predicate, but AWS documents an empty `Filter` as "applies to every object in the bucket", and it is the ordinary way to write an unconditional rule — rejecting it would break valid configurations, so this implements "at most one".

Delete-marker tag matching is untouched.

### S3 error mapping

Schema-shape violations carry a distinct `ErrorKind` (`LIFECYCLE_MALFORMED_XML_ERROR_KIND`) and the PUT boundary answers them with `MalformedXML`. Every other validation failure keeps the `InvalidArgument` this path has always returned; no existing error changes code.

## Verification

Conflict resolution against current main: `cargo +stable fmt --all --check`, `cargo +stable test -p rustfs-lifecycle` (151 passed), and `git diff --cached --check` passed. Both the count-only cases and main's independent expiration/transition retention regressions are preserved.

- `cargo +1.98.0 test -p rustfs-lifecycle` — 140 passed (127 pre-existing, unchanged).
- `cargo +1.98.0 test -p rustfs --lib put_bucket_lifecycle_validation_errors_keep_their_s3_code` — drives the real validator and asserts `MalformedXML` for a two-predicate filter and `InvalidArgument` for a negative count, so the mapping cannot drift from the validator.
- Table-driven filter cases: empty filter, single predicate, two top-level predicates, prefix + `And`, one-member `And`, two-member `And`, two tags, repeated tag key, empty/missing tag key, key at and past the limit, oversized value, negative sizes, inverted range, valid boundary range.
- Revert-proof: with the negative-count rejection removed, `validate_rejects_negative_newer_noncurrent_versions` fails; with the evaluation skip removed, `eval_inner_takes_no_action_for_a_corrupt_retention_count` fails **by deleting** the version; with filter validation removed, `validate_covers_filter_invariants` fails.
- XML round-trip test pins that the count-only form parses and re-serializes without gaining an age condition.
- `make pre-commit`, `cargo +1.98.0 fmt --all --check`, and the six repository guard scripts pass.

## Impact

A lifecycle configuration with a negative `NewerNoncurrentVersions` is now rejected with `InvalidArgument` instead of being accepted and silently disabling expiry. Existing valid configurations are unaffected: all 127 pre-existing lifecycle tests pass unchanged, and no accepted rule changes its action or priority.

Count-only rules become usable (previously rejected), matching MinIO.

Malformed `Filter` documents that were silently accepted are now rejected — `MalformedXML` for shape, `InvalidArgument` for values. This is a behaviour change for clients that were sending such documents; they were never honoured as written, since only one predicate was ever consulted.

## Additional Notes

Deviation from the issue text: "exactly one top-level predicate" is implemented as "at most one", to keep the AWS-documented empty filter valid. Flagged here for review.

Conflict resolution review: correctness and compatibility passed for count-only, count-plus-age, independent transition, and corrupt-count cases. Security checks retain the invalid-input and destructive-action guards. Simplicity review kept one expiration condition; coverage review confirmed both branches' regression tests remain and pass.


crates/lifecycle/src/core.rs
```diff
@@ -199,6 +276,79 @@ impl RuleValidate for LifecycleRule {
...
+
+    Ok(())
+}
+
+/// S3 requires a tag to carry a key; both key and value are length-bounded.
+/// The DTO makes both optional, so a keyless tag has to be rejected here
+/// rather than silently matching nothing.
+fn validate_lifecycle_tag(tag: &s3s::dto::Tag) -> Result<(), std::io::Error> {
+    let key = tag.key.as_deref().unwrap_or_default();
+    let value = tag.value.as_deref().unwrap_or_default();
```
overtrue: This accepts a missing or empty Value as an empty string, but tag_matches returns false when s3s has decoded it to None. The PUT succeeds and persists a rule that never matches, even for env=. Please preserve the MinIO-compatible empty value at matching time and add an XML round-trip-to-action regression.

crates/lifecycle/src/core.rs
```diff
@@ -611,18 +765,44 @@ impl Lifecycle for BucketLifecycleConfiguration {
                     continue;
                 }
 
+                // A retention count that PUT validation would have rejected can
+                // only come from older persistence or an import. Take no action
+                // for the rule instead of reading the negative value as "retain
+                // (almost) everything", which is how such a rule silently
+                // stopped deleting versions (backlog#2201).
+                if !obj.is_latest && rule_has_corrupt_retention_count(rule) {
```
overtrue: This skips the corrupt rule only for noncurrent versions. An old or imported rule with a negative retention count plus an expired current Expiration still produces DeleteAction; the heal path calls lc.eval directly and can enqueue that deletion without the Evaluator error gate. Please skip the entire corrupt rule and cover the latest-object heal path.

crates/lifecycle/src/core.rs
overtrue: A tag without `<Value>` reaches here as `None`, but this accepts it as empty while `Filter::test_tags` rejects `None`. The key-only form should match an empty tag value; an exact-head regression currently returns no match. Reject the missing value or normalize it consistently.

crates/lifecycle/src/core.rs
overtrue: This guard only covers noncurrent objects. An imported rule combining a due current expiration with a negative retention count reaches direct `lc.eval` callers and returns `DeleteAction`; an exact-head regression reproduces it. Guard the corrupt rule before evaluating any destructive action.

PR #7139, opened 2026-09-04
ci(upgrade): render upgrade matrix in report; fix from default (404)

majinghe:
## Background

The upgrade suite has been dying during OLD-package installation in the nightly chain for several nights, leaving an empty `Total: 0` report:

- The workflow defaulted `from_version=1.0.0-rc.4-preview.1`, whose release has no .deb asset (`rustfs_1.0.0.rc.4.preview.1_amd64.deb` returns 404); release `1.0.0-rc.4` ships no deb either (only zips/sbom/provenance)
- The report only echoed the requested URLs plus a case table — no summary of which versions were upgraded or whether it succeeded

The auto-testing side (rustfs/auto-testing#36) already changed the script default FROM to `1.0.0-rc.3` (which ships a deb asset) and made the suite capture the real versions via `rustfs --version` around the upgrade, emitting `[UPG-TOPO] <topology> <backend> <old> <new> PASS=n FAIL=m` lines.

## Changes (.github/workflows/rustfs-upgrade-test.yml)

1. `from_version` default `1.0.0-rc.4-preview.1` → `1.0.0-rc.3` (aligned with auto-testing #32/#36); the description now notes it "must ship a .deb asset"
2. The report generator parses the `[UPG-TOPO]` lines and emits an **Upgrade Matrix** section before the Case Summary:

| Topology | KMS Backend | From Version | To Version | Result |
| --- | --- | --- | --- | --- |
| single-single | local | 1.0.0-rc.3 | 0.1.0-nightly | PASS (PASS=15 FAIL=0) |

   When the suite dies before upgrading (no UPG-TOPO lines), the matrix shows a single `NOT RUN (suite failed before upgrade)` row instead of silently showing nothing.

## Verification

- The report generator python was tested locally against simulated logs: normal logs produce the matrix rows (FAIL verdicts correct), an empty log produces the NOT RUN placeholder
- YAML parses cleanly

A companion PR on rustfs/dashboard renders this matrix above the case table.


majinghe:
Follow-up: matrix table headers and the not-run placeholder row are now English (From Version / To Version / Result / NOT RUN). Re-verified the report generator against simulated logs and YAML parsing.

PR #7141, opened 2026-09-04
ci: render step results and version in workflow reports

majinghe:
## Background

The heal and pool-expansion reports were only a raw log tail: no structured indication of which steps passed, no overall verdict, and no version information for the cluster under test. On the dashboard both suites rendered as unstructured markdown.

The suites (rustfs/auto-testing#37) now emit machine-readable lines:

```
[HEAL-STEP] <n> <desc> PASS|FAIL      [POOL-STEP] <n> <desc> PASS|FAIL
[HEAL-VERSION] <ver> (node <n>)       [POOL-VERSION] <ver> (node <n>)
[HEAL-RESULT] PASS|FAIL <detail>      [POOL-RESULT] PASS|FAIL <detail>
```

## Changes

`.github/workflows/rustfs-heal-test.yml` and `rustfs-pool-expand-test.yml`: the report generators parse those lines and emit a `## Step Results` section before the log tail —

- Version under test (captured in place via `rustfs --version` on a node)
- Overall result (PASS/FAIL with detail)
- A per-step table (Step | Description | Result)

When a run dies before any step reports (old script revision or an early crash), the table shows a single `NOT RUN (no step result lines found)` row instead of silently showing nothing.

## Verification

Both generator scripts were extracted and tested locally against simulated logs:

- A partial heal run (steps 1-4 PASS, step 5 FAIL) renders the version line, `FAIL — step 5 failed`, and the 5-row table with the FAIL on step 5
- A passing pool run renders `PASS — all steps passed` with its step table
- An empty log renders the NOT RUN placeholder
- Both workflows parse as valid YAML

Companion: rustfs/auto-testing#37 (suite scripts), rustfs/dashboard (frontend rendering of the same section).

### DIFF
--- a/.github/workflows/rustfs-heal-test.yml
+++ b/.github/workflows/rustfs-heal-test.yml
@@ -152,6 +152,60 @@ jobs:
           else
             PACKAGE_SOURCE="${RUSTFS_NIGHTLY_PACKAGE_URL}"
           fi
+          STEPS_TABLE="/tmp/rustfs-heal-steps.md"
+          python3 - "${LOG_FILE}" "${STEPS_TABLE}" <<'PY'
+          import re
+          import sys
+
+          log_file, out_file = sys.argv[1], sys.argv[2]
+          ansi = re.compile(r'\x1b\[[0-9;]*m')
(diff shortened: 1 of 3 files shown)


overtrue (approved):
LGTM

PR #7148, opened 2026-09-04
fix(site-replication): bound and order outage recovery

cxymds:
## Related Issues
Closes https://github.com/rustfs/backlog/issues/2071

## Summary of Changes
- Split outage recovery into a lightweight 30-second retry drain and the existing 600-second full reconciliation pass. The lightweight pass rotates across at most four peers and replays one complete current-state bucket chain per peer: make, replayable metadata, then replication configuration.
- Persist every peer's destructive bucket-delete reservation before deleting the local bucket. A full or unwritable non-evictable queue now fails the local delete closed, and destructive, IAM, and bucket-metadata liabilities are never FIFO-evicted.
- Fence local bucket-set changes during `replicate add` with a distributed admission lock, while retaining distributed per-bucket ordering for create, delete, and backfill work.
- Fence replay settlement by the exact observed retry revision and fence peer edits with a distributed generation. Peer delivery revalidates current topology and credentials before sending.
- Record failed IAM deletion bodies for faithful replay, then resend a stable current snapshot without clearing a newer same-entity failure.
- Rebuild missed bucket creation from authoritative current metadata. Default versioning, derived-only replication rules, and the empty object-lock sentinel are recreated by make/configure and are not sent redundantly; operator-authored metadata remains in the ordered replay chain.

## Verification
- `make pre-commit`
- `cargo test -p rustfs --lib site_replication` (297 passed)
- `NO_PROXY=127.0.0.1,localhost no_proxy=127.0.0.1,localhost cargo test -p e2e_test 'replication_extension_test::test_site_replication_replays_bucket_created_during_peer_outage_real_dual_node' -- --exact --nocapture` (1 passed)

## Impact
- A recovered peer can converge a bucket created during its outage on the 30-second retry cadence, including current bucket metadata and replication configuration.
- Local bucket deletion now depends on durable retry-intent persistence and returns a retryable failure before changing the namespace if that reservation cannot be committed.
- Public bucket create/delete operations can wait behind a site-replication add transition; the transition fails closed if its preflight bucket set changes.
- The lightweight drain sends at most 11 sequential requests per selected peer and selects at most four peers per round. The full drain retains the existing operator-repair path for larger snapshots and topology work.
- New persisted fields use serde defaults, and peer-edit fencing is carried in optional query parameters, preserving rolling-upgrade readability. No new public endpoint or configuration knob is introduced.

Adversarial review verdicts:
- Correctness: pass. Current-state bucket replay is ordered make -> metadata -> configure, and every settlement is scoped to the observed revision or generation.
- Concurrency and durability: pass. Delete intent is committed before local mutation; add admission and per-bucket locks follow the documented lock order; no new lock cycle was identified.
- Security: pass. Peer membership and current credentials are revalidated at delivery, and secrets are neither persisted in retry state nor logged.
- Compatibility: pass. Legacy state defaults safely, old peers ignore the optional fence query, and destructive events remain operator-only rather than being replayed without an incarnation fence.
- Performance: pass with the bounded costs described above; peer work is concurrent and the lightweight peer window rotates to prevent cross-peer starvation.

## Additional Notes
Known residuals:
- A live destructive request that times out after reaching a peer still has no persisted bucket-incarnation fence. It remains operator-visible and is never automatically replayed, avoiding deletion of a later same-name bucket.
- The full 600-second drain still holds repair-execution coordination across its network backlog. Lock-acquisition failures remain visible and retryable.

Rollback is a normal revert of this PR. Persisted additive fields are ignored by older readers; any already-recorded liabilities remain available for operator repair.

---


overtrue (approved):
LGTM

rustfs/src/site_replication/retry.rs
```diff
@@ -1005,7 +1022,9 @@ pub(crate) fn deferred_site_replication_retry_events(
         .filter(|event| !site_replication_retry_backoff_elapsed(event, now))
         .filter(|event| {
             event.updated_at.is_none_or(|updated_at| {
-                now.unix_timestamp().saturating_sub(updated_at.unix_timestamp()) >= SITE_REPLICATION_RETRY_DRAIN_BASE_BACKOFF_SECS
+                event.peer_unreachable
```
overtrue, line 1271: Promoting a failed make event here does not restore the configure step that the original hook skipped after the transport error. Replay sends only `bucket_make_ops`, then settles the event; the receiver make arm does not install replication rules, so the bucket stays unconfigured until the 600-second reconciler. Replay the matching configure operation before settling make.

overtrue (changes requested):

rustfs/src/site_replication/retry.rs
cxymds: Fixed in `2975949c8`. A failed `make-with-versioning` replay now rebuilds the current bucket plan and sends the matching `configure-replication` operation before settlement. The 30-second path uses a targeted single-bucket plan and an observed-event settlement fence, so a newer same-path failure is preserved. The original 600-second full drain continues to handle snapshot and peer-edit recovery.

rustfs/src/app/bucket_usecase.rs
```diff
@@ -1397,16 +1405,31 @@ impl DefaultBucketUsecase {
...
+                    &operation_bucket,
+                    &DeleteBucketOptions {
+                        force,
+                        ..Default::default()
+                    },
+                )
+                .await;
+            if delete_result.is_ok() {
+                crate::storage::invalidate_bucket_validation_cache(&operation_bucket);
+                if let Err(err) = site_replication_delete_bucket_hook(&operation_bucket, force).await {
```
overtrue: The local delete commits before any durable retry intent exists, and this error is only logged. A crash, cancellation, or state/runtime read failure after `delete_bucket` succeeds leaves the peer bucket behind with no queue record. The per-bucket lock orders live requests but does not close this durability window; reserve a recoverable intent before the local delete and fail closed if it cannot be persisted.

rustfs/src/site_replication/retry.rs
```diff
@@ -227,6 +249,7 @@ pub(crate) fn upsert_site_replication_retry_event(
         last_error: detail,
         updated_at: Some(now),
         edit_generation: generation,
+        peer_unreachable,
         deletions_recorded: false,
     });
     if queue.len() > SITE_REPLICATION_RETRY_QUEUE_LIMIT {
```
overtrue: This FIFO eviction can silently discard an older destructive delete liability when the 257th failure is recorded. Since destructive entries are intentionally not auto-replayed, that row is the only repair evidence. Evict only non-destructive entries and preserve/fail closed when the queue is entirely destructive.

rustfs/src/site_replication/hooks.rs
```diff
@@ -329,6 +350,47 @@ pub(crate) fn site_replication_bootstrap_plan(info: &SRInfo) -> S3Result<SiteRep
...
+    created_at: Option<OffsetDateTime>,
+    lock_enabled: bool,
+) -> SiteReplicationBootstrapPlan {
+    let bucket = SRBucketInfo {
+        bucket: bucket.to_string(),
+        created_at,
+        object_lock_config: lock_enabled.then(String::new),
+        ..Default::default()
+    };
+    SiteReplicationBootstrapPlan {
```
overtrue: The targeted recovery plan omits the matching bucket metadata. If the initial make request failed, metadata was never sent; this two-request replay then makes and configures the bucket and settles the event, so tags, policy, quota, and other metadata never converge. Include the ordered metadata tasks and count them, or leave this event for the complete heavy replay when it exceeds the lightweight budget.

rustfs/src/site_replication/hooks.rs
```diff
@@ -406,7 +538,43 @@ pub async fn site_replication_delete_bucket_hook(bucket: &str, force_delete: boo
             .append_pair("operation", operation)
             .finish()
     );
-    broadcast_site_replication_json(&path, &serde_json::json!({})).await
+    let runtime = {
+        // Serialize the peer snapshot with topology commits, then release the
+        // global lock before waiting for repair coordination or peer I/O.
+        let _bucket_op_guard = SITE_REPLICATION_BUCKET_OP_LOCK.read().await;
+        let Some(runtime) = runtime_site_replication_targets().await? else {
```
overtrue: The new per-bucket lock does not fence this topology commit. A joining peer can already own bucket X while this site still reports replication disabled; a concurrent local delete then skips its hook, this commit succeeds, and the post-add listing omits X, leaving the joining peer's copy behind. Add-in-progress mutation admission or a recoverable post-commit generation is still needed across this transition.

overtrue (changes requested):
Found four blockers on this head; details inline.

rustfs/src/app/bucket_usecase.rs
```diff
@@ -1331,23 +1332,34 @@ impl DefaultBucketUsecase {
...
-                    ..Default::default()
-                },
-            )
-            .await;
+        // Keep the local namespace mutation and its peer hook ordered across
+        // every node in this site. Otherwise a delete waiting for repair
+        // coordination can arrive after this create on remote sites.
+        let operation_bucket = bucket.clone();
+        let operation_store = store.clone();
+        let make_result = with_site_replication_bucket_mutation_lock(store, &bucket, move || async move {
```
overtrue, line 1340: This sender-side lock does not fence a request that timed out after the peer accepted it. The old peer handler can resume after a later same-name make/delete and apply out of order because the wire operation has no persistent incarnation or generation fence. Carry a monotonic bucket-mutation generation and reject stale operations at the receiver; taking the same lock in the receiver alone risks a cross-site lock cycle. Add a timeout/barrier regression proving a delayed old operation cannot overwrite the newer incarnation.

rustfs/src/site_replication/retry.rs
```diff
@@ -1167,27 +1485,118 @@ pub(crate) async fn drain_site_replication_retry_queue_inner() -> S3Result<()> {
...
     // replayable bucket-op entry in that window changes the preflight token
     // and execute fails safe with "preflight is stale" — the operator
-    // re-runs the dry-run. Lock order matches repair: lifecycle guard (held
-    // by the reconcile tick) -> repair execution lock -> state object lock
-    // inside the send bookkeeping. An operator repair holding the lock makes
-    // this tick skip after the lock-acquire timeout.
+    // re-runs the dry-run. The lock also elects one server to probe
+    // and replay the queue; after acquiring it, reload state so a settled
+    // event or deleted bucket cannot be replayed from this admission snapshot.
     with_config_object_write_lock(store, SITE_REPLICATION_REPAIR_EXECUTION_LOCK_PATH.to_string(), move || async move {
```
overtrue, line 1569: This heavy drain still holds the repair-execution lock across the entire network replay. One peer can process up to 256 events serially at a 10-second request timeout, blocking operator repair on the same lock for tens of minutes. Use a separate drain-election lock or release the repair lock between bounded persisted batches; test that manual repair can enter while a heavy backlog is stalled.

overtrue (changes requested):
The existing unresolved blockers still apply; these two ordering and availability gaps also need to be closed.

rustfs/src/app/bucket_usecase.rs
overtrue: Leaving this as a documented residual: it predates the recovery change, and this head does not expand the delete durability window.

rustfs/src/site_replication/retry.rs
overtrue: Leaving this as a documented queue-cap residual; the current change does not alter eviction behavior.

rustfs/src/app/bucket_usecase.rs
overtrue: Leaving this unchanged: the stale completion is plausible, but there is no deterministic pre-lock overtaking path or failing barrier test on this head.

rustfs/src/site_replication/retry.rs
overtrue: Leaving this as the documented full-drain residual; lock acquisition fails visibly and retryably, and no deterministic correctness failure is shown here.

rustfs/src/site_replication/hooks.rs
overtrue: Fixed in c00f7e6e6. Targeted recovery now rebuilds current bucket metadata, replays make then metadata then configure within the bounded chain, and keeps the event queued if metadata cannot be loaded.

rustfs/src/site_replication/hooks.rs
cxymds: Fixed in `1170ef666` and included in head `9b6ba84f1`. `replicate add` now holds a distributed bucket-mutation admission write lock from before the preflight bucket snapshot through remote joins and the local topology commit. Public S3 create/delete take the matching admission read lock plus the per-bucket write lock, so the bucket set cannot change in the transition window. The final bucket-set check remains as a fail-closed invariant. Covered by `add_admission_starts_before_preflight_and_rejects_bucket_set_changes`; the full site-replication suite passes (297 tests).

rustfs/src/app/bucket_usecase.rs
cxymds: Follow-up: this residual is now closed in `1170ef666` and head `9b6ba84f1`. Delete hooks persist an all-peer destructive reservation before the local namespace delete; a full or unwritable non-evictable queue fails the delete closed. Local delete failure cancels only the exact observed reservation revision, while successful local deletion keeps failed peer deliveries as non-auto-replayed operator liabilities.

rustfs/src/site_replication/retry.rs
cxymds: Follow-up: this residual is now closed in `1170ef666` and head `9b6ba84f1`. Queue insertion no longer FIFO-evicts destructive, IAM, or metadata liabilities. Only reconstructable peer-edit and non-destructive bucket make/configure entries are eligible for eviction; when all 256 slots are non-evictable, persistence returns a retryable failure instead of silently dropping a liability.

PR #7158, opened 2026-09-05
test(e2e): add distributed cluster regression coverage

loverustfs:
## Description

Pool decommission is not reliable after expanding a cluster to four nodes and four pools.

A strict distributed end-to-end test either receives HTTP 500 when starting decommission or observes the operation enter a failed or indefinitely running state.

## Environment

- Current RustFS `main`
- 4 RustFS nodes
- 4 drives per node
- 4 server pools
- Each pool uses four isolated filesystem-backed drives
- Objects are written to pool 0 before adding the remaining pools

## Steps to Reproduce

1. Start RustFS with one four-drive pool.
2. Create a bucket and upload 96 objects.
3. Add three additional four-drive pools, resulting in four nodes, four pools, and sixteen drives.
4. Restart the cluster and wait for every node to become ready.
5. Confirm that all previously uploaded objects remain readable.
6. Start decommission for pool 0:

   ```http
   POST /rustfs/admin/v3/pools/decommission?pool=0&by-id=true
   ```

7. Poll:

   ```http
   GET /rustfs/admin/v3/decommission/status
   ```

## Expected Behavior

- The start request returns a successful response.
- Pool 0 enters an active decommission state.
- Objects are moved to the remaining pools.
- The operation reaches a clean terminal state with:
  - `complete=true`
  - `failed=false`
  - `canceled=false`
  - no unresolved entries
  - non-zero object and byte movement counters
- All object contents and checksums remain unchanged.

## Actual Behavior

The strict test has observed an immediate HTTP 500 response:

```text
InternalError
```

Other runs accepted the start request but later failed or stalled with server-side errors including:

```text
decommission bucket completion save failed:
stale pool metadata update rejected
```

and:

```text
decommission target mutation reservation identity is stale
```

Reducing decommission entry concurrency to one did not resolve the failure.

The same expanded topology successfully completes rebalance with non-zero movement, which indicates that the cluster, drives, and expanded pool map are otherwise operational.

## Suspected Cause

The failures appear to involve inconsistent decommission metadata or capacity-reservation state across nodes.

Possible failure paths include:

- A worker updates decommission progress from an older in-memory `PoolMeta` snapshot after another node has persisted a newer revision.
- Capacity reservation identity changes or is not propagated atomically before target mutations begin.
- Workers start before every peer has reloaded the newly persisted decommission metadata.
- Bucket-completion persistence replaces a complete pool entry instead of rebasing a monotonic progress update onto the latest durable state.

There is also an error-mapping problem: decommission metadata and capacity conflicts can fall through to the generic `InternalError` response, hiding the actionable server-side reason from the administrator.

## Data Integrity

No confirmed object loss has been observed. The strict test stops on the control-plane failure and does not report decommission success without a clean terminal state and post-operation checksum verification.

However, the issue prevents administrators from reliably decommissioning a valid expanded pool and may leave the operation blocked or partially completed.

## Additional Notes

The distributed test intentionally does not accept HTTP 500, missing status fields, zero-progress completion, failed counters, canceled state, or unresolved entries as successful decommission.


crates/e2e_test/src/distributed/harness.rs
```diff
@@ -0,0 +1,1108 @@
...
+    F: FnMut() -> Fut,
+    Fut: std::future::Future<Output = TestResult<bool>>,
+{
+    let deadline = Instant::now() + timeout;
+    let mut delay = Duration::from_millis(50);
+    loop {
+        let last_error = match probe().await {
+            Ok(true) => return Ok(()),
+            Ok(false) => format!("{label} still false"),
+            Err(error) => error.to_string(),
```
overtrue, line 451: [P2] Propagate fatal probe errors. `retrying_put` and `retrying_get_equals` return `Err` for failures such as InternalError, but this branch turns every error into another retry. An InternalError followed by a successful request therefore passes the movement test; an extracted-function probe confirms this. Return probe errors here and let callers explicitly map retryable availability/convergence states to `Ok(false)`.

crates/e2e_test/src/distributed/harness.rs
```diff
@@ -0,0 +1,1108 @@
...
+    }
+    let info = pool
+        .get("decommissionInfo")
+        .ok_or_else(|| format!("pool {pool_id} has no decommissionInfo: {pool}"))?;
+    let complete = info.get("complete").and_then(serde_json::Value::as_bool) == Some(true);
+    let status_text = pool.get("status").and_then(serde_json::Value::as_str).unwrap_or("");
+    let pool_status = pool.get("poolStatus").and_then(serde_json::Value::as_str).unwrap_or("");
+    let terminal = status_text.eq_ignore_ascii_case("complete") && pool_status.eq_ignore_ascii_case("decommissioned");
+    let moved_data = nonzero_u64(info.get("objectsDecommissioned")) || nonzero_u64(info.get("bytesDecommissioned"));
+    Ok(complete && terminal && moved_data)
```
overtrue, line 754: [P2] Require the promised completion fields and both movement counters. With the expected status strings, `decommissionInfo` containing only `complete: true` and `objectsDecommissioned: 1` passes this check: missing failure flags/counters are treated as clean, and the OR accepts missing or zero bytes. All these fixtures write non-empty user objects, so this can hide a broken status response or byte counter. Validate the required boolean/numeric fields and require positive object and byte counts; add negative fixtures for missing fields and zero bytes.

crates/e2e_test/src/distributed/concurrent_data_movement_test.rs
```diff
@@ -0,0 +1,68 @@
...
+    dist.create_bucket(&bucket).await?;
+    let baseline_client = dist.client(0)?;
+    let inventory = put_inventory_retrying(&baseline_client, &bucket, 96, 256 * 1024, Duration::from_secs(30)).await?;
+    dist.expand_to_four_pools().await?;
+
+    start_decommission(&dist.cluster, DECOMMISSION_POOL_ID).await?;
+    wait_for_decommission_active(&dist.cluster, DECOMMISSION_POOL_ID, Duration::from_secs(30)).await?;
+
+    let clients = Arc::new(dist.clients()?);
+    let barrier = Arc::new(Barrier::new(16));
```
overtrue: [P2] Establish overlap between writes and decommission. The active helper also accepts queued work, and the barrier only synchronizes writers with each other. All 16 small PUTs can finish before migration starts, or migration can finish between this status read and the first PUT; every later assertion still passes. Keep the workload running while checking movement progress and assert successful writes within that interval, or synchronize with a migration hook. The same gap affects the sequential operations in `s3_during_data_movement_test.rs`.

overtrue (changes requested):
Three gaps in the new tests can hide regressions or miss concurrent data movement. I reproduced the helper behavior with extracted-function probes; actionlint and the test-wiring guard pass. The full distributed suite was not run locally.

crates/e2e_test/src/distributed/harness.rs
overtrue, line 451: Propagate fatal probe errors here. `retrying_put` and `retrying_get_equals` already map only explicit availability errors to `Ok(false)`, but this branch turns `InternalError` into another retry, so an error followed by success passes. Add an `Err` then success regression.

crates/e2e_test/src/distributed/harness.rs
```diff
@@ -0,0 +1,1108 @@
...
+        return Err(format!("{reason}: {pool}").into());
+    }
+    let info = pool
+        .get("decommissionInfo")
+        .ok_or_else(|| format!("pool {pool_id} has no decommissionInfo: {pool}"))?;
+    let complete = info.get("complete").and_then(serde_json::Value::as_bool) == Some(true);
+    let status_text = pool.get("status").and_then(serde_json::Value::as_str).unwrap_or("");
+    let pool_status = pool.get("poolStatus").and_then(serde_json::Value::as_str).unwrap_or("");
+    let terminal = status_text.eq_ignore_ascii_case("complete") && pool_status.eq_ignore_ascii_case("decommissioned");
+    let moved_data = nonzero_u64(info.get("objectsDecommissioned")) || nonzero_u64(info.get("bytesDecommissioned"));
```
overtrue: Require the promised completion fields and both movement counters. A terminal response with `objectsDecommissioned: 1` and `bytesDecommissioned: 0` passes this OR despite the non-empty fixtures, while missing failure fields are treated as clean. Add negative fixtures for missing fields and zero bytes.

crates/e2e_test/src/distributed/concurrent_data_movement_test.rs
overtrue: Prove overlap with decommission. `decommission_active` also accepts queued work, and this barrier only synchronizes writers, so every PUT can finish before movement starts and the test still passes. Synchronize with actual movement progress and assert successful writes inside that interval; the sequential decommission/rebalance case has the same gap.

overtrue (changes requested):

crates/e2e_test/src/distributed/harness.rs
overtrue: Leaving this unchanged: wait_until is a convergence poll and cannot false-pass; decommission uses a separate poller that already propagates predicate errors.

overtrue: Leaving this unchanged: wait_until is a convergence poll and cannot false-pass; decommission uses a separate poller that already propagates predicate errors.

crates/e2e_test/src/distributed/harness.rs
overtrue: Leaving this unchanged: the production response always serializes the failure flags, and object-or-byte progress intentionally supports zero-byte objects.

crates/e2e_test/src/distributed/harness.rs
overtrue: Leaving this unchanged: the production response always serializes the failure flags, and object-or-byte progress intentionally supports zero-byte objects.

overtrue: This nonzero workload writes 96 128-KiB objects, so both movement counters must advance. Using OR lets a bookkeeping regression that only updates one counter pass, and the later inventory/body checks do not inspect these counters. Require both values here, or pass an explicit expectation for zero-byte workloads, and add regressions for each one-sided counter state.

overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #7164, opened 2026-09-05
fix(ecstore): reject stalled ODM pagination before merging pages

overtrue:
## Related Issues

Refs rustfs/backlog#2254 and rustfs/backlog#2255. Part of rustfs/backlog#2248.

## Summary of Changes

ODM now rejects truncated source pages whose next cursor is missing, empty, or unchanged. The merger also rejects a cursor cycle within a request. Validation happens before accepting source entries, so a stalled page cannot silently truncate a successful listing.

The local-only fallback clears buffered source state and resumes from local data. Empty pages with an advancing opaque cursor remain valid, including folded-prefix probing. The existing two-fetch limit remains unchanged.

Tests exercise the actual source SDK and handler, both error policies, folded prefixes, empty advancing pages, and an independent listing oracle with 256 generated cases under a fixed seed and a separate 5,300-key corpus.

## Verification

The final integrated CI-profile run at [e1f8cb9b8](https://github.com/rustfs/rustfs/commit/e1f8cb9b890ffb617d7506a21d2bb40d962552c2), based on `b33693fc1`, passed all 208 selected tests on macOS with no failures or retries. The JUnit report contains those 208 tests plus one successful setup script. This combined run covers the companion changes; each PR still needs its own branch CI.

Native validation used `cargo nextest run` with the three affected library targets, `rustfs-ecstore/test-util`, `--profile ci`, and the combined focused selection. All seven handler tests passed, including actual SDK requests and both error policies. `cargo fmt --all --check`, `git diff --check`, `scripts/check_s3s_footprint.sh`, and `scripts/check_layer_dependencies.sh` passed. The tests use the existing S3 facade; no dependency baseline or assertion was relaxed.

Independent oracle mutation checks failed as expected when cursor validation or fallback buffer clearing was removed. Two independent reviews passed: correctness, compatibility, and performance; concurrency, errors, test coverage, and simplicity. The branch includes the test-only map-type fix from #7161.

The CI follow-up renames the private cursor error variants and collapses an equivalent test condition to satisfy existing Clippy rules. At integrated head `0b8ccae25`, all 23 selected ECStore tests passed, including all 21 list-through/source-page tests. `CARGO_BUILD_JOBS=2 cargo clippy -p rustfs-ecstore --lib --tests --features test-util -j 2 -- -D warnings` also passed after the cursor fix. The new branch head still requires its own complete CI.

## Impact

Malformed source pagination follows the configured error policy: propagate the source failure or return a local-only page. Continuation tokens remain opaque; no lexical ordering, new cursor format, or extra fetch budget is introduced. This covers the ODM portion of the differential corpus; walker/index equivalence remains separate work.

## Additional Notes

Keep this as a draft until branch CI and PR review are complete.


crates/ecstore/src/bucket/on_demand_migration/list_through.rs
```diff
@@ -387,15 +421,19 @@ impl ListThroughMerger {
             Some(last) => last.next_token.clone(),
             None => state.start.token.clone(),
         };
-        // A truncated page without a cursor cannot be continued; treating the
-        // side as finished is the only alternative to looping on it forever.
-        state.more = is_truncated && next_token.is_some();
+        validate_list_page(is_truncated, token.as_deref(), next_token.as_deref())?;
+        // Also reject a cycle through an earlier page in this bounded fetch.
```
overtrue, line 425: The two-fetch cap does not bound a cursor cycle across client requests. Empty truncated source pages with A→B→C→A produce another empty truncated merged page on every call; per-request history is reset and `last_key` never advances, so a normal paginator loops forever. Carry a bounded no-progress count or cursor proof in the merged token and fail or degrade after the limit; add a regression that chains the returned token across A/B/C calls.

overtrue: Follow-up #7197 adds a cross-request no-progress budget with strict v2 readers and default-off first issuance. The real bucket-usecase cycle regression fails on the old implementation and passes with issuance enabled; the legal empty-page control is retained. Fleet rollout is deliberately still open: every reader must be upgraded before issuance is enabled, so this does not claim the default v1 loop is already eliminated.

PR #7169, opened 2026-09-05
fix(lifecycle): correct expiration and transition evaluation

loverustfs:
## Related Issues

N/A.

## Summary of Changes

Fix lifecycle action selection, version-retention handling, and configuration validation. An object whose transition and expiration deadlines have both passed now selects expiration, so an unavailable tier cannot indefinitely postpone deletion. Event selection retains a deterministic total order and uses deadlines to choose within each action class.

- Evaluate noncurrent expiration and transition retention counts independently. An unmet expiration count no longer skips the transition in the same rule, and `NoncurrentVersionTransition.NewerNoncurrentVersions` now gates transition eligibility.
- Reject more than one current or noncurrent transition per rule, including disabled rules. Existing stored multi-entry arrays are also excluded from transition evaluation instead of silently executing their first entry. Independent expiration actions remain eligible.
- Reject `Expiration` containing both `Days` and `Date`.
- Allow restore-copy expiration under Object Lock while preserving the logical version and remote data, including fallback when retention blocks a competing logical expiration. Require a completed transition before scheduling restore-copy cleanup, and retain replication blocking.
- Add nine regression tests covering the six reviewed issues and their boundaries. Update the independent property-test oracle to assert expiration precedence, and document supported configuration and execution limits.

## Verification

- `cargo test -p rustfs-lifecycle --lib --no-fail-fast` — **136 passed, 0 failed, 0 ignored**.
- `bash scripts/check_doc_paths.sh` — passed.

The tests cover retention counts below, at, and above the threshold; absent, zero, negative, and maximum count values; independent age and count conditions; single-object evaluation without a version group; both transition-array orders and enabled/disabled rules; restored current and noncurrent versions under legal hold, with and without competing logical expiration; pending/failed replication; and incomplete or unknown transition states.

No live remote-tier integration tests or full-workspace tests were run locally for this change. The focused library target compiles the changed Rust implementation and tests; broader verification remains with CI.

## Impact

Configuration compatibility changes intentionally: multiple transition entries and simultaneous expiration Days/Date now return `InvalidArgument` through the existing S3 validation boundary. Operators with stored multi-entry transition arrays must select a single destination. Multi-stage transition chains remain unsupported, including chains expressed as overlapping rules.

Counted noncurrent transitions require complete version-group evaluation. The existing single-object manual-transition and immediate-enqueue paths conservatively defer positive retention-count rules to the lifecycle scanner. This preserves the retention constraint but can delay those transitions until scanning occurs.

No public Rust signatures, persisted metadata formats, dependencies, remote request defaults, synchronization, or queue formats change. No tests are ignored and no gates are bypassed. Rollback is a code revert; previously expired objects cannot be recovered by reverting the evaluator.

Two sequential adversarial passes covered the final implementation:

- Correctness: deadline conflicts, retention boundaries, independent action eligibility, and deterministic selection are covered by regression and property tests.
- Simplicity: changes stay in the existing validator and evaluators; no new scheduling or persistence abstraction.
- Test coverage: the six reproductions now pass, with unsupported arrays tested for explicit rejection and safe handling of stored configurations.
- Compatibility: unsupported configurations and single-object limitations are documented; valid single-stage transitions remain covered.
- Security: temporary-copy cleanup and its fallback preserve the logical-version retention guard; incomplete transitions and pending replication still block cleanup.
- Concurrency and durability: no storage commit or locking changes; existing execution-layer identity revalidation remains in place. Crash recovery was not exercised locally.
- Performance: added eligibility checks are constant-time and reuse the existing candidate traversal; no benchmark or throughput claim.

## Additional Notes

The action-precedence and independent-expiration behavior was cross-checked against [MinIO's lifecycle evaluator](https://github.com/minio/minio/blob/7aac2a2c5b7c882e68c1ce017d8256be2feea27f/internal/bucket/lifecycle/lifecycle.go#L412), while retaining RustFS's total-order selection rather than copying the upstream comparator. [AWS documents the noncurrent-transition retention count](https://docs.aws.amazon.com/AmazonS3/latest/API/API_NoncurrentVersionTransition.html); the compared MinIO type does not implement that field.

Broader queue persistence, multi-stage tier migration, and performance work are outside this change.

---


crates/lifecycle/src/core.rs
```diff
@@ -4159,6 +4159,270 @@ mod tests {
...
+                successor_mod_time: Some(datetime!(2020-01-02 00:00:00 UTC)),
+                version_id: Some(Uuid::from_u128(1)),
+                size: 1024 * 1024,
+                ..Default::default()
+            }
+        }
+
+        #[test]
+        #[serial]
+        fn noncurrent_transition_retains_the_requested_newer_versions() {
```
overtrue, line 4212: These regression tests are intentionally red: the PR reports all seven failing, so this head makes every lifecycle test run fail. Please keep this draft until the production fixes land; do not merge failing tests.

overtrue (changes requested):

crates/lifecycle/src/core.rs
```diff
@@ -4159,6 +4159,270 @@ mod tests {
...
+                        IlmAction::TransitionVersionAction
+                    ],
+                    "the two newest noncurrent versions must remain in their current storage class"
+                );
+            });
+        }
+
+        #[test]
+        #[serial]
+        fn expiration_retention_does_not_skip_an_independent_transition() {
```
overtrue, line 4349: This ready PR adds seven enabled regressions that the reported exact-head run still fails (127 passed, 7 failed), including this accepted lifecycle case. Please fix the production evaluator/validator behavior, or explicitly reject unsupported configurations, and make all seven regressions pass before merging.

overtrue (changes requested):

crates/lifecycle/src/core.rs
loverustfs: Addressed in 8943f3839. The PR now includes the production evaluator and validator fixes: `cargo test -p rustfs-lifecycle --lib --no-fail-fast` passes all 136 tests with zero failures or ignored tests. The two unsupported multi-stage cases now assert explicit rejection in both array orders, including disabled rules. The PR was returned to draft during implementation, and its title and description are being updated to describe the fixes and compatibility limits. No failing regression is intentionally left enabled.

crates/lifecycle/src/core.rs
loverustfs: Addressed in 8943f3839. Noncurrent expiration and transition now apply their retention counts independently, so an unmet expiration count no longer skips a separately eligible transition. The accepted combined-policy regression passes, with additional below/at/above-threshold coverage. Unsupported multi-entry transition arrays are rejected at validation and excluded from runtime transition selection. All 136 lifecycle tests pass; formatting and the documentation path check also pass.

crates/lifecycle/src/core.rs
```diff
@@ -394,11 +404,20 @@ impl Lifecycle for BucketLifecycleConfiguration {
                 }
             }
             if let Some(transitions) = &r.transitions {
+                if transitions.len() > 1 {
```
overtrue, line 407: This rejects valid multi-stage lifecycle rules. S3 allows repeated Transition entries, including the standard 30-day STANDARD_IA then 90-day GLACIER shape, and NoncurrentVersionTransitions is also a list. Please keep deterministic stage selection instead of changing accepted XML into InvalidArgument.

overtrue (changes requested):

overtrue (approved):
LGTM

PR #7174, opened 2026-09-05
chore(deps): refresh SDKs and verify clock skew behavior

houseme:
## Related Issues

Refs rustfs/backlog#2240. Shared dependency prerequisite for Scanner/Heal V2 batch 1.

## Summary of Changes

Run cargo update followed by cargo upgrade and refresh the compatible AWS SDK, tokio-rustls and transitive lockfile selections. Preserve the existing outbound checksum and retry policies.

Add deterministic production-path clock-skew regression tests: two error codes with positive/negative offsets, retry re-signing, cross-operation cache reuse, and disabled/single-attempt policies. The existing aws-smithy-async package is referenced only as an ECStore dev-dependency to inject a controllable clock.

## Verification

- `cargo update`, then `cargo upgrade`: completed.
- `cargo metadata --no-deps --locked --format-version 1`: passed.
- `cargo check --locked -p rustfs-scanner -p rustfs-heal -p rustfs-tls-runtime -p rustfs-kms -j 4`: passed.
- `cargo test --locked -p rustfs-ecstore --lib remote_s3_clock_skew -j 4`: 2 passed on a fresh target and rerun after integration of main.
- `cargo build --locked --bin rustfs -j 4`: passed on the final integrated baseline.
- `cargo test --locked -p e2e_test --lib replication_target_matrix_test -j 4 -- --test-threads=1`, using the explicitly selected freshly built binary and localhost proxy bypass: 5 passed, covering all 24 modeled target/object cells.
- Formatting and diff checks passed. Initial localhost permission failures were rerun with the required local-listener permission; no test was skipped or relaxed.

## Impact

Verified local modeled target classes: source-version adoption, rejection of aws-chunked uploads, checksum requirements for Object Lock, and targets minting their own version IDs. Each target must accept and store the expected bytes; the matrix expectations were not changed.

The AWS runtime update changes default clock correction and retry classification, not just additive configuration. Real AWS/MinIO/SeaweedFS/Impossible Cloud endpoints, KMS/STS endpoints, Windows, and performance/real-target soak remain unverified; this is not a release-readiness claim.

Two independent implementation reviews: correctness and dependency resolution clear; compatibility covered by deterministic signing and local target contracts; security retains the actual SDK signer and explicit integrity policy; performance has no measured improvement claim; simplicity/test coverage clear after removing the real-clock timing flake.

## Additional Notes

[...]

PR #7181, opened 2026-09-05
test(scanner): add bounded ABBA validation harness

houseme:
## Related Issues

Refs rustfs/backlog#2266 and rustfs/backlog#2240. Shared prerequisite: [dependency PR #7174](https://github.com/rustfs/rustfs/pull/7174).

## Summary of Changes

Add a bounded ABBA validation runner and deployment-adapter contract around the existing scanner collector. Validate provenance, independent oracles, workload/target isolation, P1/P2 work accounting, front-end latency, noise and failure propagation without claiming synthetic performance results.

## Verification

- `bash scripts/test_scanner_validation_harness.sh`: passed with Python 3.12 via `RUSTFS_PYTHON`, including legacy collector checks and 13 Python unit tests.
- Synthetic 120-leg matrix and negative cases passed.
- Exact threshold, just-over-threshold, and unstable P1 walk/cold-walk control regressions passed.
- Shell/Python syntax, documentation-path and diff checks passed.
- The existing collector's SC2016 shellcheck diagnostic was reproduced on its unchanged baseline; no suppression was added.

## Impact

No storage runtime or scheduler change. Real deployment adapters, EC8+4 measured ABBA, crash and mixed-version experiments remain pending. Synthetic runs explicitly report performance as pending, not passed.

Two independent implementation reviews: Correctness, simplicity, compatibility, performance and test-coverage reviews found no unresolved issue.

## Additional Notes

This PR targets main so the repository PR CI runs. It includes the shared dependency commits until the prerequisite is merged; merge that PR first. The task-only change can be reviewed against the dependency branch. Delivery integrates main at 123967e729c74903a7f3d6dcc0a388736acd4f6c; the affected dependency/metadata checks were rerun. CI and formal PR approval are separate from local verification. Do not close the backlog task before its required evidence and merge are confirmed.


overtrue (approved):
LGTM

scripts/scanner_abba.py
```diff
@@ -0,0 +1,329 @@
...
+    inconclusive = False
+    failed = False
+    for offset in range(0, len(cells), 4):
+        group = cells[offset:offset + 4]
+        require([cell["leg"] for cell in group] == list(LEGS), "incomplete ABBA group")
+        a1, b1, b2, a2 = (cell["result"]["metrics"] for cell in group)
+        control = group[0]["comparison"] == "background"
+        drift = max(abs(a2[k] / a1[k] - 1) for k in ("p99_ms", "throughput_ops"))
+        repeat_drift = max(abs(b2[k] / b1[k] - 1) for k in ("p99_ms", "throughput_ops"))
+        noise = max(drift, repeat_drift) > 0.05
```
overtrue: These raw float ratios misclassify exact boundaries. For example, 105/100 produces slightly more than 5%, and 97/100 produces slightly less than -3%, so conforming measurements can become noise or failure. Please compare the thresholds robustly and cover exact-limit plus just-over-limit cases.

scripts/scanner_abba.py
```diff
@@ -0,0 +1,329 @@
...
+        a = {key: (a1[key] + a2[key]) / 2 for key in METRICS}
```
overtrue: P1 averages the walk counters without checking their repeatability. A1/A2 walk counts of 1000 and 10 with stable latency/throughput, plus B1/B2 of 100, passes the reduction gate even though the candidate is 10x worse than A2. Please treat unstable walk/cold-walk controls as inconclusive and add a regression for this case.

overtrue (changes requested):

scripts/scanner_abba.py
houseme: Fixed in 7f4bf4f8747204546f4ad0ab905c296e69c242e4 by using Decimal-based threshold comparisons and adding exact-limit plus just-over-limit coverage. Verified with the scanner harness tests.

scripts/scanner_abba.py
houseme: Fixed in 7f4bf4f8747204546f4ad0ab905c296e69c242e4 by marking unstable P1 walk and cold-walk controls inconclusive, with a regression matching the A1/A2 instability case. Verified with the scanner harness tests.

overtrue (approved):
LGTM

overtrue (approved):
Fixed two issues in 1dcbdda4a: an exact P1 boundary could fail after Decimal division, and live collection accepted missing or failed per-node metrics. Added regressions that fail before the fixes. All 15 Python tests and the existing collector self-tests pass, along with syntax and diff checks. No remaining blockers from this review. Real deployment performance is still pending.

overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #2303, opened 2026-03-27
build(deps): bump the dependencies group with 6 updates

houseme:
## Type of Change
- [ ] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [ ] Test/CI
- [ ] Refactor
- [x] Other: dep

## Checklist
- [x] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [x] Passed `make pre-commit`
- [ ] Added/updated necessary tests
- [ ] Documentation updated (if needed)
- [ ] CI/CD passed (if applicable)

## Impact
- [ ] Breaking change (compatibility)
- [ ] Requires doc/config/deployment update
- [ ] Other impact:

## Additional Notes

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

PR #7187, opened 2026-09-05
fix(storage): harden ODM and scanner publication

overtrue:
## Related Issues

Refs rustfs/backlog#2259 and rustfs/backlog#2288. Both remain open: this is partial release-audit remediation, not a GA-readiness declaration.

Includes the existing upstream repair from merged PR #7190 (head `5439aad3f8`) for main's duplicate local rename definitions; no separate repair implementation is introduced. Its original commits and unchanged canonical `local/commit.rs` are preserved.

Fixes rustfs/backlog#2283. Fixes rustfs/backlog#2285. Fixes rustfs/backlog#2286. Fixes rustfs/backlog#2296. Fixes rustfs/backlog#2301.

## Summary of Changes

- Make ODM inline, queued, and multipart write-back create-only at the storage commit boundary. Preserve configured local delete markers and reject unsupported namespace-lock topologies before mutation.
- Share completion results between coalesced pulls. Backfill observes failures and cancellation, reports incomplete outcomes, and never checkpoints beyond the first newly failed page. A takeover compares failures with its initial cumulative count, allowing a successfully replayed historical failure to advance without resetting failure history.
- Frame ODM LIST tokens so JSON-shaped object keys remain ordinary keys. Reject malformed framed tokens and corrupt source LIST entries. Preserve main's stalled-pagination checks before merging pages.
- Distinguish missing source keys from missing buckets, versions, and ambiguous failures; resolve bodyless HEAD 404s with a bounded HeadBucket probe. Revalidate conditional GETs against the actual GET representation, including wildcard precedence and missing validators.
- Fence authoritative scanner usage against ordinary PUT rename fanouts, quorum-acknowledged tails, and rollback work. Retain namespace ownership across caller cancellation and rollback coordinator failure, track begin/end generations, check pending state after the walk, reject publication leases while pending, and re-prove activity after remote leases are granted. Keep scan/cache admission movement-only so overlapping writes do not prevent namespace walks. Sample RPC generations after waiting for movement state.
- Integrate with main's typed local rename outcomes, rollback receipts, recovery evidence, and explicit tail-drained completion option without changing those contracts. The existing rollback coordinator and each undo child retain the namespace guard; no new production task, lock, or timeout is introduced.
- Preserve main's structural scoped-cache baseline while giving bucket work and completed set snapshots an activity-sensitive execution identity. A tail completing after a rejected scan invalidates both same-cycle bucket reuse and the persisted set-root fast path. Set snapshots also retain their starting cache revisions: a slower different execution cannot overwrite a snapshot committed while it was walking. Local and remote bucket requests use the same execution identity.
- Repair fixture defects exposed by full validation: wait for multipart and heal PUT rename tails instead of inferring completion from elapsed time or physical copy counts; make decommission's global injection budget survive restarted attempt counters while preserving the last real attempt; isolate manual tier-manifest pagination and rollback retries from automatic recovery and same-bucket contention, with a separate startup-recovery ownership test.

The decommission integration deliberately changes main's test-hook expectation for a successful final attempt: after a real failure consumes attempt 2, injecting again at attempt 3 would trigger the terminal-failure branch despite a successful migration. The existing factory and matrix are reused and extended with reset, real-failure, max, and max+1 cases; production retry behavior and strict failure-count assertions are unchanged.

## Verification

High-risk validation on macOS with Rust 1.98.0, nextest's default no-retry profile, a task-owned target directory, and localhost proxy bypass. Checks ran from `6e8f409b8936` with the local changes now committed as `a26025d2db`. Final head `5bf0fc11d6a7` includes current main `f053862aade5`; that final merge changes only the PR template relative to the verified source commit, with no runtime/build/test change.

- `cargo fmt --all --check`: passed.
- `make architecture-migration-check test-wiring-check`, `./scripts/check_logging_guardrails.sh`, and `make doc-paths-check planning-docs-check`: passed after the relevant final edits.
- `make clippy-check`: passed (all targets, `-D warnings`), including a 27.69s rerun after the final test-only assertion tightening.
- Final focused scanner nextest: 9 passed, run `37751f9b-c25e-42b2-ba8d-b1a0b3563a84`.
- `cargo nextest run --all --exclude e2e_test --status-level fail --final-status-level fail`: 16,687 passed in 630.438s, run `534f0a48-8f02-412f-a8c0-f73a752deb54`; 118 pre-existing skips, no retries or reported leaky tests.
- `cargo test --all --doc`: 46 passed, 30 pre-existing ignored examples; exit 0.
- `git diff --check`: passed.

The required `make pre-pr` components were completed incrementally, not represented as a new uninterrupted passing umbrella run. Unchanged guard/script components and the dedicated/production offline-enrollment E2E retain their earlier integration results; enrollment sources, manifests, lockfile, and gate inputs are unchanged. The production diff received fresh full nextest/doctest coverage. The subsequent test-only assertion tightening passed focused scanner tests and Clippy; no production behavior changed after the full nextest run. Script runner self-tests are not an actual fuzz campaign.

The prior published head passed all 16,588 selected nextest tests (run `efccc302-e306-44c1-9bd9-22415e916e1a`), 46 doctests, and the required gate components. Earlier focused runs passed 210 ODM/S3 tests, 32 scanner/ownership tests, two scanner fixture regressions, all six versioned/unversioned heal regressions, and five decommission/tier tests. Earlier failures and their diagnostic evidence remain in the linked backlog issues; they are not relabeled as passing runs. No guard, assertion, retry policy, or quarantine was weakened to obtain green results.

The new rollback regressions cover normal and early-ACK paths, cancellation while undo is blocked, and coordinator panic after child dispatch. The panic case waits for the original call to finish, proves a blocked child alone still prevents publication, then releases it and verifies the old object body on all four reopened disks. Removing per-child ownership makes the pending/generation assertion fail.

Failing-before/passing-after evidence: `coordinator_walks_during_pending_put_without_persisting_or_acknowledging_usage` failed in run `21ebdf74-8ee5-4e93-a2a4-c6de56d64d97`: traversal repeated, but persisted size was 5 rather than the expected 20 bytes. With the set-execution fence, it passes in the full and final focused runs above. It changes physical size after a deferred walk, holds the pending guard until the change completes, proves dirty generation is unchanged, and retries the same cycle. A second real two-cycle test pins stable structural baseline identity and changing full publication proofs. Backfill takeover separately reads persisted token progress from 2 to 4 after historical failure recovery, then proves a new failure pins token 4 through a later successful page.

The set-cache regression separately exercises a legacy snapshot with no execution identity, a newer commit followed by a stale writer, an unreadable starting revision, identical-execution reuse, and an empty set replacing a nonempty snapshot. The unidentified candidate uses current CAS revisions, so only the execution-identity guard can reject it. Existing MessagePack compatibility tests now cover the optional execution field in both directions; legacy readers still receive a map and ignore the added key.

Not run: real cloud-provider, mixed-version cluster, distributed fault-injection E2E lanes, or throughput/latency benchmarks. Local tests exercise the post-lease proof predicate and dirty-ACK rejection; successful multi-node lease acquisition followed by re-probe is statically traced, not execution-tested. Existing fixtures expose concrete authenticated RPC clients without a successful-lease transport hook, and that new harness is outside this batch. Local/mock success does not establish these release guarantees.

## Impact

- Write-back requires enabled namespace locking, one pool, and one erasure set. Unsupported topologies fail before mutation; source reads can still work, and backfill reports write-back failures. Cross-pool atomic commits are not implemented.
- Source HEAD misses require HeadBucket permission. Inaccessible buckets remain upstream errors rather than negative-cache hits. Required GET validators missing from the actual response fail closed; missing/negative LIST sizes are rejected while zero-byte objects remain valid.
- Upgrade every node and discard existing merged-list continuation tokens. Mixed-version continuation is unsupported. Ordinary local listing keys and default non-ODM PUT semantics are preserved.
- Scanner protection requires every writer node to be upgraded. It does not prove failed replicas have healed or track multipart and all other namespace mutations. Older verified snapshots can precede new writes; intermediate per-set/prefix caches retain approximate semantics. A prolonged unchanged pending tail can still defer convergence, cycle advancement, and rescans of Current caches, so indefinitely stalled I/O has no new lifecycle-progress guarantee.
- Disable ODM before rollback and restart merged listings; old binaries do not retain these create-only or token guarantees. Scanner set-cache metadata adds an optional map-encoded execution digest; missing legacy identities require refresh instead of same-cycle reuse. External Rust consumers constructing `DataUsageCacheInfo` with exhaustive struct literals must account for the added field. No scanner RPC or credential-encryption format migration is introduced. Older writers do not enforce the new cache fence, so rolling-upgrade safety is not claimed.

## Additional Notes

Independent adversarial review, including the main integration and final regression; the final set-cache follow-up received separate sequential role passes:

- Correctness: attacked concurrent PUT/delete, cross-domain commits, validator precedence, overlapping tails, rollback ownership, retry exhaustion, and same-cycle cache substitution; findings resolved with atomic checks, fail-closed topology checks, retained ownership, and execution identities.
- Simplicity: attacked helper duplication, one-off abstractions, and fixture scope; reused existing builders, rollback coordinator, fault registry, barriers, and decommission factory/matrix, with no unresolved finding.
- Security: attacked conditional bypass, token envelopes, error disclosure, internal flags, and publication leases; no unresolved finding. The final panic seam is test-only.
- Concurrency/durability: attacked coalescing, cancellation, checkpoint ordering, tombstone races, detached undo, rollback accounting, counter exhaustion, RPC sampling, recovery ownership, and a stale set writer arriving after a newer commit. Starting-revision comparison preserves existing lock/CAS fences; no unresolved finding within the stated scope limits.
- Compatibility: checked S3 ACK behavior, internal metadata, typed rename results, tail-drained completion, rollback evidence, and optional map-encoded cache identity with old-array/new-map reader tests. ODM token framing intentionally changes; required full-fleet upgrades and topology restrictions are explicit.
- Performance: ordinary PUT adds one guard allocation, four counter/generation updates, and O(disks) Arc operations; rollback reuses existing tasks with per-child Arc ownership. No new foreground hot-path lock, disk I/O, fsync, task spawn, or success log. Set scans capture the existing main revision plus a backup revision lookup once per set, not per object. Authoritative distributed publication adds one activity re-probe, and source HEAD misses add one bounded bucket probe; performance remains unbenchmarked.
- Test coverage: attacked commit-window reachability, topology rejection, pagination, retry boundaries, cache admission, caller cancellation, coordinator failure, and empty/legacy/stale set snapshots. The per-child ownership coverage finding was fixed with the barrier-controlled panic regression; the unchanged multi-node execution gap is explicitly documented above.

Credential sealing, safe cross-domain write-back, refresh/CAS semantics for `skip_existing=etag_or_size`, deletion/fallback policy questions, and the real-provider/mixed-version release matrix remain tracked in rustfs/backlog#2259 and related issues.


overtrue:
The published head `a1b1eec21f1ab59772d313fad3b56060f2d6eb58` passed the previously reported 16,588 nextest tests and 46 doctests. Main advanced during that validation, so the PR currently shows conflicts.

The conflicts with `19a29a70270507a0df931083d85ba1ab6f124cd7` are resolved locally, preserving the upstream typed rename/rollback contracts and stalled ODM pagination checks. Scanner namespace ownership now also follows the existing detached rollback coordinator and its undo children. Independent review identified a missing per-child ownership regression; a test-only coordinator-panic seam and barrier-controlled old-body readback test now cover it in both early-ACK modes.

Fresh integration validation has passed static guards, the offline enrollment E2E root gate, Clippy, and script self-tests. Workspace nextest is currently compiling; the integration update will be pushed after required validation completes. The final test-only addition also passed formatting, architecture, and test-wiring checks. No checks or assertions were relaxed.

Main has advanced again in the meantime. After this fixed snapshot is verified, I will check the current PR head/base and mergeability and handle any additional overlap. The broader release-audit and scanner-consistency limits remain open in rustfs/backlog#2259 and rustfs/backlog#2288.

crates/ecstore/src/bucket/on_demand_migration/backfill.rs
```diff
@@ -1204,10 +1204,13 @@ impl Job {
...
                 self.tick(false).await?;
             }
-            // Only advance the cursor once every pull of this page reported
-            // back, so a takeover re-lists at most this page.
+            // A persisted cursor certifies successful work, not just listing
+            // progress. Keep it at the first failed page for crash recovery.
             self.drain_all().await?;
-            self.checkpoint.continuation_token = page.next_continuation_token.clone();
+            cursor = page.next_continuation_token;
+            if self.checkpoint.failed == 0 {
```
overtrue: This checks the cumulative failure count, so after one failed page is persisted and taken over, `failed` stays nonzero forever. Even if that page and every later page succeed, no continuation token is durably advanced; repeated restarts keep rescanning from the first failure. Compare the failure count before and after each page, or track unresolved failures separately, and add a takeover regression proving a successfully retried page advances the persisted token.

overtrue:
The integration update is committed locally at `6e8f409b89`, based on main `7ba5cd6888`, but is not pushed yet. It includes the reviewed backfill takeover correction: cumulative historical failures no longer permanently block cursor progress, while any newly failed page still pins all later checkpoints. A persisted-checkpoint takeover regression covers both outcomes.

Integration review also found and fixed a same-cycle scanner cache issue: namespace completion must invalidate bucket work even while the structural scoped baseline remains reusable. The regression changes actual object size while publication is deferred, keeps dirty generation unchanged, then requires fresh traversal and the updated persisted size on the same cycle. Local and remote bucket requests share the activity-sensitive cache identity.

Final static checks passed. Clippy then exposed main's duplicate local rename definitions (rustfs/backlog#2300), so this branch now includes the existing repair commits from #7190 at `5439aad3f8`; canonical `local/commit.rs` and its tests remain unchanged. #7190 should land first. Clippy is rerunning, followed by focused and full tests. No failed, interrupted, or not-yet-run check is treated as passing. The earlier tier test ownership failure is tracked in rustfs/backlog#2301.

overtrue:
Main e2a921bc has a reproduced cursor compatibility regression after this merge: the retained progress_tokens_preserve_v1_bytes_and_validate_v2_counts test compiles and fails at line 1093 because the new encoder adds NUL framing to the old v1 bytes. Old bare v1/v2 cursors are also classified as Local, so a mixed-reader listing can restart and lose its existing budget. The compatibility repair and acceptance matrix are tracked in [backlog #2303](https://github.com/rustfs/backlog/issues/2303). The old wire golden stays intact.

Two independent reviews also found that the new namespace owner can end at an async timeout while the actual metadata rename continues. This is a separate static concurrency finding, tracked with exact source and the physical regression plan in [backlog #2304](https://github.com/rustfs/backlog/issues/2304). The physical reproduction is being prepared; it is not yet a demonstrated test failure. Passing results from earlier SHAs do not cover either merged boundary.

overtrue:
The upstream rename build repair #7190 has merged. This branch already contains its original repair commits; the final synchronization will preserve identical code.

The strengthened scanner retry regression caught an additional stale set-root fast path after bucket traversal had been fixed. The local follow-up now fences both bucket and set reuse with execution identity and rejects a different execution whose starting cache revision has changed. The original 5-to-20-byte assertion now passes, together with legacy cache compatibility, stale-writer, identical-execution, empty-set, and scoped-baseline tests (9 selected tests passed, run `24188c03-b72d-4187-9d05-412db160099a`). Detailed evidence is in rustfs/backlog#2288.

Final-source Clippy passed in 6m 35s. The full no-retry nextest run `534f0a48-8f02-412f-a8c0-f73a752deb54` is now executing 16,687 tests across 139 binaries, with 118 pre-existing skips; doctests follow if it succeeds. No full-suite success is claimed yet. These follow-up changes, including the backfill review fix, remain local until required verification completes.

crates/ecstore/src/bucket/on_demand_migration/backfill.rs
overtrue: Fixed by capturing the cumulative failure count once at resume and allowing checkpoints while it remains at that starting value. Historical failures remain visible, but a successfully replayed page can advance the durable cursor. The baseline is intentionally not reset per page: after a new failure, later successful pages must not checkpoint beyond that unresolved page.

`recovery_advances_past_historical_failures_but_pins_new_failures` takes over an expired checkpoint with a historical failure, proves the persisted token advances from 2 to 4, then injects a new failure and verifies token 4 remains pinned through a later successful page. It also checks cumulative counts, final state, owner/job identity, page requests, and hashes. The focused backfill run passed; final workspace verification is recorded in the PR description.

crates/ecstore/src/store/init.rs
```diff
@@ -2535,6 +2542,348 @@ mod tests {
...
+            assert!(
+                ctx.scanner_publication_state_allowed(),
+                "pending PUT tails must not disable scanner namespace walks"
+            );
+            let (active, blocked, observed_movement_generation) = store.scanner_data_movement_activity().await;
+            assert!(!active, "ordinary PUT tails are not decommission or rebalance work");
+            assert!(!blocked, "ordinary PUT tails must not block the movement-only scan baseline");
+            assert_eq!(observed_movement_generation, movement_generation);
+            assert!(store.scanner_data_usage_publication_blocked().await);
+            assert!(store.scanner_data_usage_publication_admission_guard().await.is_some());
```
overtrue, line 2625: This assertion exposes a publication race: while `namespace_commits_pending()` is true, the final publication scope is also admitted because it checks only the movement gate. A PUT can begin its post-ACK rename tail after the scanner's last probe and stale authoritative usage can commit. Please serialize namespace commits with final publication admission, revalidate generation under that gate, and cover a commit starting between the probe and scope acquisition.

overtrue:
The integration of #7197 is committed locally as `9c703f7748`; it is not pushed yet. It retains the strict v2 reader, default-off first issuance, and continued budgeting for existing v2 chains alongside #7187's NUL framing. The automatic merge exposed a stale upstream golden fixture: the initial reader/handler run had 35 passes and one failure because it expected unframed v1 JSON (`f1708718-7bbf-444a-9d73-5c54298776ff`). The corrected framed v1/v2 fixtures and the expanded malformed-token/handler matrix passed all 36 selected tests using the full workspace feature graph (`e6d505ec-cf34-4d09-85b7-f0cda0fc0a5b`). No malformed-count assertion was relaxed.

Two final wrong/missing envelope-tag inputs were added after that test binary was compiled. They are included in the pending final-source gate, not claimed as covered by the earlier 36-test run.

The rollout guide now separates framing from v2 issuance: disabling first issuance does not make framed v1 tokens readable by older unframed nodes, and reverting to an unframed reader requires restarting outstanding framed v1 listings as well. Fleet rollout and real-service acceptance remain open in rustfs/backlog#2298.

Main then advanced to `0a92a7d98c`, adding zstd 0.14, the multipart registry-lock repair and tier transport timeouts. That merge is now local with no textual conflicts; upstream's multipart lock-release regression and the existing task-owned fixture/ownership fixes are retained. The previous old-base Clippy chain was explicitly interrupted, not counted as passing. A fresh uninterrupted `make pre-pr` is running on the composed source and new dependency graph; static guards passed and the isolated offline-enrollment boundary is currently building. The published PR head remains `5bf0fc11d6a7`; fresh full-gate success is not yet claimed.

crates/ecstore/src/store/init.rs
overtrue: Confirmed: a movement-only scope and a prior pending probe do not serialize the final publication against namespace commits. The PR was merged before this finding was closed. The exact race is now tracked in rustfs/backlog#2322 under rustfs/backlog#2304, including the final local admission and remote physical-owner/undo/expiry boundaries. Fixed generation checks alone do not resolve it. I am leaving this review unresolved and handing the follow-up to that existing work rather than pushing a competing patch from the obsolete integration worktree.

overtrue:
Post-merge verification: #7187 was merged externally as `e2a921bc1608823c8efec955d7463ab8350a8a01`. That commit is preserved in fetched main `72e85210f136da936b27f83faf4a4f0054f1bb32`. Reconstructing the merge of the published head with the upstream parent produces exactly the merged tree, `900bf0aee618918313d387d8f7efe5a6cedaf217`.

The unpublished `9c703f7748` integration must not be mistaken for a verified follow-up. Its local `make pre-pr` passed static guards, offline enrollment E2E, Clippy and script self-tests, but nextest run `08ed39e5-56e1-46cd-a705-08a7484ba038` stopped with two `DiskFull` failures: 1,635 passed out of 1,637 run, with 15,074 selected tests not run; doctests were not reached. The previously reported successful runs retain their original SHA attribution. Remote default/rio-v2/swift checks passed, but cancelled integration checks do not constitute complete release acceptance.

The final-admission review at https://github.com/rustfs/rustfs/pull/7187#discussion_r3940653605 is not resolved by the merge. Its concrete race and the required physical/remote ownership boundaries are now explicitly tracked in rustfs/backlog#2322 under rustfs/backlog#2304. The cursor compatibility follow-up #7238 has also merged; remaining rollout acceptance stays in rustfs/backlog#2303 and rustfs/backlog#2298. The obsolete local framed-golden adjustment will not be pushed over that repair.

This PR's terminal monitor is being stopped. Unpublished source is retained; `cargo clean` removed 19,846 task-owned generated files (35.1 GiB). The broader release audit remains open in rustfs/backlog#2259 and rustfs/backlog#2288.

PR #7190, opened 2026-09-05
fix(ecstore): remove duplicate local rename implementation

houseme:
## Related Issues

Refs rustfs/backlog#2240. Repairs the mainline integration of #7163 and #7166.

## Summary of Changes

Remove the duplicate LocalRenamePreflightRejection, rename_data_inner and rename_data_observed definitions accidentally reintroduced into disk/local.rs by #7163 after #7166 had moved them to disk/local/commit.rs. Keep the canonical commit implementation, its old tracing target, re-export and borrowed-call adapter.

## Verification

- The uncorrected mainline produced 27 ECStore compiler errors during the Scanner/Heal check; the corrected integration passed `cargo check --locked -p rustfs-scanner -p rustfs-heal -j 4`.
- `cargo nextest run --locked -p rustfs-ecstore --lib -E 'test(observed_rename_) or test(write_all_meta_)' -j 4`: 5 passed.
- `cargo nextest run --locked -p rustfs-ecstore --lib -E 'test(tail_drained_put) or test(rename_rollback_incomplete) or test(put_incomplete_rollback_preserves_staging_and_old_version_backup) or test(rename_preflight_evidence)' -j 4`: 11 passed.
- `cargo fmt --all --check` and `git diff --check`: passed.
- The corrected local.rs is byte-identical to the file at bbd7b9ef1. Checks ran on the batch integration containing this identical source fix and a separate compatible lockfile refresh; that dependency change is not in this PR. The subsequent main merge only changed CI files. Exact-head CI and Windows execution remain pending.

## Impact

Restores compilation without replacing storage behavior. #7166 already contains the control-write drain, owned mutation, per-disk rollback receipt and uncertain rollback changes. No quorum, lease, fsync or cleanup policy is weakened.

Two independent reviews: correctness/concurrency/durability found no missing control-write or rollback behavior; simplicity/compatibility/performance/test coverage confirmed equivalent method bodies after module-path/visibility normalization and retained logging targets. The focused tests above exercise the preserved paths.

## Additional Notes

The main-relative diff only deletes the redundant block from one file. No public API, disk format, configuration or dependency changes are included. No merge is performed by this task.


overtrue (approved):
Independently verified this exact 997-line deletion against main 2d159635e: the parent file returns byte-for-byte to the canonical E08 version, with the child implementation and original tests unchanged. Two independent reviews found no lost preflight proof, guard, fsync, completion or rollback behavior. On the equivalent merge tree, 102 existing rename, rollback and PUT-cleanup regressions passed with zero retries. Current-head CI and platform-specific checks still need to complete before merge.

overtrue:
A separate current-main CI blocker is already confirmed in #7184: Rust 1.98 Clippy rejects the redundant ConfigNotFound guard in `rustfs/src/app/object/shared.rs:312`; SFTP was cancelled after Swift failed. The existing one-line fix is commit `742f4b789d5a2c1c89f17b724e373f5adf84e64d` from #7187. It is independent of the other ODM/scanner changes in that PR. The MPU/heal follow-up branches are being validated with this fix and the duplicate-definition removal together.

overtrue (approved):
Reviewed the new-head delta as well: it contains only the existing equivalent SSE match and three test-only Clippy fixes. The duplicate-definition removal is unchanged, and the fixtures and assertions retain their meaning. No additional code-review blockers; current-head CI still needs to pass before merge.

overtrue (approved):

PR #7194, opened 2026-09-05
fix(ci): share quick checks and lint workflows

overtrue:
## Related Issues

Refs rustfs/backlog#2297.

## Summary of Changes

The main and docs-only workflows maintained separate Quick Checks implementations, and the docs companion had already missed the error-format ratchet. Both jobs now call one local composite action, preserving every existing main-workflow check and adding workflow linting with pinned actionlint and ShellCheck. Install actionlint from its official checksum-pinned release because the existing tool installer does not support it.

Require ShellCheck availability before actionlint so missing tooling cannot silently omit shell analysis. Extend the existing test-wiring guard to protect the shared calls and critical commands against conditional skips, ignored failures, quoted control keys, and folded YAML scalar continuations.

## Verification

- Executed all 17 validation `run` bodies locally using `bash --noprofile --norc -e -o pipefail`, with Rust 1.98.0 and Python 3.13. All passed, including `cargo fmt --all --check` and every preserved guard. Independently compared all 16 original run bodies and confirmed exact preservation and ordering.
- `shellcheck --version && actionlint` passed with actionlint 1.7.12 and ShellCheck 0.11.0; separately linted the composite shell bodies with ShellCheck. An environment without ShellCheck correctly exited 127.
- `python3 scripts/check_test_wiring.py --self-test`, `python3 scripts/check_scheduled_validation_freshness.py --self-test`, and `python3 scripts/test_security_workflow.py` passed. Regressions cover caller/action mutations, guard reachability from `validate`, and independent failure injection into ShellCheck, actionlint, and the ratchet.
- Cold installer download, hash verification, and extraction passed on macOS using a SHA256 CLI compatibility wrapper. Download failure, corrupt archives, and extraction failure stop before publishing `GITHUB_PATH`; Linux installation and execution passed in [Quick Checks](https://github.com/rustfs/rustfs/actions/runs/33958705160/job/101286616087) on the installer-fix head (85 seconds).
- The reviewed `needs` bypass is covered by 20 cases across both callers and plain, quoted, list, and folded forms; restoring the old guard makes all 20 fail. The review thread is resolved. After merging #7190 at `0ff03c596c0773b30a75dbffb39c194f0d502ecc`, all 15 wiring self-tests, the actual wiring guard, actionlint, Cargo formatting, and `git diff --check` passed.
- `git diff --check` passed. Independent correctness, simplicity, security, and coverage review findings were fixed and rechecked.

## Impact

CI validation only. Existing event filters, required check names, timeouts, and the Rust compilation/test matrix are preserved. The docs companion gains its missing ratchet, and both workflows run workflow syntax and embedded shell-script analysis. This does not resolve the separate required-status aggregation issue described in the audit.

## Additional Notes

Validation was scoped to the changed CI/tooling behavior; no Rust build or full `make pre-pr` was needed. The main-branch duplicate rename compilation regression was repaired by merged #7190. This branch now includes that repair, and its current-head remote matrix is running; earlier failures on the broken base are not being treated as successful validation.


scripts/check_test_wiring.py
```diff
@@ -520,6 +524,67 @@ def workflow_step_block(job_lines: list[str], action: str) -> tuple[int, list[st
...
+    errors: list[str] = []
+    bypass_key = r'''(?:if|continue-on-error|"if"|"continue-on-error"|'if'|'continue-on-error')\s*:'''
+    for name in ("ci.yml", "ci-docs-only.yml"):
+        relative = f".github/workflows/{name}"
+        path = root / relative
+        job = yaml_block(path.read_text().splitlines(), "quick-checks", 2) if path.is_file() else None
+        if job is None:
+            errors.append(f"{relative}: missing Quick Checks job")
+            continue
+        conditions = [index for index, line in enumerate(job) if re.match(rf"^    {bypass_key}", line)]
```
overtrue, line 544: This guard rejects if and continue-on-error, but it still accepts a needs dependency. If that dependency fails or is skipped, GitHub skips Quick Checks and reports the skipped required job as successful, so none of these checks run. I reproduced this by adding needs: [prerequisite] to ci-docs-only.yml; check_quick_checks returned no errors. Please reject needs here, or validate a fail-closed dependency condition, and add the mutation case.

overtrue (commented):
Quick Checks can still be skipped through a failed or skipped needs dependency.

scripts/check_test_wiring.py
overtrue: Confirmed and fixed in a74919db8e. Quick Checks now rejects job-level needs in scalar, inline-list, block-list, and quoted forms. Both callers have regressions for failed and skipped prerequisites. Reverting only the guard to the previous implementation produces exactly 20 failing dependency mutations; the fixed guard and all 15 wiring self-tests pass. The full Rust CI baseline remains blocked by the separate mainline rename duplication tracked in #7190.

houseme (approved):

PR #7195, opened 2026-09-05
fix(replication): close IAM snapshot, marker purge and broadcast gaps

reatang:
## Related Issues

- Fixes rustfs/backlog#2289 (R6.2: IAM snapshot resend drops user credentials and service accounts)
- Fixes rustfs/backlog#2290 (R6.1: MRF delete-marker purge intent loses the target-assigned version id)
- Fixes rustfs/backlog#2291 (R6.3: policy / policy-mapping / group-info receive paths have no timestamp gate)
- Fixes rustfs/backlog#2292 (R6.4: bucket config staleness judged by local write time)
- Fixes rustfs/backlog#2293 (R6.5: generic peer broadcast stops at the first failing peer)
- Verification plan and VM results: rustfs/backlog#2080 (R6 group)
- Follow-up split out: rustfs/backlog#2305 (retry drain not promoted after peer recovery, not addressed here)

## Summary of Changes

Five defects found by the adversarial review of the site-replication and bucket-replication code and reproduced on the two-site VM lab (rustfs/backlog#2080, R6.1-R6.5). The fixes are integrated in one PR so the combined build could be validated on the real VMs once; commits stay one per ticket.

| Ticket | Commits | Change |
|---|---|---|
| rustfs/backlog#2289 | `b56ff5503`, `850639c67` | The IAM snapshot used by retry resend, repair and site-add bootstrap now carries user secret keys, user status and service accounts. Secrets are read from the IAM store at snapshot time so `SRInfo` stays secret-free. The receiver applies a disabled status after creating a service account, and the retry snapshot tombstones removed service accounts. Repair task ids and retry-snapshot fingerprints hash key-sorted JSON: service-account claims are a `HashMap` and `serde_json` is built with `preserve_order`, so the repair preflight token went stale between dry-run and execute (412 on the VMs). |
| rustfs/backlog#2290 | `18670b269`, `6c37ebb95` | MRF purge intents persist the per-ARN target-assigned delete-marker version ids plus a corrupt flag behind a new MRF capability bit; replay addresses the target's version instead of the source marker id. A source-side `DELETE ?versionId=<marker>` replicates as a version purge and now addresses the target's recorded marker version (generic S3 targets answered the source id with 204 and kept the marker); it no longer spawns a duplicate delayed-purge watcher. |
| rustfs/backlog#2291 | `68499b654`, `e83422892`, `9032adcb1` | `policy`, `policy-mapping` and `group-info` receive paths gate on the source `updatedAt` against the local record. A bounded persisted map of IAM deletion marks covers the "record already deleted" case so a stale grant cannot follow a revoke; service-account items are judged against the marks too. |
| rustfs/backlog#2292 | `c5e6b7259` | Replicated bucket configs are stamped with the source `updatedAt` (new `*_at` write entries in ecstore bucket metadata, expand-only). Staleness is judged source time against source time; lifecycle config keeps its expiry axis. |
| rustfs/backlog#2293 | `2518fb5cd`, `b98b4821a` | Generic peer broadcast attempts every peer, records a retry event per failed peer (transport construction included) and returns the first error. Make-bucket runs both broadcast steps on every peer. |
| review round 2 | `cc68d1878` | The group write paths keep the source stamp on `GroupInfo::update_at` only and publish the IAM cache with the local clock: `LockedCache::exec` drops a publication whose time is behind the entity's load time, so a replicated group edit older than this node's startup was persisted but unreadable, and the receiver's following status write failed with `NoSuchGroup`. |
| review round | `10575e851` | The receive-side gate compared a source `updatedAt` against a stamp the *local* write had put on the record, and verdict, write and deletion mark were three separate steps. Every replicated IAM write now goes through an explicit-stamp variant (`set_policy_at`, `policy_db_set_at`, group and user `*_at`, `new_service_account_at`, `update_service_account_at`; `PolicyDoc::new_at` / `update_at`) so the record carries its source time; `apply_iam_item` runs verdict, write and mark commit inside one site-replication state transaction (the distributed state-object lock), so a concurrent older grant and newer revoke are ordered on every node of the site; a replicated service account is created with its source status in the single create write (`NewServiceAccountOpts::status`); deletion marks are pruned by age (30 days, `SITE_REPLICATION_IAM_DELETION_MARK_RETENTION`) instead of by count; bucket-config deletes persist the source stamp (`delete_if_incarnation_at`). |

`encode_service_account_replication_policy` moved from the admin handler into `rustfs/src/site_replication/hooks.rs` so the snapshot builder and the live hook share one encoder.

## Verification

Tested commit: `cc68d1878` (review round 2 on top of `680c63ea1`, which merges the review-round commit `10575e851` with `origin/main`; main carries the `SourceConfig` fixture fix for #7211). The unit rows below were re-run on `cc68d1878` for `rustfs`, `rustfs-iam` and `make pre-commit`, and on the tree-identical `e397015a0` for the crates round 2 does not touch, with no local changes. The VM rounds ran on the CI debug binaries of the branch commits named in each round, before the review rounds; the review-round changes are covered by the receiver-level regressions listed below and will be re-run on the VMs on the next lab pass.

Unit and gate:

| Command | Result |
|---|---|
| `make pre-commit` (fmt, unsafe, architecture, s3s footprint, quick-check) | passed |
| `cargo test -p rustfs --lib -- site_replication::tests admin::handlers::site_replication::tests` | 298 passed, 0 failed |
| `cargo test -p rustfs-replication` | 193 + 2 passed, 0 failed |
| `cargo test -p rustfs-iam` | 339 + 1 + 1 passed, 0 failed |
| `cargo test -p rustfs-policy` | 614 + 3 + 7 + 29 + 7 passed, 0 failed, 1 ignored |
| `cargo test -p rustfs-ecstore --lib -- bucket::metadata_sys bucket::replication bucket::metadata` | 279 passed, 0 failed |

Review-round regressions, all through the real receiver (`apply_iam_item` against a real IAM store) and failing on the previous head: `apply_iam_item_applies_delayed_in_order_updates_through_the_receiver` (T1 < T2 both delivered late, for policy body, policy-mapping, group membership and user status; previously the T2 revision was dropped), `apply_iam_item_serializes_a_concurrent_older_grant_and_newer_revoke` (both spawned concurrently, both arrival orders, the revoke must win every round), `apply_iam_item_rejects_a_stale_recreate_after_a_replicated_delete` (delete, stale replay of the older create rejected with the mark committed by the same transaction, newer create lands), `apply_iam_item_creates_a_replicated_service_account_with_its_status` (created disabled in one write, source stamp kept), `group_writes_stamped_before_the_cache_load_time_still_publish` (`rustfs-iam`: after a startup-shaped cache load, add / status / removal with hour-old source stamps are readable at once with the stamp on the record), `explicit_updated_at_is_persisted_by_a_delete` (ecstore: the delete keeps the source stamp, a later-timestamped re-create is not stale against it) and `test_record_iam_deletion_marks_newest_wins_and_expires_by_age_only` (no count eviction, age pruning at the retention edge). The source-shape guard now also pins that no apply path uses a locally-stamped write and that the transaction wraps verdict, write and mark.

New tests cover each defect: snapshot carries credentials and service accounts and never leaks them through `SRInfo`; MRF intent round-trips the target marker ids under the new capability bit and legacy entries decode unchanged; policy / mapping / group items older than the local record (or than a deletion mark) are rejected; a bucket config with an older source `updatedAt` is dropped while a newer one lands even if written locally later; a failing first peer no longer hides the remaining peers and each failure records a retry event.

Real VMs (two-site lab vm002/vm003 plus a third site, SeaweedFS as generic S3 target, `r6_review_test.py --strict r6_3 r6_4 r6_2 r6_1 r6_5`, results also posted on rustfs/backlog#2080):

- Round 1 on the CI debug binary of `9032adcb1`: R6.2, R6.3a/b/c, R6.4a/c, R6.5 PASS. R6.2b FAIL (repair execute returned 412 "preflight is stale", the HashMap-order fingerprint instability fixed in `850639c67`). R6.1 FAIL (marker left on the generic target; traces showed the source-side purge replicating by the source marker id, fixed in `6c37ebb95`).
- Round 2 on the CI debug binary of `6c37ebb95`: R6.3a/b/c, R6.4a/c, R6.5, R6.2b PASS. R6.1 PASS 2/2 once the harness waits for the MRF persister's 10 s flush before restarting the source; the trace confirms the purge addresses the target's marker id. R6.2 automatic retry drain after the outage: 1 PASS / 3 FAIL, the failed deliveries are recorded but the reconciler tick does not promote the entry; the drain scheduling is untouched by this PR and is tracked as rustfs/backlog#2305.

Not run: full workspace clippy and the e2e suites beyond what PR CI executes; the review-round changes have not yet been through the VM lab. Remaining risk: the IAM deletion-mark map is a `serde(default)` field in the persisted site-replication state, and the bucket config stamps reuse the existing `*_config_updated_at` fields; the MRF journal change is the only format change that an older reader refuses (see Impact).

### Concurrent cache reload follow-up

`6e3059114` samples group cache publication time inside the cache write lock after storage completes. A periodic reload during the save can no longer discard a successful add or removal; the record and return value still keep the source timestamp.

`cargo test --locked -p rustfs-iam --lib group -j4`: 32 passed, including two new add/remove regressions that both fail on `cc68d1878` and the existing old-source-timestamp regression. The tests run real IamSys/cache methods with controlled Store save/reload barriers; they are not a real-disk or distributed test. `cargo fmt --all --check` and `git diff --check` passed.

Final fix review: correctness and test coverage passed with the two deterministic regressions; concurrency/durability passed for local cache publication; security and compatibility passed with unchanged source stamps and policy semantics; simplicity and performance passed with no new production locks or tasks.

## Impact

- MRF journal: purge intents gain a capability bit (`1 << 4`) and a per-ARN target version map. Legacy entries decode with an empty map and replay exactly as before. A journal that advertises the new bit is refused by a reader from before this change, the same rule the earlier optional MRF fields follow; rolling back a node after it persisted such intents means handling that journal under the existing MRF capability policy.
- Site replication IAM snapshot: user secret keys and service accounts now travel inside the peer-to-peer snapshot payload (already an authenticated admin channel). The admin-facing `SRInfo` response is unchanged and still carries no secrets.
- Bucket metadata: replicated bucket configs are stamped with the source `updatedAt` through new `*_at` write entries in ecstore and the storage API; the on-disk fields are unchanged, and a peer running an older build keeps stamping its local write time and interoperates.
- Retry ledger: one event per failed peer instead of one per broadcast; operators will see more entries after a partial outage.
- IAM replication receive path: every gated IAM item (policy, policy-mapping, group-info, iam-user, service-account) is now applied under the site-replication state-object lock, one item at a time per site. IAM replication volume is low and the lock is already taken by the deletion-mark commit, but a burst (bootstrap of a large IAM set) is serialized on it. When the state object cannot be read or persisted the item now fails and the sender retries it, instead of being applied without its mark.
- `rustfs-iam` API: explicit-stamp write variants are additive; `NewServiceAccountOpts` gains a `status` field (constructors that spell out every field need `status: None`; those using `..Default::default()` are unaffected).
- Site-replication state: `iam_deletion_marks` is no longer capped at 1024 entries; it holds one entry per IAM deletion of the last 30 days.
- No configuration or deployment changes.

## Additional Notes

- The automatic drain of the IAM retry entry after a peer outage did not converge reliably on the VMs (R6.2). The scheduling code is not touched by this PR; the investigation continues in rustfs/backlog#2305.
- Graceful shutdown does not flush the MRF persister queue, so a restart inside the 10 s flush window loses a not-yet-persisted purge intent. Pre-existing behaviour, to be filed as a follow-up.
- `cargo clippy -p rustfs -p rustfs-iam -p rustfs-policy --lib --tests -- -D warnings` on `cc68d1878`: no findings in the changed crates.
- PR CI on `680c63ea1`: Test and Lint and Test and Lint (swift) abort on `app::object::on_demand_migration_put::tests::write_back_multipart_completion_preserves_a_client_put_after_staging` (SIGABRT). `origin/main` at `e1608fbd9` fails its own Test and Lint on the same test; it is inherited from main and needs a rerun once main is green.


rustfs/src/admin/handlers/site_replication.rs
```diff
@@ -5762,27 +5802,87 @@ async fn apply_iam_item(item: SRIAMItem) -> S3Result<()> {
...
+    // older body (or delete) cannot overwrite a newer edit; once the document
+    // is deleted, its deletion mark stands in for it (backlog#2291).
+    let local_updated_at = match iam_sys.get_policy_doc(name).await {
+        Ok(doc) => Some(doc.update_date.unwrap_or(OffsetDateTime::UNIX_EPOCH)),
+        Err(err) if rustfs_iam::error::is_err_no_such_policy(&err) => {
+            local_iam_deletion_mark(&[iam_policy_deletion_mark_entity(name)]).await
+        }
+        Err(err) => return Err(ApiError::from(err).into()),
+    };
+    if judge_iam_item_staleness(local_updated_at, incoming_updated_at) == IamItemVerdict::SkipStale {
```
overtrue, line 5901: [P1] Preserve source timestamps when applying IAM changes. If two source policy edits T1 < T2 both predate delivery, set_policy stamps the first with receiver time, so this gate silently skips the newer revoke. A regression through real apply_iam_item leaves the PutObject grant in the stored policy after the narrower update succeeds. Preserve incoming updatedAt in the IAM write paths and test delayed, in-order updates through the real receiver.

overtrue (changes requested):
The new IAM timestamp gate can discard a newer revoke.

overtrue (changes requested):
I found a few ordering gaps that can still restore stale replication state:

- Bucket-config deletes at `site_replication.rs:5681` do not persist the source `updatedAt`. A delete from T1 can stamp local receive time R, then a newer T2 create is skipped because T2 < R.
- IAM writes around `site_replication.rs:5876` compare source timestamps with local apply timestamps written by the IAM manager. Even in-order T1 then T2 delivery can reject T2 when the T1 receive time is later than T2.
- The IAM staleness check, mutation, and deletion-mark commit are not one atomic entity operation. Concurrent older grant and newer revoke requests can both pass the old check and let the grant win after the revoke.
- A missing disabled service account is created enabled at `site_replication.rs:6141`, then disabled in a second write. If that write fails, the replicated secret remains enabled.
- Deletion marks are evicted after 1024 entries in `site_replication/state.rs:79`. Once a mark is evicted, a delayed older grant can recreate the deleted entity.

These need source-version persistence and atomic apply semantics, plus regressions for delayed and concurrent grant/revoke and delete/recreate delivery.

rustfs/src/admin/handlers/site_replication.rs
reatang: Fixed in 10575e851. The receiver now writes through `set_policy_at` (and the matching `*_at` variants for mappings, groups, users and service accounts), so the stored `UpdateDate` is the item's source `updatedAt`, never the receive time; the plain `set_policy` keeps stamping the local clock for local edits. `apply_iam_item_applies_delayed_in_order_updates_through_the_receiver` replays exactly this case through the real `apply_iam_item` on a real IAM store: T1 grant then T2 narrower body, both delivered after T2, and asserts the stored policy is the T2 body with `UpdateDate == T2`. It fails on the previous head with the PutObject grant still in the document.

reatang:
Thanks — all six findings were real and share one root cause: the receiver judged a source timestamp against a stamp the *local* IAM/bucket write had put on the record, and the verdict, the write and the deletion mark were three separate steps. Addressed in 10575e851 (pushed head 680c63ea1) as follows.

**Source-version persistence (P1 inline, `site_replication.rs:5876`, `:5681`).** Every IAM write the receiver performs now goes through an explicit-stamp variant that persists the item's source `updatedAt` on the record itself: `set_policy_at`, `policy_db_set_at`, `add_users_to_group_at` / `remove_users_from_group_at` / `set_group_status_at`, `create_user_at` / `set_user_status_at`, `new_service_account_at` / `update_service_account_at` (`rustfs-iam`), backed by `PolicyDoc::new_at` / `update_at` (`rustfs-policy`). The plain entry points delegate with the local clock, so local edits are unchanged and the stamp a source returns to its broadcast is now exactly the stamp it stored. Bucket-config deletes take the same path as writes via `delete_if_incarnation_at` (`ecstore`), so a T1 delete applied at receive time R no longer makes a T2 re-create with T2 < R look stale.

**Atomic apply.** `apply_iam_item` runs the verdict, the IAM write and the deletion-mark commit inside one site-replication state transaction (`with_site_replication_state_transaction`, the distributed state-object write lock). Two deliveries for the same entity — on this node or on a sibling node of the site — are therefore applied one after the other, each judged against what the other left behind. The mark is written in the same commit as the delete; if the state cannot be persisted the item fails and the sender retries the idempotent deletion, instead of the previous "apply, then hope the mark lands". STS items stay outside the transaction (no source revision, no mark).

**Service accounts (`:6141`).** `NewServiceAccountOpts` gained `status`; a replicated create persists the source status in the single create write. No second update, nothing to fail half-way.

**Deletion marks (`state.rs:79`).** The count bound is gone. Marks are pruned by age only, with a 30-day retention (`SITE_REPLICATION_IAM_DELETION_MARK_RETENTION`); the constant's doc lists the delivery paths a mark has to outlive and why thirty days clears all of them (the retry drain never replays a stale grant — it resends snapshots of the current records plus the recorded deletion bodies — so the only stale sources are in-transit delays, the concurrency race now closed by the transaction, and a peer snapshot built before our deletion reached it, which is bounded by our own retry queue towards that peer). Growth is one entry per deletion within the window, not per entity.

**Regressions, through the real receiver (`apply_iam_item` on a real IAM store):**
- `apply_iam_item_applies_delayed_in_order_updates_through_the_receiver` — T1 < T2 both delivered late, for policy body, policy-mapping, group membership and user status; fails on the previous head with the T2 revision dropped.
- `apply_iam_item_serializes_a_concurrent_older_grant_and_newer_revoke` — older grant and newer revoke spawned concurrently, both arrival orders, the revoke must win every round.
- `apply_iam_item_rejects_a_stale_recreate_after_a_replicated_delete` — delete, stale replay of the older create (rejected, mark committed with the delete), newer create (lands).
- `apply_iam_item_creates_a_replicated_service_account_with_its_status` — created disabled in one write, source stamp kept.
- `explicit_updated_at_is_persisted_by_a_delete` (ecstore) — the delete keeps the source stamp; a later-timestamped re-create is not stale against it.
- `test_record_iam_deletion_marks_newest_wins_and_expires_by_age_only` — no count eviction, age pruning at the retention edge.
- The source-shape guard now also pins that no apply path uses a locally-stamped write and that the transaction wraps verdict, write and mark.

PR description updated accordingly (Verification / Impact).

crates/iam/src/manager.rs
```diff
@@ -1693,6 +1787,10 @@ where
             }
         }
 
+        // The group's own timestamp moves with every membership or status
+        // change: site replication judges an incoming group item against it
+        // (backlog#2291), so it must reflect the last change, not creation.
+        let now = updated_at;
```
overtrue: [P1] Keep source time separate from cache publication time. After startup loads existing groups, a delayed source timestamp predates the cache's load_time. LockedCache::exec then silently drops these group and membership updates. A focused IamSys initialization/add test returns success but cannot read the new group; the receiver's following set_group_status_at consequently sees NoSuchGroup. Publishing the cache with local time makes the same test pass while preserving GroupInfo.update_at. Apply that separation to the status and removal paths too.

overtrue (changes requested):
Replicated group updates can disappear from the receiver's cache after startup or reload.

crates/iam/src/manager.rs
reatang: Fixed in cc68d1878. `add_users_to_group_at`, `set_group_status_at` and `remove_members_from_group_at` now keep `updated_at` for `GroupInfo::update_at` only and publish the cache (group entity and membership index) with the local clock, so `LockedCache::exec` no longer drops a replicated edit whose source time predates this node's load time. `group_writes_stamped_before_the_cache_load_time_still_publish` (crates/iam/src/sys.rs) initialises `IamSys`, replaces the group entities as a startup load would, then adds, disables and removes with source stamps an hour in the past: each write must be readable through `get_group_info` immediately, with `update_at` equal to the source stamp and the membership index following. It fails on the previous head at the first `get_group_info`.

reatang:
Second round, cc68d1878: the group write paths (`add_users_to_group_at`, `set_group_status_at`, `remove_members_from_group_at`) now keep the source stamp on `GroupInfo::update_at` only and publish the cache with the local clock — `LockedCache::exec` rejects a publication time behind the entity's load time, so a replicated edit older than this node's startup was written to the store but never reached the cache. Regression `group_writes_stamped_before_the_cache_load_time_still_publish` drives `IamSys` through a startup-shaped load and then add / status / removal with hour-old source stamps, asserting each is readable right away.

On the two failing checks (Test and Lint, Test and Lint (swift)): both die on `app::object::on_demand_migration_put::tests::write_back_multipart_completion_preserves_a_client_put_after_staging` with SIGABRT. `origin/main` at `e1608fbd9` fails its own Test and Lint on the same test (job 101360617709), so this is inherited from the on-demand-migration work on main, not from this branch; the checks need a rerun once main is green.

overtrue (approved):

PR #7201, opened 2026-09-05
fix(ci): isolate functional evidence and preserve every result

overtrue:
## Related Issues

Part of rustfs/backlog#2297. Builds on the merged failure-propagation prerequisite #7199. This PR remains draft until the companion private diagnostic fix rustfs/auto-testing#41 is available. The base remains main so the normal CI matrix runs.

## Summary of Changes

Bind functional reports, logs, intermediate case tables, and performance results/version files to a new directory for this run and attempt. Previously, fixed `/tmp` paths could publish a previous run's evidence after an early failure. Initialization rejects an existing directory, and report publication and failure-issue fallback require successful initialization.

Replace five duplicated case parsers with one small parser that preserves every execution and its topology/backend context. A failing KMS case followed by a passing execution of the same ID now remains two results with a failure. Preserve S3C identifiers, actual newline formatting, and the upgrade matrix. Missing logs, unfinished or orphaned case results, and missing upgrade matrices cannot produce a successful report.

Validate heal's actual two-phase execution by requiring all seven recorded steps to pass; an earlier step or overall failure cannot be overwritten by a later pass. Failed or incomplete suites produce an explicit failure report for the dashboard, while partial tables and logs remain in the current run's artifact.

Keep process-wide scratch state in a sibling directory outside the published evidence. Upload explicit report/log/result paths, including heal's existing warp log, so failed Vault initialization cannot expose root tokens or unseal keys. Preserve security stdout/stderr through a pipefail-protected log while excluding individual scratch case files. OIDC negative-response and missing-field diagnostics retain status/field context without printing STS credentials.

## Verification

Validated the final task-owned diff at `f4ac6014e4fd2b1e928a0253a12f0a455a6afb4c` against main `447f3c704b`. All 11 PR files retain the previously reviewed bytes; main's failure-propagation tests remain present in one test class.

- `./scripts/python_bin.sh -B scripts/test_security_workflow.py`: 18 tests passed. Coverage includes success/failure/skipped/cancelled/missing/incomplete reports; all eight suites reject pre-existing evidence or scratch directories; explicit upload paths retain required diagnostics and exclude synthetic credentials; actual OIDC guards reject unexpected STS success and missing required fields without printing issued credentials. All commands use local substitutes; no shared machines were touched.
- Independent final review passed the same 18 tests and caught 37 deliberate regressions, including every suite pointing scratch back into artifacts, directory-wide uploads, scratch reuse, missing logs, swallowed security-suite exit status, and all three OIDC response diagnostic regressions. A separate full private KMS-script fixture reproduced failure after Vault initialization: both versions exit 42 and produce a failure report, but only the original artifact selection includes the synthetic Vault initialization secret.
- Ten independent parser mutations failed explicit assertions: hiding FAIL/RUNNING/orphan results, missing upgrade matrix, S3C rejection, literal escaped newlines, ignored INFO-prefixed contexts, requiring a nonexistent single-invocation heal verdict, and overwriting step or overall failures.
- Independent review compared real auto-testing producer formats and workflow invocations. It found and fixed the INFO prefix mismatch and the split heal invocation mismatch. Fixtures execute the producer's printf formats and the workflow's actual phase arguments. Each missing heal step and both failure/pass orders are covered.
- Existing wiring guard and its 15 self-tests, `actionlint`, `cargo +1.98.0 fmt --all --check`, and `git diff --check` passed. Cleanup and downstream dispatch failure-path regressions remain green. Independent merge review also passed six focused tests; restoring the fixed log override or removing the evidence-success report condition is caught by existing regressions.

The compilation and suite selection matrices are unchanged, so no local workspace build or `make pre-pr` was needed. Full remote functional suites were not dispatched; the tested behavior is local workflow execution, report validation, and evidence isolation.

## Impact

KMS, storage, S3 compatibility, replication, upgrade, heal, performance, and security evidence can no longer borrow the fixed files from an earlier run. Reports retain per-execution results and identify the workflow commit, run, and attempt. Existing test commands, topology choices, cleanup, and handoffs remain in place; tier and pool's existing isolated evidence paths are unchanged.

## Additional Notes

This patch does not establish immutable package identity, selected/expected case completeness across the whole chain, or successful execution of every nightly node. Those remain separate follow-ups. Confirmed sensitive response diagnostics in the private KMS, upgrade, and security suites are fixed by rustfs/auto-testing#41, which must land before enabling this report chain; explicit artifact paths alone do not certify log contents as secret-free. Existing UNSUPPORTED case policy is preserved and its counts remain visible. Performance's false-zero exit is addressed by rustfs/auto-testing#39.


.github/workflows/rustfs-kms-test.yml
```diff
@@ -49,10 +49,28 @@ env:
...
+        run: |
+          set -euo pipefail
+          umask 077
+          FUNCTIONAL_ARTIFACTS_DIR="${RUNNER_TEMP}/rustfs-kms-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          mkdir -- "${FUNCTIONAL_ARTIFACTS_DIR}"
+          {
+            printf 'FUNCTIONAL_ARTIFACTS_DIR=%s\n' "${FUNCTIONAL_ARTIFACTS_DIR}"
+            printf 'LOG_FILE=%s/suite.log\n' "${FUNCTIONAL_ARTIFACTS_DIR}"
+            printf 'REPORT_FILE=%s/report.md\n' "${FUNCTIONAL_ARTIFACTS_DIR}"
+            printf 'TMPDIR=%s\n' "${FUNCTIONAL_ARTIFACTS_DIR}"
```
overtrue: This makes the uploaded evidence directory the process-wide `TMPDIR`. The pinned suite writes `rustfs-vault-data/init.json` there with the Vault root token and unseal key; if a later setup step aborts before `vault_stop`, the `always()` artifact step uploads that file before cleanup. Please keep scratch state outside the artifact root, upload an explicit evidence allowlist, and add a failure-after-Vault-init regression.

overtrue: Confirmed and fixed in 8d4bca08b8. All eight suites now use sibling scratch directories and explicit artifact allowlists. The real KMS-script failure-after-Vault-init fixture preserves exit 42 and the failure report while excluding the synthetic init.json secret. Existing diagnostics, including heal warp logs and security stdout/stderr, remain available; OIDC failure paths no longer print credentials. Independent validation passed 18 tests and detected all 37 targeted regression mutations. Companion private-suite diagnostic fixes are tracked before enabling the chain.

Issue #7108, opened 2026-09-03
Census never becomes correct on a continuously-written cluster: every cycle is Superseded, per-bucket cache never seeded — datausageinfo reports 36 objects for a 151k-object bucket (rc.5)


sir-alex636:
## Environment

- RustFS **1.0.0-rc.5** (official image, digest `c36b3efe…`), Docker, 3 nodes × 2 drives, EC 4+2, TLS
- Same clusters as #6907 / #6958 / #6218. Scanner state was fully reset on 2026-09-02 (bootstrap from scratch, see #6907), so there is no inherited state involved here.
- Workload: one bucket (`loki-dati`, **150,994 objects / 85.3 GB** by `mc du`) receives continuous Loki writes — new chunks every few seconds, around the clock. `RUSTFS_SCANNER_CYCLE=3600` (explicit cadence); everything else default.

## Symptom

`datausageinfo` (and therefore the console and quota accounting) reports the cluster as **6,498 objects / 0.4 GB**, with `loki-dati` at **36 objects / 0.01 GB** — off by ~23× on count and ~200× on size — and it has stayed like this since the reset, while the S3 data plane is perfectly healthy:

```
datausageinfo: objects_total_count 6498 | objects_total_size 0.4 GB | buckets_count 14
               usage_snapshot_complete: true | usage_snapshot_converged: false | scanner_epoch 2 | scanner_cycle 30
  loki-dati        36 objects   0.01 GB      (real: 150,994 objects / 85.3 GB)
  ads-loki       6013 objects   0.02 GB      (real: 2,223 objects — the census counts versions here, fine)
```

`scanner/status` on the leader:

```
cycle_schedule: effective_interval 3600, superseded_cycles 28, superseded_retry_backoff_enabled true
metrics: last_cycle_result "superseded", last_cycle_duration_seconds 44.6, last_cycle_objects_scanned 6215,
         cycles_completed_at [], usage_freshness.last_publication_state "success"
```

Every cycle since the reset ended `superseded`; **no cycle has ever ended `Complete`** (`cycles_completed_at` is empty). Each hourly cycle lasts ~45 s and visits ~6k objects — the dirty subset — and the per-bucket usage cache for `loki-dati` (`.rustfs.sys/buckets/loki-dati/.usage-cache.bin`) is tiny (xl.meta ≈ 885 B, rewritten every cycle), i.e. it was never seeded with a full walk.

## Why (from the rc.5 source)

[...]

PR #7206, opened 2026-09-05
feat(scanner): reuse complete observed scan candidates

marshawcoco:
## Related Issues

Refs #5950
Refs #7108

## Summary of Changes

When the persisted primary usage snapshot is a bootstrap or legacy-compatible identity rather than a reusable complete baseline, permit the next scan plan to reuse a newer observed snapshot only after it proves complete, non-partial, non-converged state and exact primary identity, per-set provenance, leader epoch, and structural plan-digest agreement. The candidate limits only the next dirty-bucket scan scope; it cannot publish usage, satisfy quota or confirmed-empty decisions, or acknowledge dirty generations.

## Verification

- `cargo test -p rustfs-scanner --lib scoped_scan_ -- --test-threads=1` passed: 3 scoped-baseline regression tests passed.
- `./scripts/check_architecture_migration_rules.sh` and `git diff --check upstream/main...HEAD` passed.

## Impact

Eligible local scan planning can avoid an unnecessary full namespace walk after a complete observed snapshot. Any missing, stale, partial, incompatible, or unproven input continues to select the existing full-scan fail-closed path. No public API, configuration, persisted format, or authoritative usage semantics change.

## Additional Notes

This is a planning-only optimization. Distributed prefix/object reuse, EC-scoped reuse, usage publication, and dirty acknowledgement remain outside its scope.


overtrue (approved):
LGTM

houseme (approved):

overtrue (approved):
LGTM

overtrue (approved):
LGTM

overtrue (changes requested):
complete_scanner_cache_baseline_plan_digest discards a verified maintenance digest and returns the structural digest instead. On head a8b8132a3, cargo test -p rustfs-scanner scanner_io::tests::scoped_scan_baseline_work_proof_requires_uniform_known_set_identity -- --exact fails at crates/scanner/src/scanner_io/tests.rs:1533: left Some(structural), right Some(full). Please return the digest produced by complete_scanner_cache_snapshot_plan_digest so cache reuse retains the actual work proof.

Issue #2302, opened 2026-03-27
Memory display issues after deploying rustfs


bestyjs:
I deployed rustfs using Docker. When I enter the 'Running Status' section of the visual interface, it shows my total capacity as 6.8 TiB, with 168 GB used. But I only created an empty bucket, and my entire server doesn't even have 6 T. What are these values based on?

<img width="2144" height="734" alt="Image" src="https://github.com/user-attachments/assets/0c1acb51-9552-40b1-8871-b8c4608fb89b" />

majinghe:
@bestyjs Please share with us your configuration, such as volume mount(volume configuration in docker or docker compose file) and disk mount(df -h).

houseme:
@bestyjs `6.8TiB` total size of the data disk, `168.1GiB` total used data on the current disk, not the project data directory usage, where `168.1 GiB` will be updated to the usage of the statistics directory in the next version.

bestyjs:
> [@bestyjs](https://github.com/bestyjs)请与我们分享您的配置，例如卷挂载（docker 或 docker compose 文件中的卷配置）和磁盘挂载（df -h）。

The disk is mounted as follows

<img width="1068" height="315" alt="Image" src="https://github.com/user-attachments/assets/bc99f046-884a-4f4a-a681-98efc86b527c" />

The volume mount in the docker-compose.yml file is as follows. The depends_on has been commented out, and only the rustfs service is started.

```
services:
  rustfs:
      volumes:
           - ./deploy/data/pro:/data
           - ./deploy/logs:/app/logs
           - ./deploy/data/certs/:/opt/tls

volumes:
  rustfs_data_0:
    driver: local
  rustfs_data_1:
    driver: local
  rustfs_data_2:
    driver: local
  rustfs_data_3:
    driver: local
  rustfs_dev_data:
    driver: local
  prometheus_data:
    driver: local
  grafana_data:
    driver: local
  tempo_data:
    driver: local
  loki_data:
    driver: local
  jaeger_data:
    driver: local
  logs:
    driver: local
  cargo_registry:
    driver: local
```

majinghe:
@bestyjs From the screenshot, I think there is one issue for your configuration. Although you claim the volume(rustfs_data_0 to 4), but you don't use those volumes, since you still have another configuration for volumes `./deploy/data/pro:/data`, which means you use directory under `docker-compose.yml` directory to persist the docker data.


By the way, please share the below info

```
    environment:
[...]

PR #7209, opened 2026-09-05
chore(deps): preserve scanner and heal validation compatibility

houseme:
## Related Issues

Refs rustfs/backlog#2240 and rustfs/backlog#2302.

## Summary of Changes

Refresh compatible lockfile selections for Scanner/Heal validation, including generic-array and platform dependency edges. Keep hotpath runtime exactly at 0.25.0: the inspected 0.25.1 focus parser replaces regex with regex-lite and panics on previously supported Unicode classes. Retain the compatible macro patch, whose source is unchanged, and keep profiling disabled by default.

The update includes crypto-common 0.1.7 to 0.1.6, not only upward version changes. Its implementation was compared and the resolver change permits generic-array 0.14.9.

## Verification

- Requested `cargo update` and `cargo upgrade` completed in isolated batch baselines. The incompatible profiler candidate was then explicitly constrained; `cargo metadata --locked --no-deps --format-version 1` and diff checks passed.
- This exact final manifest/lockfile pair was compiled in Scanner integration 1aafe477ec773eb86ca2dd77fe8a1cfb3faed57f on main c589fd243; 114 focused scanner/cache/publication/reader tests passed. The earlier corrected baseline also passed Scanner/Heal checks and 16 ECStore tail/rollback/fsync tests.
- Independent source review confirmed unchanged default profiler and macro implementations. Opt-in profiler execution, Windows selections and full platform CI have not been verified for this delivery; #2302 records the remaining compatibility gate.

## Impact

No storage format, wire protocol or production profiler enablement is introduced by this PR. The exact profiler constraint preserves existing focus-expression behavior until its migration or upstream compatibility fix is validated.

## Additional Notes

This is a shared dependency prerequisite for the remaining Scanner/Heal task PRs. Main integration includes the already-merged rename compilation repair #7190. Focused and integration evidence is separate from exact-head CI and release approval.


pawurb:
Hi @houseme, I noticed this issue. Can you explain why the `regex-lite` is blocking an update for you? App only panics if non unicode characters are passed to `HOTPATH_FOCUS` ENV. I did not expect function names to contain special characters.

houseme:
> Hi @houseme, I noticed this issue. Can you explain why the `regex-lite` is blocking an update for you? App only panics if non unicode characters are passed to `HOTPATH_FOCUS` ENV. I did not expect function names to contain special characters.

Thank you for your concern. Currently, we are focusing on maintaining stability and will address this specific issue after completing the current round of optimizations.

PR #7176, opened 2026-09-05
fix(scanner): require complete publication coverage

houseme:
## Related Issues

Refs rustfs/backlog#2261 and rustfs/backlog#2240. Shared prerequisite: [dependency PR #7174](https://github.com/rustfs/rustfs/pull/7174).

## Summary of Changes

Validate complete set/bucket coverage and bind usage candidates to the requested cycle, leader, plan and tier generation before authoritative publication. Reject orphaned entries and invalid root aggregates. Disable unselected-bucket cache reuse until durable incarnation binding exists; creation timestamps are not identity proofs.

## Verification

- `cargo test -p rustfs-scanner --lib scanner_io::publish_gate_tests -j4`: 58 passed.
- `cargo test -p rustfs-scanner --lib scoped_set_scan -j4`: 4 passed.
- Focused quota-ledger-preserving reset, primary-save failure, and deleted-bucket CAS regressions: 3 passed.
- `cargo fmt --all --check` and cumulative `git diff --check`: passed.

## Impact

No new writable ledger or on-disk format. Existing quota reservations are preserved. Unselected scoped reuse remains disabled until its identity proof is implemented; this intentionally does not claim a performance gain. Distributed and mixed-version validation remains separate.

Two independent implementation reviews: Correctness/concurrency/durability and simplicity/compatibility/performance/test-coverage reviews found no unresolved issue after rejecting timestamp-based identity reuse.

## Additional Notes

This PR targets main so the repository PR CI runs. It includes the shared dependency commits until the prerequisite is merged; merge that PR first. The task-only change can be reviewed against the dependency branch. Delivery integrates main at 123967e729c74903a7f3d6dcc0a388736acd4f6c; the affected dependency/metadata checks were rerun. CI and formal PR approval are separate from local verification. Do not close the backlog task before its required evidence and merge are confirmed.

PR #7210, opened 2026-09-05
fix(scanner): bind resumable scans and cache publication coverage

houseme:
## Related Issues

Refs rustfs/backlog#2267 (W03), rustfs/backlog#2270 (W04) and rustfs/backlog#2240. Prerequisites: #7176 and [shared dependency validation](https://github.com/rustfs/rustfs/pull/7209).

## Summary of Changes

Retain partial bucket work across changing dirty plans without promoting old coverage to a complete snapshot. Bind resume receipts to the bucket incarnation, layout, fences, scan mode and actually completed prefix. Keep stable completed scans on their normal sampling path.

Fix same-cycle retry publication: after bucket/set persistence succeeds but root delivery fails, a later Normal or Deep scan must not replace newly scanned counts with an older Current set cache. Separate structural/work identity from full activity/inventory coverage and validate it at bucket, set persistence and complete aggregation boundaries.

W03 and W04 are delivered together because their cache proofs and optional map fields must remain consistent. Preserve W02's rejection of unselected caches without a durable incarnation binding.

## Verification

- On tested HEAD 1aafe477ec773eb86ca2dd77fe8a1cfb3faed57f, 17 checkpoint fixtures passed, including corrupted existing cursors, a stable-plan partial, Normal-to-Deep changes and the combined optional-field round trip. Fixed-budget retained coverage advanced 3 to 6 to 9 and completed at exactly 25 objects, 2 versions and 34 bytes; no unbudgeted finishing scan was used.
- `scoped_scan_same_cycle_maintenance_rewalks_after_root_delivery_failure` first failed with old count 1 instead of 2 on the Normal-to-Normal branch. The unchanged assertion now passes, together with same-cycle Deep/full-maintenance cases. The complete integration ran 114 distinct focused tests: checkpoint 17, scoped scan 8, scoped set 4, publication 60, legacy prepare 13, folder 8 and reader/wire 4.
- Formatting, diff, documentation, layer, logging and architecture guards passed. Tests compiled against main c589fd243 with zstd 0.14 and the exact shared dependency baseline. Real process-restarted raw enumeration, arbitrary-width compaction convergence, distributed crash, actual bitrot repair and ABBA have not been validated.

## Impact

- Correctness: incomplete or mismatched coverage cannot be reused as complete; fresh same-cycle counts survive set persistence.
- Concurrency/durability: existing locks, publication fences, revision CAS and newer-cycle/leader rejection remain intact.
- Compatibility: optional fields retain hand-written map serialization; old readers can ignore them, but old coordinators do not gain the new proof guarantees.
- Security: missing or corrupt scope/receipt identities fail closed; no new authority is inferred from an existing path or cursor alone.
- Simplicity: one existing cache state model, without a separate writable baseline ledger.
- Performance: stable sampling is preserved and no per-object I/O was added. Prefix hashing/encoding costs and foreground latency still require measurement.
- Test coverage: two independent final-diff reviews found no unresolved issue after the focused regressions above; full release lanes remain separate.

## Additional Notes

The main-relative diff temporarily includes W02 and the shared dependency changes until their PRs merge. This PR does not enable unproven cold-bucket/segment reuse or complete the R-E raw-directory enumeration gate. Keep the parent and child issues open for their remaining activation and validation requirements.


overtrue (approved):
LGTM

overtrue (approved):
LGTM

Issue #2237, opened 2026-03-19
Tracing: RustFS does not continue incoming trace context for S3 HTTP requests


ghostman2013:
## Problem Description

When `rustfs` is used behind a **user service**, `rustfs` spans appear in `Jaeger`, but they are not linked to the incoming trace context:

- upstream sends `traceparent` / `baggage` (and optionally B3);
- `rustfs` creates a new `trace_id`;
- `Jaeger` shows two separate traces instead of one chain.

## Expected Behavior

If an incoming request carries valid trace context, `rustfs` should:

1. Extract trace context from request headers.
2. Create the server span as a child of that context.

If headers are missing, `rustfs` should continue creating a local root span.

## Actual Behavior

- `rustfs` exports spans to `Jaeger`.
- Spans do not continue incoming context and use a different `trace_id`.

## Environment (example)

- `rustfs/rustfs:latest`
- `Jaeger all-in-one` with OTLP (`4317`, `4318`)
- `RUSTFS_OBS_ENDPOINT=http://jaeger:4318` (or `http://jaeger:4317` depending on exporter mode)

## Reproduction Steps

1. Start `jaeger` and `rustfs`.
2. Configure an upstream service that sends request headers to `rustfs`:
   - `traceparent`, `baggage`
   - (optionally) `b3`, `x-b3-*`
3. Execute an S3 operation (PUT/GET) through upstream.
4. Open `Jaeger` and compare `trace_id` of upstream and `rustfs` spans.

## Minimum Expected Fix

- Add incoming trace context extraction in `rustfs` HTTP layer before creating request span.
- Support at least W3C (`traceparent`), preferably B3 as well.
- Keep current behavior when headers are absent.

## Additional Context

In our integration setup, upstream already sends both W3C and B3 headers in outgoing requests to `rustfs`.
Therefore, the issue appears to be missing server-side extraction/parenting in `rustfs`.

houseme:
Thank you for your feedback. After reviewing, we will make improvements

loverustfs:
Hello @ghostman2013 ,

Fixed.

PR #7213, opened 2026-09-05
fix(tier): drain cleanup before tier removal

cxymds:
## Related Issues

Refs https://github.com/rustfs/backlog/issues/2237

## Summary of Changes

- Persist and install the exact Prepared mutation admission fence before waiting, so new operations cannot acquire a target-tier lease while already-issued generation leases remain current long enough to finish their local commit.
- Drain those existing leases before peer Prepare and the authoritative reference proof, then revoke the generation immediately before publication. Exact mutation ID plus runtime fence revision makes a concurrent recovery or fence change fail closed.
- Preserve the original early-revocation path for no-intent stale-manager reconciliation, and include local-only stale target tiers in the process-local Prepared fence/drain set.
- Add a real ECStore overlap regression at the remote-DELETE/local-marker boundary, focused coordinator/recovery regressions, structured internal blocker logging, and update the persistence/lock-order contract.

## Verification

Validated PR commit `681a3228d` on a clean worktree based on `origin/main@cc5060ac2`.

- `cargo test -p rustfs-ecstore --features test-util tier_remove_waits_for_inflight_free_version_local_commit -- --nocapture`: 1 passed after rebase; exercises the actual remote DELETE -> local free-version marker commit overlap with real tier removal.
- `cargo test -p rustfs-ecstore --features test-util services::tier::tier::tests:: -- --test-threads=1`: 212 passed after rebase; includes force/legacy/recovery/peer/refresh matrices and the new admission, no-intent, cancellation, and revision races. `cargo test -p rustfs --lib tier_backend_error_response -- --nocapture`: 1 passed on the final source diff, preserving the public MinIO-compatible response by error code.
- `cargo check -p rustfs-ecstore`; `cargo check -p rustfs-ecstore --features test-util`; `cargo clippy -p rustfs-ecstore --all-targets --features test-util -- -D warnings`; `cargo fmt --all --check`; `git diff --check`; and `./scripts/check_logging_guardrails.sh`: passed. Formatting, diff and logging checks were repeated after rebase. The test targets emitted only the existing macOS `__eh_frame section too large` linker warning.

Not run: `make pre-pr`, per task instruction. Remaining acceptance risk is the real multi-node ILM workflow; #2237 stays open until a build containing this change passes the directed case and full ILM repetitions.

## Impact

No admin route, authorization, public error code/status/message, force semantics, configuration schema, or on-disk format changes. Existing and ambiguous references remain fail closed. The only added wait is a bounded 30-second drain on tier configuration mutation paths; ordinary object GET/PUT and lifecycle worker hot paths gain no scans or persistence work.

## Additional Notes

Final-diff adversarial review 1 (correctness, security, compatibility, test coverage): found and fixed the no-intent stale-manager admission regression, the local target-superset gap, and a test-only mutex-across-yield deadlock. The final diff keeps canonical external errors, emits no credentials, retains authoritative reference gates, and has no remaining blocker.

Final-diff adversarial review 2 (concurrency/durability, performance, simplicity): verified config namespace WRITE -> admin update -> manager -> runtime lock ordering; no manager/runtime guard crosses lease waits or network I/O; Prepared durability precedes admission closure and drain; exact ID/revision rejects recovery races; abort restores admission; no extra authoritative scan, format version, or hot-path overhead. No remaining blocker.

The separate release-validation evidence-preservation change is rustfs/rustfs-release-validation#7; it does not substitute for product or real-machine acceptance of this PR.


overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #7219, opened 2026-09-05
feat(heal): add pending legacy MRF migration staging

houseme:
## Related Issues

Refs rustfs/backlog#2268 and rustfs/backlog#2240. This prerequisite does not close W13 or fix the production replay deletion window.

## Summary of Changes

Add explicit pending staging for legacy MRF sources that disagree, change during capture, or contain responsibilities still needed after replay. Capture both journal paths on every configured local disk, retaining original bytes, source identities, digests, and distinct record encodings. Publish a bounded COW candidate, revalidate sources, commit its manifest, and read back the committed evidence before returning success.

Successors inherit earlier source responsibilities. Damaged, unsupported, or over-budget lineage blocks writing; exact retry candidates are validated after inheritance, and orphan payloads require matching committed evidence or the complete retry candidate. Reuse the existing manifest codec and storage-owner conditional-file API. Pending artifacts are separate from the active reader, and there is no automatic consumer integration or source reclamation.

## Verification

- `cargo test --locked -p rustfs-heal --lib heal::mrf_queue::snapshot:: -j4`: 29 passed, 0 failed, 0 ignored on final HEAD `3ebc65cc4589bf4b855e76505f2c33957c6b4158`, including 16 migration fixtures and 13 existing snapshot-reader tests. This was a fresh run after merging main, not a reused result from an earlier HEAD.
- `rustfmt --check --edition 2024 crates/heal/src/heal/mrf_queue/snapshot.rs`, `git diff --check a6b5da64f27c2c0c82fe8ce5592a0e9e80138655...HEAD`, `bash scripts/check_doc_paths.sh`, and `bash scripts/check_no_planning_docs.sh` passed. The merge left the three task files, complete `crates` tree, Cargo manifests, lockfile, and toolchain byte-identical to the preceding delivery HEAD.
- Two independent final-base reviews found no new issue. Not run: real process-kill, disk-full, directory-sync failure, power-loss, mixed-version, and 100k-record replay/write-amplification experiments. Boundary injection and same-process reopen do not substitute for them.

## Impact

Only explicitly requested staging writes the reserved pending-import paths. The production legacy consumer, its current replay/flush behavior, the active snapshot writer, normal request paths, and source deletion policy remain unchanged. A successful staging result is not a durable-admission, cluster-quorum, activation, or GC receipt. Interrupted claims remain blocked until a separately fenced recovery procedure exists.

- Correctness: strict lineage checks and inherited source evidence prevent fallback from silently omitting a newer responsibility.
- Concurrency/durability: ordered conditional claims, source revalidation, manifest commit, and readback retain recovery anchors; metadata durability follows the existing storage policy.
- Compatibility: legacy raw bytes, kind/scope distinctions, and nil/absent version encodings are retained without inventing incarnation identity.
- Security: configured local-disk coverage and bounded strict parsing fail closed; inspection evidence never grants deletion rights.
- Simplicity: reuse the existing owner CAS and manifest codec.
- Performance: byte, record, and source-history limits are explicit; no throughput or long-running memory result is claimed.
- Test coverage: focused fixtures exercise subset/superset sources, changed sources, corrupt or oversized newer generations, empty-source inheritance, claim release failures, and exact retries across replica and commit boundaries.

## Additional Notes

The PR is limited to the pending-staging implementation and its testing reference, based on main at `a6b5da64f27c2c0c82fe8ce5592a0e9e80138655`. Dependency refreshes already present on main are absent from the final diff. Rollback preserves pending and legacy artifacts. Production activation still requires legacy-writer coordination, bounded recoverable ingress, exact disposition/successor receipts, and W14/W21 crash and compatibility evidence.


crates/heal/src/heal/mrf_queue/snapshot/migration.rs
```diff
@@ -0,0 +1,1293 @@
...
+    let (sequence, slot) = previous.map_or((1, 0), |old| {
+        if old.manifest.payload_digest == digest {
+            (old.manifest.sequence, old.slot)
+        } else {
+            (old.manifest.sequence.saturating_add(1), 1 - old.slot)
+        }
+    });
+    let manifest = Manifest::encode(owner, sequence, &payload)?;
+    for disk in disks {
+        install(disk, PAYLOADS[slot], &payload, limits.max_bytes).await?;
```
overtrue, line 488: [P2] Keep reused-slot writes retryable. After generations 1 and 2 commit, generation 3 overwrites slot 0 while its generation-1 manifest remains. An ordinary error before the new manifest releases the claims, but retrying the same candidate fails in read_staging_lineage before exact-retry validation. A real-disk regression through capture_legacy_migration/stage_legacy_migration with the existing AfterPayload hook returns Snapshot(TooLarge), despite unchanged sources, owner and limits. Preserve a recoverable slot transition and cover the third-generation retry.

overtrue (changes requested):
Please keep the third-generation write retryable after a recoverable error.

crates/heal/src/heal/mrf_queue/snapshot/migration.rs
overtrue, line 488: [P2] Preserve retryability when reusing a committed slot. This head still overwrites the generation-1 payload before replacing its manifest. A real-disk regression commits A and B, stages C while the legacy source changes to D, observes SourceChanged and released claims, then restores C. Retrying the same candidate fails with Snapshot(TooLarge) because read_staging_lineage applies the old manifest length to the new payload before retry reconciliation. Keep the reused slot recoverable across this ordinary error and cover the third-generation retry.

overtrue (changes requested):

crates/heal/src/heal/mrf_queue/snapshot/migration.rs
houseme: Status: Fixed in 79b132a3 and covered further in 422f4ab. read_staging_lineage now treats a reused slot with a stale manifest and newer payload as an orphan, so exact retry can complete while plain recovery still fails closed on unresolved orphan evidence. Added third-generation AfterPayload/BeforeManifest retry coverage and verified cargo test -p rustfs-heal migration_ plus cargo clippy -p rustfs-heal --lib --tests -- -D warnings.

houseme: Status: Fixed in 422f4ab. The new regression commits generations A and B, interrupts generation C with a source change to D, restores C, and verifies retry commits sequence 3 with only A/B/C responsibilities. The lineage reader keeps the reused-slot payload recoverable as exact retry evidence but recover_pending_migration still reports a conflict until the retry completes.

houseme:
Conclusion: Fixed and verified. I adopted both review comments by preserving retryability for reused staging slots when a newer payload is paired with a stale committed manifest, while keeping unresolved orphan evidence conflict-safe for plain recovery. The source-change variant is also covered: after generations A and B commit, generation C can hit SourceChanged, restore the same C payload, and retry to sequence 3 without importing the changed D source. Verification completed: cargo fmt --all --check, git diff --check, cargo test -p rustfs-heal migration_third_generation, cargo test -p rustfs-heal migration_, and cargo clippy -p rustfs-heal --lib --tests -- -D warnings. Follow-up commits are 79b132a3 and 422f4ab.

overtrue (approved):
LGTM

PR #7228, opened 2026-09-05
test(scanner): diagnose raw enumeration across restarts

houseme:
## Related Issues

Refs rustfs/backlog#2260, rustfs/backlog#2267, and rustfs/backlog#2240, especially https://github.com/rustfs/backlog/issues/2240#issuecomment-5549222480. This diagnostic does not close the raw-enumeration restart (R-E) gate.

## Summary of Changes

Add a test-only observation hook at the real scanner's raw directory-entry delivery point, a healthy scanner/cache-codec control, and a bounded subprocess diagnostic. The driver starts a fresh OS scanner test-worker process each round against the same synthetic disk and cache. It verifies the exact test selection and child PID, enforces the fixed entry limit, and reports observed enumeration separately from processed and retained object coverage.

The strict oracle requires complete, exact object/version/byte coverage. It does not assert that starvation must remain present, run a final unbudgeted sweep, use an assumed stable `readdir` order, or manufacture a persisted enumeration cursor. No production algorithm, dependency, on-disk format, or CI wiring changes are included. The guide identifies the missing storage-owner capability and its first-build/restart/budget requirements.

## Verification

Tested clean commit `ed026a0f1f3aa843f3f2b1499d68c0a0dcbb86f8`. The real scanner was compiled and tested at `f25df09f6177d71fccad961e8610c464e9b2a4ff`; the final commit changes only Python report validation and its tests. Both real CLI scenarios were repeated on the final clean commit with that unchanged Rust test executable.

1. `RUST_MIN_STACK=4194304 cargo test -p rustfs-scanner --lib enumeration_restart_worker -j4 -- --nocapture`: 1 passed, 0 ignored. The default worker performs a real scan of four versioned `xl.meta` fixtures whose metadata declares `size=1`, verifying the resulting 4/4/4 accounting oracle plus codec reload. No S3 object body is uploaded or verified. The observation uses the canonical path returned by the disk owner, including temporary-directory aliases.
2. `python3 -m unittest discover -s scripts -p test_diagnose_scanner_enumeration_restart.py`: 8 passed, including rejection of complete coverage with zero observed entries so a missed budget hook cannot produce a false pass. The real diagnostic with `--objects 128 --raw-entry-budget 8 --rounds 8` returned exit 1: eight separate worker processes each observed eight raw entries but processed and retained zero objects, with completeness false. This is a reproduced unmet convergence oracle, not a passing R-E gate. The small control with `--objects 4 --raw-entry-budget 16 --rounds 2` returned exit 0 after its first worker completed exact 4/4/4 coverage.
3. `cargo fmt --all --check`, `git diff --check`, and `bash scripts/check_doc_paths.sh` passed. The macOS linker emitted its existing large `__eh_frame` warning; no test was ignored or assertion relaxed.

The strict command uses the scanner libtest executable built from the tested commit; its full invocation is documented in `docs/testing/scanner-checkpoint-fixture.md`. Whole-daemon restart, crash/fsync durability, EC quorum persistence, remote RPC, and throughput/performance gates were not run.

## Impact

Test/tooling-only. The hook is absent from non-test builds and only observes the selected synthetic disk during tests. The entry limit is deterministic injected work accounting, not a wall-clock performance limit; kernel prefetch, metadata probes, cache I/O, and name bytes are not independently budgeted. Raw entries observed are not durable enumeration progress.

The existing storage APIs still do not provide a restartable raw-directory snapshot/index with budgeted first construction. This PR supplies executable evidence and a bounded positive oracle, not that missing implementation or a production convergence guarantee. Retained artifacts contain only the generated synthetic namespace and diagnostic reports.

## Additional Notes

Exit 0 means exact completion for the selected fixture, exit 1 means the convergence oracle is unmet, and exit 2 means invalid input, worker failure, or invalid evidence. No public API or deployment configuration changes are introduced. Two independent final-diff reviewers passed correctness, simplicity, bounded-resource and test-coverage checks after the zero-observation false-pass finding was fixed and both reviewers rechecked the affected code. Their verdict does not make the R-E convergence gate pass.


overtrue (approved):
LGTM

overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #7232, opened 2026-09-05
test(scanner): verify restart evidence against tested builds

houseme:
## Related Issues

Part of rustfs/backlog#2269 (W21). This branch includes the W20 code prerequisite from #7181, which remains open; it does not claim that prerequisite has merged or close the full W21 acceptance task.

## Summary of Changes

Register one existing background-target-restart E2E case and emit bounded evidence only after its real process restart, per-node listings, complete S3 bodies, and physical shard checks succeed. Reuse the existing test-wiring tool to bind receipts to compiled source/lock/features, actual binaries, nextest selection, JUnit execution timestamps, and strictly typed oracle fields.

Keep the remaining release requirements explicitly pending. A focused case pass cannot approve writer activation, mixed-version rollout, power-loss durability, EC8+4 coverage, or ABBA performance. The included W20 harness keeps synthetic validation separate from measurements.

Fix the included ABBA adapter/collector cleanup so an exited parent cannot hide surviving children. Retain the session leader until the final process-group signal, preserve the ten-second grace period, and guarantee group KILL/reap after cleanup interruption or gate-initialization failure. Successful prepare actions keep their services until adapter stop.

## Verification

- Final cleanup commit `be82ea7b7`: `scripts/test_scanner_validation_harness.sh` passed, including all 16 ABBA tests. Four real-child failure cases were observed failing before their fixes and passing afterward; controls also verify successful prepare service lifetime and that a reaped process-group ID is never signalled.
- `scripts/python_bin.sh scripts/check_test_wiring.py --self-test` passed all 21 tests; the default wiring gate, Python resolver checks, 9 nightly-candidate tests, `actionlint`, documentation guards, Rust formatting for the changed files, and `git diff --check` passed. `cargo check --locked -p e2e_test --tests -j 4` passed on `05cbac10`; the later cleanup commit changes only Python and documentation.
- At the frozen integration baseline, `scripts/test_security_workflow.py` has 5 failing chain-handoff subcases because `LOG_FILE` is not initialized before the replication step runs. The inherited test and workflow were not modified by this slice. Main has since changed that test; its newer result was not verified here. No assertions or CI failure propagation were weakened.

No real E2E run or ABBA measurement was performed on this delivery head. Process-group regressions ran on macOS; Linux execution remains for CI. The earlier `aca241b17` restart run and its original receipt/JUnit/oracle files were verified present, but are historical evidence only and are not validation of the updated runtime baseline.

## Impact

This is a registry/receipt delivery slice. The E2E evidence path is opt-in through `RUSTFS_SCANNER_HEAL_RUN_DIR`, with explicit expected crate features through `RUSTFS_E2E_EXPECTED_FEATURES`. The registry continues to block release for all 21 pending gate groups. Production scanner/heal policy and persistent writer activation are unchanged by the task-owned diff.

The new main functional CI contracts, including script-contract execution and suite failure propagation, are preserved. Existing workflows are not treated as proof that the scanner-specific oracle matrix completed.

Two independent final-diff reviews passed after all four process-cleanup counterexamples were fixed. Correctness and evidence integrity: receipts cannot approve unexecuted gates. Concurrency/resource ownership: group signals precede leader reap, including initialization and cleanup interruption. Compatibility: successful prepare services survive and the existing adapter contract remains intact. Simplicity/performance: the same bounded-lifetime helper serves both callers without a supervisor or pipe/logging redesign. Test coverage: real child-process regressions and positive controls cover the reported failures; Linux and full release lanes remain unverified locally.

## Additional Notes

The frozen integration base is `8f763fb1a214a4050649758e4d45257b23aa4d5a`. W20 prerequisite code is included explicitly while #7181 is open. Reverting this slice removes the evidence tooling and optional E2E output without deleting scanner floors, journals, intents, or runtime data.


scripts/scanner_abba.py
```diff
@@ -0,0 +1,403 @@
...
+        drift = max(abs(a2[k] / a1[k] - 1) for k in ("p99_ms", "throughput_ops"))
+        repeat_drift = max(abs(b2[k] / b1[k] - 1) for k in ("p99_ms", "throughput_ops"))
+        noise = max(drift, repeat_drift) > 0.05
+        a = {key: (a1[key] + a2[key]) / 2 for key in METRICS}
+        b = {key: (b1[key] + b2[key]) / 2 for key in METRICS}
+        p99 = b["p99_ms"] / a["p99_ms"] - 1
+        throughput = b["throughput_ops"] / a["throughput_ops"] - 1
+        thresholds = {"p99_regression": 0.10 if control else 0.05,
+                      "throughput_loss": 0.05 if control else 0.03}
+        passed = p99 <= thresholds["p99_regression"] and throughput >= -thresholds["throughput_loss"]
```
overtrue, line 295: [P2] Accept measurements exactly at the documented limits. With identical A legs at p99=100 and throughput=100, identical B legs at p99=105 and throughput=97, valid oracle/P2 evidence and no drift, validate_result accepts every cell but evaluate returns fail: floating-point division produces 0.050000000000000044 and -0.030000000000000027. These are exactly the allowed 5% and 3% boundaries. Use a boundary-safe comparison and add exact-limit tests; the current #7181 prerequisite already includes this correction.

overtrue (changes requested):
Please preserve the documented inclusive ABBA limits.

houseme:
Thanks, fixed in 6e7909b75c.

I preserved the documented inclusive ABBA limits by switching the ratio comparisons to `Decimal`-backed boundary checks while keeping the W21 process-group cleanup path intact. The synthetic adapter now covers exact p99/throughput thresholds, just-over-threshold failure, and the fractional P1 boundary.

Verification run:

- `scripts/python_bin.sh -m py_compile scripts/scanner_abba.py scripts/test_scanner_abba.py`
- `scripts/python_bin.sh scripts/test_scanner_abba.py`
- `scripts/test_scanner_validation_harness.sh`
- `git diff --check`

overtrue (approved):
LGTM

overtrue (approved):
LGTM

PR #7245, opened 2026-09-05
fix(ecstore): retain namespace owners through local disk completion

overtrue:
## Related Issues

Relates to [rustfs/backlog#2320](https://github.com/rustfs/backlog/issues/2320), [rustfs/backlog#2324](https://github.com/rustfs/backlog/issues/2324), and [rustfs/backlog#2304](https://github.com/rustfs/backlog/issues/2304).

## Summary of Changes

When a disk deadline expires while a blocking rename is still running, namespace pending can reach zero before the filesystem finishes publishing. Keep the existing namespace owner in the physical lease through local rename, rollback, undo and post-publication fsync. Undo shares the existing version-delete implementation, and namespace ownership stays separate from quota and scanner guards.

Grouped fsync keeps every batch owner's lease in its physical closure. Its cleanup guard is captured as a whole value before the worker's first poll. Idle cleanup also checks the registered Arc, so an old worker cannot remove a successor group with the same key. The change touches six ECStore source files and their regression tests.

## Verification

- Focused native runs used `cargo nextest run --locked --offline -p rustfs-ecstore --lib --features test-util --build-jobs 2 --profile ci --retries 0 --test-threads 1 --no-fail-fast --no-tests=fail -E "$FILTER"`. The selections cover rename/undo deadlines, successful ACK with a cancelled caller and a physical tail, quota coexistence, rollback recovery, and grouped fsync. Both new group-identity tests failed at their intended assertions on `ed100103d03e71aa4b8f7da3ae24a28b2caf2bbc` and passed on `5e33186cf660b48810149ac272cb4d31a14f8a57`.
- The 93-test run on `5e33186c` passed 87 tests; six new fixture body assertions failed because their inline seeds lacked the normal metadata marker. Only those seed preparations changed in `58f18406307492e50ed71e814765ab15d4ff2812`. All eight tests sharing the edited helpers then passed, including the two previously passing controls. The other 85 passes remain attributed to `5e33186c`; this is not a fresh 93-test run on the final head. JUnit UUIDs: `3cd7d64a-2691-42fb-a59a-5ee64094adb9` and `ff85d16d-1189-451d-b853-326c2d4beb07`. Each run used zero retries and had one separately passing setup.
- On `58f184063`: `cargo clippy --locked --offline -p rustfs-ecstore --all-targets --features test-util -j2 -- -D warnings` and `cargo fmt --all --check` passed. After merging main at `bde4b78a9`, the six task files, complete task diff and all 56 build inputs remain unchanged; only upstream CI and three ODM test fixtures changed. `git diff --check bde4b78a9 3c54ac1de` passed.

- Follow-up on `3ae29aab2`: the rio-v2 CI lane failed `undo_fresh_version_keeps_physical_namespace_owner_after_timeout` at `assert!(!ctx.namespace_commits_pending())`. The tests treated a non-upgradable `Weak` probe or a reacquired mutation lease as proof that the owner had been released, but `Arc` drops its strong count before running `Drop` and the lease releases its locks before its owner field, so the pending counter could still be one. Every drain wait now also requires `namespace_commits_pending()` to be false. The 14 affected tests passed 20 consecutive runs with `--features rio-v2,test-util --retries 0`; `cargo clippy -p rustfs-ecstore --all-targets --features rio-v2,test-util -- -D warnings` and `cargo fmt --all --check` passed.

The first group test attempt failed compilation and ran zero tests; it is not RED evidence. That compilation also exposed the incomplete async guard capture, which was corrected and independently checked before the native runs above.

## Impact

Ordinary deadlines, quota/scanner guards, health errors, rollback receipts and ACK policy remain unchanged. There is no RPC or disk-format change, new configuration, or extra filesystem operation. Existing Arc ownership lasts until the corresponding physical work ends. Remote instance binding and recovery remain separate B/C tasks; this PR does not complete F2.

Independent final and affected reviews: correctness passed; concurrency/durability passed after correcting whole-value capture; compatibility passed for the stated local scope; test coverage passed with the fixture correction and eight-case rerun; simplicity passed with one existing execution path; performance review found no added I/O or executor, with no throughput benchmark claimed.

## Additional Notes

Local validation ran on macOS. Windows, Linux-specific paths, process crashes and remote or mixed-version behavior were not exercised locally. No data migration is needed. Reverting restores the earlier local ownership gap.

### DIFF
--- a/crates/ecstore/src/disk/local.rs
+++ b/crates/ecstore/src/disk/local.rs
@@ -13171,10 +13171,12 @@ mod test {
             let object_dir = disk.io_get_object_path(bucket, &object).expect("object IO path");
             let xl_path = object_dir.join(STORAGE_FORMAT_FILE);
             let old_version = Uuid::new_v4();
-            let old = test_file_info(&object, old_version, None, Some(Bytes::from_static(b"old-payload")));
+            let mut old = test_file_info(&object, old_version, None, Some(Bytes::from_static(b"old-payload")));
+            old.set_inline_data();
             let old_meta = test_meta(old.clone());
             let new_version = if case == "backup" { old_version } else { Uuid::new_v4() };
             let mut new = test_file_info(&object, new_version, None, Some(Bytes::from_static(b"new-payload")));
+            new.set_inline_data();
             let rollback_dir = Uuid::new_v4();
             let mut opts = DeleteOptions {
                 undo_write: true,
@@ -13367,10 +13369,13 @@ mod test {
         let xl_path = object_dir.join(STORAGE_FORMAT_FILE);
         fs::create_dir_all(&object_dir).await.expect("object directory");
         let version_id = Uuid::new_v4();
-        let old_meta = test_meta(test_file_info(object, version_id, None, Some(Bytes::from_static(b"old-payload"))));
+        let mut old = test_file_info(object, version_id, None, Some(Bytes::from_static(b"old-payload")));
+        old.set_inline_data();
+        let old_meta = test_meta(old);
         fs::write(&xl_path, &old_meta).await.expect("old metadata");
         set_rename_data_fail_after_metadata_commit(object);
-        let new = test_file_info(object, version_id, None, Some(Bytes::from_static(b"new-payload")));
+        let mut new = test_file_info(object, version_id, None, Some(Bytes::from_static(b"new-payload")));
+        new.set_inline_data();
         let (entered_tx, entered_rx) = tokio::sync::oneshot::channel();
         let (release_tx, release_rx) = std::sync::mpsc::channel::<()>();
         let _hook = hooks::install_at(hooks::Stage::Rollback, &xl_path, move || {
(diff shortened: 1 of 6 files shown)


crates/ecstore/src/disk/local.rs
```diff
@@ -12971,6 +13155,461 @@ mod test {
...
+                        "marker" | "fresh" => !xl_path.exists(),
+                        "remaining" => std::fs::read(&xl_path)
+                            .ok()
+                            .and_then(|data| FileMeta::load(&data).ok())
+                            .is_some_and(|meta| {
+                                meta.find_version(Some(old_version)).is_ok() && meta.find_version(Some(new_version)).is_err()
+                            }),
+                        _ => unreachable!(),
+                    };
+                    if published && owner_probe.upgrade().is_none() {
```
overtrue: [P2] Wait for the namespace counter to finish dropping. Weak::upgrade() can return None after the last strong count reaches zero but before NamespaceCommitGuard::drop decrements pending on the blocking worker. The exact-head undo_fresh_version_keeps_physical_namespace_owner_after_timeout test already fails at the final pending assertion in the rio-v2 run. Include !ctx.namespace_commits_pending() in this completion wait and retain the ownership assertions.

PR #7252, opened 2026-09-06
feat(ilm): persist recovery controls for legacy tier journals

cxymds:
## Related Issues

Refs rustfs/backlog#2206.

## Summary of Changes

- Discover canonical tier-delete journal v1/v2 records and persist create-only recovery controls classified as `retained_ambiguous` with a bounded `remote_version_unknown` reason.
- Classify undecodable or path-mismatched legacy records as `corrupt` without retaining raw journal fields in the control or admin view.
- Treat existing non-retrying controls as durable scheduler fences so repeated scans avoid full source observation and write attempts, while preserving the original source generation for later operator-side validation.
- Return before tier backend acquisition for every legacy and corrupt path; no remote PUT, GET, probe, or DELETE is attempted.

## Verification

- `cargo test -p rustfs-ecstore --features test-util --lib legacy_tier_delete_journals_create_redacted_recovery_controls_without_remote_calls` passed 1/1 after rebasing onto the tested head. It covers pinned v1/v2 JSON, corrupt bytes, concurrent discovery, equivalent source rewrites, redacted views, stable controls, and an empty backend operation log.
- `cargo test -p rustfs-ecstore --features test-util --lib bucket::lifecycle::tier_delete_journal::tests` passed 24/24, and `cargo test -p rustfs-ecstore --features test-util --lib bucket::lifecycle::recovery_control::tests` passed 7/7.
- `cargo check -p rustfs-ecstore`, `cargo fmt --all -- --check`, `git diff --check`, and `./scripts/check_logging_guardrails.sh` passed. `make pre-pr` was not run per maintainer direction; scoped compilation and tests cover the changed crate and recovery path.

## Impact

Existing v1/v2 and corrupt legacy tier-delete journals remain on disk but become visible through the existing authenticated recovery-control list and inspect APIs. The change adds no configuration, wire API, or remote-tier behavior. Compatibility remains fail closed: v3-v6 processing is unchanged, mixed or replaced source generations cannot authorize remote work, and control output contains only canonical digests, bounded schemas, classifications, counters, and copy counts.

Adversarial review verdicts: correctness found and fixed repeated local recovery I/O after a control exists; simplicity found no smaller design that preserves create-only reconciliation and durable fencing; test coverage exercises the real recovery and admin-view data path; security found no raw remote object, version, tier, or credential exposure; concurrency/durability verified create-only races and lost-response reconciliation remain fail closed; compatibility verified pinned v1/v2 input while leaving v3-v6 unchanged; performance confines the all-set observation and control write to first discovery.

## Additional Notes

This is the first of three scoped deliveries for rustfs/backlog#2206. Immutable export remains deferred until the next PR can bind it to all-member topology proof and an observation receipt. Destructive abandon disposition remains deferred to the final PR with generation fences, CAS, quotas, restart recovery, and garbage collection.

---

### DIFF
--- a/crates/ecstore/src/bucket/lifecycle/tier_delete_journal.rs
+++ b/crates/ecstore/src/bucket/lifecycle/tier_delete_journal.rs
@@ -35,6 +35,10 @@ use crate::bucket::lifecycle::config_boundary;
 use crate::bucket::lifecycle::durable_namespace::{
     TIER_DELETE_JOURNAL_NAMESPACE, TIER_DELETE_JOURNAL_V6_NAMESPACE, validate_durable_ilm_record,
 };
+use crate::bucket::lifecycle::recovery_control::{
+    IlmRecoveryClassification, IlmRecoveryControl, IlmRecoveryControlIdentity, IlmRecoveryErrorCode, IlmRecoveryProtocol,
+    load_recovery_control, observe_recovery_source, save_recovery_control_if_absent,
+};
 use crate::bucket::lifecycle::runtime_boundary;
 use crate::bucket::lifecycle::tier_sweeper::{
     Jentry, TierDeleteDispatchBinding, TierDeleteJournalState, TierDeleteSourceIdentity,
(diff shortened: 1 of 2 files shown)


crates/ecstore/src/bucket/lifecycle/tier_delete_journal.rs
```diff
@@ -5509,6 +5520,127 @@ enum TierDeleteJournalEntryRecoveryOutcome {
...
+        && control.owner.is_none()
+        && control.attempt_count == 0
+        && control.consecutive_failure_count == 0
+}
+
+fn legacy_tier_delete_control_is_scheduler_fence(control: &IlmRecoveryControl, identity: &IlmRecoveryControlIdentity) -> bool {
+    control.identity == *identity && control.owner.is_none() && !control.classification.permits_automatic_attempt()
+}
+
+async fn persist_legacy_tier_delete_recovery_control(
```
overtrue, line 5562: [P2] Keep the new helper within the enforced Clippy limit. This eight-argument function fails the existing `cargo clippy ... -- -D warnings` gate with `clippy::too_many_arguments`; both the rio-v2 and sftp jobs for this head stop with exit 101 before tests. Please reduce the argument list and rerun the affected checks. Evidence: https://github.com/rustfs/rustfs/actions/runs/34008643565/job/101421044582

overtrue (changes requested):

crates/ecstore/src/bucket/lifecycle/tier_delete_journal.rs
cxymds: Verified the fix in bb27dd59d: grouping the recovery descriptor reduces the helper to seven arguments without a lint allowance. Both `cargo clippy -p rustfs-ecstore --all-targets --features rio-v2 -- -D warnings` and `cargo clippy -p rustfs-ecstore --all-targets -- -D warnings` pass, and the focused legacy-journal recovery test passes (1/1). The full two-package rio-v2 command also progresses past this finding; it is independently blocked later by the pre-existing `chunks_exact_to_as_chunks` lint in `rustfs/src/on_demand_migration/azure.rs`, outside this PR diff.

overtrue (approved):

PR #7309, opened 2026-09-06
fix(s3-tables): preserve encoded paths in client signatures

GatewayJ:
## Related Issues

N/A

## Summary of Changes

The S3 Tables smoke clients sign an already encoded namespace path such as `sales%1Ftax` as `sales%251Ftax`, causing signature mismatches. RustFS profiles now use S3 path canonicalization in both direct REST requests and the PyIceberg adapter, and the adapter signs the complete wire URL without losing repeated or empty query parameters. Other vendor profiles keep their generic SigV4 signer.

Install the adapter during REST session initialization so the first `/config` request is signed with its payload hash. PyIceberg, DuckDB's PyIceberg setup, and sessions recreated with vended data credentials share this initialization. Preserve named catalog configuration, TLS certificates, and explicit property precedence.

## Verification

Tested commit: `515d81fba` (based on `22bff27ae`).

- From `scripts/table-catalog`: `python3.12 -B -m unittest test_pyiceberg_smoke test_duckdb_smoke test_engine_compatibility` — 89 tests passed.
- `python3 -B -m unittest discover -s scripts/table-catalog -p test_pyiceberg_signing.py` — 5 tests passed with PyIceberg 0.10.0 and botocore 1.42.97. The encoded adapter/direct-request tests fail in 19 subcases against the unmodified baseline. Coverage includes first config, recreated sessions, named TLS configuration, literal percent escapes, queries, request mutation, and vendor signing profiles.
- `make -j4 fmt-check` and `git diff --check` passed. No server code changes; Rust compilation and full workspace tests were not required. A live server smoke run was not performed for this PR; transport tests use real PyIceberg/requests/botocore with a captured HTTP transport.

## Impact

Changes are confined to S3 Tables client scripts and their tests. Server signature verification, authorization, storage behavior, dependencies, and wire formats are unchanged. The initialization hook is covered against PyIceberg 0.10.0; future PyIceberg upgrades must pass the bootstrap regression tests.

## Additional Notes

Independent review covered these boundaries:

- Correctness: first request, repeated query parameters, and session reconstruction preserved.
- Simplicity: shared signing and initialization replace duplicated client paths.
- Security: signed method/path/body remain bound; vended data credentials do not replace Catalog signing credentials.
- Concurrency: adapter/session changes remain local to each catalog instance.
- Compatibility: named TLS configuration and non-RustFS signer selection preserved.
- Performance: no additional network requests or shared synchronization introduced.
- Test coverage: failing-before encoding checks and real-library bootstrap checks pass.

PR #7310, opened 2026-09-06
perf(s3-tables): avoid duplicate registration prefix scans

GatewayJ:
## Related Issues

N/A

## Summary of Changes

Creating a table checks warehouse prefix overlap twice while holding the bucket publication fence. Both checks list table entries and warehouse indexes and read their contents, so the duplicate work increases lock hold time as the catalog grows.

Reuse the registration path's completed overlap check when reserving its warehouse index. Keep the existing reservation checks for commit, backfill, and failed-drop restoration, including the scan-free fast path for an already reserved warehouse. The publication fence, conditional writes, and rollback behavior remain unchanged.

## Verification

Tested commit: 50125ce7a, based on origin/main at 22bff27ae, with no additional source changes.

- Failing before/passing after: `object_catalog_registration_scans_warehouse_prefixes_once` observes four LIST calls on the base implementation and the expected two with this change. It also checks that each additional existing table adds only two reads. The existing commit test now asserts zero LIST calls for an unchanged warehouse and passes both before and after.
- `cargo nextest run --locked -j 4 -p rustfs --lib -E 'test(table_catalog)'`: 577 passed. `cargo nextest run --locked -j 4 -p rustfs --lib -E 'test(table_data_plane_publication) | test(multipart_parts_skip_table_publication)'`: 2 passed, covering content mutation and multipart publication fence behavior.
- `make -j4 fmt-check` and `git diff --check`: passed. Validation is scoped to the changed catalog implementation and its publication fence dependents; a full workspace gate was not run.

The original live eight-writer/100-table workload was not rerun. These checks establish removal of duplicate work and preserve catalog behavior; they do not establish a throughput target or prove that every lock timeout is eliminated.

## Impact

Reduces catalog reads inside the existing bucket publication fence during registration. One complete overlap scan remains, so registration still scales with existing catalog entries. No API, storage format, configuration, lock timeout, or lock ordering changes.

## Additional Notes

Independent from the S3 Tables client signing correction in #7309.

Two independent reviewers covered all seven high-risk review roles:

- Correctness: attacked overlapping prefixes, stale exact reservations, and missing indexes; no unresolved findings.
- Simplicity: checked existing helper reuse and the smaller local change; no unresolved findings.
- Security: attacked warehouse isolation and untrusted paths; no unresolved findings.
- Concurrency/durability: checked fence ownership, cancellation, conditional writes, and rollback; no unresolved findings.
- Compatibility: checked API, stored indexes, and existing reservation callers; no unresolved findings.
- Performance: checked registration scan counts and unchanged-warehouse commit fast paths; no unresolved findings.
- Test coverage: checked failing-before evidence and catalog/fence regressions; no unresolved findings.


GatewayJ:
@codex

PR #7208, opened 2026-09-05
feat(scanner): reuse clean local bucket prefixes

marshawcoco:
## Related Issues

N/A

## Summary of Changes

Tracks bounded, process-local top-level object mutation hints for dirty buckets. When a complete local baseline is available, a normal non-erasure scanner can reuse known-clean direct child subtrees and rescan only the changed top-level entries. Ambiguous paths, hint overflow, restart, remote or distributed scanning, lifecycle, replication, Object Lock, and erasure-mode scans retain the existing whole-bucket behavior.

## Verification

- `rustfs-cargo-safe test -p rustfs-scanner --lib dirty_usage_object -- --test-threads=1` passed 2 targeted dirty-hint tests.
- `rustfs-cargo-safe test -p rustfs-scanner --lib scoped_ -- --test-threads=1` passed 12 scoped-cache and erasure-fallback tests.
- `rustfs-cargo-safe check -p rustfs`, `rustfs-cargo-safe fmt --all --check`, `git diff --check`, and `./scripts/check_architecture_migration_rules.sh` passed.

## Impact

No configuration or public API changes. Eligible local usage scans reduce unnecessary subtree walks without changing usage authority or dirty-generation acknowledgement semantics.

## Additional Notes

Distributed and erasure-code paths intentionally remain conservative until a durable, peer-verified mutation journal exists.

PR #7313, opened 2026-09-06
fix(replication): forward single-part object checksums as headers

reatang:
## Related Issues

rustfs/backlog#2340 (finding: replicas of single-part objects carry no `x-amz-checksum-*`). Not Wasabi specific: it reproduces between two local RustFS sites and with bucket replication.

## Summary of Changes

`replication_put_object_options` decoded the stored checksum record with `decrypt_checksums`, whose map is keyed by algorithm name (`CRC32`, `SHA256`, ...), and inserted those keys into the outbound user metadata as they were. `PutObjectOptions::header()` only passes `x-amz-*` keys through as headers, so every checksum left the source as `x-amz-meta-crc32: <base64>` user metadata and the target never received a checksum. A replica's `HEAD ?ChecksumMode=ENABLED` therefore returned nothing while the source HEAD returned the value.

The record is now translated to the `x-amz-checksum-<algorithm>` header names (`ChecksumType::key`) on the single-PUT route, which is what MinIO `getCRCMeta` sends. On the multipart route the object-level record is no longer forwarded at all: a multipart replica is rebuilt through CreateMultipartUpload/UploadPart, and announcing a checksum there that the parts do not carry would be rejected by the target. Previously that route sent the same bogus `x-amz-meta-<algorithm>` metadata, so a multipart replica loses one stray user-metadata key. Managed-SSE objects are unchanged (their checksum record is not exposed as plaintext headers).

With `RUSTFS_REPLICATION_STREAMING_CHECKSUMS=true` the SDK adds a CRC32 trailer to every upload; a target that receives the forwarded SHA256 header and that trailer keeps only the trailer's algorithm, so the forwarded checksum vanished again (observed on the first cut of this PR). `TargetClient::put_object` now pins a PUT that forwards a source checksum to `WhenRequired`, so nothing is sent beside the source's own checksum; uploads without a forwarded checksum keep the knob's trailer.

## Verification

Tested commit: this branch head on top of `origin/main` `22bff27ae`; after the final rebase onto `4d1ce9618` (taking the neighbouring tests from #7315 and #7307, the own-version-id re-drive fix from #7323, and the scanner build fix from #7329) the checksum e2e and all five matrix tests were re-run against a binary built from the rebased head: 6 passed; workspace Clippy clean.

- `cargo test -p rustfs-ecstore --lib replication_put_object_options`: 2 passed.
- `cargo test -p rustfs-ecstore --lib put_object`: 77 passed; the new `streaming_put_object_with_forwarded_checksum_sends_no_sdk_checksum` records a `WhenSupported` client sending a forwarded SHA256 header with no `x-amz-trailer`, and the control upload without a forwarded checksum still carrying the trailer. `replication_put_object_options_forwards_new_algorithm_checksums_like_classic` now asserts the `x-amz-checksum-*` key for all ten algorithms and the absence of the bare name (it asserted the bare name before); `replication_put_object_options_keeps_multipart_checksum_records_off_the_wire` asserts a composite and a full-object multipart record put nothing checksum-related into the outbound metadata.
- `cargo nextest run -p e2e_test -E 'test(test_bucket_replication_forwards_single_part_object_checksums)'`: bucket replication between two local servers, CRC32 and SHA256 single PUTs; before the fix FAIL (`replica lost the CRC32 checksum`, observed with the source hunk reverted), after the fix PASS.
- Site replication (two paired local sites, the report's own `test_replication_checksum[CRC32|CRC32C|CRC64NVME|SHA1|SHA256]` harness): 5 passed (CRC32, CRC32C, CRC64NVME, SHA1, SHA256) against a pair built from the first cut of this branch (the second cut only changes the streaming-checksum mode, which the pair did not use); the same five cases failed on the report's baseline `0b72f3902` (`results-2026-09-06/wasabi-site-control.xml` in the report) with `None == '<checksum>'` on the replica HEAD.
- Outbound target matrix (`replication_target_matrix_test`, four rows, new `Checksummed` shape): all four rows PASS after the fix; before the fix `matrix_baseline_target` FAILs on `baseline/Checksummed: a PutObject went out without the source's x-amz-checksum-sha256 header` (observed with the source hunk reverted).
- The bucket-replication e2e also PASSes with `RUSTFS_REPLICATION_STREAMING_CHECKSUMS=true` exported to the spawned servers: PASS (the first cut of this PR failed it with `replica lost the SHA256 checksum`).
- `make pre-commit`, `cargo clippy --all-targets -- -D warnings`: clean.

Not run: the full `rustfs-ecstore` test suite. Remaining risk: a third-party target that rejects a forwarded `x-amz-checksum-*` header on PutObject; AWS S3 and Wasabi accept them (the report's direct probes), and the value is the one the source verified on upload.

## Impact

Outbound client default change (high risk for every target class per the replication checksum postmortem SOP): a single-part replication PUT of an object that was uploaded with a checksum now carries `x-amz-checksum-<algorithm>` and no longer carries `x-amz-meta-<algorithm>`. Objects uploaded without a checksum, multipart objects, and managed-SSE objects send no checksum header, as before. No new environment knob; `docs/operations/replication-outbound-transport.md` now states the single-part-only forwarding (its previous wording claimed every object-level checksum was forwarded) and lists the checksum-storing target class with its matrix shape. Verified target classes: RustFS (bucket replication e2e, site replication), the fake S3 target matrix. Unverified: AWS S3, Wasabi, other third-party targets.

Adversarial review (high risk: replication outbound defaults), two sequential passes:

- Correctness: traced the record shapes through the changed block: single-part record (header forwarded), composite multipart record (multipart route, nothing forwarded), full-object record on a single-part object (`is_multipart` cleared, header forwarded without the `-N` suffix because `read_checksums` omits it for FULL_OBJECT), full-object record on a multipart object (multipart route, nothing forwarded), unknown algorithm name (`ChecksumType::key` returns `None`, entry dropped instead of leaking as metadata). No finding.
- Compatibility (outbound targets): target-side rules the default still satisfies, with the matrix cell that proves each: plain signed body, never `aws-chunked` (all four rows, every shape including the new `Checksummed`); `Content-MD5` or `x-amz-checksum-*` with Object Lock parameters (`RequireChecksumWithObjectLock` row; a forwarded header now also satisfies it, and `object_lock_put_integrity_for` then adds nothing); version-id adoption unchanged (`MintOwnVersionIds` row); ETag equals content MD5 unchanged. The forwarded value is the source's stored checksum of the plaintext body, which is exactly what the target recomputes. No finding.
- Test coverage: the unit test asserts the header key and the absence of the bare name for all ten algorithms; the e2e asserts the replica's HEAD checksum for CRC32 and SHA256 and that no `sha256` user-metadata key leaked; the matrix `Checksummed` cell asserts the wire header on every PutObject of the shape. E2e and matrix cell fail with the source hunk reverted (observed). No finding.

## Additional Notes

The other findings from the same report are handled in separate PRs.

### DIFF
--- a/crates/e2e_test/src/replication_extension_test.rs
+++ b/crates/e2e_test/src/replication_extension_test.rs
@@ -3988,6 +3988,93 @@ async fn test_bucket_replication_version_purge_of_non_inline_object_releases_sou
     Ok(())
 }
 
+/// Regression for rustfs/backlog#2340 (not Wasabi specific): a single-part
+/// object uploaded with `x-amz-checksum-*` must reach the target with the same
+/// checksum. The outbound options keyed the stored record by algorithm name,
+/// which the target client sent as `x-amz-meta-*` user metadata, so a replica
+/// never carried a checksum although the source HEAD returned one.
+#[tokio::test]
+async fn test_bucket_replication_forwards_single_part_object_checksums() -> TestResult {
+    init_logging();
+
+    let mut source_env = RustFSTestEnvironment::new().await?;
+    let mut source_env_vars = replication_fast_env();
+    source_env_vars.extend_from_slice(LOOPBACK_REPLICATION_TARGET_ENV);
+    source_env.start_rustfs_server_with_env(vec![], &source_env_vars).await?;
+
+    let mut target_env = RustFSTestEnvironment::new().await?;
+    target_env.start_rustfs_server_without_cleanup(vec![]).await?;
+
+    let source_bucket = "replication-checksum-src";
+    let target_bucket = "replication-checksum-dst";
+    let source_client = source_env.create_s3_client();
+    let target_client = target_env.create_s3_client();
+
+    source_client.create_bucket().bucket(source_bucket).send().await?;
+    target_client.create_bucket().bucket(target_bucket).send().await?;
+    enable_bucket_versioning(&source_env, source_bucket).await?;
+    enable_bucket_versioning(&target_env, target_bucket).await?;
+    let target_arn = set_replication_target(&source_env, source_bucket, &target_env, target_bucket).await?;
+    put_bucket_replication(&source_env, source_bucket, &target_arn).await?;
+
+    let body = b"123456789";
+    let crc32_key = "checksum-crc32.txt";
+    let sha256_key = "checksum-sha256.txt";
+    let crc32_put = source_client
+        .put_object()
+        .bucket(source_bucket)
+        .key(crc32_key)
+        .body(ByteStream::from_static(body))
+        .checksum_algorithm(aws_sdk_s3::types::ChecksumAlgorithm::Crc32)
+        .send()
+        .await?;
+    let expected_crc32 = crc32_put.checksum_crc32().ok_or("source PUT omitted CRC32")?.to_string();
+    let sha256_put = source_client
+        .put_object()
+        .bucket(source_bucket)
+        .key(sha256_key)
+        .body(ByteStream::from_static(body))
+        .checksum_algorithm(aws_sdk_s3::types::ChecksumAlgorithm::Sha256)
+        .send()
+        .await?;
+    let expected_sha256 = sha256_put.checksum_sha256().ok_or("source PUT omitted SHA256")?.to_string();
+
+    for key in [crc32_key, sha256_key] {
+        wait_for_source_replication_status(&source_client, source_bucket, key, "COMPLETED", false).await?;
+    }
+
+    let replica = target_client
+        .head_object()
+        .bucket(target_bucket)
+        .key(crc32_key)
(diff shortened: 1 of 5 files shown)


reatang:
`Test and Lint (sftp)`/`(swift)` fail on the same three `ilm_transition` source-marker tests that fail on `main` `a04ddc237`; fixed by #7321, not touched here.

reatang:
Rebased onto `main` `a2bad0953` to resolve the test-file conflict with #7307. `main` currently fails to compile `rustfs-scanner` (#7208 and #7322 merged against each other; fix in #7329), so the `Build`/`Test and Lint` lanes here will stay red until that lands; the branch's own checks (boundary and target-client unit tests, checksum e2e, outbound matrix) were re-run locally after the rebase.

PR #7340, opened 2026-09-07
fix(heal): retain MRF replay journal until retry anchor

houseme:
## Related Issues
Related: rustfs/backlog#2240, rustfs/backlog#2268, rustfs/backlog#2278.

## Summary of Changes
MRF replay now keeps the startup journal on disk until every replayed intent has a durable successor anchor. When replay decodes an existing journal, it temporarily raises the startup queue limits to cover that already-bounded snapshot, so a later flush cannot accidentally overwrite the old journal with only a prefix of the replayed work.

If the heal manager returns `Full`/`QueueFull` and a replayed intent is left pending for retry, the old journal is retained instead of being deleted immediately. The periodic tick still deletes the stale journal once the queue is empty and the replayed snapshot has been fully consumed.

The integration harness now also drives a real ECStore-backed journal through replay with a one-slot heal-manager queue. It verifies that `Full` replay keeps the authoritative journal on disk and that a simulated restart can replay the same two-record journal again instead of losing the refused tail.

## Verification
Tested head: e715fa17eeafb106fe53bb9dfcabae41d68a12ea.

- `cargo +nightly-aarch64-apple-darwin fmt --package rustfs-heal --check` passed.
- `cargo +nightly-aarch64-apple-darwin test --locked -q -p rustfs-heal --lib replay_cleanup_retains_journal_for_unarmed_or_refused_records -j4` passed.
- `cargo +nightly-aarch64-apple-darwin test --locked -q -p rustfs-heal --lib replay_can_arm_more_records_than_live_queue_budget -j4` passed.
- `cargo +nightly-aarch64-apple-darwin test --locked -q -p rustfs-heal --lib tick_action_table -j4` passed.
- `cargo +nightly-aarch64-apple-darwin test --locked -q -p rustfs-heal --lib heal::mrf_queue:: -j4` passed, covering 47 MRF queue tests.
- `cargo +nightly-aarch64-apple-darwin test --locked -q -p rustfs-heal --test mrf_pipeline_test journal_replay_retains_file_when_manager_is_full -j4` passed.
- `cargo +nightly-aarch64-apple-darwin test --locked -q -p rustfs-heal --test mrf_pipeline_test -j4` passed, covering 4 MRF pipeline integration tests.
- `git diff --check` passed.

Full workspace fmt was not used as the gating result because current `origin/main` has an unrelated rustfmt diff in `crates/crypto/src/encdec/aes.rs`; this PR does not touch that file.

## Impact
No API, configuration, or disk-format change. A startup journal may remain on disk for one more retry/flush cycle when replayed MRF work is still pending, favoring duplicate-safe replay over loss. Replay-only queue limit raising applies to a snapshot that has already been loaded from the bounded journal.

The new integration test uses real local-disk journal files and a fresh manager instance to exercise the restart-visible anchor behavior, but it is still an in-process restart simulation rather than an OS process-kill test.

## Additional Notes
Adversarial review focus: crash after replay but before successor snapshot, low live queue budget versus larger replay journal, manager `Full`/`QueueFull`, duplicate-safe replay through the queue coalescing path, stale journal deletion after the queue becomes empty, and scoped-versus-legacy epoch selection.

Null verdict after root-side correctness, durability/concurrency, compatibility, performance, and test-coverage lenses: no blocking finding remains. The remaining backlog risk is outside this focused PR: real-process crash injection, mixed-version binary replay, and long-running MRF throughput evidence are still required before closing the V2 backlog items.

### DIFF
--- a/crates/heal/src/heal/mrf_queue.rs
+++ b/crates/heal/src/heal/mrf_queue.rs
@@ -186,6 +186,11 @@ impl MrfQueue {
         MrfQueuePushResult::Enqueued
     }
 
+    fn raise_limits_for_replay(&mut self, intents: usize, bytes: usize) {
+        self.capacity = self.capacity.max(self.pending.len().saturating_add(intents));
+        self.byte_budget = self.byte_budget.max(self.bytes.saturating_add(bytes));
+    }
+
     /// Bool compatibility adapter: only a newly executable queue item is
     /// reported as accepted; a coalesced duplicate is not durable admission.
     #[cfg(test)]
@@ -655,9 +660,10 @@ pub fn spawn_mrf_consumer(manager: Arc<HealManager>) {
 
 /// Replay the durable journal into a fresh pending queue and submit whatever
 /// it armed. Returns the number of intact intents replayed. Duplicates are
-/// merged by the manager's dedup key; the journal file is removed once read
-/// (torn tails truncate via the per-record CRC). Public for integration tests;
-/// the live consumer invokes this through [`replay_into`] at startup.
+/// merged by the manager's dedup key; the journal is retained whenever replay
+/// cannot fully hand off a successor in-memory snapshot (torn tails truncate
+/// via the per-record CRC). Public for integration tests; the live consumer
+/// invokes this through [`replay_into`] at startup.
 pub async fn replay_journal_once(manager: &Arc<HealManager>) -> usize {
     let config = MrfConsumerConfig::default();
     let mut queue = MrfQueue::new(config.queue_capacity, config.journal_max_bytes);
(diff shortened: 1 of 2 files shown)


houseme:
W13 follow-up status update: we have now landed the accepted/merged-retain window fix on `release` as commit `a8bb53218`.

Delta vs prior `retain until retry anchor` behavior:
- replay judge now includes `Accepted`/`Merged` outcomes as retention-worthy (in addition to queue retry/backoff leftovers).
- `tick_action` now takes `retain_replay_journal` state from startup replay.
- tests updated in `crates/heal/src/heal/mrf_queue.rs`.

Validation already run on current release-base:
- `cargo test -p rustfs-heal --lib tick_action_table`
- `cargo test -p rustfs-heal --lib replay_cleanup_retains_journal_for_unarmed_or_refused_records`
- `cargo test -p rustfs-heal --lib mrf_queue::tests` (14 passed)
- `cargo fmt --all --check`
- `git diff --check`

This is to keep this PR family consistent: W13 #2268 no longer risks deleting startup journal immediately after Accepted/Merged replay admission.

PR #7342, opened 2026-09-07
fix(ecstore): hide and reclaim delete residue in prefix listings

### LINKED ISSUE #6898: Stale / orphaned objects after upgrade to 1.0.0-rc.4

overtrue:
## Related Issues
Fixes #6898.

## Summary of Changes
Since 1.0.0-rc.4, never-versioned buckets list through the bounded prefix fast path added in #6473 (`skip_hidden_prefix_check`): a directory without `xl.meta` is emitted as a common prefix as soon as it is non-empty, without the recursive probe rc.3 used to prove that something listable exists under it. That made a different kind of directory visible: the data dir a deleted version leaves behind (`<key>/<uuid>/part.N`, plus a `delete-data.<uuid>` marker when the delete transaction committed). The reporter sees `mc ls` and the console show a `288ed293-...-1e2c98ad8fb4/` folder under a deleted `.parquet` key, `HEAD` on it returns 404, listing inside it is empty, and `mc rm -r --force` succeeds without changing anything because the recursive listing that `mc` drives never returns directory entries. Nothing on that path could reclaim the residue either: the rc.5 purge from #6974 only fires for an exact recursive listing with `max-keys=1`, a request shape no client sends.

Two coordinated changes close this:

- The fast path in `LocalDisk::scan_dir` now runs a bounded delete-residue probe before emitting a metadata-less directory as a prefix. A directory is residue when it is itself a non-nil UUID directory holding only `part.N` files and delete-transaction markers, or when every child is one. The first non-UUID child, stray file, or subdirectory inside a UUID child proves a genuine prefix, so user keys under UUID-named prefixes and objects keyed by a UUID stay visible. Reads are bounded to an 8-entry batch with a complete read only when the batch is full, so the probe cost on real prefixes stays constant and #6473's amplification does not return.
- `should_purge_empty_directory_listing` now accepts any complete first page of a prefix listing, with or without a delimiter and at any `max-keys`, as the trigger for `purge_orphan_dir_object`. That is the request shape of a console browse, `mc ls`, `mc stat`, and the recursive listing behind `mc rm -r --force`, so a user who lists or removes a phantom folder now reclaims committed residue under it. The purge itself is unchanged: it re-scans every directory on every disk and fails closed on anything that is not an empty directory or a committed delete data dir, so residue without a committed marker and any in-flight write are never touched.
- `metadata_less_part_file` moves from `set_disk/ops/heal.rs` to `disk/local.rs` so the walk and heal classify data-dir entries with one predicate.

## Verification
Tested at the PR head (`fix/listing-delete-residue` on top of `origin/main` a70c96d52), local build.

- `cargo +1.98.0 test -p rustfs-ecstore --lib -- test_scan_dir_nonrecursive` — 6 passed. New: `test_scan_dir_nonrecursive_fast_path_hides_delete_residue` builds `residue/2026/object.parquet/<uuid>/part.1`, the same shape with a committed marker, `uploads/<uuid>/file` (real object under a UUID prefix), `named/<uuid>` (object keyed by a UUID), and residue beside a live child, then asserts the fast path lists nothing under the two residue prefixes while the other three keep their entries. Before this change the residue data dirs are emitted as prefixes, which is the reported symptom.
- `cargo +1.98.0 test -p rustfs-ecstore --lib -- purge` — 61 passed; new `empty_delimiter_listing_hides_and_purges_committed_delete_residue` lists `metrics/2026/` with `delimiter=/` and `max-keys=1000` over committed residue on every disk, asserts no prefix is returned, and asserts the residue tree is gone afterwards. `empty_recursive_listing_purges_committed_delete_residue` and the `purge_orphan_dir_object_*` fail-closed tests (uncommitted residue preserved, live data preserved) still pass. `empty_directory_listing_purge_requires_complete_exact_recursive_request` is renamed to `..._requires_complete_first_page_of_prefix` and now pins the widened trigger.
- `cargo +1.98.0 test -p rustfs-ecstore --lib -- metadata_less` and `-- orphan` — 10 and 22 passed.
- `cargo fmt --all --check`, `git diff --check` — clean.

Not verified: the reporter's on-disk residue may predate the delete-transaction markers; such residue is now hidden from delimiter listings but is not reclaimed by listing (the purge fails closed on it by design) and still needs `rc admin heal --remove`, which already reclaims metadata-less `part.N` data dirs.

## Impact
- S3-visible: delimiter listings on never-versioned buckets no longer show folders that contain only deleted-version data dirs. Recursive listings are unchanged (they never returned directory entries).
- Delete residue with committed markers is reclaimed by ordinary listing traffic on never-versioned buckets instead of only by an exact `max-keys=1` recursive request. Versioned buckets are unaffected; the purge stays gated on authoritative never-versioned metadata.
- Cost: one to a few bounded `list_dir` reads per metadata-less directory on the fast path, only when that directory is non-empty. Empty prefix listings on never-versioned buckets now issue one `list_dir` per disk of the prefix root (cheap `FileNotFound` when the prefix does not exist).
- No configuration, wire, or on-disk format change.

## Additional Notes
Adversarial validation (high risk: S3-visible listing semantics), two sequential passes:
- Correctness: a UUID child that contains `xl.meta` (an object keyed by a UUID) or any subdirectory is never classified as residue; a full 8-entry batch falls back to a complete read so a listable child cannot hide behind the bound.
- Concurrency/durability: the walk change only affects what is emitted; deletion happens solely in the unchanged `scan_orphan_dir`/`purge_orphan_dir_object`, which requires committed markers and uses non-recursive directory removal so a racing PUT fails with `DirectoryNotEmpty` rather than losing data.
- Compatibility: `WalkDirOptions` is unchanged; old peers keep `skip_hidden_prefix_check = false` and the conservative probe.
- Performance: the probe short-circuits on the first non-residue entry, so prefixes with many UUID-named children cost one bounded read of the parent plus one bounded read of the first child.
Residual: ancestors of a residue object dir remain ordinary non-empty directories and still list as prefixes until an empty listing beneath them reclaims the tree bottom-up, or a recursive listing of the ancestor reclaims it in one pass.

### DIFF
--- a/crates/ecstore/src/disk/local.rs
+++ b/crates/ecstore/src/disk/local.rs
@@ -98,6 +98,35 @@ const INLINE_METADATA_ROLLBACK_DIR_XOR: u128 = 0x7275737466735f696e6c696e655f726
 const DELETE_MARKER_ROLLBACK_FILE: &str = "xl.meta.delete-marker.rollback";
 pub(crate) const DELETE_DATA_DIR_MARKER_PREFIX: &str = "delete-data.";
 pub(crate) const RESERVED_DELETE_DATA_DIR_MARKER_PREFIX: &str = "reserve-delete-data.";
+/// Largest directory read the delete-residue probe issues before it must
+/// fall back to a complete read. Residue holds one or two data dirs, so an
+/// under-filled batch settles the common case without materializing large
+/// child sets; a full batch cannot prove no listable child hides behind it.
+const DELETE_RESIDUE_PROBE_LIMIT: i32 = 8;
+
+/// A `part.N` file with a positive part number, the shape erasure data takes
+/// inside a version data dir.
+pub(crate) fn metadata_less_part_file(entry: &str) -> bool {
+    entry
+        .strip_prefix("part.")
+        .is_some_and(|part_number| part_number.parse::<usize>().is_ok_and(|part_number| part_number > 0))
+}
+
+fn is_delete_transaction_marker(entry: &str, prefix: &str) -> bool {
+    entry
+        .strip_prefix(prefix)
+        .is_some_and(|transaction| Uuid::parse_str(transaction).is_ok_and(|uuid| !uuid.is_nil()))
+}
+
+/// Whether a `list_dir` entry inside a UUID data dir is erasure data or a
+/// delete-transaction marker. Anything else (a subdirectory, an `xl.meta`, an
+/// unknown file) means the directory is not plain delete residue.
+fn is_metadata_less_data_dir_entry(entry: &str) -> bool {
+    !entry.ends_with(SLASH_SEPARATOR)
+        && (metadata_less_part_file(entry)
+            || is_delete_transaction_marker(entry, DELETE_DATA_DIR_MARKER_PREFIX)
+            || is_delete_transaction_marker(entry, RESERVED_DELETE_DATA_DIR_MARKER_PREFIX))
+}
 const STARTUP_CLEANUP_WAIT_TIMEOUT: Duration = Duration::from_secs(2);
 const ENV_BITROT_SIZE_MISMATCH_RETRY_COUNT: &str = "RUSTFS_BITROT_SIZE_MISMATCH_RETRY_COUNT";
 const ENV_BITROT_SIZE_MISMATCH_RETRY_DELAY_MS: &str = "RUSTFS_BITROT_SIZE_MISMATCH_RETRY_DELAY_MS";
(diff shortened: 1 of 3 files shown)


agelwarg:
@overtrue This is referenced as fixing #6898, but I am still experiencing the issue on rc6.  What else should I provide to aid in troubleshooting?

PR #7366, opened 2026-09-07
fix(replication): carry the compression layout through SSE-C passthrough

overtrue:
## Related Issues

Fixes rustfs/backlog#2363. Refs rustfs/backlog#2147 and rustfs/rustfs#7334. Also unblocks the `Test and Lint` lanes on `main`, failing since c9c6bb7a2 with `error: unused import: Digest` in `crates/scanner/src/scanner_folder/item_actions.rs`.

## Summary of Changes

This PR carries two commits. The first is the one-line scanner lint fix: `Sha256::digest` no longer needs the `Digest` trait in scope, so the unused `Digest as _` import fails the crate under `-D warnings`; the import is dropped and the two call sites are unchanged. The second is the SSE-C passthrough fix below, originally opened as rustfs/rustfs#7372 stacked on this branch and merged into it by the repository automation.

SSE-C ciphertext passthrough replicates an object's stored bytes verbatim, but the sender stripped every `x-rustfs-internal-*` key, so a target never learned that those bytes were compressed. Against a real RustFS target a compressed single-PUT SSE-C object replicated `COMPLETED` and then failed to read back (the replica waited for the plaintext length while holding only the compressed ciphertext), and a compressed SSE-C multipart object was rejected at CompleteMultipartUpload with `EntityTooSmall` because the target validated the 5 MiB minimum against the stored bytes of the first part. rc.5 fails the same way as a sender.

- Sender: for an SSE-C object stored compressed, `replication_put_object_options` adds `x-rustfs-replication-compression` (the stored scheme) and `x-rustfs-replication-compression-actual-size` (the plaintext size) to the PutObject and CreateMultipartUpload requests, and the multipart transport adds `x-rustfs-replication-part-actual-size` to every UploadPart of a passthrough object. Internal keys still never leave the source as plain metadata; the compression headers travel only with SSE-C passthrough.
- Target: an authorized passthrough request restores the scheme and plaintext size into the stored metadata; the single-PUT path records the plaintext size as the part's logical length while still reading the body at its wire size; UploadPart on a passthrough session takes the part's logical length from the header and no longer adds a second compression layer on top of ciphertext; CompleteMultipartUpload therefore validates the 5 MiB minimum against the plaintext length.
- The fake S3 target honours the per-part header the same way RustFS does.
- Headers are additive. An older target ignores them and behaves as before; a non-passthrough replication request never carries them.

## Verification

Tested commit: the PR head (`main` c9c6bb7a2 plus both commits; the passthrough commit was verified at 42906c11e before the squash into this branch, with an identical tree).

- Failing-before: `replication_extension_test::test_bucket_replication_sse_c_compressed_passthrough` replicates a compressed SSE-C single PUT and a two-part multipart object to a second RustFS instance and reads both replicas back with the customer key. Against the unfixed server binary it fails with `replica body read failed: streaming error` for the single PUT and `source reports FAILED` (EntityTooSmall) for the multipart object; with this change both replicas are byte-identical.
- Unit: `put_opts_from_headers_restores_the_compression_layout_only_for_ssec_passthrough` (target restore is gated on SSE-C passthrough and authorization, malformed sizes are dropped), `compressed_ssec_objects_declare_their_compression_layout_on_the_wire` (sender emits the headers only for SSE-C, never an internal key), and `multipart_transport_declares_passthrough_part_lengths` (a hyper fixture target observes the per-part length on every UploadPart and the scheme on CreateMultipartUpload; the decrypted transport carries neither).
- `cargo nextest run -p e2e_test -E 'test(/^replication_extension_test::(test_bucket_replication_sse_c_compressed_passthrough|test_bucket_replication_sse_c_multipart_passthrough|test_bucket_replication_sse_c_contract)$/)'`: 3 passed, so uncompressed SSE-C passthrough is unchanged.
- With the pinned rc.5 binary, `upgrade_compatibility_test::direct_upgrade_from_rc5_preserves_multipart_layouts` (rustfs/rustfs#7334) replicates the rc.5-written compressed SSE-C multipart object `COMPLETED` with parts 1 and 2; `rc5_baseline_replicates_multipart_layouts` still records `FAILED` for the same layout when rc.5 is the sender, which is the expected pre-fix behaviour.
- `cargo clippy --locked -p rustfs-utils -p rustfs-scanner -p rustfs-ecstore -p rustfs -p e2e_test --all-targets -- -D warnings` clean with the import fix (without it the same command fails exactly as the `main` run 34091718733 does); `cargo fmt --all --check` clean; the six architecture guard scripts pass.

Not run: the full unit suite, the nightly replication lane, and a mixed-version pair where only one side has this change (an old target still rejects or misreads compressed passthrough replicas, as today). Remaining risk: the passthrough part length is trusted from an authorized replication sender; it only affects the logical length recorded for the part and the minimum-size check, never the stored bytes.

## Impact

Compressed SSE-C objects now replicate to RustFS targets on the current build and the replicas decrypt and decompress to the source bytes. Three additive replication transport headers are introduced; no configuration or on-disk format change.

## Additional Notes

The backlog#2362 fix (rustfs/rustfs#7376, in-flight dedupe) is stacked on this branch because its regression test needs both fixes.

### DIFF
--- a/crates/e2e_test/src/fake_s3_target/mod.rs
+++ b/crates/e2e_test/src/fake_s3_target/mod.rs
@@ -584,6 +584,10 @@ struct MultipartPart {
     body: Bytes,
     e_tag: String,
     digest: [u8; 16],
+    /// Plaintext length declared by an SSE-C passthrough sender
+    /// (`x-rustfs-replication-part-actual-size`); RustFS validates the 5 MiB
+    /// minimum against it rather than against the stored bytes.
+    actual_size: Option<usize>,
 }
 
 #[derive(Clone)]
@@ -2624,6 +2628,11 @@ impl S3 for FakeBackend {
 
     async fn upload_part(&self, req: S3Request<UploadPartInput>) -> S3Result<S3Response<UploadPartOutput>> {
         let fault = request_fault(&req);
+        let declared_actual_size = req
+            .headers
+            .get("x-rustfs-replication-part-actual-size")
+            .and_then(|value| value.to_str().ok())
+            .and_then(|value| value.parse::<usize>().ok());
         let _body_permit = timeout(MAX_FAULT_DURATION, Arc::clone(&self.body_limit).acquire_owned())
             .await
             .map_err(|_| s3s::s3_error!(RequestTimeout, "fake target body limiter wait exceeded 30 seconds"))?
@@ -2665,6 +2674,7 @@ impl S3 for FakeBackend {
                 body,
                 e_tag: e_tag.clone(),
                 digest,
+                actual_size: declared_actual_size,
             },
         );
         Ok(apply_response_fault(
@@ -2734,7 +2744,9 @@ impl S3 for FakeBackend {
                 if requested_etag != &stored.e_tag {
                     return Err(s3s::s3_error!(InvalidPart, "part ETag does not match"));
                 }
-                if index + 1 != requested_parts.len() && stored.body.len() < MIN_MULTIPART_PART_BYTES {
+                if index + 1 != requested_parts.len()
+                    && stored.actual_size.unwrap_or(stored.body.len()) < MIN_MULTIPART_PART_BYTES
+                {
                     return Err(s3s::s3_error!(EntityTooSmall, "non-final multipart part is smaller than 5 MiB"));
                 }
                 selected.push((*number, stored.clone()));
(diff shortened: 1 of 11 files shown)


overtrue:
Note: the first commit here is the same one-line scanner import removal as #7374 (identical change, no conflict either way); the rest of this PR is the SSE-C passthrough fix that the automation merged into this branch from #7372.

Issue #6824, opened 2026-08-28
Bucket replication: ordinary create/delete can duplicate versions and delete markers on a generic S3 destination


jesseryangraham:
## Describe the bug

On rustfs/rustfs 1.0.0-rc.3 (`b15928220`), ordinary create/delete replication to a generic S3 destination (Wasabi) can duplicate destination versions and delete markers.

A combined create-then-delete test reproduced this in 3/3 fresh runs. Isolated create behavior was timing-dependent. Exactly one rule and one target were configured. The duplication mechanism remains unestablished pending request-level tracing.

This is not [#1742](https://github.com/rustfs/rustfs/issues/1742). That report was extra versions on the healthcheck interval, and it was closed against [#1747](https://github.com/rustfs/rustfs/pull/1747). The rc.3 behavior we measured is ordinary create/delete, not healthcheck.

## Environment

Live tests used a locally compiled binary, not a container image.

- Commit: [`b15928220f53818164c2cd4e2c236d5c1f5aada6`](https://github.com/rustfs/rustfs/tree/b15928220f53818164c2cd4e2c236d5c1f5aada6) (`1.0.0-rc.3`)
- Binary reports: `rustfs @b15928220`
- Build: debug, linux-x86_64
- Rust compiler: 1.98.0
- Binary SHA-256: `224b8141b9151db698831adf35b3e47581f8325b932565b2b10499fee437efe2`
- Destination: Wasabi generic S3
- Wasabi region: `us-west-1`
- Wasabi endpoint: `https://s3.us-west-1.wasabisys.com`

## To reproduce

1. Enable versioning on source and destination (Wasabi `us-west-1`, `https://s3.us-west-1.wasabisys.com`).
2. Configure bucket replication RustFS -> Wasabi. Exactly one rule and one target.
3. PUT a new object. Wait for replication.
4. Delete the object (current-version delete, delete-marker path).
5. List object versions on the destination.

## Expected

One destination version for the create. One delete marker for the delete.

## Actual

The combined create/delete path duplicated destination versions and delete markers in 3/3 fresh runs. Isolated create was timing-dependent. Mechanism not established.

reatang:
Thank you for the rigorous, well-controlled testing. We have confirmed the duplication behavior internally and are working on fixes as part of the same replication work stream. We will update this issue as it progresses.

jesseryangraham:
Retested on 1.0.0-rc.6 against a live generic S3 destination (Wasabi).

[...]

Issue #6823, opened 2026-08-28
Bucket replication: a failed version-purge task continues retrying after the source version is gone while destination delete-marker count grows


jesseryangraham:
## Describe the bug

On rustfs/rustfs 1.0.0-rc.3 (`b15928220`), after a non-current-version purge against a generic S3 destination (Wasabi) fails with `Version not found`, the failed purge task continues retrying after the source version is gone while the destination delete-marker count grows.

The two test keys reached 20 and 24 markers. We did not trace every retry request closely enough to claim that each retry created exactly one marker.

Companion outcome ([#6822](https://github.com/rustfs/rustfs/issues/6822)): RustFS passes a source-side version identifier to a destination that assigned its own version identifiers, so the requested destination version is not found and the current Wasabi key becomes hidden by a delete marker. Request-level tracing is still needed to establish the exact marker-creation path.

## Environment

Live tests used a locally compiled binary, not a container image.

- Commit: [`b15928220f53818164c2cd4e2c236d5c1f5aada6`](https://github.com/rustfs/rustfs/tree/b15928220f53818164c2cd4e2c236d5c1f5aada6) (`1.0.0-rc.3`)
- Binary reports: `rustfs @b15928220`
- Build: debug, linux-x86_64
- Rust compiler: 1.98.0
- Binary SHA-256: `224b8141b9151db698831adf35b3e47581f8325b932565b2b10499fee437efe2`
- Destination: Wasabi generic S3
- Wasabi region: `us-west-1`
- Wasabi endpoint: `https://s3.us-west-1.wasabisys.com`

## To reproduce

1. Enable versioning on source and destination (Wasabi `us-west-1`, `https://s3.us-west-1.wasabisys.com`).
2. Configure bucket replication RustFS -> Wasabi, including version and delete-marker replication. Exactly one rule and one target.
3. PUT an object twice, then delete a non-current source version.
4. Wait for replication retries.
5. List object versions / delete markers on the destination key.

## Expected

A purge removes the corresponding destination version and completes. Any retry is idempotent and does not add destination delete markers.

## Actual

The failed purge task continued retrying throughout the observation period after the source version was gone. Destination marker count grew (20 and 24 on the two keys we watched). We do not claim a 1:1 retry-to-marker mapping.

## Source (pinned)

Same purge path as [#6822](https://github.com/rustfs/rustfs/issues/6822). The source UUID is sent as the destination `versionId`; the RustFS server log reports `Version not found` for the destination operation:

[...]

Issue #6822, opened 2026-08-28
Bucket replication: deleting a non-current source version can hide the current object on a generic S3 destination


jesseryangraham:
## Describe the bug

On rustfs/rustfs 1.0.0-rc.3 (`b15928220`), bucket replication to a generic S3 destination (Wasabi) mishandles a non-current-version purge.

In 2/2 tests, deleting an old (non-current) source version made the current Wasabi key return 404. Logs reported `Version not found`.

RustFS passes a source-side version identifier to a generic S3 destination that assigned its own version identifiers. The requested destination version is therefore not found. The current Wasabi key becomes hidden by a delete marker. Request-level tracing is still needed to establish the exact marker-creation path.

This is not [#6218](https://github.com/rustfs/rustfs/issues/6218) (local LIST vs GET). The damage is on the destination bucket.

## Environment

Live tests used a locally compiled binary, not a container image.

- Commit: [`b15928220f53818164c2cd4e2c236d5c1f5aada6`](https://github.com/rustfs/rustfs/tree/b15928220f53818164c2cd4e2c236d5c1f5aada6) (`1.0.0-rc.3`)
- Binary reports: `rustfs @b15928220`
- Build: debug, linux-x86_64
- Rust compiler: 1.98.0
- Binary SHA-256: `224b8141b9151db698831adf35b3e47581f8325b932565b2b10499fee437efe2`
- Destination: Wasabi generic S3
- Wasabi region: `us-west-1`
- Wasabi endpoint: `https://s3.us-west-1.wasabisys.com`

## To reproduce

1. Enable versioning on a RustFS source bucket and a Wasabi destination bucket (`us-west-1`, `https://s3.us-west-1.wasabisys.com`).
2. Configure bucket replication (RustFS -> Wasabi), including version and delete-marker replication. Exactly one rule and one target.
3. PUT an object, then PUT it again so the destination has a current version.
4. Delete a non-current source version (not the current one).
5. GET the current destination key. Check destination versions and server logs for `Version not found`.

## Expected

The named old version is removed on the destination. The current destination object stays readable.

## Actual

The current destination key returns 404. Logs report `Version not found`. The current Wasabi key is hidden by a delete marker.

## Source (pinned)

For a version purge, `resolve_delete_api_version_id` passes the source `version_id` through (it omits `versionId` only when propagating a delete-marker creation). That source-side UUID does not identify the version Wasabi assigned independently:

[...]

PR #7368, opened 2026-09-07
fix(replication): resolve drifted replicas via a target version ledger

reatang:
## Related Issues

Step 2 ("target-side version ledger, P2") of the fix plan in rustfs/backlog#2340, including new finding 1 of its 2026-09-07 comment (the unresolvable purge retried every scanner cycle without backoff). Product boundary: rustfs/backlog#2085. Same root as rustfs/rustfs#6822, rustfs/rustfs#6823, rustfs/rustfs#6824. Builds on rustfs/rustfs#7323. The two follow-ups that were stacked on this PR — rustfs/rustfs#7373 (`?replication-check` ledger-addressed phases) and rustfs/rustfs#7377 (pending-purge lifecycle, new finding 2 of the 2026-09-07 comment) — were merged into this branch by the maintainers, so this PR now carries all three changes; their descriptions are folded in below.

### 1. Target-version ledger (original scope of this PR)

A replication target that mints its own version ids (Wasabi, AWS S3, Impossible Cloud) never answers to the source uuid. #7323 stopped the duplicate re-drives; everything after the initial PUT still failed: a permanent version delete answered `NoSuchVersion` on every heal pass, and a tag, retention or legal-hold update re-PUT the object and minted one more target version per update (confirmed on Wasabi 2026-09-07: `test_permanent_version_delete_preserves_other_generation`, `test_tag_update_does_not_create_version`, `test_retention_and_hold_updates` all red).

Decision on the ledger shape, as the plan asked:

- (a) persist the id the target assigned in the source's xl.meta and address it directly;
- (b) do not persist; resolve at mutation time by exact key + ETag (what `replica_head_fallback` already did for heal).

(b) cannot tell two generations of the same key with the same bytes apart: two source versions with identical content are two target versions with the same ETag, and a permanent delete of the older one would remove whichever the listing returns first — on a newest-first listing, the live generation. A wrong pick destroys data on the target, so (b) is unsafe as the primary handle. The PR takes (a) as the ledger and keeps (b) as a bounded fallback for replicas written before the ledger existed, with the ambiguous case refused rather than guessed. While writing the e2e test this also surfaced that the #7323 heal fallback could declare a same-bytes newer generation converged on the older generation's replica; content-identity candidates are now disowned when another version of the key already claims them.

What changes:

- The ledger is one internal metadata key per drifting target, `x-rustfs-internal-replication-target-version-<arn>` (dual `x-minio-internal-` twin; same shape as the delete-marker ledger from backlog#2290), written through the existing replication-status writeback from the PUT / CompleteMultipartUpload response, and when heal locates a pre-ledger replica. Adopting targets record nothing, so xl.meta only grows on drifting targets, and only when the recorded id changes.
- `replica_head_fallback` tries the ledger id before content identity; a ledger hit also teaches the runtime identity verdict (it survives restarts, the cache does not). A fresh write (`ReplicationType::Object`) no longer consults content identity: it has no replica to find, and an older same-bytes generation would be mistaken for one.
- `TargetClient::replica_candidates_by_etag` returns every same-content candidate (bounded); callers drop the ones other versions of the same source key claim through their ledgers (`SiblingLedger`, one listing of the exact key, paid only when a candidate exists) and classify the rest as `ReplicaLocation::{Missing, Unique, Ambiguous}`.
- Tag, retention and legal-hold updates on a located drifted replica go through PutObjectTagging / PutObjectRetention / PutObjectLegalHold against the assigned id (`Metadata` attempts always apply tags; heal/resync only touch what the HEAD proves diverged). Retention is only ever extended: shortening or clearing needs a governance bypass the replication client does not hold, logged at debug and left as is.
- A permanent version delete reads the purged version's ETag and ledger once, resolves the target id (ledger → unique disowned content match), treats a missing replica or a `NoSuchVersion`/`NoSuchKey` answer as purged, and refuses an ambiguous match with the same one-hour backoff the object-lock denial uses (`object_lock_denied_purge_*` generalized to `purge_backoff_*`). A `NoSuchVersion` for the source id on a target whose contract is still unknown re-resolves by content identity once before it counts as purged. New event `replication_purge_replica_unresolved` (warn) and counter `rustfs_replication_version_purge_replica_total{resolution}`.
- Fake target: Wasabi shape `reject_unknown_version_deletes` (404 `NoSuchVersion` on an unknown id) and per-version retention / legal hold with the four Object Lock APIs; the `MintOwnVersionIds` matrix row uses both.

### 2. `?replication-check` probes mutations by the assigned id (was #7373)

`GET /BUCKET?replication-check` skipped its `DeleteMarker` and `VersionDelete` phases on a target that mints its own version ids: they addressed the source id, which such a target never had, so on Wasabi the report ended at `VersionFidelity: FAILED` with both phases `SKIPPED`. Since #7368 the replication worker addresses the id the target assigned (the ledger), and the probe already has that id — it is the version the target answered the probe PUT with. Both phases now gate on a successful `Put` that reported a version id instead of on `VersionFidelity == OK`, and run against `probe_version_id`: on a drifting target they report whether the ledger-addressed purge path the worker uses works against this endpoint. `VersionFidelity` still fails with `BucketRemoteTargetVersionMismatch` and the target stays `FAILED`, so callers that branch on the code see no change; they get two more phases of evidence.

- `execute_replication_probe`: mutation phases gated on `put == OK && probe_version_id.is_some()`.
- Unit tests: the minting-target probe asserts both phases ran with the assigned id (the scripted probe records the ids it was handed) and that cleanup addresses the marker it created; a drifting target that refuses the ledger-addressed delete reports `VersionDelete: FAILED` on its own.
- e2e `test_replication_check_flags_version_minting_target`: expects `DeleteMarker` and `VersionDelete` `OK` on the fake minting target.
- `docs/operations/replication-check.md`: VersionFidelity section rewritten around the ledger; the two phases are `SKIPPED` only when the probe PUT failed or reported no version id.

### 3. Pending purges outlive the replication configuration (was #7377)

A permanent version delete whose replication keeps failing leaves the version in xl.meta as a PENDING purge, hidden from listings, until every target confirms it. Once the operator removes the bucket's replication configuration, or the rule naming that target, nothing ever confirms it: the scanner skips buckets without active rules, the purge is never revisited, and `DeleteBucket` answers `BucketNotEmpty` forever (`bucket_delete_blocked, blocker=visible_version`) for a residue the client can neither list nor delete. Observed on Wasabi: the source bucket of `test_permanent_version_delete_preserves_other_generation` stayed undeletable after its teardown removed the configuration.

Two options were weighed:

- (A) let the delete worker settle the purge locally when the configuration no longer names the target — the same path that would have confirmed it, driven by the scanner heal pass that already revisits pending purges; one metadata write per version, nothing new persisted (the stored decision names the targets), so it survives restarts; observable through an event and a counter; also covers a rule edit that drops one destination.
- (B) sweep the bucket synchronously at `DeleteBucketReplication` / target removal — immediate, but a one-shot walk of the whole bucket that a crash mid-way silently loses, and it does not see a rule edit.

(A) is implemented; if the Owner wants (B) as well, the sweep can call the same `abandoned_purge_target_info` path. The purge is *abandoned*, not executed: the replica stays on the former target (that is the operator's data now); `replication_purge_abandoned` (warn) and `rustfs_replication_version_purge_abandoned_total` are the record.

- `ReplicationConfigurationExt::configured_target_arns`: every destination ARN the configuration still names (role, or every rule whatever its status) — the set a pending delete may still be owed to.
- `replicate_delete_with_outcome`: for version purges, a target outside that set is settled `Complete` locally through the normal writeback (`abandoned_purge_target_info`) instead of failing on a missing client. An unreadable or invalid configuration abandons nothing: the decision needs positive evidence.
- Heal admission (`queue_replication_heal`, `queue_replication_heal_internal`) and the scanner's `heal_replication` let a version with a pending or failed purge through even when the bucket has no configuration or active rules; the delete worker then decides against the current configuration. Objects without a purge keep the old skip.
- `get_heal_replicate_object_info`: a version purge is owed to the targets its persisted purge state names. The replicate-decision string is not persisted, so a heal after a restart, or after the configuration was removed or edited, derived an empty decision and skipped the purge forever — which is why the scanner never revisited the stuck version in the first place (found while writing the e2e test: the abandonment only ever ran from the in-memory MRF retry, never from the scanner).
- e2e `matrix_removed_replication_config_abandons_pending_purge`: a refused purge hides the version and blocks `DeleteBucket`; after `DeleteBucketReplication` the bucket becomes deletable within the scanner cycle, the replica on the former target is untouched, and purge attempts stop.

## Verification

Tested commit: the PR head after merging `origin/main` (the merge also carries the same one-line `Digest` import fix as rustfs/rustfs#7366, without which the merge commit does not pass `-D warnings`); `target/debug/rustfs` rebuilt from it before every e2e run.

Ledger (part 1):
- `cargo test -p rustfs-ecstore --lib replication_resyncer::tests`: 81 passed. New cases: purge addresses the ledger id (source uuid never sent); no ledger → unique content match; ambiguous match → FAILED, no DELETE sent, `purge_backoff_active`; missing replica → Complete; `NoSuchVersion` → Complete; source-id `NoSuchVersion` on an unknown contract re-resolves by ETag then deletes the located id; a sibling-claimed candidate is disowned for both purge (Complete without DELETE) and heal (replicates instead of converging on the sibling); corrupt ledger refused; a `Metadata` attempt on a drifted replica sends PutObjectTagging and no PutObject; heal records the ledger.
- `cargo nextest run -p e2e_test -E 'test(/^replication_target_matrix_test::/) | test(test_replication_check_flags_version_minting_target) | test(test_delayed_delete_marker_purge_uses_target_assigned_version) | test(/read_proxy/)'`: 9 passed, including the new `matrix_mint_own_version_ids_addresses_mutations_through_the_ledger` (tag update, retention extension, legal hold ON then OFF, permanent delete of the older of two same-bytes generations — each lands on the existing target version, one PutObject per key). Before the fix the same test failed at the legal-hold and the same-bytes-generation steps.
- `cargo test -p rustfs-utils --all-features --lib target_versions`, `cargo test -p rustfs-replication --lib` (194), `cargo test -p rustfs-filemeta --lib replication`, `cargo test -p e2e_test --lib fake_s3_target` (10): all passed.
- `make pre-commit` passed; `cargo clippy --all-targets -- -D warnings` passed.
- Real Wasabi bucket lane of the compatibility suite: run pending on the stacked branch (results will be added in a comment together with the two stacked PRs).

Not run: `make pre-pr`; the full `e2e-full` lane. Remaining risk: AWS S3 and Impossible Cloud share the Wasabi contract shape but were not exercised.

replication-check (part 2):
- `cargo test -p rustfs --lib replication_probe`: 12 passed (before the change, `replication_probe_flags_version_minting_target` pinned the phases as `SKIPPED`; the renamed test now pins them `OK` with the assigned id).
- `cargo nextest run -p e2e_test -E 'test(test_replication_check_flags_version_minting_target)'`: passed against the fake minting target (which also answers 404 `NoSuchVersion` to unknown ids since #7368).
- `make pre-commit` passed; `cargo clippy --all-targets -- -D warnings` passed.
- Real Wasabi: `test_replication_check_diagnoses_target` in the compatibility suite bucket lane, to be run on the full stack (results in a follow-up comment).

Not run: `make pre-pr`, `e2e-full`.

Pending-purge lifecycle (part 3):
- `cargo test -p rustfs-ecstore --lib replication_resyncer::tests`: 83 passed (new: abandoning marks the target's purge Complete and leaves a finished target untouched; a failed purge without a configuration still classifies as a delete owed to the ARN its purge state names).
- `cargo test -p rustfs-replication --lib config`: 43 passed.
- `cargo nextest run -p e2e_test -E 'test(matrix_removed_replication_config_abandons_pending_purge) | test(matrix_mint_own_version_ids_addresses_mutations_through_the_ledger) | test(test_delayed_delete_marker_purge_uses_target_assigned_version)'`: passed (the first is new and fails before this change at "DeleteBucket to succeed once the orphaned purge is abandoned"; the other two guard the heal-admission change).
- `make pre-commit` passed; `cargo clippy --all-targets -- -D warnings` passed.
- Real Wasabi: the teardown of `test_permanent_version_delete_preserves_other_generation` (DeleteBucketReplication then DeleteBucket) on the full stack — results in a follow-up comment.

Not run: `make pre-pr`, `e2e-full`.

After the stack was merged into this branch and `origin/main` merged in, the full set was rerun on the combined tree: `cargo test -p rustfs-ecstore --lib replication_resyncer::tests`, `cargo test -p rustfs --lib replication_probe`, the matrix / replication-check / delete-marker / read-proxy e2e selection, `make pre-commit`, `cargo clippy --all-targets -- -D warnings` (results in the follow-up comment together with the real-Wasabi run).

## Impact

Ledger (part 1):
Outbound client defaults unchanged for every target class: no new header on the replication PUT, no new knob. The added calls (PutObjectTagging / PutObjectRetention / PutObjectLegalHold / DELETE by assigned id / ListObjectVersions) run only against a target that has proven it mints its own version ids (PUT response, ledger entry, or replication-check verdict). Verified target classes: the four fake-target matrix rows; unverified: AWS S3, Impossible Cloud.

Behavior change: a version purge whose target answers `NoSuchVersion`/`NoSuchKey` for the addressed id now completes instead of failing forever (RustFS/MinIO answer 204 for the same "already gone" case). A permanent delete of a replica that cannot be identified is refused and retried hourly instead of every scanner cycle.

On-disk: new dual-prefixed internal metadata keys on versions replicated to drifting targets; older readers ignore them. `docs/operations/replication-outbound-transport.md` updated.

Adversarial lenses (replication, S3-visible semantics): correctness — the ambiguous case is refused with backoff, never guessed, and sibling claims are removed before any candidate is trusted; compatibility — ledger keys are ordinary internal metadata; concurrency — the ledger rides the existing CAS'd status writeback, and the purge source (ETag, ledger, sibling claims) is read once under the replicate lock before the per-target fan-out.

replication-check (part 2):
Same mutations on the same reserved probe key as before, now also on drifting targets: one replicated delete marker and one version delete under `.rustfs.sys/replication-check/`, both cleaned up as before. No outbound default, header or knob change. Response shape unchanged; only the phase statuses on drifting targets differ.

Pending-purge lifecycle (part 3):
Behavior change: a bucket whose replication configuration was removed, or whose rule no longer names a target, with version purges still pending now finishes those purges locally and becomes deletable; before, it was stuck until the configuration was restored. Buckets whose configuration still names the target are unaffected (same failure and retry as today). The replica on the former target is never touched. New warn event and counter; no S3-visible change beyond `DeleteBucket` succeeding.

## Additional Notes

Originally the first of a three-PR stack (#7368 → #7373 → #7377); the maintainers merged the two follow-ups into this branch, so this PR is now the whole series. The real-Wasabi compatibility run against this head is posted as a comment.


reatang:
Real-Wasabi run of the compatibility suite's bucket lane (RustFS 19000 source with a 1 s scanner cycle → a fresh one-off Wasabi bucket) against this branch after the stack was merged in.

| Lane | This head | Round 2 (main `4c4dcb6f5`, before this PR) |
| --- | --- | --- |
| Bucket lane, full selection | **30 passed / 4 failed / 1 skipped** (604 s) | 25 passed / 8 failed / 1 teardown error |

Turned green: `test_tag_update_does_not_create_version`, `test_retention_and_hold_updates`, `test_permanent_version_delete_preserves_other_generation` (including its teardown: DeleteBucketReplication followed by DeleteBucket, the new-finding-2 shape), and `test_delete_marker_and_undelete` (intermittent in round 2). The source log carries no `replication_purge_replica_unresolved` / `replication_purge_abandoned`: every mutation resolved through the ledger without the content-identity fallback.

Still red, all by design or Wasabi-side: `test_source_version_addressing_contract[False|True]` (the source-id contract itself), `test_managed_sse_preserves_encryption_intent[aws:kms]` (Wasabi answers 501, RustFS fails closed), and `test_replication_check_diagnoses_target` (`Status: FAILED` because `VersionFidelity` fails with `BucketRemoteTargetVersionMismatch`).

The replication-check run did surface two probe-shape bugs that only appear once the mutation phases run on a drifting target, both fixed in the follow-up commit on this branch: the DeleteMarker phase sent the assigned version id on its DELETE, which a generic S3 target executes as a permanent delete (RustFS peers read the `source-deletemarker` header instead), so VersionDelete then answered `NoSuchVersion`; and cleanup treated `NoSuchVersion` on an already-removed probe version as a failure. After the fix the phases on Wasabi are Put / DeleteMarker / VersionDelete / Cleanup `OK` with `VersionFidelity` `FAILED`, and the target holds no probe artifacts.

Results are archived in the (gitignored) suite directory `tmp/site-replication-wasabi/results-2026-09-07-p2/`.

houseme (approved):

Issue #7363, opened 2026-09-07
Lock RPC timeouts against one erasure set's lock endpoints spiral into a reset/GOAWAY/reconnect loop across the cluster, with no backoff


stevapple:
**Describe the bug**

For about an hour, and again for a shorter window six hours later, every node in a 4-node cluster timed out its lock RPCs against the lock endpoints of one erasure set — overwhelmingly `node3:/mnt/rustfs11` — at the 3 s client budget, and the client-side reaction turned a slow lock server into a self-sustaining storm: each timeout cancels the gRPC stream, the accumulated `RST_STREAM`s make the server answer `GOAWAY too_many_resets`, the client evicts the cached connection and reconnects, and the next lock request re-arms the cycle. There is no backoff anywhere in the loop. Lock *releases* time out too, leaving `lock not found for release` and `distributed unlock left unreleased entries after retry` behind. The storm subsided on its own both times (07:48 the second time) with no intervention.

The storm's most expensive consequence — one node permanently latching all writes off — is reported in #7361; this report is about the storm itself, because it recurred, it hit every node, and the loop is what turns "one lock endpoint is slow" into cluster-wide lock failure.

**To Reproduce**

We cannot reproduce the server-side slowness on demand. What we can give is the shape, measured from the plain-text hourly logs on all four nodes (times local, +08:00):

- 00:07:57 first `Remote lock RPC timed out` (`timeout_ms=3000`, [`rpc_timeout`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/cluster/rpc/remote_locker.rs#L181), default [`DEFAULT_OBJECT_LOCK_RPC_TIMEOUT_MS = 3000`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/config/src/constants/object.rs#L526)); 00:07:58 first `GoAway(b"too_many_resets", ENHANCE_YOUR_CALM)` received; 00:07:59 the first slow inbound `NodeService/Lock` on node3 (5,756 ms).
- In that first hour, lock timeouts by client node: node1 9,736 · node2 4,979 · node4 6,524 (node3's own client-side count not collected). By target, from node4's log: `node3:/mnt/rustfs11` 6,393 · `node1:/mnt/rustfs11` 106 · `node2:/mnt/rustfs11` 25 — i.e. the lock endpoints of one set, one of them very sick.
[...]

PR #7387, opened 2026-09-07
fix(ecstore): recover interrupted pool metadata writes

### LINKED ISSUE #7361: A single timed-out read of pool.bin permanently blocks every write on the node until restart, while /health/ready keeps returning 200
**Describe the bug**

One node of a healthy 4-node cluster (EC 12+4, 48/48 disks `ok`/`online` on every node) started failing **every** write it was given with `500 InternalError` and kept doing so for 7.5 hours — ~2.2 million 500s — until we restarted it. Reads on the same node were fine throughout, its peers were fine, and it kept answering `200` on `/health/ready`, so the load balancer kept routing writes to a node that could not accept any.

The cause is an in-process latch in the pool-metadata layer. Every object mutation runs a capacity-admission probe that reads `pool.bin` from the pool's disks under a distributed read lock; during a cluster-wide lock-RPC storm (reported separately), one of those reads timed out. The pool-metadata loader classifies a timed-out read as an *unreadable replica*, an unreadable replica sets `write_blocked`, and `write_blocked` is never cleared. From that second on, the same admission probe rejected every PUT and DELETE on the node with the (unlogged) error "pool metadata writes remain blocked after a recovery-required replica state; restart after all replicas are readable and consistent, with compatible formats". The replicas were readable and consistent again within minutes; the process never re-checked.

The chain, all on `40a2470f`:

1. Every object mutation admits itself through [`acquire_external_decommission_capacity_fence_with_active_source`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/core/pools.rs#L8340-L8352) — from PUT via [`run_decommission_capacity_admitted_mutation_with_capacity_lease`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/store/object.rs#L3738) and the [external-mutation helper](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/store/object.rs#L2741-L2760), from DELETE via [`reconcile_decommission_capacity_before_exact_delete`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/store/rebalance.rs#L945). Reads do not.
2. That fence takes [`acquire_pool_meta_read_guard`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/core/pools.rs#L8308-L8330), which calls [`ensure_write_safe`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/core/pools.rs#L4260-L4266) before and after re-reading the replicas — so once the flag is set, admission fails before any data work (our failures took 9-213 ms).
3. The replica read: [`read_pool_meta_replica`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/core/pools.rs#L4554-L4583) turns *any* error other than not-found — including a transport `Timeout` — into `PoolMetaReplica::Unreadable`. The observing selector then [calls `block_writes()` if any replica is Unreadable](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/core/pools.rs#L4801-L4806), before even choosing a replica, and [`select_pool_meta_replica_reads` sets `repair_write_safe = false`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/core/pools.rs#L4611-L4615), which [`observe_replicas` ORs into `write_blocked`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/core/pools.rs#L4156-L4158).
4. [`block_writes`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/core/pools.rs#L4160-L4162) only ever sets the flag; nothing clears it, and nothing logs the transition. A second latch of the same kind exists next to it: [`PoolMetaTransactionArm::drop`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/core/pools.rs#L4282-L4288) sets `aborted_transaction` whenever an armed pool-metadata save is dropped — including by an ordinary `?` error return inside `save_no_lock_armed_scoped` — and `ensure_write_safe` treats it identically.
5. The refusal surfaces as `Error::other(...)`, i.e. `StorageError::Io`, which the API layer maps to a bare `500 InternalError` with the generic text ([error.rs:424-431](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/rustfs/src/error.rs#L424-L431)); the only trace at `warn` level is the bucket-metadata refresher failing every 15 minutes ([interval](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/bucket/metadata_sys.rs#L51), [message](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/bucket/metadata_sys.rs#L1467)) because bucket heal is gated by the same flag.
6. Readiness never consults it: [`collect_node_readiness_report`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/rustfs/src/server/readiness.rs#L721-L731) reports storage ready whenever the object-store handle exists.

**To Reproduce**

We do not have a from-scratch reproducer of the timed-out read (it needs the lock storm of the companion issue), but the state machine is deterministic and the field timeline pins each step to the second. All times local (+08:00), one node (`object04`), plain-text hourly logs:

- 23:00-23:59: zero lock timeouts, zero 500s, zero pool-metadata messages (baseline).
- 00:07:57 the lock-RPC storm begins (first `Remote lock RPC timed out` to `node3:/mnt/rustfs11`; GOAWAY `too_many_resets` one second later). For the next 33 minutes: thousands of lock timeouts, **zero** 500s.
- 00:40:27-00:40:45: six sporadic 500s whose logged cause is a lock timeout (table-catalog publication guard, `.bucket-incarnation` reads at the 5 s lock-acquire budget).
- **00:40:46.472** `read_config_with_metadata: err: Timeout, file: pool.bin` ([the warn site](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/ecstore/src/config/com.rs#L572)) — the only `pool.bin` read error in the hour. In the same minute the node was timing out lock *releases* on `.rustfs.sys/pool.bin@latest` and logging `lock not found for release` for it: the admission probes were holding pool.bin locks through the storm.
- 00:40:46-00:41:30: 1,994 500s (1,740 DELETE, 254 PUT), then a steady ~85/s.
- 00:41:29: first `Unable to load bucket metadata, will be retried: ... bucket heal cannot run while pool metadata requires recovery: pool metadata writes remain blocked ...` — 62 lines, repeating at :11/:26/:41/:56 for the rest of the outage, on this node only.
- Per hour, `lock timeouts / 500s`: 00h 6,524 / 83,805 · 01h 1,392 / 247,492 · 02h 34 / 423,447 · 03h 38 / 358,153 · 04h 260 / 376,913 · **05h 0 / 222,153** · 06h 3,183 / 379,208 · 07h 105 / 123,141. The 05h hour has no lock timeouts and no GOAWAYs at all and still fails every write: the storm was the trigger, not the cause.
- 08:16:23 restart; last 500 at 08:16:30; zero 500s, zero recovery-required lines since.

The storm itself was cluster-wide — this node logged 6,524 lock timeouts in the first hour and two of its peers logged 9,736 and 4,979 — but only this node's admission probe happened to time out its `pool.bin` read, so only this node latched (its three peers logged 1, 0 and 1 500s in that hour and no recovery-required lines at all). A controlled 4 MB PUT against `127.0.0.1:9000` on each node, three rounds, reproduced the node-locality exactly: `object04 put=ERR status=WEDGED_WRITE` every round, `inline`/`get` fine on the same node, all three peers `OK`; every put path and size class on the node failed, including 0-byte objects.

**Expected behavior**

1. **A timed-out replica read is not a recovery-required replica.** `Unreadable` currently conflates "the bytes are bad" with "the RPC did not answer in time"; the fail-closed block makes sense for the former and is wrong for the latter. A transport error should be retried or fail the one operation, not latch the process.
2. **Either flag must have an exit.** If the replicas become readable and consistent again — as ours were within minutes — the next successful replica read should clear the block. As it stands the error text's own remedy, "restart", is the only one, and nothing tells the operator that.
3. **A node that refuses all writes must not be ready.** This is the highest-value fix: it turns a silent 7.5-hour outage into an ordinary drain. It is the same class as #5921 — a node serving nothing while `/health/ready` stays 200 — and it is the case #5936's fix cannot catch: that watchdog withdraws readiness when an in-flight object request makes no progress for [`RUSTFS_HEALTH_OBJECT_PROGRESS_TIMEOUT_MS`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/config/src/constants/health.rs#L34) (30 s), and [progress is any completion](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/rustfs/src/app/object_traffic_health.rs#L149-L161) — a write rejected in 9-213 ms completes instantly and never looks stalled. Readiness needs to consult the latch directly, not infer it from timing. The latch is also the shape #6285's residual describes — a node-local fault that a restart fixes and nothing restarts — so a locally-determined liveness signal would have caught it too.
4. **Log the transition, once, at error level**, naming the flag and the replica error that set it. Today the flip is silent and the only trace is a bucket-heal retry message every 15 minutes (which also lacks the bucket name).
5. **Do not surface this as a bare `InternalError`.** It is a known, named, deterministic server state; `ServiceUnavailable` (or any distinguishable code) would let clients back off and operators recognise it.

**Environment**

- RustFS `1.0.0-rc.5`, git `40a2470feb567201165a5b809b7598bb4b1f68f5` (confirmed via `rustfs --version` on the node; every link above is pinned to it).
- 4 nodes, EC 12+4, 48 disks; Rocky Linux; `RUST_LOG=warn,rustfs_ecstore::disk=info,rustfs_ecstore::store::init=info,rustfs_scanner=info,rustfs_iam=info`; `RUSTFS_OBJECT_LOCK_ACQUIRE_TIMEOUT=30` (the 5 s figure above is the internal [`DEFAULT_OBJECT_LOCK_ACQUIRE_TIMEOUT`](https://github.com/rustfs/rustfs/blob/40a2470feb567201165a5b809b7598bb4b1f68f5/crates/config/src/constants/object.rs#L500) used by `read_config`).
- Workload: five JuiceFS filesystems, continuous write + delete traffic. Affected node had been up 4 d 17 h; no restarts on any node in that period; no decommission or rebalance ever run on this cluster.

**Additional context**

*What we ruled out.* The storage layer on the node was healthy and indistinguishable from its peers: `storageinfo` reported `('ok','online',False) -> 48` on the affected and a control node; in the affected node's logs the counts for `write quorum`, `Erasure decode failed`, `DiskNotFound`, `FaultyDisk`, `insufficient` and `quorum` were all zero; the per-minute FaultyDisk short-circuit counters were flat zero on all four nodes.

*Why every mutation touches pool.bin.* Worth a look in its own right: with no decommission or rebalance ever configured, every PUT and DELETE still reads `pool.bin` from disk and takes a distributed read lock on `.rustfs.sys/pool.bin@latest` (the "target capacity admission" probe). That is what put pool.bin's lock and read on the critical path of a lock storm in the first place, and it makes pool.bin's lock the hottest lock in the cluster under write load.

*Companion reports.* The lock-RPC storm that supplied the timed-out read is reported separately (client-side 3 s timeout → stream reset → GOAWAY `too_many_resets` → evict/reconnect loop against one erasure set's lock endpoints), as is the diagnosability gap that turned this into hours of investigation (no request counters, the mapped rather than the underlying error logged, silent latch).

Happy to provide the full log windows; the failing state is reproducible on demand for us by the companion issue's storm plus one unlucky admission probe, but we would rather not induce it on production.

cxymds:
## Related Issues

Fixes #7361. Builds on the read-only admission and single-pool fixes in #7359 and #7365.

## Summary of Changes

- A runtime metadata/identity read failure before write dispatch now fails that operation with a typed, retryable error without installing a persistent write block. Startup validation retains its existing fail-closed/degraded-read behavior.
- Arm transaction cancellation protection immediately before the first identity, legacy, V3 prepare, or commit CAS. A rejected first CAS is retryable; an uncertain dispatched write or abandoned publication remains blocked.
- Recover known interrupted pool metadata transactions under the local movement gates and distributed write fence. Drain old decommission workers, validate identity/topology and durable authority, conditionally repair replicas, verify convergence, publish the durable snapshot, then clear the original block. Preserve the observed V3 format floor, including prepare-only migration recovery.
- Return S3 `503 ServiceUnavailable` with a safe public message while preserving the typed local root cause. Bound metadata readiness inspection to 100 ms and distinguish a write block from inspection timeout. Add first-transition logs/counters and update the recovery runbook and readiness matrix.

## Verification

Behavioral regressions passed on `63e1eb6c750bed47752dc778fd143bb21d1a94d4`, based on main `4c2a0cdf9a7fbf75edf5cd9da24350cd861de1db`, with no local source changes. Static-check evidence below is reused from the identical task patch before the clean rebase; `git range-diff` reports an unchanged commit patch.

1. `cargo test --locked -p rustfs-ecstore --lib pool_meta --no-fail-fast`: **129 passed**. Covers retry after read failure, identity-write cancellation, first/later CAS rejection, interrupted V1/V2/V3 repair, corrupt/missing/incompatible replicas, worker drain, outstanding outcomes, canceled recovery, fence/publication failures, and bounded readiness. The single/multi-pool mutation regression performs PUT, DELETE, multipart completion/abort, and exact byte read-back after preflight failure. Existing startup and format-heal fail-closed regressions also pass. The upstream regression `cargo test --locked -p rustfs-ecstore --lib single_pool_public_writes_skip_decommission_capacity_admission --no-fail-fast` also passed after rebase.
2. `cargo test --locked -p rustfs --lib pool_metadata --no-fail-fast`: **2 passed**, covering all four typed failure classes, safe 503 mapping, source-preserving error cloning, and distinct readiness reasons. `cargo test --locked -p rustfs --lib server::readiness::tests --no-fail-fast`: **36 passed**, including cache/single-flight behavior, sanitized probe timeout, read/write quorum separation, and startup readiness.
3. `cargo clippy --locked -p rustfs-ecstore -p rustfs --lib --tests --no-deps -- -D warnings`: **passed** on pre-rebase commit `2bfd70bf45a35d4190659e99c928aba22e4a61dd`; the unchanged task patch preserves that lint evidence. `cargo fmt --all --check` and `git diff origin/main...HEAD --check` passed after rebase. `bash scripts/check_logging_guardrails.sh`, `bash scripts/check_doc_paths.sh`, `bash scripts/check_layer_dependencies.sh`, and `bash scripts/check_architecture_migration_rules.sh` passed on the unchanged relevant diff.

A direct baseline run of the new recovery regressions was unavailable because they exercise APIs introduced here. Prior sticky-read test expectations were changed only for pre-dispatch runtime read failures; startup, corrupt metadata, missing authority, and lost-fence protections remain covered. The original four-node long-running workload/RPC storm and throughput benchmarks were not replayed. Full-workspace gates were not run locally; verification is scoped to the changed storage, HTTP, readiness, and architecture boundaries.

## Impact

This is a high-risk metadata/recovery and S3-visible change. Unavailability is retryable at the HTTP boundary, but corruption, conflicting identity/epoch/transaction, unsafe topology changes, and startup or storage-format-heal blocks still require operator reconciliation. The worker does not force-clear a block or initialize all-missing metadata, and it waits for an attached rebalance worker to quiesce without canceling it. See `docs/operations/pool-metadata-recovery.md` for the restart/escalation boundary.

Healthy background checks do not read disk metadata. Recovery polls every five seconds, backs off to sixty seconds on failure, has a thirty-second attempt budget, and stops on shutdown. No new dependencies, environment variables, administrative routes, persistent formats, or storage/RPC error numbers are introduced. Rollback must use a build that supports any V3 metadata already persisted; never delete `pool.bin` or identity markers to clear readiness.

## Additional Notes

Two sequential final-diff review passes covered the following lenses; no independent external review is claimed. The error-chain issues found during review were corrected and covered by regression tests.

- Correctness: preflight failures do not arm writes; original CAS errors survive the caller and latch paths; recovery selects durable authority rather than speculative memory.
- Simplicity: reuse existing selection, CAS, fencing, and decommission supervision; no parallel persistence protocol or force-clear API.
- Test coverage: fault-injection and real single/multi-pool mutation checks cover the affected state transitions; distributed stress remains unverified.
- Security: validate every identity/topology/metadata copy before repair; fail closed on missing authority; do not expose internal error details over S3.
- Concurrency/durability: retain the block through worker drain, outstanding transaction completion, repair verification, and fenced publication; canceled recovery cannot clear it.
- Compatibility: preserve legacy selection, startup policy, V3 format floors, wire error codes, and outer I/O error kind/message when cloning typed context.
- Performance: healthy recovery checks avoid disk I/O and readiness waits are bounded; no throughput improvement claim is made.

RPC storm mitigation (#7363) and general multi-pool idle capacity-admission lock optimization are outside this PR.

### DIFF
--- a/crates/ecstore/src/core/pools.rs
+++ b/crates/ecstore/src/core/pools.rs
@@ -5228,7 +5228,7 @@ where
     R: Into<PoolMetaReplicaRead>,
 {
     let replicas = replicas.into_iter().map(Into::into).collect::<Vec<_>>();
-    if let Some(err) = pool_meta_replica_read_failure(&replicas) {
+    if let Some(err) = pool_meta_replica_read_failure(&replicas, "pool_read") {
         return Err(err);
     }
     match select_pool_meta_replica_reads(replicas) {
@@ -5248,13 +5248,14 @@ fn block_pool_meta_validation(write_state: &mut PoolMetaWriteState, err: Error, 
     Error::other(context)
 }
 
-fn pool_meta_replica_read_failure(replicas: &[PoolMetaReplicaRead]) -> Option<Error> {
+fn pool_meta_replica_read_failure(replicas: &[PoolMetaReplicaRead], operation: &str) -> Option<Error> {
     replicas.iter().find_map(|read| match &read.replica {
-        PoolMetaReplica::Unreadable(source) => Some(Error::other(pool_metadata_error(
-            crate::error::PoolMetadataFailure::ReadUnavailable,
-            "pool_read",
-            Some(Arc::clone(source)),
-        ))),
+        PoolMetaReplica::Unreadable(source) => {
+            let mut context =
+                pool_metadata_error(crate::error::PoolMetadataFailure::ReadUnavailable, "pool_read", Some(Arc::clone(source)));
+            context.operation = operation.to_owned();
+            Some(Error::other(context))
+        }
         _ => None,
     })
 }
@@ -5268,7 +5269,7 @@ where
     R: Into<PoolMetaReplicaRead>,
 {
     let replicas = replicas.into_iter().map(Into::into).collect::<Vec<_>>();
-    if let Some(err) = pool_meta_replica_read_failure(&replicas) {
+    if let Some(err) = pool_meta_replica_read_failure(&replicas, operation) {
         return Err(err);
     }
     let selection = select_pool_meta_replica_reads(replicas)?;
@@ -6349,7 +6350,7 @@ where
             write_state.observe_identity(&identity)?;
         }
         let reads = read_pool_meta_replicas(pools, true).await;
-        if let Some(err) = pool_meta_replica_read_failure(&reads) {
+        if let Some(err) = pool_meta_replica_read_failure(&reads, "pool metadata transaction recovery") {
             return Err(err);
         }
         if reads
@@ -6771,7 +6772,7 @@ impl PoolMeta {
             write_state.observe_identity(&identity)?;
         }
         let reads = read_pool_meta_replicas(pools, true).await;
-        if let Some(err) = pool_meta_replica_read_failure(&reads) {
+        if let Some(err) = pool_meta_replica_read_failure(&reads, "pool metadata load") {
             write_state.block_with_context(pool_metadata_error(
                 crate::error::PoolMetadataFailure::RecoveryRequired,
                 "startup_read",
--- a/crates/ecstore/src/core/pools.rs
+++ b/crates/ecstore/src/core/pools.rs
@@ -4344,7 +4344,7 @@ enum PoolMetaReplica {
     },
     Corrupt(String),
     Incompatible(String),
-    Unreadable(String),
+    Unreadable(Arc<Error>),
 }
 
 #[derive(Debug, Clone)]
@@ -4390,6 +4390,11 @@ impl PoolMetaReplicaState {
 pub(crate) struct PoolMetaWriteState {
     write_blocked: bool,
     aborted_transaction: Arc<AtomicBool>,
+    block_context: Option<crate::error::PoolMetadataError>,
+    transaction_failure: Arc<std::sync::Mutex<Option<crate::error::PoolMetadataError>>>,
+    active_transactions: Arc<AtomicUsize>,
+    recovery_failure: Option<Arc<Error>>,
+    block_started_at: Arc<std::sync::Mutex<Option<OffsetDateTime>>>,
     expected_cluster_id: Option<uuid::Uuid>,
     cluster_epoch: Option<u64>,
     pool_meta_absent: bool,
@@ -4456,6 +4461,19 @@ impl PoolMetaWriteState {
     pub(crate) fn independent_clone_for_test(&self) -> Self {
         let mut cloned = self.clone();
         cloned.aborted_transaction = Arc::new(AtomicBool::new(self.aborted_transaction.load(Ordering::Acquire)));
+        cloned.transaction_failure = Arc::new(std::sync::Mutex::new(
+            self.transaction_failure
+                .lock()
+                .unwrap_or_else(std::sync::PoisonError::into_inner)
+                .clone(),
+        ));
+        cloned.active_transactions = Arc::new(AtomicUsize::new(0));
+        cloned.block_started_at = Arc::new(std::sync::Mutex::new(
+            *self
+                .block_started_at
+                .lock()
+                .unwrap_or_else(std::sync::PoisonError::into_inner),
+        ));
         cloned
     }
 
@@ -4470,21 +4488,46 @@ impl PoolMetaWriteState {
     }
 
     pub(crate) fn observe_replicas(&mut self, replica_state: PoolMetaReplicaState) {
-        self.write_blocked |= !replica_state.repair_write_safe;
+        if !replica_state.repair_write_safe {
+            self.block_writes();
+        }
     }
 
     fn block_writes(&mut self) {
+        self.block_with_context(pool_metadata_error(
+            crate::error::PoolMetadataFailure::RecoveryRequired,
+            "validation",
+            None,
+        ));
+    }
+
+    fn block_with_context(&mut self, mut context: crate::error::PoolMetadataError) {
+        if !self.write_blocked {
+            record_pool_meta_block_once(&self.block_started_at, &mut context);
+            self.block_context = Some(context);
+        } else if let Some(previous) = &self.block_context
+            && previous.source.is_none()
+            && context.source.is_some()
+        {
+            context.since = previous.since;
+            self.block_context = Some(context);
+        }
         self.write_blocked = true;
     }
 
     pub(crate) fn block_writes_after_fence_loss(&mut self) {
-        self.block_writes();
+        self.block_with_context(pool_metadata_error(crate::error::PoolMetadataFailure::FenceLost, "format_heal", None));
     }
 
     fn arm_transaction(&self) -> PoolMetaTransactionArm {
+        self.active_transactions.fetch_add(1, Ordering::SeqCst);
         PoolMetaTransactionArm {
             aborted_transaction: Arc::clone(&self.aborted_transaction),
-            armed: true,
+            failure: Arc::clone(&self.transaction_failure),
+            active_transactions: Arc::clone(&self.active_transactions),
+            block_started_at: Arc::clone(&self.block_started_at),
+            phase: None,
+            source: None,
         }
     }
 
@@ -4549,11 +4592,8 @@ impl PoolMetaWriteState {
         if !self.pool_meta_absent {
             return Ok(());
         }
-        let result = self.validate_missing_metadata_can_initialize();
-        if result.is_err() {
-            self.block_writes();
-        }
-        result
+        self.validate_missing_metadata_can_initialize()
+            .map_err(|err| block_pool_meta_validation(self, err, "metadata_absence"))
     }
 
     fn validate_missing_metadata_can_initialize(&self) -> Result<()> {
@@ -4579,29 +4619,99 @@ impl PoolMetaWriteState {
         if !self.write_blocked && !self.aborted_transaction.load(Ordering::SeqCst) {
             return Ok(());
         }
-        Err(Error::other(format!(
-            "{operation}: pool metadata writes remain blocked after a recovery-required replica state; restart after all replicas are readable and consistent, with compatible formats"
-        )))
+        let mut context = self
+            .block_context
+            .clone()
+            .or_else(|| {
+                self.transaction_failure
+                    .lock()
+                    .unwrap_or_else(std::sync::PoisonError::into_inner)
+                    .clone()
+            })
+            .unwrap_or_else(|| pool_metadata_error(crate::error::PoolMetadataFailure::FenceLost, "external_fence", None));
+        context.operation = operation.to_owned();
+        Err(Error::other(context))
     }
+}
+
+fn pool_metadata_error(
+    kind: crate::error::PoolMetadataFailure,
+    phase: &'static str,
+    source: Option<Arc<Error>>,
+) -> crate::error::PoolMetadataError {
+    crate::error::PoolMetadataError {
+        kind,
+        operation: phase.to_owned(),
+        phase,
+        since: OffsetDateTime::now_utc(),
+        source,
+    }
+}
+
+fn record_pool_meta_block_once(
+    started_at: &std::sync::Mutex<Option<OffsetDateTime>>,
+    context: &mut crate::error::PoolMetadataError,
+) {
+    let mut started_at = started_at.lock().unwrap_or_else(std::sync::PoisonError::into_inner);
+    if let Some(since) = *started_at {
+        context.since = since;
+        return;
+    }
+    *started_at = Some(context.since);
+    error!(
+        event = EVENT_DECOMMISSION_STATE,
+        component = LOG_COMPONENT_ECSTORE,
+        subsystem = LOG_SUBSYSTEM_POOLS,
+        state = "pool_metadata_blocked",
+        reason = context.kind.as_str(),
+        phase = context.phase,
+        blocked_since = %context.since,
+        "Pool metadata writes blocked"
+    );
+    metrics::counter!("rustfs_pool_metadata_blocks_total", "reason" => context.kind.as_str()).increment(1);
 }
 
 #[derive(Debug)]
 struct PoolMetaTransactionArm {
     aborted_transaction: Arc<AtomicBool>,
-    armed: bool,
+    failure: Arc<std::sync::Mutex<Option<crate::error::PoolMetadataError>>>,
+    active_transactions: Arc<AtomicUsize>,
+    block_started_at: Arc<std::sync::Mutex<Option<OffsetDateTime>>>,
+    phase: Option<&'static str>,
+    source: Option<Arc<Error>>,
 }
 
 impl PoolMetaTransactionArm {
+    fn failure(&mut self, err: Error) -> Error {
+        let Some(phase) = self.phase else {
+            return err;
+        };
+        let source = Arc::new(err);
+        self.source = Some(Arc::clone(&source));
+        Error::other(pool_metadata_error(
+            crate::error::PoolMetadataFailure::TransactionUnknown,
+            phase,
+            Some(source),
+        ))
+    }
+
     fn disarm(&mut self) {
-        self.armed = false;
+        self.phase = None;
     }
 }
 
 impl Drop for PoolMetaTransactionArm {
     fn drop(&mut self) {
-        if self.armed {
-            self.aborted_transaction.store(true, Ordering::SeqCst);
+        if let Some(phase) = self.phase {
+            let mut context =
+                pool_metadata_error(crate::error::PoolMetadataFailure::TransactionUnknown, phase, self.source.take());
+            let mut failure = self.failure.lock().unwrap_or_else(std::sync::PoisonError::into_inner);
+            if !self.aborted_transaction.swap(true, Ordering::SeqCst) {
+                record_pool_meta_block_once(&self.block_started_at, &mut context);
+                *failure = Some(context);
+            }
         }
+        self.active_transactions.fetch_sub(1, Ordering::SeqCst);
     }
 }
 
@@ -4897,7 +5007,7 @@ where
             cas: PoolMetaCasToken::Missing,
         },
         Err(err) => PoolMetaReplicaRead {
-            replica: PoolMetaReplica::Unreadable(err.to_string()),
+            replica: PoolMetaReplica::Unreadable(Arc::new(err)),
             cas: PoolMetaCasToken::Unsafe,
         },
     }
@@ -5118,25 +5228,35 @@ where
     R: Into<PoolMetaReplicaRead>,
 {
     let replicas = replicas.into_iter().map(Into::into).collect::<Vec<_>>();
-    if replicas
-        .iter()
-        .any(|replica| matches!(&replica.replica, PoolMetaReplica::Unreadable(_)))
-    {
-        write_state.block_writes();
+    if let Some(err) = pool_meta_replica_read_failure(&replicas) {
+        return Err(err);
     }
     match select_pool_meta_replica_reads(replicas) {
         Ok(selection) => {
             if let Err(err) = write_state.observe_selection(&selection) {
-                write_state.block_writes();
-                return Err(err);
+                return Err(block_pool_meta_validation(write_state, err, "pool_selection"));
             }
             Ok(selection)
         }
-        Err(err) => {
-            write_state.block_writes();
-            Err(err)
-        }
+        Err(err) => Err(block_pool_meta_validation(write_state, err, "pool_selection")),
     }
+}
+
+fn block_pool_meta_validation(write_state: &mut PoolMetaWriteState, err: Error, phase: &'static str) -> Error {
+    let context = pool_metadata_error(crate::error::PoolMetadataFailure::RecoveryRequired, phase, Some(Arc::new(err)));
+    write_state.block_with_context(context.clone());
+    Error::other(context)
+}
+
+fn pool_meta_replica_read_failure(replicas: &[PoolMetaReplicaRead]) -> Option<Error> {
+    replicas.iter().find_map(|read| match &read.replica {
+        PoolMetaReplica::Unreadable(source) => Some(Error::other(pool_metadata_error(
+            crate::error::PoolMetadataFailure::ReadUnavailable,
+            "pool_read",
+            Some(Arc::clone(source)),
+        ))),
+        _ => None,
+    })
 }
 
 fn select_pool_meta_replicas_for_read_probe<R>(
@@ -5147,7 +5267,11 @@ fn select_pool_meta_replicas_for_read_probe<R>(
 where
     R: Into<PoolMetaReplicaRead>,
 {
-    let selection = select_pool_meta_replica_reads(replicas.into_iter().map(Into::into).collect())?;
+    let replicas = replicas.into_iter().map(Into::into).collect::<Vec<_>>();
+    if let Some(err) = pool_meta_replica_read_failure(&replicas) {
+        return Err(err);
+    }
+    let selection = select_pool_meta_replica_reads(replicas)?;
     write_state.validate_selection(&selection)?;
     selection.replica_state.ensure_write_safe(operation)?;
     if selection.absent && (write_state.expected_cluster_id.is_some() || write_state.identity_initialized.is_some()) {
@@ -5185,7 +5309,17 @@ where
     S: EcstoreObjectIO,
 {
     let replicas = read_pool_meta_replicas(pools, no_lock).await;
-    select_pool_meta_replicas_for_read_probe(write_state, replicas, operation)
+    select_pool_meta_replicas_for_read_probe(write_state, replicas, operation).map_err(|err| {
+        if err.pool_metadata_failure().is_some() {
+            err
+        } else {
+            Error::other(pool_metadata_error(
+                crate::error::PoolMetadataFailure::RecoveryRequired,
+                "read_probe",
+                Some(Arc::new(err)),
+            ))
+        }
+    })
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -5228,7 +5362,7 @@ enum PoolMetaIdentityReplica {
     Valid(PersistedPoolMetaIdentity),
     Corrupt(String),
     Incompatible(String),
-    Unreadable(String),
+    Unreadable(Arc<Error>),
 }
 
 #[derive(Debug)]
@@ -5320,7 +5454,7 @@ where
             cas: PoolMetaCasToken::Missing,
         },
         Err(err) => PoolMetaIdentityRead {
-            replica: PoolMetaIdentityReplica::Unreadable(err.to_string()),
+            replica: PoolMetaIdentityReplica::Unreadable(Arc::new(err)),
             cas: PoolMetaCasToken::Unsafe,
         },
     }
@@ -5396,6 +5530,16 @@ where
     S: EcstoreObjectIO,
 {
     let reads = join_all(pools.into_iter().map(read_pool_meta_identity_replica)).await;
+    if let Some(source) = reads.iter().find_map(|read| match &read.replica {
+        PoolMetaIdentityReplica::Unreadable(source) => Some(Arc::clone(source)),
+        _ => None,
+    }) {
+        return Err(Error::other(pool_metadata_error(
+            crate::error::PoolMetadataFailure::ReadUnavailable,
+            "identity_read",
+            Some(source),
+        )));
+    }
     select_pool_meta_identity(reads, expected_cluster_id)
 }
 
@@ -5410,13 +5554,17 @@ where
     let selection = match load_pool_meta_identity(pools, expected_cluster_id).await {
         Ok(selection) => selection,
         Err(err) => {
-            write_state.block_writes();
-            return Err(err);
+            if err
+                .pool_metadata_failure()
+                .is_some_and(|context| context.kind == crate::error::PoolMetadataFailure::ReadUnavailable)
+            {
+                return Err(err);
+            }
+            return Err(block_pool_meta_validation(write_state, err, "identity_selection"));
         }
     };
     if let Err(err) = write_state.observe_identity(&selection) {
-        write_state.block_writes();
-        return Err(err);
+        return Err(block_pool_meta_validation(write_state, err, "identity_selection"));
     }
     Ok(selection)
 }
@@ -5519,6 +5667,7 @@ async fn save_pool_meta_object_cas<S>(
     token: &PoolMetaCasToken,
     fence: &PoolMetaPersistenceFence<'_>,
     phase: &'static str,
+    transaction_arm: &mut PoolMetaTransactionArm,
 ) -> Result<crate::object_api::ObjectInfo>
 where
     S: EcstoreObjectIO,
@@ -5532,11 +5681,30 @@ where
         ..Default::default()
     };
     fence.add_to_options(&mut opts);
+    // Cancellation can happen at the very first poll of the storage future.
+    // Arm before dispatch, but not during read/encode/fence preflight.
+    let previous_phase = transaction_arm.phase;
+    transaction_arm.phase = Some(phase);
     let result = save_config_with_opts_and_metadata(pool, object, data, &opts).await;
     if matches!(&result, Err(Error::PreconditionFailed)) {
         record_pool_meta_stale_write_rejection(phase);
+        transaction_arm.phase = previous_phase;
     }
-    let object_info = result?;
+    let object_info = match result {
+        Ok(info) => info,
+        Err(err) => {
+            let source = Arc::new(err);
+            transaction_arm.source = Some(Arc::clone(&source));
+            if matches!(source.as_ref(), Error::PreconditionFailed) {
+                return Err(Error::PreconditionFailed);
+            }
+            return Err(Error::other(pool_metadata_error(
+                crate::error::PoolMetadataFailure::TransactionUnknown,
+                phase,
+                Some(source),
+            )));
+        }
+    };
     fence.ensure_held()?;
     Ok(object_info)
 }
@@ -5546,6 +5714,7 @@ async fn persist_pool_meta_identity<S>(
     write_state: &mut PoolMetaWriteState,
     initialized: bool,
     fence: &PoolMetaPersistenceFence<'_>,
+    transaction_arm: &mut PoolMetaTransactionArm,
 ) -> Result<()>
 where
     S: EcstoreObjectIO,
@@ -5588,7 +5757,17 @@ where
         let data = encode_pool_meta_identity(identity)?;
         let mut conflict = false;
         for (pool, token) in pools.iter().cloned().zip(&selection.cas_tokens) {
-            match save_pool_meta_object_cas(pool, POOL_META_IDENTITY_NAME, data.clone(), token, fence, "identity_cas").await {
+            match save_pool_meta_object_cas(
+                pool,
+                POOL_META_IDENTITY_NAME,
+                data.clone(),
+                token,
+                fence,
+                "identity_cas",
+                transaction_arm,
+            )
+            .await
+            {
                 Ok(_) => {}
                 Err(Error::PreconditionFailed) => {
                     conflict = true;
@@ -5631,7 +5810,17 @@ pub(crate) async fn persist_pool_meta_identity_for_startup<S>(
 where
     S: EcstoreObjectIO,
 {
-    persist_pool_meta_identity(pools, write_state, initialized, &PoolMetaPersistenceFence::Distributed(None)).await
+    let mut transaction_arm = write_state.arm_transaction();
+    persist_pool_meta_identity(
+        pools,
+        write_state,
+        initialized,
+        &PoolMetaPersistenceFence::Distributed(None),
+        &mut transaction_arm,
+    )
+    .await?;
+    transaction_arm.disarm();
+    Ok(())
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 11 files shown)


houseme:
Follow-up CI fix pushed in `84bdfd4cc`.

Conclusion:

- Updated the stale `error(format!)` ratchet baseline after the pool metadata recovery diff reduced call sites.
- Local `scripts/check_error_other_format_ratchet.sh`: PASS.
- Local `git diff --check`: PASS.
- No runtime code was changed in this follow-up commit.

The new GitHub checks are running on the updated head.

houseme:
CI follow-up pushed after rebasing on the latest `main`.

Conclusion:

- The previous failures in `Test and Lint` and `Test and Lint (rio-v2)` both reduced to `rustfs-ecstore::store::heal::tests::unscoped_heal_object_suspended_owner_semantics`.
- The fix preserves the caller operation in `PoolMetadataError` for pool metadata read-probe failures while keeping the phase as `pool_read`.
- The suspended-owner regression now asserts the typed `ReadUnavailable` failure, the admission operation, and that a read-only admission failure does not latch the pool metadata writer.

Validation:

- `cargo test --locked -q -p rustfs-ecstore --lib unscoped_heal_object_suspended_owner_semantics -j4`: PASS, 1/1.
- `cargo test --locked -q -p rustfs-ecstore --features rio-v2 --lib unscoped_heal_object_suspended_owner_semantics -j4`: PASS, 1/1.
- `cargo fmt --all --check`: PASS.
- `git diff --check`: PASS.

Current status: new head `b5a05bd7400734127f6633e880e6f3cea219b546`; GitHub CI has restarted and is still running.

houseme:
CI follow-up:

- Rebased the branch onto current `origin/main` (`cc91483fac`) so the PR now carries only the pool metadata recovery changes plus the error-format baseline update.
- Fixed the pool metadata read-probe context used by heal capacity admission: read-only replica unavailability still reports `PoolMetadataFailure::ReadUnavailable`, and the caller operation remains `target capacity admission failed`.
- Updated the focused heal regression to assert the typed failure context instead of matching unstable rendered text.

Validation:

- `cargo test --locked -q -p rustfs-ecstore --lib unscoped_heal_object_suspended_owner_semantics -j4`: PASS
- `cargo test --locked -q -p rustfs-ecstore --features rio-v2 --lib unscoped_heal_object_suspended_owner_semantics -j4`: PASS
- `cargo test --locked -q -p rustfs-ecstore --lib pool_meta_read_probe -j4`: PASS
- `cargo fmt --package rustfs-ecstore --check`: PASS
- `git diff --check`: PASS
- `cargo clippy --locked -q -p rustfs-ecstore --lib --tests --no-deps -- -D warnings`: PASS

The new GitHub CI run is in progress on head `b5a05bd7400734127f6633e880e6f3cea219b546`; this PR is still not treated as complete until that run finishes successfully and the branch is merged.

houseme:
Follow-up verification after the latest main refresh: the previously failing focused test now passes locally on the current PR head. Additional local result: `cargo test --locked -q -p rustfs-ecstore --lib unscoped_heal_object_suspended_owner_semantics -j4` PASS, `cargo test --locked -q -p rustfs-ecstore --features rio-v2 --lib unscoped_heal_object_suspended_owner_semantics -j4` PASS, and `cargo test --locked -q -p rustfs-ecstore --lib pool_meta -j4` PASS (129/129). Current PR checks have no completed failures, but several CI jobs are still pending, so this remains open.

houseme:
CI triage update: the current `Test and Lint` failure is `rustfs-scanner::scanner_io::tests::scoped_scan_production_entry_preserves_deep_and_full_maintenance_work`, and it reproduces on current main as a scanner test-oracle issue rather than a pool metadata recovery regression. I opened rustfs/rustfs#7404 with the isolated oracle fix; the focused scanner test passes there. This PR remains open until CI is rerun on a base that contains that fix.

crates/ecstore/src/core/pools.rs
```diff
@@ -10275,8 +10668,174 @@ impl ECStore {
     /// Read-only admission probes do not change this state; startup and real
     /// metadata transactions still latch it on unrecoverable conditions.
     pub async fn pool_meta_writes_ready(&self) -> bool {
-        let write_state = self.pool_meta_save_gate.lock().await;
-        !write_state.write_blocked && !write_state.aborted_transaction.load(Ordering::SeqCst)
+        self.pool_meta_write_status().await.is_ok()
+    }
+
+    /// A health probe must not wait behind a metadata transaction's disk I/O.
+    pub async fn pool_meta_write_status(&self) -> Result<()> {
```
overtrue, line 10675: **Blocking (finding 1).** `pool_meta_writes_ready()` is now `pool_meta_write_status().await.is_ok()`, so the 100 ms `Error::Timeout` is indistinguishable from a latched block for the only consumers (`rustfs/src/server/readiness.rs:847-868`, not in this diff), which report it as `pool_meta_write_blocked` and strip `StorageQuorumUnavailable`. The multi-pool admission path holds the save gate across a distributed read-lock wait of up to `get_lock_acquire_timeout()` (pools.rs:9570-9573), so a healthy cluster under write load flips readiness with the sticky-block reason. `pool_metadata_check_timeout` from the docs exists nowhere in code. Readiness needs to call `pool_meta_write_status()` and treat `Timeout` as unknown/last-known, not blocked.

crates/ecstore/src/core/pools.rs
```diff
@@ -10275,8 +10668,174 @@ impl ECStore {
...
+                canceler.cancel();
+                draining = true;
+            }
+            if draining {
+                return Err(Error::OperationCanceled);
+            }
+        }
+        let movement_gate = self.ctx.data_movement_operation_gate();
+        let _movement_guard = movement_gate.write().await;
+        let mut state = self.pool_meta_save_gate.lock().await;
```
overtrue, line 10718: **Blocking (finding 2).** The save gate is held from here through `get_write_lock(get_lock_acquire_timeout())` (up to 5 s), the multi-pool reads, `repair_pool_meta_transaction` and publication, bounded only by the 30 s budget in `store/init.rs`. Every multi-pool PUT/DELETE queues on this mutex before it can fail fast (pools.rs:9570 precedes `ensure_write_safe`), so while a block is installed mutations go from a fast 503 to a stall of up to 30 s on every attempt of the 5 s→60 s cycle; `movement_gate.write()` above also makes bucket-heal `try_read` fail for the same window. Take the namespace write lock before the save gate, or expose a recovery-in-progress flag readable without the mutex.

crates/ecstore/src/core/pools.rs
```diff
@@ -10275,8 +10668,174 @@ impl ECStore {
...
+
+    pub(crate) async fn recover_pool_meta_transaction(&self) -> Result<bool> {
+        {
+            let Ok(state) = self.pool_meta_save_gate.try_lock() else {
+                return Ok(false);
+            };
+            if state.ensure_write_safe("pool metadata recovery").is_ok() {
+                return Ok(false);
+            }
+            if state.write_blocked
```
overtrue, line 10690: **Major (finding 3).** Refusing every `write_blocked` state means two runtime paths still turn a transient fault into a restart-only outage: `store/heal.rs:268-376` latches `block_writes_after_fence_loss()` on any `is_lock_lost()` (a lease refresh blip), and `decommission_cancel_transaction` (pools.rs:11606-11624) saves through a throwaway `PoolMetaWriteState::default()` (via `save_no_lock_with_fence`, :6889-6904) and then calls `save_guard.block_writes()` on *any* error — including the `ReadUnavailable` preflight and first-CAS `PreconditionFailed` this PR defines as retryable — returning an untyped error (500). Either classify no-dispatched-CAS failures as retryable on those paths too, or name both residual latches in the Impact section.

crates/ecstore/src/store/init.rs
```diff
@@ -769,6 +766,33 @@ impl ECStore {
             });
         }
 
+        let recovery_store = self.clone();
```
overtrue, line 769: **Major (finding 4).** This spawn is the only production caller of `recover_pool_meta_transaction`; every test calls the method directly, so deleting this block leaves the suite green. Nothing covers cancellation on `recovery_rx`, the 30 s timeout → `Error::Timeout` mapping, the 5 s reset / doubling / 60 s cap, or that `record_pool_meta_recovery_failure` is reached. Extract the tick into a testable `async fn` returning the next delay and add one integration test that latches a real store and polls `pool_meta_writes_ready()` until the worker flips it. Consider jitter as well: latched nodes in a cluster-wide blip retry the same `pool.bin` write lock on the same schedule.

crates/ecstore/src/core/pools.rs
```diff
@@ -5532,11 +5682,30 @@ where
...
     if matches!(&result, Err(Error::PreconditionFailed)) {
         record_pool_meta_stale_write_rejection(phase);
+        transaction_arm.phase = previous_phase;
     }
-    let object_info = result?;
+    let object_info = match result {
+        Ok(info) => info,
+        Err(err) => {
+            let source = Arc::new(err);
+            transaction_arm.source = Some(Arc::clone(&source));
```
overtrue, line 5698: **Minor (finding 5).** `source` is set on every `Err` (including `PreconditionFailed`, whose `phase` is then restored) and never cleared on `Ok`. `prepare_cas` pool0 → `PreconditionFailed`; retry → `Ok`; future dropped → `Drop` records `TransactionUnknown{phase:"prepare_cas", source: PreconditionFailed}`, blaming a precondition rejection for an unknown-outcome abort. Set `source = None` on `Ok` and skip it for `PreconditionFailed`.

crates/ecstore/src/core/pools.rs
```diff
@@ -10275,8 +10668,174 @@ impl ECStore {
...
+        }
+        let outcome = repair_pool_meta_transaction(self.pools.clone(), &mut candidate, selection, &fence).await?;
+        let mut pool_meta = self.pool_meta.write().await;
+        fence.ensure_held()?;
+        self.ctx.advance_data_movement_operation_epoch();
+        *pool_meta = outcome.committed.clone();
+        fence.ensure_held()?;
+        outcome.disarm();
+        // No await between final fence validation, publication, and clearing.
+        candidate.aborted_transaction = Arc::clone(&state.aborted_transaction);
```
overtrue, line 10779: **Minor (finding 9).** Struct-update at :10748 followed by hand-relinking four Arcs here before `*state = candidate`: a sixth shared field later needs a fifth line, or an outstanding `PoolMetaTransactionArm::drop` writes into an orphaned latch nobody reads. One `Arc<Mutex<PoolMetaLatch { aborted, failure, blocked_since }>>` plus the counter removes this and most of the 14 `PoisonError::into_inner` sites; the clear becomes a single `*latch = default()`.

docs/architecture/readiness-matrix.md
```diff
@@ -40,6 +40,9 @@ an unknown or unsupported peer-health snapshot degrades readiness with
 - Liveness reports process availability and must not depend on storage, IAM,
   lock quorum, or peer health.
 - Node readiness reports local dependency readiness.
+- A blocked pool metadata writer degrades node and cluster-write readiness with
+  `pool_metadata_blocked`. Metadata save-gate inspection is bounded to 100 ms;
+  contention reports `pool_metadata_check_timeout` without installing a block.
```
overtrue, line 45: **Finding 1 (docs).** Neither reason string exists: the emitted reason is `pool_meta_write_blocked` (`rustfs/src/shared_types.rs:65`), and `pool_metadata_check_timeout` appears nowhere in code (same sentence in `docs/operations/pool-metadata-recovery.md:52`). Today the timeout *is* reported as `pool_meta_write_blocked`. Either implement the distinct reason or correct both docs.

crates/ecstore/src/error/mod.rs
```diff
@@ -23,17 +23,59 @@ use s3s::S3ErrorCode;
...
+            Self::ReadUnavailable => "read_unavailable",
+            Self::RecoveryRequired => "recovery_required",
+            Self::TransactionUnknown => "transaction_unknown",
+            Self::FenceLost => "fence_lost",
+        }
+    }
+}
+
+/// Local control-plane context. Keep the existing storage error wire codes;
+/// the HTTP boundary recognizes this typed source, not an error-message prefix.
```
overtrue, line 56: **Nit.** This `#[error]` interpolates `operation` and the full source `Display`, which is exactly the high-cardinality `Io` message the `::other(format!` ratchet exists to keep out of quorum bucketing. Not live today (these errors do not reach `reduce_errs`), but rendering only `kind`/`phase` here and leaving detail to `#[source]` is the free fix the ratchet script itself recommends.

overtrue (changes requested):
## Adversarial review — head `b5a05bd74` (base `cc91483fa`)

High-risk tier (metadata persistence, locking, S3-visible semantics). Two independent reviewers split the seven lenses; each finding below was re-verified against the head before posting. Overall: the core mechanics are sound — the phase/arm bookkeeping, the CAS-conflict handling, the fence checks, the no-await window between disarm and latch clear, and the lock order all held up under attack — but the readiness half of the PR is not what the body and docs claim, and the recovery worker changes the latency profile of the multi-pool mutation path while a block is installed. Requesting changes for the two items marked blocking; everything else is scoped or minor.

### Blocking

**1. Readiness: the 100 ms save-gate timeout is reported as a hard `pool_meta_write_blocked`, and the documented `pool_metadata_check_timeout` reason does not exist.** `rustfs/src/server/readiness.rs` is not in this diff. Its two consumers (`node_pool_meta_write_readiness` :847-856, `collect_storage_write_readiness_uncached` :859-868) call `pool_meta_writes_ready()`, which is now `pool_meta_write_status().await.is_ok()` (pools.rs:10670-10679), so `Err(Error::Timeout)` after 100 ms of mutex wait becomes `ready: false, pool_meta_write_blocked: true`; `degraded_reasons_with_pool_meta_status` then inserts `PoolMetaWriteBlocked` at position 0 and strips `StorageQuorumUnavailable`. Trigger needs no fault: the ordinary multi-pool PUT/DELETE admission path holds the save gate across `acquire_pool_meta_read_guard` (pools.rs:9570-9573), which waits on a distributed read lock for up to `get_lock_acquire_timeout()` (5 s default, 30 s in the reporter's config) plus a pool.bin fan-out read; decommission/rebalance metadata saves hold it across multi-pool CAS. Under write load a probe sample lands inside that window, the node is reported not-ready with the sticky-block reason, and cluster-write readiness caches that sample for `DEFAULT_HEALTH_READINESS_CACHE_TTL_MS` (1 s). Before this PR the probe waited and reported the truth. `grep -rn pool_metadata_check_timeout` returns only `docs/architecture/readiness-matrix.md:45` and `docs/operations/pool-metadata-recovery.md:52`; the emitted reason string is `pool_meta_write_blocked` (`rustfs/src/shared_types.rs:65`), not `pool_metadata_blocked` as both docs state. The PR body's "36 readiness tests passed including sanitized probe timeout" cannot cover this: those tests are unmodified and never see the new timeout. Fix: make readiness call `pool_meta_write_status()` and distinguish `Error::Timeout` from `err.pool_metadata_failure().is_some()`; on timeout keep the last known state (or add the reason the docs promise) and do not set `pool_meta_write_blocked`; correct both docs to the real reason strings; add a readiness test that holds `pool_meta_save_gate` and asserts the report is not `pool_meta_write_blocked`.

**2. Recovery holds the process-wide `pool_meta_save_gate` across the distributed lock acquisition and the whole repair.** pools.rs:10718 takes the gate and keeps it through `get_write_lock(get_lock_acquire_timeout())` (:10745, up to 5 s), the multi-pool identity/pool.bin reads, `repair_pool_meta_transaction` (per-pool `recovery_cas`, identity persist, full reread) and publication, bounded only by the 30 s budget in `store/init.rs:775-777`. Every multi-pool mutation queues on that mutex before it can fail fast (pools.rs:9570 precedes the `ensure_write_safe` check). While a block is installed, PUT/DELETE therefore go from a fast 503 to a stall of up to 30 s on every recovery attempt, repeated on the 5 s → 60 s cycle for the life of the block; the same window holds `data_movement_operation_gate.write()`, so bucket-heal `try_read` fails and scanner publication admission blocks. Fix: acquire the namespace write lock before the save gate, or publish a recovery-in-progress flag readable without the mutex so admission probes fail fast, and add a test that measures mutation latency while recovery is mid-flight.

### Scope gaps against the PR's own claim (major, resolve or re-word)

**3. Two runtime paths still turn a transient fault into a permanent, unrecoverable `write_blocked`.** The worker refuses any `write_blocked` state (pools.rs:10690-10693, :10721). (a) Format heal: `store/heal.rs:268-275`, :295-300, :358-365, :369-376 call `block_writes_after_fence_loss()` whenever `is_lock_lost()` fires — a lease refresh blip during an admin/background format heal latches every later multi-pool mutation until restart. (b) Decommission cancel: `decommission_cancel_transaction` saves through `save_no_lock_with_fence` (pools.rs:6889-6904), which uses a throwaway `PoolMetaWriteState::default()` so the arm's latch goes nowhere, and then the caller does `save_guard.block_writes(); return Err(err);` on **any** error (pools.rs:11606-11624) — including the `ReadUnavailable` preflight failure and the first-CAS `PreconditionFailed` this PR defines as retryable — installing `RecoveryRequired/"validation"` on the shared gate and returning an untyped error (500, not 503). The docs acknowledge (a) as out of scope but the excluded condition is transient, so the headline "a single timed-out read no longer bricks the node" is only true for the object-mutation entry point. Either classify a lost lease / cancel-save failure with no dispatched CAS as retryable, or state both residual latches explicitly in the PR Impact section.

**4. The background recovery worker has no test.** `store/init.rs:769-793` is the only production caller of `recover_pool_meta_transaction`; every test calls it directly. Deleting the `tokio::spawn` block leaves the suite green, yet "recovers without a restart" lives only there. Nothing covers cancellation on `recovery_rx`, the 30 s timeout → `Error::Timeout` mapping, the 5 s reset / doubling / 60 s cap, or that `record_pool_meta_recovery_failure` is reached. Fix: extract the tick into a testable `async fn` returning the next delay and unit-test the three transitions, plus one integration test that latches a real store and polls `pool_meta_writes_ready()` until the worker flips it.

### Minor

- **5. Stale `transaction_arm.source` misattributes an abort.** pools.rs:5697-5698 sets `source` on every `Err` (including `PreconditionFailed`, which then restores `previous_phase`) and never clears it on `Ok`. Sequence: `prepare_cas` pool0 → `PreconditionFailed`; retry `prepare_cas` pool0 → `Ok`; future dropped → `Drop` records `TransactionUnknown{phase:"prepare_cas", source: PreconditionFailed}`, blaming a precondition rejection for an unknown-outcome abort. Set `source = None` on `Ok` and skip it for `PreconditionFailed`.
- **6. Typed context flattened at two save sites.** pools.rs:11062-11068 wraps the decommission save error with `Error::other(format!(...{err}))`, so `pool_metadata_failure()` returns `None` downstream and the 503 mapping and the recovery classifier both fall to the generic arm; `peer_s3_client.rs:793` (`narrow_to_disk().unwrap_or_else(Error::other)`) does the same on the bucket-heal gate. Use `Error::other_with_context(...)`, which is chain-preserving.
- **7. Write-path tests exercise the identity read, not pool.bin.** On a real store `expected_cluster_id` is always set, so `acquire_pool_meta_write_guard_with_lock_error` fails at `load_pool_meta_identity_observing` (:9391) before reaching the pool.bin read (:9404). Both `pool_meta_writer_retries_after_identity_read_failure_without_restart` and `pool_metadata_preflight_recovery_preserves_single_and_multi_pool_public_mutations` assert only `kind == ReadUnavailable`; the write-path pool.bin branch is covered solely by the unit test on `select_pool_meta_replicas_observing`. Pin `phase` in both and add a case with identity readable and pool.bin failing (`phase == "pool_read"`). Also untested: the topology-mismatch refusal (:10763-10775), an `Incompatible` (newer-version) replica in `pool_meta_recovery_rejects_missing_corrupt_and_incompatible_replicas_without_writes` (its fixtures decode only to Missing/Corrupt), and any `legacy_cas` phase assertion.
- **8. Destructive cancellation before the refusal gates.** pools.rs:10705-10714 cancels every active decommission worker before the rebalance-quiescence, corruption and topology checks, so a cluster whose recovery is permanently refused still loses its workers on the first attempt (one-shot, since the watchdog is gated on `ensure_pool_meta_side_effects_safe`). Move the quiescence check ahead of the cancellation. The backoff is also unjittered; latched nodes in a cluster-wide blip contend for the same `pool.bin` write lock on the same schedule.
- **9. Simplicity.** `PoolMetaWriteState` gained five fields including two `Arc<std::sync::Mutex<..>>`; `PoisonError::into_inner` sites went 5 → 14; `recover_pool_meta_transaction` builds `candidate` by struct-update and then hand-relinks four Arcs (:10779-10782) before `*state = candidate` — a sixth shared field later needs a fifth line or an outstanding `Drop` writes into an orphaned latch. Smaller: one `Arc<Mutex<PoolMetaLatch { aborted, failure, blocked_since }>>` plus the counter; recovery's clear becomes one `*latch = default()`. Also: `operation` vs `phase` render the same token twice when `block_pool_meta_validation` returns without a caller override (`"pool_selection: ... (recovery_required, pool_selection)"`); `block_writes()` survives as an eight-caller wrapper that erases the cause it should record; the identity-`Unreadable` scan closure is written out three times (:5534, :6337, :6759).

### Nits

- `PoolMetadataError`'s `#[error(...)]` interpolates `operation` and the full source `Display` (error/mod.rs:56); that is exactly the high-cardinality `Io` message the `::other(format!` ratchet exists to keep out of quorum bucketing. Not live today (these errors do not reach `reduce_errs`), but render only `kind`/`phase` and leave detail to `#[source]`. The ratchet decrease itself (182→181, 12→10) is correct and shrink-only.
- pools.rs:5232 passes the literal `"pool_read"` as the write-path `operation`; the read-probe path threads the real one.
- `!repair_write_safe` branches at :4585-4587, :5728-5733, :7024-7029 are now unreachable at runtime (every `Unreadable` short-circuits earlier); the `ensure_write_safe` fallback `FenceLost/"external_fence"` (:4631) is likewise unreachable but would mislabel — a `debug_assert!` or an `Unknown` kind is more honest.
- A peer receiving a `StorageError::Io(PoolMetadataError)` over RPC gets a generic `Io` code and would map to 500; no RPC surface reaches a pool-metadata write today, but worth one line in Impact.

### Lens verdicts

- **Correctness:** findings 1, 3, 5, 6. Null on: phase-restore matrix across `legacy_cas`/`prepare_cas`/`commit_cas`/`identity_cas`/`recovery_cas`; `ReadUnavailable`/`TransactionUnknown` cannot become success or absent on any traced caller; `block_with_context` upgrade branch cannot demote a kind; first-transition log fires once per episode and again after a recovery; error chain survives `Clone` and `other_with_context`.
- **Concurrency/durability:** finding 2, 8. Null on: global lock order `start_gate < movement < save_gate < rebalance_meta < ns(pool.bin) < ns(rebalance.bin) < pool_meta` holds across recovery, bucket heal, `refresh_pool_status_meta`, `save_current_pool_meta`, decommission start, rebalance, format heal — no ABBA; the 30 s drop points leave the original block installed and the repair CAS-idempotent, and `block_started_at` is shared so a dropped attempt does not re-log; `ensure_held()` before and after publication with no await between disarm and clear; `active_transactions` cannot race a live arm; decommission resume path exists behind `ensure_pool_meta_side_effects_safe`.
- **Performance:** null. OK path unchanged (one bool + one atomic load); `Arc<Error>` replaces a formatted `String` per replica; healthy worker tick is one uncontended `try_lock`; logging is once per transition, recovery warn ≤ 1/min. The real cost is finding 2.
- **Test coverage:** findings 4, 7. Null on: startup fail-closed coverage intact (`test_store_init_pool_meta_stays_blocked_after_all_replicas_were_unreadable` unchanged); the fence-loss expectation flip is justified (`!storage.wrote` asserted immediately before); the 503 test's extra `io::Error` wrap is redundant, not wrong — the production single-wrap shape resolves through `pool_metadata_failure()` and `store/heal.rs:2360-2368` asserts it on a real error; polling loops use readiness polling with caps, no fixed sleeps.
- **Simplicity:** finding 9.
- **Compatibility:** finding 1 (docs vs code). Null on: 503 is consistent with `StorageFull`/write-quorum/lock mappings; no e2e or test asserted the old 500 text; startup start/refuse behavior preserved; `StorageErrorCode::Io` wire mapping unchanged; `#[cfg(test)]` façade re-exports consistent with the existing pattern.
- **Security:** null. Public message is the fixed `ServiceUnavailable` constant, source never reaches `<Message>`; no `{:?}`/source logged per request; `recover_pool_meta_transaction` is `pub(crate)` with no admin route or RPC caller; every repair write is preceded by identity/corruption/absence/topology fail-closed checks under the distributed fence.

### Local verification (head `b5a05bd74`)

- `cargo test --locked -p rustfs-ecstore --lib pool_meta --no-fail-fast`: 129 passed.
- `cargo fmt --all --check`, `git diff --check`, `check_logging_guardrails.sh`, `check_doc_paths.sh`, `check_layer_dependencies.sh`, `check_architecture_migration_rules.sh`: pass.
- `cargo clippy --locked -p rustfs-ecstore --lib --tests --no-deps -- -D warnings`: clean.
- `cargo test --locked -p rustfs --lib -- pool_metadata server::readiness::tests`: 38 passed (note: these readiness tests are unchanged by the PR and do not exercise the new 100 ms timeout, see finding 1).
- CI at review time: fail=4 pass=16; failing: Test and Lint,Alert on scheduled failure,Cancel siblings after Test and Lint failure,Test and Lint (rio-v2).

overtrue:
CI attribution for the red `Test and Lint` lane on `b5a05bd74`: the only failing test is `rustfs-scanner scanner_io::tests::scoped_scan_production_entry_preserves_deep_and_full_maintenance_work` (`crates/scanner/src/scanner_io/tests.rs:428`). The same single test fails on `main` at this PR's base commit `cc91483fa` (run 34121340369, 17463 passed / 1 failed), and this PR does not touch the scanner crate, so it is a pre-existing baseline failure, not introduced here. `Test and Lint (rio-v2)` shows as failed only because "Cancel siblings after Test and Lint failure" cancelled it; on `main` that lane is green. Rebasing once `main` is fixed should clear both.

houseme:
CI triage update:

The current failing `Test and Lint` result is not caused by the pool metadata recovery change. The same focused scanner test fails on current `main` with `left: 1, right: 2`.

I isolated that as a scanner test-oracle issue and opened rustfs/rustfs#7404. That PR keeps production code unchanged and aligns the oracle so the ordinary dirty scoped cycle preserves the unselected cold bucket baseline, while Deep/full maintenance cycles still require full cold-bucket coverage.

Local validation for #7404:
- PASS: `cargo test --locked -q -p rustfs-scanner --lib scoped_scan_production_entry_preserves_deep_and_full_maintenance_work -j4`
- PASS: `cargo test --locked -q -p rustfs-scanner --lib scoped_set_scan_reuses_unselected_buckets_with_matching_incarnations -j4`
- PASS: `cargo fmt --package rustfs-scanner --check`
- PASS: `git diff --check`

Once #7404 is in `main`, this PR should be rebased/rerun so remaining CI reflects the pool metadata recovery patch itself.

houseme:
Update: I reproduced the current `Test and Lint` failure locally and confirmed it is not caused by this PR's pool metadata diff.

The failing test is `rustfs-scanner::scanner_io::tests::scoped_scan_production_entry_preserves_deep_and_full_maintenance_work`. It also fails on current `main` with the same `cold-bucket` count assertion. I opened #7405 as a separate test-invariant fix so this PR can stay scoped to pool metadata recovery.

Once #7405 lands, this branch should be rebased or rerun against the updated base to verify whether any pool-metadata-specific CI failures remain.

houseme:
Correction: #7405 has been closed as a duplicate. The canonical fix for the current-main scanner scoped maintenance oracle failure is now #7404.

#7404 was updated with the stricter fixture-based correction and the scoped scanner validation, so this PR should wait for #7404 to land before rerunning or rebasing the pool metadata recovery checks.

PR #7405, opened 2026-09-07
test(scanner): separate dirty scope and maintenance walks

houseme:
## Related Issues
rustfs/backlog#2240
rustfs/backlog#2268

## Summary of Changes
The scoped scanner regression test mutated `cold-bucket` during the dirty-only cycle while also asserting that the production resolver should select only `hot-bucket`. That made the test expect a fresh cold storage walk in a cycle whose purpose is to prove dirty-scope baseline reuse.

This updates the fixture so the dirty-only cycle mutates the selected hot bucket, while the later Deep and full-maintenance cycles mutate cold storage and still prove that maintenance semantics force a full walk instead of reusing the stale cold baseline.

## Verification
Validated commit: 346ba4503 with a clean worktree.

- `cargo test --locked -q -p rustfs-scanner --lib scoped_scan_production_entry_preserves_deep_and_full_maintenance_work -j4` passed: the previously failing regression now matches the scanner invariant.
- `cargo test --locked -q -p rustfs-scanner --lib scoped_scan_ -j4` passed: 14 scoped scanner tests passed.
- `git diff --check` passed.

Remaining risk: this is a test-only fix. It restores CI coverage for the existing scoped scanner behavior but does not add distributed, mixed-version, or long-running scanner/heal evidence.

## Impact
No runtime, API, configuration, compatibility, or deployment impact.

## Additional Notes
This keeps dirty-scope performance coverage separate from Deep/full maintenance correctness coverage so future failures point to the intended invariant.

### DIFF
--- a/crates/scanner/src/scanner_io/tests.rs
+++ b/crates/scanner/src/scanner_io/tests.rs
@@ -365,14 +365,16 @@ async fn scoped_scan_production_entry_preserves_deep_and_full_maintenance_work()
     .enumerate()
     {
         if index > 0 {
+            let mutated_bucket = if index == 1 { "hot-bucket" } else { "cold-bucket" };
             let mut reader = ScannerPutObjReader::from_vec(b"maintenance".to_vec());
             store.pools[0].disk_set[0]
-                .put_object("cold-bucket", &format!("added-{index}"), &mut reader, &ScannerObjectOptions::default())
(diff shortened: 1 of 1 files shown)


houseme:
Closing as a duplicate of #7404.

#7404 was opened first for the same current-main scanner scoped maintenance oracle failure. I updated #7404 with the stricter fixture-based fix and matching validation, so keeping this second PR would only add noise.

PR #2378, opened 2026-04-03
feat(admin): support config KV and history APIs

weisd:
## Type of Change
- [x] New Feature
- [x] Bug Fix
- [ ] Documentation
- [ ] Performance Improvement
- [x] Test/CI
- [ ] Refactor
- [ ] Other:

## Related Issues
#https://github.com/rustfs/backlog/issues/608

## Summary of Changes
- Added admin config KV, config history, and full config APIs for the RustFS admin surface.
- Added config validation, runtime snapshot refresh, dynamic reload, and peer snapshot refresh handling for config updates.
- Aligned `mc admin config` compatibility for help/get output, env override visibility, history restore, and empty-body restore requests.
- Added route, contract, unit, and compatibility-focused tests for the new admin config flows.

## Checklist
- [x] I have read and followed the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines
- [x] Passed `make pre-commit`
- [x] Added/updated necessary tests
- [ ] Documentation updated (if needed)
- [ ] CI/CD passed (if applicable)

## Impact
- [ ] Breaking change (compatibility)
- [ ] Requires doc/config/deployment update
- [x] Other impact:
  Improves RustFS admin config compatibility and runtime reload behavior for existing client workflows.

## Additional Notes
Verification commands run:
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- `make pre-commit`
- `cargo test -p rustfs admin_restore_config_history_without_content_length_is_normalized -- --nocapture`

Manual validation performed:
- Verified `mc admin config get`, `set`, `reset`, `history`, `restore`, `export`, and `import` against a local RustFS instance.
- Confirmed `mc admin config restore` now succeeds for empty-body chunked PUT requests and restores the expected config value.

---

Thank you for your contribution! Please ensure your PR follows the community standards ([CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)) and sign the CLA if this is your first contribution.

PR #7472, opened 2026-09-08
fix(ecstore): preserve online writes during pool retirement

cxymds:
## Related Issues

Refs https://github.com/rustfs/backlog/issues/2377, https://github.com/rustfs/backlog/issues/2379 and https://github.com/rustfs/backlog/issues/2380. The companion harness fixes for #2378/#2379 are in https://github.com/rustfs/rustfs-release-validation/pull/11. These issues remain open pending merged, full-fleet validation.

## Summary of Changes

- Reconcile an identical native scanner backlog replica after cross-pool PreconditionFailed despite replica-local mtime differences. Restrict the exception to the exact internal key, known unversioned non-delete identity and unchanged payload/metadata; retain source revalidation and durable pending-capacity settlement.
- Route native scanner CAS through a storage-owned replica writer. A stale node must not write a source that another node has durably retired. Fence the actual replica set, including non-hash-routed replicas, with fixed-object → durable pool metadata → replica namespace ordering. Preserve failed/canceled native membership repair, healthy reserved-target writes and CAS on every surviving set.
- Retain the native writer's outer fences in an owned task until the full rename tail drains, including after caller cancellation. Propagate lock-loss signals to the set commit. No change to write quorum or fsync policy.
- Preserve typed InvalidArgument and actionable reasons for deterministic Pool request rejections. Active conflicts, storage/quorum failures and fleet-proof retry markers keep their existing behavior. Tighten the legacy-error baseline rather than relaxing the gate.
- Fence late ordinary writes to suspended sources; allow mixed batch deletion against actual marker destinations while retaining exact-version release. Preserve the separately contributed existing-multipart drain behavior in `0e487e35c804d93bff637cad0a8ee41ea528ea87`; new uploads remain excluded from retiring sources.
- Keep healthy reserved targets open to foreground writes under the selected MinIO shared-capacity policy. Reservations budget migration/recovery, not an exclusive foreground quota. Preserve owner/pending accounting, source-cleanup identity checks and conservative repair admission.

## Verification

CI follow-up tested commit: `daabee319d0efb1ed46b4c4059262e982b5e75d8`. The default and rio-v2 [CI jobs](https://github.com/rustfs/rustfs/actions/runs/34212597433) failed the same two suspended DELETE initialization tests: they still expected a healthy reserved target to reject foreground deletion with SlowDown. Those two tests were not included in the earlier 484-test selection. Both original failures reproduced locally before editing.

The follow-up is test-only: a successful single/batch DELETE must publish one null marker on the healthy target, leave the retiring source's original null version, timestamp and full bytes untouched before the worker, then converge source cleanup without changing the target marker generation. Public storage GET and HEAD must return ObjectNotFound both before and after the real entry worker runs. Batch native null identities accept None/nil UUID, not a non-null version.

Final verification on that commit's exact source tree:

```sh
cargo nextest run --profile ci -p rustfs-ecstore --lib -E 'test(/^store::init::tests::(suspended_(batch_)?delete_marker_then_decommission_worker_converges_null_source|versioned_(batch_delete_marker_skips_decommission_source|delete_marker_survives_decommission_source_cleanup))$/) | test(/^data_movement::tests::test_precondition_conflict_/)'
# Repeat with --features rio-v2.
```

Default: **10/10 passed** (2.964 s). rio-v2: **10/10 passed** (2.957 s), covering the same 10 distinct tests in both configurations. These include the two repaired tests, two existing versioned DELETE regressions and six target-conflict positive/negative controls. Native macOS Rust 1.98.0 / nextest 0.9.140; the unchanged ci-profile setup supplied a 4 MiB test stack, with no manual stack override, retries or policy changes. `cargo fmt --all --check` and `git diff --check` passed. The follow-up changes one test file (+77/-23), diff SHA-256 `9a6520e78864bb9a13acda4c5cb5a1c4897e3b80032a674fba654220f4c4b7e2`. This is not Linux io_uring or live-VM validation; the new GitHub CI run must still pass.

Earlier product verification at commit `97a406b41ef330254fb6f61712fff2bd8221953a`, identical to the final pre-commit source tree. The latest scanner-fence diff has SHA-256 `f073122452a99e056916ecf779e7b2f8f3a602935c293c90355c53c1be8fc62a`. All **484 distinct scoped tests passed**, without ignored tests:

1. ECStore: `cargo test -p rustfs-ecstore --features test-util --lib <filter> --quiet`, using `core::pools::pools_tests::` (300), `core::pools_test::decommission_lock_order_tests::` (45, 56.46 s), `data_movement::` (106), and each of `replacement_format_heal_coexists_with_capacity_read_fence`, `targeted_heal_is_blocked_by_exact_fit_decommission_reservation`, `pool_meta_heal_keeps_per_target_capacity_admission` (one each). The new native tests cover stale-source CAS, each actual target set, stale revisions, invalid indices, canceled membership repair, caller cancellation through quorum/tail, both outer fences and injected metadata-lease loss. Original conflict recovery, live PUT/MPU, mixed deletion and capacity-loss regressions remain green.
2. Scanner: `cargo test -p rustfs-scanner --lib <filter> --quiet`, using `scanner::backlog::` (27), `native_backlog_replica_writes_preserve_cas_across_writer_restart` (1), `restarted_main_loop_completes_durable_pause_backlog_catch_up` (1), and `running_main_loop_catches_up_pause_cleared_after_startup_observe` (1). These include real controller CAS/restart and single-/multi-pool main-loop recovery on temporary disks, not just fabricated ledger identity structs.
3. Red/green: the original native direct-set CAS returned success after another node retired its source; the new admission rejects it and preserves the source bytes. Keeping TailDrained but removing only the owned task makes the cancellation regression fail when its capacity fence releases early; restoring ownership passes. Earlier identity, typed-error, late ordinary-source PUT, all-pool batch and blanket healthy-target rejection negative controls remain documented. `cargo fmt --all --check`, `git diff --check`, `scripts/check_doc_paths.sh`, `scripts/check_architecture_migration_rules.sh`, `scripts/check_no_planning_docs.sh` and `scripts/check_error_other_format_ratchet.sh` passed; ratchet count remains 586.

Native macOS Rust 1.98.0, `CARGO_BUILD_JOBS=2`, `RUST_MIN_STACK=67108864`; existing large unwind-table linker warnings remain. The branch merge base is main `72e85210f136da936b27f83faf4a4f0054f1bb32`. Main advanced during this work to `7e0c67111b97703d47e23719b0264a739c8acea8`; the read-only merge-tree check is conflict-free, but the local results above describe the tested head, not an untested combined tree. CI must validate integration with the latest base.

Not verified: patched Linux release artifact, jump-host/Azure VM 13-case acceptance, or large-fleet performance. Historical VM snapshots lack conflict-time identities; controlled local reproductions do not establish the unique cause of every historical failure. Test capacity and lease failures are injected, not physical disk exhaustion or a real network partition.

## Impact

The CI follow-up changes no production code, wire/disk format, workflow, ignore list or retry policy. Mechanical final-diff review: correctness covers response, source bytes/metadata, target marker and public reads before/after convergence; simplicity reuses the existing fixture, worker and cleanup assertion with one shared test-only marker assertion. No unresolved findings in the follow-up.

No new disk/RPC format, dependency, configuration flag, or quorum exception. Scanner native membership repair on failed/canceled sources is distinct from new business writes, which remain blocked until clear. Healthy capacity is shared, not guaranteed foreground priority or exclusive migration space. Existing capacity-blocked/recovery and canceled queuedBuckets semantics remain.

The native scanner writer now performs background metadata admission and uses an owned task per replica, with serialization on its fixed internal key. It does not reroute ordinary PUT/GET, but this is additional background I/O/locking; no scale or throughput benchmark is claimed. Older scanner writers still use direct set CAS, and older target nodes may retain blanket foreground rejection; deploy all scanner-capable nodes before treating these protections as fleet-wide.

The product-diff review at `97a406b41` used two fresh sequential passes. Correctness: exact-key/set/index/state boundaries, unchanged CAS and source retention; no unresolved findings. Simplicity: reuse storage admission, namespace guards and existing replica protocol without a new ledger or schema; no unresolved findings. Test coverage: falsifiable native red/green, independently probed fences, lease loss and real scanner restart/main-loop recovery; subsequent CI exposed the two incompatible initialization-test oracles now addressed above. Concurrency/durability: fixed → metadata → actual replica ordering, lock-loss propagation, owned full-tail cancellation and unchanged membership recovery/quorum; no unresolved findings. Compatibility: existing MPU contribution, terminal native-repair behavior, old data formats and full-fleet rollout caveat retained; no unresolved findings. Performance: additional background work and serialization explicitly recorded; no measured regression or unsupported no-impact claim.

## Additional Notes

The user selected MinIO-aligned shared foreground/migration capacity, superseding the earlier exclusive-reservation/new-ordinary-ledger proposal. Behavioral reference: [pinned MinIO decommission documentation](https://github.com/minio/minio/blob/7aac2a2c5b7c882e68c1ce017d8256be2feea27f/docs/distributed/DECOMMISSION.md). This does not claim identical admin terminal-state/retry semantics. Current work is issue code repair; no new live deployment was launched.

Rollback: revert the relevant commits; no metadata migration is required. Reverting scanner-fence commit `97a406b41` restores direct native scanner CAS and its stale-source/cancellation gaps; reverting the test-only follow-up restores the incompatible DELETE expectations. Do not close the linked issues solely from local results; final acceptance still requires a release-based candidate and the batch-frozen latest released rc on both existing nodes through the jump host.

### DIFF
--- a/crates/ecstore/src/core/pools.rs
+++ b/crates/ecstore/src/core/pools.rs
@@ -840,15 +840,21 @@ fn is_decommission_start_active_pool(pool: &PoolStatus) -> bool {
     decommission_start_pool_state(Some(pool)) == DecommissionStartPoolState::Active
 }
 
+fn invalid_decommission_request(reason: impl Into<String>) -> Error {
+    Error::InvalidArgument("decommission".to_string(), "pool-state".to_string(), reason.into())
+}
+
 fn ensure_decommission_start_allowed(state: DecommissionStartPoolState) -> Result<()> {
     match state {
-        DecommissionStartPoolState::Missing => Err(Error::other("failed to start decommission: target pool was not found")),
+        DecommissionStartPoolState::Missing => {
+            Err(invalid_decommission_request("failed to start decommission: target pool was not found"))
+        }
         DecommissionStartPoolState::Active | DecommissionStartPoolState::Retryable => Ok(()),
         DecommissionStartPoolState::Decommissioning => Err(StorageError::DecommissionAlreadyRunning),
-        DecommissionStartPoolState::Decommissioned => {
-            Err(Error::other("failed to start decommission: target pool is already decommissioned"))
-        }
-        DecommissionStartPoolState::Blocked => Err(Error::other(
+        DecommissionStartPoolState::Decommissioned => Err(invalid_decommission_request(
+            "failed to start decommission: target pool is already decommissioned",
+        )),
+        DecommissionStartPoolState::Blocked => Err(invalid_decommission_request(
             "failed to start decommission: target pool decommission is blocked; clear failed or canceled metadata before starting again",
         )),
     }
@@ -865,7 +871,7 @@ fn ensure_decommission_start_keeps_active_pool(meta: &PoolMeta, indices: &[usize
         .filter(|idx| meta.pools.get(**idx).is_some_and(is_decommission_start_active_pool))
         .count();
     if active_count.saturating_sub(active_target_count) == 0 {
-        return Err(Error::other(
+        return Err(invalid_decommission_request(
             "failed to start decommission: at least one active pool must remain after decommission start",
         ));
     }
@@ -4268,7 +4274,7 @@ fn should_retry_decommission_cancel_reload(changed: bool, already_canceled: bool
 
 fn ensure_decommission_cancel_allowed(pool_present: bool, decommission_present: bool, terminal: bool) -> Result<()> {
     if !pool_present {
-        return Err(Error::other("failed to cancel decommission: target pool was not found"));
+        return Err(invalid_decommission_request("failed to cancel decommission: target pool was not found"));
     }
 
     if !decommission_present || terminal {
@@ -4287,7 +4293,7 @@ fn ensure_decommission_clear_allowed(
     unresolved_entries: usize,
 ) -> Result<()> {
     if !pool_present {
-        return Err(Error::other("failed to clear decommission: target pool was not found"));
+        return Err(invalid_decommission_request("failed to clear decommission: target pool was not found"));
     }
 
     if !decommission_present {
@@ -4303,7 +4309,7 @@ fn ensure_decommission_clear_allowed(
     }
 
     if unresolved_entries > 0 {
-        return Err(Error::other(format!(
+        return Err(invalid_decommission_request(format!(
             "failed to clear decommission: {unresolved_entries} unresolved listing entries must be reconciled by retrying decommission"
         )));
     }
@@ -4313,7 +4319,7 @@ fn ensure_decommission_clear_allowed(
 
 fn ensure_decommission_terminal_operation_supported(single_pool: bool, operation: &str) -> Result<()> {
     if single_pool {
-        return Err(Error::other(format!(
+        return Err(invalid_decommission_request(format!(
             "failed to {operation}: single pool deployments do not support decommission"
         )));
     }
@@ -4323,7 +4329,9 @@ fn ensure_decommission_terminal_operation_supported(single_pool: bool, operation
 
 fn validate_start_decommission_request(indices: &[usize], single_pool: bool) -> Result<()> {
     if indices.is_empty() {
-        return Err(Error::other("failed to start decommission: no target pools were provided"));
+        return Err(invalid_decommission_request(
+            "failed to start decommission: no target pools were provided",
+        ));
     }
 
     ensure_decommission_terminal_operation_supported(single_pool, "start decommission")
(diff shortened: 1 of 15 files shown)


overtrue:
The rio-v2 lane exposed a real contract regression at `store::init::tests::suspended_decommission_source_multipart_remains_operable_until_drained`: `put_object_part` now returns `SlowDown`. An active suspended decommission source must remain writable until drain completes; the new `meta.is_suspended(target_pool_index)` mutation fence is too broad. Please narrow it to terminal/retired pools and keep this existing test unchanged.

overtrue:
Fixed in `0e487e35c`: existing multipart writes are admitted only while the source is actively draining; terminal suspended pools still reject them. Both the rio-v2 regression test and the focused admission matrix pass unchanged.

PR #7508, opened 2026-09-08
fix(targets): reject trailing batch items

cuishuang:
## Related Issues

N/A

## Summary of Changes

Reject persisted queue batches whose payload contains more JSON items than the `item_count` encoded in the filename.

`QueueStore::get_multiple()` previously deserialized exactly `item_count` items and returned successfully without checking whether the stream contained additional items. If a three-item payload was stored under a filename claiming two items, the third event could be silently ignored. A caller could then delete the queue entry and permanently lose that event.

The reader now validates the stream after the expected items have been deserialized:

- trailing whitespace is accepted;
- additional valid JSON items return `StoreError::Deserialization`;
- malformed trailing data also returns `StoreError::Deserialization`.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rustfs-targets store::tests` 35 tests passed
- Includes the existing truncated-batch regression test and the new trailing-item regression test
- `cargo clippy -p rustfs-targets --all-targets -- -D warnings`  Passed

## Impact

Corrupted or mismatched queue entries are now rejected instead of being treated as successfully delivered.

Valid queue entries are unaffected. Trailing whitespace after the expected JSON objects remains accepted.

## Additional Notes

This complements the existing protection against truncated batches, ensuring that both fewer-than-advertised and more-than-advertised payloads are rejected.

### DIFF
--- a/crates/targets/src/store.rs
+++ b/crates/targets/src/store.rs
@@ -919,6 +919,22 @@ where
             }
         }
 
+        // The filename's item count is untrusted. Reject a payload that contains
+        // more items than advertised instead of returning success and allowing the
+        // caller to delete the entry with trailing events still in the file.
+        match deserializer.next() {
+            None => {}
+            Some(Ok(_)) => {
+                return Err(StoreError::Deserialization(format!(
+                    "Batch for key {key} contains more than {} items",
+                    key.item_count
+                )));
+            }
+            Some(Err(e)) => {
+                return Err(StoreError::Deserialization(format!("Failed to deserialize trailing batch item: {e}")));
+            }
+        }
+
         if items.is_empty() && key.item_count > 0 {
             return Err(StoreError::Deserialization("No items found".to_string()));
         }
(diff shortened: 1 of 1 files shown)


houseme (approved):

houseme (approved):
Reviewed the exact diff from base `73957d0f` to head `58595a43`.

No findings.

I traced `QueueStore::get_multiple()` through the persisted batch reader, filename count parsing, raw replay/delete boundaries, and the adjacent truncated/huge-count tests. The new stream check runs before returning `Ok`, accepts trailing whitespace through serde_json stream EOF handling, rejects both extra JSON values and malformed trailing data, and keeps the queue entry recoverable because the delete path is only reached after a successful read/delivery path.

Risk lenses:

- Correctness: checked exact count, truncated batch, advertised-too-low batch, huge advertised count, and single-item `get()` behavior; I did not find a remaining partial-success path in `get_multiple()`.
- Simplicity: the change is local, uses the existing stream deserializer, and does not add a new abstraction or widen ownership boundaries.
- Test coverage: the new trailing-item regression fails if the added EOF/extra-item check is removed; existing truncated and huge-count tests cover the adjacent boundaries.
- Concurrency/durability: no new locks, awaits, writes, or delete-before-validation path; failed reads leave the persisted entry on disk for retry or inspection.
- Compatibility: valid concatenated JSON entries and trailing whitespace remain accepted; corrupt/mismatched persisted entries now fail closed as intended.

Local checks I ran:

- `cargo test -p rustfs-targets store::tests` (35 passed; integration test binaries compiled with 0 selected by the filter)
- `cargo fmt --all --check`
- `git diff --check`

GitHub checks at review time: CLA, Quick Checks, Typos, and io_uring are passing. Broader CI still has pending jobs, including Build RustFS Debug Binary, Test and Lint, ILM Integration, and rio-v2 lanes, so this approval is for the code review and focused local validation, not a claim that full CI has completed.

PR #7608, opened 2026-09-09
test: report tier reference proof setup failures without hanging

overtrue:
## Related Issues

Related to rustfs/backlog#2400.

## Summary of Changes

The reference-proof regression waited only for a test barrier. If the update exited before reaching it, the test could hang indefinitely; release acceptance observed it still waiting after 16,920 seconds.

Wait for the barrier, the update result, or a bounded deadline. Report early failures with their original result, and release the artificial barrier on timeout. Production code and the fence-revision assertions are unchanged.

## Verification

At 212950fe5, both focused tests passed with retries disabled: the original fence-revision test and a new real update-path test that injects an initialization failure before the barrier. Formatting and whitespace checks passed. The exact cause of the historical hang was not captured; the new test reproduces the unobserved early-exit path.

## Impact

Test diagnostics only. No production behavior changes.

## Additional Notes

Reverting removes the bounded wait and early-failure diagnostic.

### DIFF
--- a/crates/ecstore/src/services/tier/tier.rs
+++ b/crates/ecstore/src/services/tier/tier.rs
@@ -17579,6 +17579,66 @@ mod tests {
         assert!(current.tiers.contains_key("COLD-B"));
     }
 
+    async fn wait_for_reference_proof_barrier(
+        barrier: &TierDriverBuildBarrier,
+        update: &mut tokio::task::JoinHandle<std::result::Result<(), TierConfigUpdateError>>,
+    ) -> std::result::Result<(), String> {
+        tokio::select! {
+            biased;
+            result = &mut *update => Err(format!("tier update exited before the reference proof barrier: {result:?}")),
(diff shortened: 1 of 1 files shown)


houseme:
Rebased this PR onto the latest `release` after the four Scanner/Heal V2 follow-up PRs landed.

Current head: `c37f24f609d42a944bd1142edb1467565e15732b`

Local validation:

- PASS `cargo fmt --all --check`
- PASS `git diff --check origin/release...HEAD`
- PASS `cargo test -p rustfs-ecstore reference_proof_barrier_reports_update_failure_before_arrival --lib`
- PASS `cargo test -p rustfs-ecstore reference_proof_rejects_a_changed_prepared_fence_revision_before_publish --lib`

This remains a test-stability fix: it reports early reference-proof setup failures directly instead of waiting on a test barrier, and it bounds the join after releasing the barrier. No production behavior change is intended.

Issue #2430, opened 2026-04-08
Does rustfs support aioboto3?


bestyjs:
<img width="1441" height="524" alt="Image" src="https://github.com/user-attachments/assets/63500ba5-192f-45f6-82f2-14f8ffcd2464" />

<img width="1378" height="531" alt="Image" src="https://github.com/user-attachments/assets/d349cd90-173b-459a-bb33-c6e41ead6d81" />
I want to ask whether asynchronous operations are supported, or do I have to build the event loop myself?

loverustfs:
We have compatibility with boto3, but we haven't tested aioboto3.

loverustfs:
Fixed. Closing this, but please reopen if you still encounter this issue.

PR #7635, opened 2026-09-10
fix(s3): reject oversize single PUT early and map body errors to 4xx

### LINKED ISSUE #7596: rc.5: single PutObject of a multi-GB object over a slow link fails with 500 partway through
## Describe the bug

On rustfs/rustfs 1.0.0-rc.5 (`40a2470fe`), a single `PutObject` of a multi-GB object over a slow link is closed with HTTP 500 while the client is still sending the body.

This is inbound (a client writing into RustFS). It is not #6825, which is outbound replication of a large object to a generic S3 target.

## Environment

- Tag: `1.0.0-rc.5`
- Commit: [`40a2470feb567201165a5b809b7598bb4b1f68f5`](https://github.com/rustfs/rustfs/tree/40a2470feb567201165a5b809b7598bb4b1f68f5)
- Deploy: single node, data on a virtio volume
- Source of the object body: Wasabi `us-west-1` (`https://s3.us-west-1.wasabisys.com`), read concurrently with the PUT into RustFS

## Evidence

- Object size: 7.68 GB
- Client: one `PutObject` with `Content-Length` set, not multipart
- Inbound rate: about 3 to 5 MB/s (WAN read from Wasabi `us-west-1` into a RustFS node in London)
- Attempt 1: HTTP 500 after 162 seconds, request id `cc486234-bfa2-4e1c-b1c7-b867ae65feaf`
- Attempt 2: HTTP 500 after 571 seconds, request id `df055a6d-b586-4453-8107-2e536a6a7cc4`
- Both responses included an XML error body. The client was still sending when RustFS closed the request.

Server log:

```
{"timestamp":"2026-09-09T01:46:11.007581525Z","level":"ERROR","message":"encode_batched err Custom { kind: UnexpectedEof, error: Custom { kind: Other, error: hyper::Error(Body, Custom { kind: UnexpectedEof, error: IncompleteBody }) } }","target":"rustfs_ecstore::set_disk::ops::object"}
{"timestamp":"2026-09-09T01:55:51.702148096Z","level":"ERROR","message":"encode_batched err Custom { kind: UnexpectedEof, error: Custom { kind: Other, error: BodySizeLimitExceeded { size: 16384, limit: 6389 } } }","target":"rustfs_ecstore::set_disk::ops::object"}
```

## To reproduce

1. Run RustFS 1.0.0-rc.5 (`40a2470fe`) as a single node with data on a virtio volume.
2. From a remote client, `PutObject` a ~7.7 GB object into RustFS as one request with `Content-Length` set (not multipart).
3. Feed the body at WAN speed (a few MB/s), for example by reading it from Wasabi `us-west-1` at the same time.
4. Watch the PUT and the RustFS server log.

## Expected behavior

The PUT completes, or RustFS rejects it up front with a size error.

## Actual

RustFS returns HTTP 500 several minutes in, while the body is still arriving. The second error names a 16384-byte body against a 6389-byte limit, which reads like an internal chunking bug rather than anything about the object size.

A mid-stream 500 on a single PUT forces the client to restart the whole object from offset 0.

## What works

- The same client and object complete against Wasabi.
- 340 MB objects complete against this RustFS on the same path in about 20 seconds.
- Local uploads into this RustFS at about 70 MB/s complete.

The failure shows up on the slow stream.

## Additional context

Large objects currently fail in both directions on rc.5: inbound on this ticket, outbound on #6825.

reatang:
## Related Issues

Fixes #7596. Upstream tracking for the s3s-side early rejection: s3s-project/s3s#811 (PR s3s-project/s3s#812). Partially addresses rustfs/backlog#2430 (the declared-size admission part only); that issue stays open for unknown-length UploadPart rejection, removal of the whole-part buffer, raw-body inactivity detection before s3s decoding, and a typed `RequestTimeout` path.

## Summary of Changes

A single `PutObject` whose declared length exceeds the 5 GiB single-request ceiling was only rejected once the client had already streamed 5 GiB: s3s's `put_object_max_size` is a read-time body budget, and RustFS never compared the declared length against it. When the budget ran out, the resulting `BodySizeLimitExceeded` surfaced from the erasure writer as `500 InternalError`. A request body whose TCP connection hit EOF before `Content-Length` bytes arrived (hyper's own `IncompleteBody`) was also mapped to `500 InternalError`. Both 500s are treated as retryable by SDKs, so a single oversize upload was retried from offset 0 five times in the report.

- `PutObject` and `UploadPart` now reject a declared length above `MAX_SINGLE_PUT_OBJECT_SIZE` (5 GiB) with `400 EntityTooLarge` at the request-header stage, before any body byte is read. The constant moves from a private `server::http` const to `rustfs_config` so the s3s streaming-body limit and the admission check share one value.
- `ApiError` maps s3s `BodySizeLimitExceeded` to `400 EntityTooLarge` and a hyper body EOF (`hyper::Error` whose cause is an `UnexpectedEof` `io::Error`) to `400 IncompleteBody`, across both the `StorageError::Io` and plain `io::Error` conversions. `BodySizeLimitExceeded` lives in s3s's private `http` module, so it is recognised by its `Display` form like the other s3s body-stream errors already handled there; this can switch to a typed downcast once s3s-project/s3s#811 exports it.

The 5 GiB ceiling itself is unchanged (AWS and Wasabi apply the same single-PUT limit; MinIO accepts up to 5 TiB). Raising it or making it configurable is a separate product decision.

## Verification

Tested commit: this branch on top of `main` 8f150d1d8, debug build of `rustfs`, single node, HTTP/1.1 path-style requests with `Content-Length` set.

Failing-before / passing-after against a real server:

| Request | Before (main 8f150d1d8) | After |
|---|---|---|
| single PUT, declared 5.5e9 bytes (5.12 GiB) | `500 InternalError` after streaming 5 GiB; log `encode_batched err ... BodySizeLimitExceeded { size: 65536, limit: 124 }` | `400 EntityTooLarge` in 0.6 s, before the body is read |
| declared 1,000,000 bytes, 300,000 sent, client closes | `500 InternalError`; log `hyper::Error(Body, UnexpectedEof, IncompleteBody)` | `400 IncompleteBody` |
| single PUT, declared 5.2e9 bytes (4.84 GiB, under the cap) | `200 OK` | `200 OK` |

Unit tests (all pass):

```
cargo nextest run -p rustfs --lib -E 'test(/body_size_limit_exceeded_maps/) | test(/hyper_body_eof_maps/) | test(/reject_oversize_single_upload_enforces/) | test(/rejects_oversize_content_length_before/) | test(/rustfs_s3_config_preserves/)'
```

- `error::tests::body_size_limit_exceeded_maps_to_entity_too_large_across_io_boundaries` uses the exact nested `io::Error` shape from the production log and also checks that an unrelated message mentioning a limit stays `InternalError`.
- `error::tests::hyper_body_eof_maps_to_incomplete_body_across_io_boundaries` drives a real hyper HTTP/1 server with a short body to obtain hyper's own error (it has no public constructor) and maps it through both conversions.
- `error::tests::real_s3s_body_size_limit_error_maps_to_entity_too_large` trips the pinned s3s `Body` budget with a four-byte limit to obtain the real `BodySizeLimitExceeded`, checks the display matcher against it, and maps it through both conversions.
- `app::object::put::tests::execute_put_object_rejects_oversize_content_length_before_reading_the_body` uses a poll-counting body with `Content-Length` of 5 GiB + 1 and asserts `EntityTooLarge` with zero body polls.
- `app::object::put::tests::execute_put_object_oversize_admission_uses_decoded_length_for_aws_chunked` sends a `STREAMING-*` aws-chunked request whose framed `Content-Length` is above the cap: a decoded length at the cap is not rejected as oversize, a decoded length of cap + 1 is, with zero body polls.
- `app::multipart_usecase::tests::execute_upload_part_rejects_oversize_declared_part_before_reading_the_body` covers `UploadPart` for cap + 1 (rejected, zero polls), exact cap and zero bytes (pass admission).
- `cargo nextest run -p rustfs --lib -E 'test(/error::tests::/) | ...'` (72 tests) passes.

Also run: `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, `make pre-commit`; `cargo clippy -p rustfs --all-targets -- -D warnings` rerun after the review-driven test additions.

Not run: the full workspace test suite and the e2e lanes; the change is confined to request admission and error classification in the `rustfs` crate plus one constant in `rustfs-config`. Remaining risk: a client-side error whose `Display` starts with `body size ` and contains ` exceeds limit ` would be classified as `EntityTooLarge`; that string is s3s's own and no other error in the chain produces it.

## Impact

- Client-facing: an oversize single `PutObject` or `UploadPart` now fails immediately with `400 EntityTooLarge` instead of `500 InternalError` after streaming 5 GiB, and a client that disconnects mid-body now receives `400 IncompleteBody`. SDKs stop retrying these as server errors.
- No configuration, API, or on-disk change. The 5 GiB limit is unchanged.

## Additional Notes

Review follow-ups deliberately left out of this PR and tracked in rustfs/backlog#2430: the raw-body inactivity guard placed before s3s decoding, the unknown-length `UploadPart` buffer, and the typed `RequestTimeout` path. The display-based `BodySizeLimitExceeded` matcher should become a typed downcast once s3s-project/s3s#812 is in the pinned dependency.

The erasure writer still logs the client-caused body errors at `ERROR` level (`encode_batched err ...`); that is unchanged here and could be lowered separately. The reporter's timing data (5 GiB reached at a claimed 3 to 5 MB/s) is not explained by this change; the issue comment asks for per-attempt byte counts.


cxymds:
Reviewed at `22fb79b6f2d8afd5edba9dea47b28754fcf7587c`.

This addresses the specific #7596 failure mode: a declared single-request size above 5 GiB is rejected before business-layer body ingestion, and the two identified body errors are classified as `EntityTooLarge` / `IncompleteBody` instead of `InternalError`. The ceiling remains unchanged, so a 7.68 GB object still needs multipart upload.

It provides **partial coverage of [rustfs/backlog#2430](https://github.com/rustfs/backlog/issues/2430)**. The [unknown-length whole-part buffer](https://github.com/rustfs/rustfs/blob/22fb79b6f2d8afd5edba9dea47b28754fcf7587c/rustfs/src/app/multipart_usecase.rs#L1222-L1238) and [capped-session-only decoded-body timeout guard](https://github.com/rustfs/rustfs/blob/22fb79b6f2d8afd5edba9dea47b28754fcf7587c/rustfs/src/app/multipart_usecase.rs#L1206-L1219) remain. A client that stops sending without EOF is therefore a different case from the premature EOF handled here; this PR does not add a typed `RequestTimeout` path.

Suggested next steps:

1. **Keep the scope explicit.** Reference backlog #2430 as partially addressed and leave it open for unknown-length rejection, buffer removal, raw-body inactivity detection, and timeout-specific cleanup/retry coverage. Those remaining changes can be delivered separately from this fix.

2. **Add direct UploadPart admission coverage and an AWS streaming boundary case.** The new call in `execute_upload_part` deserves a regression test that asserts `EntityTooLarge` for a declared part size of 5 GiB + 1 and verifies zero body polls. A poll-counting stream would make the no-ingestion guarantee explicit for PutObject too. At the S3S/service boundary, cover a genuinely signed AWS chunked request whose encoded `Content-Length` exceeds the cap but decoded length is within it, plus decoded length above the cap. This checks that admission uses the logical part/object size rather than framing overhead. Preserve exact-cap and zero-length behavior.

3. **Test the real S3S body-limit error alongside the mock.** A tiny `Body` budget can produce an actual `BodySizeLimitExceeded` without uploading gigabytes. Feed that error through the same nested I/O shape and both `ApiError` conversions. This guards the display-based compatibility matcher against dependency changes; the current mocked display only verifies the expected string. An isolated probe against the pinned S3S dependency confirmed that the new matcher recognizes the real error and its nested I/O wrapper.

4. **Use the typed error when integrating [s3s-project/s3s#812](https://github.com/s3s-project/s3s/pull/812).** Replace the display heuristic with a typed downcast once the export is available in the RustFS dependency. That integration also matters for authoritative length normalization: the current dependency can backfill a missing `Content-Length` from an unrelated `x-amz-decoded-content-length` header.

5. **Place the follow-up inactivity guard before S3S decoding.** The existing guard measures decoded output, while signed AWS chunks are collected and verified before yielding. In a virtual-time probe with real S3S verification, a valid 64 KiB signed chunk arriving as 8 KiB raw fragments every 60 seconds failed at 300 seconds with the decoded-body guard, but completed at 480 seconds with the guard before decoding. The follow-up should observe raw per-request byte progress, exclude admission waits and storage backpressure, preserve the capped-session timeout floor, and fail through the reader so existing temporary cleanup and same-part retry semantics remain intact. Verify the typed timeout through intermediate I/O wrappers, while retaining storage/server-side timeout classifications.

No additional blocking code issue was identified in the reviewed size-admission and body-error changes. Verification here consisted of source review and an isolated probe using the actual pinned S3S body-limit error; the full RustFS integration tests and the reported multi-GB server experiments were not independently rerun.

reatang:
Thanks for the review. Addressed in 51928eedc and in the PR description:

- Scope: the description now states that this PR partially addresses rustfs/backlog#2430 (declared-size admission only) and lists the pieces that stay open there (unknown-length UploadPart rejection, whole-part buffer removal, raw-body inactivity detection before s3s decoding, typed RequestTimeout). No inactivity-guard change is made here.
- UploadPart admission test: `execute_upload_part_rejects_oversize_declared_part_before_reading_the_body` asserts EntityTooLarge for cap + 1 with zero body polls (new poll-counting test body), and that exact-cap and zero-byte parts pass admission. The PutObject test now uses the same poll-counting body.
- aws-chunked boundary: `execute_put_object_oversize_admission_uses_decoded_length_for_aws_chunked` sends a STREAMING-* request whose framed Content-Length is above the cap; decoded length at the cap is not rejected as oversize, decoded cap + 1 is, with zero polls. This is at the usecase level; a fully signed request through the s3s service layer is not covered here.
- Real s3s error: `real_s3s_body_size_limit_error_maps_to_entity_too_large` trips the pinned s3s Body budget with a four-byte limit and runs the actual BodySizeLimitExceeded through the matcher and both ApiError conversions, alongside the mocked string.
- Typed downcast: left as a follow-up for when s3s-project/s3s#812 is in the pinned dependency; noted in the code comment and the description.

reatang:
CI note: the `Test and Lint (rio-v2)` failure on the merge commit 84464f567 is `rustfs-ecstore store::init::tests::batch_transitioned_delete_uses_free_version_per_item` panicking at `crates/ecstore/src/store/init.rs:15594` with `InsufficientReadQuorum` inside its cleanup polling loop. This PR does not touch ecstore; the test arrived with the rc.6 merge into main (2f5db4e8f). main's own Continuous Integration run at its current head 94d875ea0 (run 34471872666, rio-v2 job 102855527129) fails the same test at the same line, plus `restored_transitioned_delete_uses_free_version_as_cleanup_owner` at init.rs:12528, while the earlier main run at 2f5db4e8f passed it, so it is a flake in that test's polling loop (a transient read-quorum error during the delete race is treated as fatal instead of as 'not yet absent'). Not fixed here to keep the PR scoped; it needs its own change on main.

reatang:
CI note, second run on 26affde51: same flake family, different members. rio-v2 failed on `store::init::tests::transitioned_delete_uses_free_version_as_cleanup_owner` (init.rs:12528) and Workspace Test and Lint on `store::init::tests::batch_transitioned_delete_post_commit_failures_roll_back_without_free_version_receipt` (init.rs:16140), both with the same `InsufficientReadQuorum` inside the cleanup polling loop's `expect`. All three tests come from the rc.6 merge into main; nothing in this PR touches ecstore. Re-ran only the failed jobs. The four polling loops in `store::init` tests that call `load_file_info_versions_exact(...).expect(...)` need to treat a transient quorum error as 'not yet absent' on main.

PR #7650, opened 2026-09-10
fix(replication): correct peer joins and remote-state reporting

jkossis:
## Related Issues

Addresses the unresolved join-map identity finding recorded in #6399, misleading peer-state diagnostics, and the first-add initial-sync ordering error in fresh three-site groups.

## Summary of Changes

Three-site setup can persist placeholder peer IDs, report the coordinator's membership as every remote site's state, and return `initialSyncErrorMessage` because the first receiver probes the last site before its shared service account exists. This PR fixes these three server behaviors in separate commits:

- Seed joins from deployment IDs verified during preflight, replace same-endpoint placeholders during finalization, and prevent legacy messages from downgrading actual IDs.
- Build `PeerStates` from each site's already-fetched metainfo, preserving divergent or empty membership and remote metadata. Omit unreachable sites and retain their `PeerErrors`.
- Complete every remote join before asking acknowledging receivers to run their initial probes and backfill, so every site already has the shared account. Reuse the signed join route and bootstrap token for a completion request that checks the committed timestamp/account and pending lifecycle operations without repeating IAM/topology writes. Preserve genuine initial-sync failures and older receivers' reported errors.

The TLS test server also shuts down its TLS stream explicitly after writing a response, flushing buffered records and sending `close_notify` before closing its socket. This final commit changes only the shared test fixture and adds a clean-closure regression; production transport and trust policy are unchanged.

## Verification

Current head: `ba549dfd3959bb91bb3ac92446ba03b90ad992d9`. The three runtime corrections were tested at `78d888d222d700293b224a829a03b4fe5c5dc10d`, based on `5cd58319ed6148ed7f09f2a4d0b4e46e429f043a`; the final commit is test-only. The runtime artifact from that head has Linux ARM64 SHA-256 `3f87b68e3a4ccec52e112e37ecefaf74ccfb1dbc3a52e5b55e35266a56303ecd`.

- CI at `78d888d2` passed the workspace, rio-v2, end-to-end, S3, ILM, build, and boundary checks. The Swift lane failed only `site_replication::tests::peer_clients_isolate_skip_and_custom_ca_trust`, with a connection reset on its second custom-CA request. The SFTP job was cancelled after both Clippy and its tests passed; the aggregate gate consequently failed. The separate CLA Check passed despite the cancelled wrapper.
- The shared TLS fixture returned immediately after `write_all`, without flushing/shutting down TLS. The exact new regression and fixture, extracted unchanged into a small harness using the repository's locked dependency versions, deterministically failed with missing `close_notify` before the fix and passed afterward. The fixture now uses `shutdown`, which sends the alert and drains pending TLS writes. Formatting and `git diff --check` passed; correctness and simplicity review found no supported issues.
- With the correction, 1,000 sequential and 1,600 concurrent custom-CA requests passed. Another 100 rounds covered trusted CA success, rejection of an unrelated CA, explicit skip verification, redirect suppression, and complete HTTP-error responses. The exact CI connection reset did not reproduce in the pre-fix request loops either; the reproducible local defect is the invalid TLS closure. The fresh native CI run must confirm the correction against the original failing lane. No test assertions, trust settings, retry policy, or quarantine configuration were weakened.

- The preceding head `14c42de3` reproduced the fresh three-site ordering error; two-site setup and an explicit post-join retry passed. With this commit, fresh two/three-site setup and setup with a pre-existing receiver bucket/object return no operation or initial-sync error. Direct deployment IDs and remote states agree; all-origin replication preserves object versions and bytes with read proxying disabled. The pre-existing object also backfills with its original version and bytes.
- The unchanged provider at `9f6bb1eecb48f48f8d6ef242eaeb7e90e588b89c`, using real OpenTofu 1.12.6 against three disposable servers with pre-existing receiver data, now succeeds on its first apply, retains untainted ownership of all three peers, backfills that object, and produces a subsequent no-change plan. No retry, untaint, state removal, or replacement was needed. Against the preceding server head, the same provider correctly failed creation, retained tainted ownership, and proposed replacement.
- Negative and compatibility API checks passed: broken reverse connectivity still reports an initial-sync error; a stale completion request with the correct account is rejected without changing state; an rc.5 two-site pairing succeeds; an rc.5 receiver in a three-site group still exposes its existing ordering error. All test credentials and Docker resources were disposable; test containers, networks, and data volumes were removed.
- Added unit coverage for optional wire-field defaults/acknowledgment and completion admission, including timestamp/account mismatches and pending operations. The existing portable three-node endpoint-edit/status E2E now starts with data on the first joining receiver and requires clean add plus backfill before exercising endpoint edits.
- `cargo fmt --all --check`, `git diff --check`, `cargo build --locked --offline --no-default-features -p rustfs --bin rustfs --config profile.dev.package.rustfs.codegen-units=512`, and `cargo check --locked --no-default-features --jobs 1 -p rustfs -p e2e_test --tests` passed with Rust 1.98.0, debug information disabled, and lld. The server build used two jobs; test typechecking used one. The initial parallel typecheck exceeded the local VM memory limit (SIGKILL, cgroup `oom_kill: 1`); serial typechecking passed. Full native test execution was not repeated locally because the earlier unit-binary build exceeded the Docker memory limit; test typechecking and real API checks are not a full native suite run.

Earlier commits also passed retained-volume rc.5 identity repair/restart checks and a failing-before/passing-after status test with divergent metadata and an offline peer. The previous head received Codex's thumbs-up at 19:30 UTC. Fresh CI and review are required for the added test-fixture commit; the earlier reaction does not approve this head.

## Impact

Fresh groups receive actual identities immediately. Existing identity mismatches still require reviewed peer-edit replay from a correct coordinator after upgrading every site; upgrading alone does not repair stored mismatches.

Status consumers requesting `peer-state=true` receive each site's own snapshot. An absent entry with `PeerErrors` means unavailable evidence, not agreement or an empty topology. A reachable site's empty map remains visible.

Join deferral uses two additive optional wire fields, `deferInitialSync` and `initialSyncDeferred`, and one additional signed request per acknowledging remote. Missing fields retain the prior behavior; old coordinators and nonacknowledging receivers use the existing join flow. Mixed-version three-site groups can therefore still report the old ordering error. No dependencies, persisted fields, or storage-schema changes are added. An interrupted operation can still be partial, and clients must expose its diagnostics and review recovery rather than replace existing topology blindly.

Rollback: revert these commits and deploy the previous binary; no schema rollback is required. Older binaries can reintroduce placeholder identities, misleading status, and the initial-sync ordering error. No production rollout has been performed.

## Additional Notes

Two solo review passes covered the join-order change; earlier identity/status reviews remain applicable. The final test-only correction received correctness and simplicity review. No supported findings remain:

- Correctness: all accounts are installed before deferred probes/backfill, which still runs before the coordinator's bucket-set validation; genuine failures remain visible.
- Simplicity: existing join transport, admission, backfill, and lifecycle helpers are reused; no new route or persisted phase is introduced.
- Test coverage: fresh/nonempty real-server and real-provider cases pass; broken connectivity, legacy behavior, and stale completion are exercised. Native execution remains a CI gate.
- Security: existing admin authorization and encrypted/signed transport remain; completion checks the committed account and state generation without repeating credential writes.
- Concurrency/durability: completion holds the local lifecycle guard; network operations remain outside the state-object transaction. Changed or pending lifecycle state rejects completion.
- Compatibility: explicit acknowledgment controls the second request; rc.5 two-site success and mixed three-site failure visibility are verified. Persisted schema is unchanged.

---

### DIFF
--- a/crates/e2e_test/src/replication_extension_test.rs
+++ b/crates/e2e_test/src/replication_extension_test.rs
@@ -6383,6 +6383,18 @@ async fn test_site_replication_edit_and_status_peer_state_real_three_node() -> R
     let relayed_key = "after-edit-from-relay.txt";
     let relayed_payload = b"site replication after endpoint edit from relay".to_vec();
 
+    // The first joining receiver owns data before the third site has the
+    // shared account. Initial probes and backfill must wait for every join.
+    target_client.create_bucket().bucket(bucket).send().await?;
+    enable_bucket_versioning(&target_env, bucket).await?;
+    target_client
+        .put_object()
+        .bucket(bucket)
+        .key(baseline_key)
+        .body(ByteStream::from(baseline_payload.clone()))
+        .send()
+        .await?;
+
     let add_status = site_replication_add(
         &source_env,
         &[
@@ -6410,7 +6422,10 @@ async fn test_site_replication_edit_and_status_peer_state_real_three_node() -> R
         ],
     )
     .await?;
-    assert!(add_status.success, "unexpected site add result: {:?}", add_status);
+    assert!(
+        add_status.success && add_status.err_detail.is_empty() && add_status.initial_sync_error_message.is_empty(),
+        "unexpected site add result: {add_status:?}"
+    );
 
     let source_info = wait_for_site_replication_enabled(&source_env, 3).await?;
     let _target_info = wait_for_site_replication_enabled(&target_env, 3).await?;
@@ -6421,19 +6436,11 @@ async fn test_site_replication_edit_and_status_peer_state_real_three_node() -> R
         .find(|peer| peer.endpoint == target_env.url)
         .ok_or("target peer missing from source site replication info")?;
 
-    source_client.create_bucket().bucket(bucket).send().await?;
-    enable_bucket_versioning(&source_env, bucket).await?;
-    wait_for_bucket_on_target(&target_client, bucket).await?;
-    wait_for_bucket_on_target(&relay_client, bucket).await?;
-    source_client
-        .put_object()
-        .bucket(bucket)
-        .key(baseline_key)
-        .body(ByteStream::from(baseline_payload.clone()))
-        .send()
-        .await?;
-    let replicated_baseline = wait_for_object_on_target(&target_client, bucket, baseline_key).await?;
-    assert_eq!(replicated_baseline, baseline_payload);
+    for client in [&source_client, &relay_client] {
+        wait_for_bucket_on_target(client, bucket).await?;
+        let backfilled = wait_for_object_on_target(client, bucket, baseline_key).await?;
+        assert_eq!(backfilled, baseline_payload);
+    }
 
     let old_target_address = target_env.address.clone();
     let new_target_port = RustFSTestEnvironment::find_available_port().await?;
--- a/rustfs/src/admin/handlers/site_replication.rs
+++ b/rustfs/src/admin/handlers/site_replication.rs
@@ -2793,6 +2793,19 @@ fn prune_in_sync_status_details(status: &mut SRStatusInfo, opts: &SRStatusOption
     }
 }
 
+fn peer_states_from_infos(
+    site_infos: BTreeMap<String, SRInfo>,
+    reachable_peers: &HashSet<String>,
+) -> BTreeMap<String, SRStateInfo> {
+    // Failed metainfo fetches leave default entries in site_infos for comparison;
+    // they must not become fabricated peer state. PeerErrors describes the failure.
+    site_infos
+        .into_iter()
+        .filter(|(deployment_id, _)| reachable_peers.contains(deployment_id))
+        .map(|(deployment_id, info)| (deployment_id, info.state))
+        .collect()
+}
+
 async fn build_status_info(state: &SiteReplicationState, local_peer: &PeerInfo, uri: &Uri) -> S3Result<SRStatusInfo> {
     let opts = sr_status_options(uri);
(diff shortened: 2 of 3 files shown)


houseme (approved):
Reviewed the current head `800e8e2d62f058c9ee15a0feb356a6694ec4b5ef` against the PR description, #6399's recorded follow-up context, the current diff, visible CI, and MinIO's site-replication join/status behavior.

Conclusion: no findings.

The peer identity fix looks right: joins are seeded from deployment IDs verified during preflight, placeholders are replaced by endpoint identity, and later legacy/placeholder updates cannot downgrade a real deployment ID.

The peer-state fix also matches MinIO's shape. MinIO builds `PeerStates` from each site's own metainfo, and this PR now does the same while keeping unreachable sites out of `PeerStates` and reporting them through `PeerErrors`.

For the three-site initial-sync ordering issue, the new two-step flow is conservative: RustFS peers that acknowledge `deferInitialSync` install the service account/topology first, then run probes and backfill through `initial-sync=true`; older RustFS or MinIO peers keep the old one-pass behavior. The committed-snapshot checks on the second request prevent stale or mismatched completion from running against a moved topology.

I also checked the TLS fixture change. It is test-only and fixes the missing TLS shutdown/close_notify issue without changing production transport or trust policy.

Visible CI is green, and the added tests cover the important failure modes: actual peer IDs, real remote peer-state reporting, deferred initial sync, stale completion rejection, join admission serialization, and the three-node pre-existing-data path. Approving.

PR #7657, opened 2026-09-11
test(kms): cover non-default Vault Transit paths

hkwi:
## Related Issues

N/A

## Summary of Changes

The Vault Transit live tests use the default Transit mount and metadata location, leaving path overrides untested against a real Vault. Extend the AppRole harness with one case that sets all three paths to non-default values:

| Environment variable | Existing case | New case |
| --- | --- | --- |
| `RUSTFS_KMS_VAULT_MOUNT_PATH` | `transit` | `kms-test/transit` |
| `RUSTFS_KMS_VAULT_TRANSIT_METADATA_KV_MOUNT` | `secret` | `kms-test/metadata` |
| `RUSTFS_KMS_VAULT_TRANSIT_METADATA_PREFIX` | `rustfs/kms/transit-metadata` | `custom/transit-metadata` |

The new case runs key creation, description, listing, and data-key wrapping/unwrapping after removing both default mounts (`transit` and `secret`) from the ephemeral Vault. It then recreates the backend and checks the persisted key description, so an in-memory cache or synthesized metadata cannot satisfy the read. AppRole permissions cover only the required paths and operations.

The harness retains the default KV2 and Transit checks and adds one all-custom case, keeping three live cases in total. The existing nightly job already invokes this script.

The rio-v2 gate also exposed a polling race in `batch_transitioned_delete_uses_free_version_per_item`: an unlocked metadata read can briefly lose quorum while asynchronous cleanup removes per-disk copies. The poll now treats only that transient quorum result as not yet converged; every other error still fails immediately.

After the base branch reverted the fault-tolerance functional suite, its workflow contract test still expected the removed workflow and replication handoff. Restore the contract assertions for the current ten-suite chain, where replication hands off directly to performance.

## Verification

Passed on base `853ae63` with Rust 1.98.0 and Vault 1.15.1:

- `bash scripts/test/vault_approle_kms_live.sh`: all three live cases passed (KV2, default Transit paths, and all-custom Transit paths); this also compiled the changed Rust test.
- `cargo fmt --all --check`
- `bash -n scripts/test/vault_approle_kms_live.sh`
- `shellcheck scripts/test/vault_approle_kms_live.sh`
- `git diff --check`

Additional rio-v2 verification with Rust 1.98.1:

- `cargo nextest run -p rustfs-ecstore --features rio-v2,test-util -E 'test(=store::init::tests::batch_transitioned_delete_uses_free_version_per_item)'`: 10 consecutive passes.
- `python3 scripts/test_security_workflow.py`
- `make script-tests`

## Impact

Test-only change. Production code, configuration defaults, APIs, and stored formats are unchanged. All mounts are created or disabled only in the script's ephemeral loopback Vault. Reverting the four changed test and test-harness files removes the added coverage and CI-race fixes.

## Additional Notes

This checks the KMS backend through the same environment configuration used by RustFS. It does not start a RustFS S3 server or exercise a full S3 PUT/GET round trip. Custom paths use generic test names and include slashes to cover nested mounts and prefixes. Mixed default/custom combinations are outside this focused coverage.


overtrue (commented):
Reviewed `d20059ab5...7c4227f11` (test-only, low risk). The KMS live coverage looks good: the custom case disables both default mounts before running, pins all three resolved paths, and re-reads the description through a fresh backend, so a fallback to `transit`/`secret` or an in-memory record cannot pass. The narrower custom AppRole policy (data path only, no `metadata/*`) matches what the exercised operations call (`kv2::read` / `set_with_options` with `cas`).

Two issues, both fixed in commits pushed to this branch:

1. **`scripts/test_security_workflow.py` conflicted with main.** #7679 landed the same ten-suite expectations on main, differing only in the test name (`test_all_suites_...` vs `test_all_ten_suites_...`). I merged `main` and took main's version, so the PR no longer touches this file.

2. **The quorum-race fix covered only one of the two poll loops** (`crates/ecstore/src/store/init.rs`, `batch_transitioned_delete_uses_free_version_per_item`). The earlier unversioned `causal` duplicate-delete poll reads `load_file_info_versions_exact` while the same asynchronous free-version cleanup removes the per-disk `xl.meta` copies. Once some disks are gone but not all, `shallow_versions.len() < read_quorum` returns `ErasureReadQuorum`, which `to_object_err` maps to `InsufficientReadQuorum`. That poll still used `.expect(...)`, so the flake could move there. Both polls now share `exact_metadata_absent_during_cleanup`, which treats only that transient result as "not converged" and panics on any other error. A permanent quorum loss still fails via the 30s convergence timeout.

Verification on `dbe310ec4`: `cargo nextest run -p rustfs-ecstore --features rio-v2,test-util -E 'test(=store::init::tests::batch_transitioned_delete_uses_free_version_per_item)'` 5/5 passes, 1/1 with `--features test-util`; `cargo clippy -p rustfs-ecstore --all-targets --features test-util -- -D warnings` clean; `cargo fmt --all --check`; `python3 scripts/test_security_workflow.py` (21 OK). The Vault live script was not re-run because the KMS files are unchanged from your head.

PR #7662, opened 2026-09-11
feat(ci): add fault-tolerance degradation suite to the functional chain

majinghe:
## What

Adds a fault-tolerance (degradation) functional suite and wires it into the nightly chain between replication and performance:

```
upgrade -> s3 -> kms -> tier -> storage -> heal -> pool -> security
       -> replication -> fault-tolerance -> performance
```

New workflow: `.github/workflows/rustfs-fault-tolerance-test.yml`. The replication workflow's chain handoff now dispatches `rustfs-chain-fault-tolerance`; the new suite forwards to `rustfs-chain-performance`.

## Scenarios (expectations derived from product source)

- **A — single-node 4 drives** (`default_parity_count(4) = 2` → read quorum 2, write quorum 3): hide 1/2/3 drive dirs (empty tmpfs bind over the drive path), then restore.
- **B — multi-node 4x1** (one set of 4, EC:2): stop 1/2/3 nodes, then restore. The 2-nodes-down tier is the externally reported divergence case (reads satisfy the EC read quorum).
- **C — multi-node 4x4** (16 drives form one erasure set; EC:4 → read quorum 12, write quorum 13): 1 node down lands exactly on the read-quorum boundary (12 online), 2 nodes down breaks it.
- **C2 — multi-node 4x4 with `RUSTFS_STORAGE_CLASS_STANDARD=EC:8`** (read quorum 8, write quorum 9, lock majority 9): 2 nodes down puts reads inside the reported divergence window — read quorum met while the lock majority is broken (external report: reads 503 with `lock_quorum_unavailable`).

Every tier snapshots `/health`, `/health/ready`, and `/minio/health/live` (codes + bodies) next to the actual S3 GET/LIST/PUT results, so health-endpoint truthfulness is captured alongside real availability.

## Failure policy

By default, reads refused despite a met read quorum are reported as `known-divergence` (recorded in the report + evidence, suite still green). `strict: true` escalates them to failures — flip it once the product behavior is settled. Anything violating the EC contract in an uncontroversial way always fails.

## Notes

- Suite script lives in rustfs/auto-testing (`rustfs-fault-tolerance-test.sh`); destructive steps are bracketed by cleanup-before/after, sharing the usual functional-tests concurrency group.
- Already live on the `release` branch (pushed directly); this PR mirrors it to `main`.


majinghe:
Superseded by a conflict-free branch based on main.

PR #7695, opened 2026-09-12
test(heal): retry recovery readback SlowDownRead

houseme:
## Related Issues

Related to rustfs/backlog#2428 and rustfs/backlog#2240.

## Summary of Changes

Retry transient retryable GET failures in the Scanner/Heal interruption oracle after replacement recovery has already converged.

The G14 multi-pool crash case can reach the final target-node readback while the recovered read path still briefly returns `SlowDownRead`. This change wraps only the final oracle readback in a bounded retry helper for `SlowDownRead` and `ServiceUnavailable`. The test still verifies exact object bytes and still fails if the object remains unreadable past the readback window.

## Verification

- `cargo check -p e2e_test` passed.
- `git diff --check` passed.
- Azure Linux current-head G14 multi-pool evidence case reproduced the pre-fix failure with `SlowDownRead` during final target-node GET readback, including a low-load retry.

## Impact

No production runtime behavior change. This only tightens the e2e recovery oracle so it treats retryable post-recovery read unavailability as a bounded convergence condition instead of an immediate hard failure.

## Additional Notes

This does not close G14 by itself. After merge, the G14 multi-pool measured case must be rerun on latest `release`, followed by G14 descriptor assembly and bundle ledger refresh.

### DIFF
--- a/crates/e2e_test/src/heal_erasure_disk_rebuild_test.rs
+++ b/crates/e2e_test/src/heal_erasure_disk_rebuild_test.rs
@@ -24,7 +24,7 @@ mod tests {
     use crate::storage_api::RUSTFS_META_BUCKET;
     use aws_sdk_s3::{
         error::{ProvideErrorMetadata, SdkError},
-        operation::{delete_object::DeleteObjectError, put_object::PutObjectError},
+        operation::{delete_object::DeleteObjectError, get_object::GetObjectError, put_object::PutObjectError},
         primitives::ByteStream,
     };
     use http::Method;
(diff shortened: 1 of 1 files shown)


houseme:
## PR-head Linux validation update

I ran the G14 multi-pool crash evidence case on the PR head to check whether the bounded readback retry was sufficient for the current release blocker.

Source and run:

- head: `72f471546e54b4f6028369bd3295b1847030a52e`
- case: `background-target-crash-ec8-4-multi-pool`
- first PR-head run: default readback window, clean build
- second PR-head run: `RUSTFS_HEAL_READBACK_TIMEOUT_SECS=180`, reused the already-built binary

Result:

- Both PR-head runs still failed with `SlowDownRead` during target-node readback.
- The 180s run failed only after the bounded readback window was exhausted, so the retry helper is active, but the current multi-pool crash recovery path still does not satisfy the G14 oracle.
- No G14 oracle JSON was produced, so this cannot be used as release evidence.

Conclusion:

This PR is still a useful oracle hardening change, but it is not sufficient to close G14. The remaining G14 blocker is now a product/recovery-path issue: after multi-pool background target crash/rejoin, the target-node GET path can remain `SlowDownRead` beyond the measured recovery readback window.

PR #7756, opened 2026-09-13
fix(replication): count a bodiless 405 as a replicated delete marker

reatang:
## Related Issues

Fixes rustfs/backlog#2479 (product half; the probe half landed in rustfs/auto-testing#66). Also the SITE-105 signature tracked in rustfs/backlog#2195.

## Summary of Changes

A site or bucket resync verifies each replicated delete marker with `HEAD ?versionId=<marker>` on the target. Every S3 target (RustFS, MinIO, AWS) answers that request with 405 and no body. The AWS SDK only synthesizes an error code for 404, so the HEAD error reached `verify_resync_head_result` with `code() == None`, and `is_retryable_delete_replication_head_error` classified it as an ambiguous failure. Result: every delete marker counted as a failed object with `errorDetail: "target service error"`, and a site resync over any bucket holding a delete marker reported the whole bucket as `failed`. This is exactly what the functional suite shows in SITE-105 on rc.6 and on every nightly since the probe fixture started creating a delete marker (latest: run 34756288056, `replicatedObjects: 2, failedObjects: 1`).

The fix reads the raw HTTP status the same way the existing 404 path already does (`head_object_not_found` / `has_raw_status`): a 405 without a code confirms the marker propagated. Ambiguous statuses (5xx, 403, ...) still count as failed, so backlog#862 / B13 semantics are unchanged.

## Verification

Tested commit: the PR head, rebased onto `origin/main` at `cc29b03a0`. The rebase kept both the `#7709` e2e case and the new one at the same insertion point. `main` was merged in again at 19948a0ed; the diff against `main` is still the three files listed below plus the review follow-up commit.

- `RUST_MIN_STACK=33554432 cargo nextest run -p rustfs-ecstore -E 'test(resync_verification)'`: 3 passed, 0 failed (`resync_verification_accepts_bodiless_405_for_delete_marker`, `resync_verification_still_fails_delete_marker_on_ambiguous_head_error`, plus the existing drifted-replica case)
- e2e `test_site_replication_resync_replicates_delete_marker` (two real rustfs processes joined as sites, one live object + one delete marker converged, then a site resync must finish with `failed_objects == 0`): PASS in 9.8s via `python3 scripts/e2e_binary.py build` then `python3 scripts/e2e_binary.py run -- cargo nextest run --profile e2e-repl-nightly -p e2e_test -E 'test(test_site_replication_resync_replicates_delete_marker)'`. On the pre-fix commit the same case failed with `target service error` (red-to-green established when the branch was first cut on 2026-09-12).
- Review follow-up (29f7b23ee): a marker-version purge (`version_purge_status` set) is verified by absence only, so the 405 mapping applies to marker creation alone. `RUST_MIN_STACK=33554432 cargo nextest run -p rustfs-ecstore -E 'test(resync_verification)'`: 5 passed, including the new `resync_verification_keeps_marker_purge_failed_when_target_still_answers_405` (Pending and Failed) and `resync_verification_accepts_marker_purge_when_target_answers_404`.
- `cargo fmt --all --check`: pass.
- `python3 scripts/check_test_wiring.py --update-profile e2e-repl-nightly <listing> darwin` refreshed `.config/e2e-repl-nightly-selection.txt` (73 tests, sha256 `e4ee3898…`); `--check-profile` passes on the same listing. `make pre-commit`: pass (fmt, architecture guards, workspace quick-check).

Not run: the full `e2e-repl-nightly` lane and the functional replication suite on VMs. Remaining risk is limited to the delete-marker branch of resync verification; live-object verification is untouched.

## Impact

Site and bucket resyncs over buckets that contain delete markers now complete instead of reporting the bucket as failed. No API, configuration, or on-disk format change. Outbound client defaults are untouched (the change only reinterprets an inbound HEAD status).

## Additional Notes

Companion product fix for the other SITE-105 symptom (`target belongs to a different active resync operation`) was superseded by #7709; the `fix/sr-target-reset-id-carryover` branch should be discarded.


cxymds (commented):
Reviewed base `0d322c5086afb938880d33d463ffacfbd4121904` and head `19948a0ed255220a4e1d918acbb5ff91124787ee` as a high-risk replication change, with two sequential passes.

One P2 finding is attached. Please distinguish delete-marker creation from version-purge verification before merging.

- Correctness and compatibility: a failed marker-version purge can now be counted as successful.
- Test coverage: the added cases omit the pending/failed marker-purge shape.
- Simplicity: no independent finding.

Validation: a focused differential probe of the original base/head verifier functions, using the actual SDK and replication types, reproduced the counter change. `git diff --check` and `cargo fmt --all --check` passed. The focused nextest run was stopped during compilation; E2E was not rerun. The CI Quick Checks failures observed during review were in code and a guard script unchanged from the base.

crates/ecstore/src/bucket/replication/replication_resyncer.rs
```diff
@@ -1815,12 +1815,17 @@ async fn verify_resync_head_result(
...
+            // code for 404; a target answering the delete marker's version
+            // with 405 (RustFS, MinIO, AWS) leaves `code()` empty and the
+            // raw status is the only evidence (rustfs/backlog#2479).
             let retryable = {
                 let (is_not_found, code) = err
                     .as_service_error()
                     .map(|se| (se.is_not_found(), se.code()))
                     .unwrap_or((false, None));
-                is_retryable_delete_replication_head_error(is_not_found, code)
+                let code = code.or_else(|| has_raw_status(&err, 405).then_some("405"));
```
cxymds: **[P2] Do not count a remaining delete marker as a successful version purge**

This arm also receives `roi.delete_marker == true` with `version_purge_status == Pending` or `Failed`. That is an admitted resync shape: `resync_rebuilt_purge_entry_keeps_versioned_delete_semantics` and `resync_keeps_marker_version_purges_separate_from_marker_creation` already cover its construction and admission.

If the version-addressed DELETE fails, the target still has the marker and its HEAD returns a bodiless 405. This new fallback accepts that as success. Since `resync_worker_process_object` discards the delete result and derives its counters from this HEAD, the bucket/site resync can report completion while the requested purge remains unapplied.

A differential probe with the original verifier functions confirms that a pending or failed purge plus a codeless 405 changes from `(replicated_count, failed_count) = (0, 1)` on the base to `(1, 0)` on this head. The 405 proves presence, which is the desired outcome for marker creation but the opposite of successful version removal.

Please accept 405 only for marker creation (`version_purge_status.is_empty()`), require absence for purge verification, and add a regression using the existing 405 fixture with Pending/Failed purge status.

reatang: Fixed in 29f7b23ee: the codeless-405 mapping now applies only to marker creation (`version_purge_status.is_empty()`); a pending or failed marker-version purge counts as replicated only when the target answers not found, and a 405 stays failed. Added `resync_verification_keeps_marker_purge_failed_when_target_still_answers_405` (Pending and Failed against the 405 fixture) and `resync_verification_accepts_marker_purge_when_target_answers_404`; 5/5 resync verification tests pass locally.

cxymds (approved):

PR #7832, opened 2026-09-14
fix(heal): scope replacement pool metadata to its owning set

hkwi:
## Related Issues

N/A

## Summary of Changes

A replacement disk in an erasure set that does not own `pool.bin` can remain marked as healing after its user data has been reconstructed. Replacement healing currently requires `pool.bin` on every target set, although metadata writes place it in one hashed set per pool. The resulting `FileNotFound` is repeatedly recorded as a transient skip.

Use ECStore's existing placement logic to require pool metadata repair only on its owning set. Owning sets still require a successful repair and target-specific physical readback. Missing or invalid scope and unsupported placement fail closed; missing shards are never used to infer placement.

Add coverage for automatic replacement and admin healing, plus a real-disk regression that repairs current and historical versions in both sets and checks healing-marker cleanup.

## Verification

Tested source contents are identical to commit `e8c5448e216d17c98429ec2be8a3af93ad040ffb`. The PR includes main through `6db5ac496881b1f543559f8cc33341eff9f2a65f`.

- `cargo nextest run -p rustfs-heal --lib --test heal_pool_metadata_scope_test -E 'test(resume_loop_tests) | binary(heal_pool_metadata_scope_test)' -j 16`: **38 passed**, 540 outside the selected filter. The real-disk test uses one pool with two sets of four disks, removes committed shards, verifies both versions and reconstructed shard bytes, checks marker removal, and rejects complete loss of the owning set's `pool.bin`.
- Negative control: removing only the placement guard makes the same integration test fail with `0 bucket(s) failed, 0 object(s) failed, 1 object(s) skipped; retry scheduled`. Restoring the guard makes it pass.
- `cargo fmt --all --check` and `git diff --check` passed. The original patch application preflight passed.

Not run with this patch: four-node deployment, independently mounted replacement devices, cold cluster restart, or encrypted beta-to-RC migration. These focused tests establish the placement/healing fix, not migration approval.

The CI follow-up replaces the regression test's grouped `storage_api` import with explicit domain imports. The architecture checker no longer reports this test, and all 38 focused tests and formatting checks pass again.

**Remaining CI blockers inherited from main:** `check_architecture_migration_rules.sh` still reports two failures: the raw-symbol test fixture in `rustfs/src/connect/diagnostics/profile_cpu.rs:725` matches the forbidden ECStore import pattern, and `crates/io-metrics/Cargo.toml:52` adds `rustfs-common` despite the leaf-crate dependency rule. Both affected implementations and the checker are identical to main `6db5ac496881b1f543559f8cc33341eff9f2a65f`; they are not changed by this PR. Full CI is not passing yet.

## Impact

Replacement healing can finish for a non-owning set without creating an extra `pool.bin` replica. Required metadata repair, target readback, final identity checks, completion proofs, and failed-task persistence remain enforced.

No disk, RPC, or encryption format changes. Existing exhausted checkpoints are retained; recovery from those saved states is not covered by this change. Reverting this patch restores the previous healing behavior and requires no format conversion introduced by this patch.

## Additional Notes

Two sequential review passes covered this high-risk repair change:

- Correctness: no unresolved findings after checking owning/non-owning sets, missing metadata, invalid scope, and error propagation.
- Simplicity: reuses the existing placement algorithm and scoped set selection; no duplicated repair path.
- Test coverage: the real storage/task regression fails without the guard; failure cases retain incomplete state.
- Concurrency/durability: the added lookup is synchronous and uses immutable topology; locking, commit ordering, identity fencing, and completion proof handling are unchanged.
- Compatibility: existing placement algorithms remain authoritative; no metadata relocation or serialization changes.


hkwi:
These three PRs address different stages of healing and should be considered together when validating recovery:

| PR | Scope |
| --- | --- |
| #7823 | Scanner request admission and identity. Empty request IDs could make requests for different objects collide before target deduplication. The fix assigns identities before publication and normalizes empty IDs at admission. |
| #7832 (this PR) | Replacement healing completion. Requiring `pool.bin` on a set that does not own it causes repeated transient skips even after user data has been reconstructed. This PR uses authoritative placement to require metadata repair only on the owning set. |
| #7839 | Administrator pool/set outcome accounting and recovery. Physical repairs could produce zero canonical outcomes, while an empty all-buckets scope failed recovery decoding. The fix persists evidence-backed outcomes and recovery scope so the original administrator token retains its results across restart. |

#7823 has no changed files in common with this PR. #7839 also changes `crates/heal/src/heal/erasure_healer.rs`, including the pool-metadata outcome path, but does not include this PR's owning-set placement check. It therefore does not replace #7832. Shared-file overlap alone does not establish a merge conflict or an implementation dependency.

Combined validation should cover scanner request independence, replacement healing in both owning and non-owning sets, and administrator outcomes retained under the original token after restart. The combined patches have not been tested here; the individual PR test results do not establish their combined behavior.

There is also a rollback distinction: #7832 introduces no persisted-format change, whereas #7839 introduces schema-8 administrator checkpoints that older binaries cannot resume. Its documented checkpoint retention and rollback constraints must be considered when validating the combined candidate.

At the time of this review, #7823 and #7839 target `release`, while #7832 targets `main`; all three are open and unmerged. Reviewed heads: #7823 `d94ab2ee6`, #7832 `de69fa690`, and #7839 `1ad6c2ab4`.

cxymds:
The fix is now tracked in #7846, targeting `release`. That PR preserves the original commits and includes the release integration-test compilation adjustment.

`main` will be synchronized from `release` later, so this PR does not need to be merged separately into `main`. Closing in favor of #7846.

PR #7936, opened 2026-09-16
docs+feat: multi-cipher plan, AES-GCM walkthroughs, demo cipher crate

chchas:
## Related Issues
N/A

## Summary of Changes
Lands the multi-cipher encryption extension design and its reference implementation groundwork. reading-record/09-multi-cipher-plan.md is the design plan for adding a second cipher (Aes256GcmDemo) on the legacy crates/rio data path via new frame type bytes 0x03/0x04 and a RUSTFS_ENCRYPTION_CIPHER env for the write side, with read-side dispatch on the on-disk frame type byte (never env) and a six-step recipe for future algorithms. Plan sections 1-4 are intentionally unimplemented; this PR lands only the design and the vendor dependency layer of the plan. reading-record/10-encryption-walkthrough-aes256-gcm.md and 11-decryption-walkthrough-aes256-gcm.md are annotated code-flow walkthroughs of the existing AES-256-GCM data-plane encryption/decryption (frame layout, nonce derivation, v2 AAD, legacy nonce fallbacks, SSE material flow). vendor/rustfs-aes256-gcm-demo is a new self-authored demo cipher crate placed outside crates/, wired into the workspace as a path dependency (workspace members + [workspace.dependencies] + crates/rio dependency) and tracked under .gitignore D1 (vendor/* with an exception for the crate); it implements aead::AeadCore + aead::Aead (aead 0.6.1) by wrapping aes_gcm::Aes256Gcm, locking the 12-byte nonce / 16-byte tag parameter contract at the type level.

## Verification
- cargo test -p rustfs-aes256-gcm-demo: 4 contract tests pass (AeadCore/Aead parameter locking, roundtrip, tampered-tag and wrong-AAD authentication failures) — verified on commit fad1e441f.
- Documentation-only changes: git diff --check passes for all doc commits (whitespace clean).
- No default-build behavior change in this PR: the demo crate is wired into the workspace but not referenced by any runtime code path yet (plan sections 1-4 unimplemented); no runtime test applies to the walkthrough/prose-only changes.

## Impact
Additive: a new vendored demo crate and three reading-record documents. No change to the default data path, frame layout, or configuration semantics; RUSTFS_ENCRYPTION_CIPHER is plan-only and no code reads it yet. Existing AES-256-GCM behavior and frame types 0x00/0x01/0x02 are untouched.

## Additional Notes
Plan sections 1-4 (config env, WritePlan dispatch, crates/rio frame and cipher changes, six-step recipe) are intentionally left unimplemented and reference this PR's vendor crate as the API contract to plug against. Commit subjects follow the repository Conventional Commits convention.


chchas:

overtrue (changes requested):
## Review: PR #7936 — reviewed base `b74dd7942` … head `fad1e441f` (risk tier: Standard — workspace/build wiring plus repository-policy docs; no runtime data-path change)

### Verdict: REQUEST_CHANGES

### Findings

**P1 — `reading-record/` commits one-shot planning/walkthrough documents that repository policy forbids, in a location that bypasses both guards.**
- `reading-record/09-multi-cipher-plan.md`, `10-encryption-walkthrough-aes256-gcm.md`, `11-decryption-walkthrough-aes256-gcm.md` (2,396 lines) are a design plan and two annotated code walkthroughs. `AGENTS.md` "Sources of Truth" states: *do not commit one-shot plans … or agent scratch notes; durable architecture belongs under `docs/architecture/`*. `docs/architecture/README.md` rule 2 additionally requires *cite a file path plus a symbol name, never a line number*; these files contain ~300 `file:line` anchors (e.g. `encrypt_reader.rs:123-173`, `sse.rs:1386-1452`) and the plan's own header says line numbers were "re-verified against origin/main a7341abdb and updated", which is exactly the staleness the rule exists to prevent.
- Failure scenario: `scripts/check_no_planning_docs.sh` only scans `docs/superpowers/*` and `.gitignore` only whitelists `docs/{architecture,operations,postmortems,testing}`; a new top-level `reading-record/` directory is invisible to both, so the guard passing locally/in CI is not evidence of compliance. Once merged it becomes a precedent for arbitrary top-level note directories.
- Fix: drop the three documents from this PR and keep the plan in a GitHub issue/discussion (the PR body already summarises it). If a durable design record is wanted after the design is accepted, write it as a symbol-cited page under `docs/architecture/` and register it in that README's routing table.

**P1 — CI is red at `cargo fmt --check`; the downstream guards never ran.**
- `vendor/rustfs-aes256-gcm-demo/src/lib.rs` is not rustfmt-formatted (5 hunks, reproduced locally with `cargo fmt --all --check`), and both `lib.rs` and the crate `Cargo.toml` lack a trailing newline. Quick Checks fails at the fmt step, Test and Lint fails on the dependency, and the 15 later guard steps (layer deps, logging, planning docs, script contract tests, …) were skipped, so the PR has not actually been exercised by the merge gate.
- Fix: `cargo fmt --all` and re-push; expect the remaining guards to run.

**P2 — The demo crate is wired into `crates/rio` as an unused dependency and placed under `vendor/`, which misdescribes it.**
- `crates/rio/Cargo.toml:67` adds `rustfs-aes256-gcm-demo`, but no file under `crates/` or `rustfs/` references `rustfs_aes256_gcm_demo`; plan sections 1–4 (the only consumers) are explicitly unimplemented. An unused dependency in the hot-path I/O crate is dead weight with no test exercising the 0x03/0x04 wiring the crate exists for.
- `vendor/` is the conventional output directory of `cargo vendor` (third-party source with `.cargo-checksum.json` per package). A first-party path crate there, plus the new `vendor/*` / `!vendor/rustfs-aes256-gcm-demo/` `.gitignore` split, means a future `cargo vendor` run will interleave third-party output with a tracked crate that lacks a checksum file and fail to validate. Workspace convention is `crates/<name>` with `version.workspace = true`; this crate ships `version = "0.1.0"`, no `publish = false`, and a `documentation` URL pointing at a docs.rs page that does not exist.
- Fix (either): (a) hold the crate until the write/read dispatch (plan §2–§3) lands in the same PR so it has a consumer and a frame-level test, or (b) if landing the contract alone is the goal, move it to `crates/aes256-gcm-demo`, inherit the workspace version, set `publish = false`, drop the docs.rs link, and do not add it to `crates/rio` until something uses it. Revert the `.gitignore` change in both cases.

**P3 — Misleading doc comment on `Aes256GcmDemo`.**
- `vendor/rustfs-aes256-gcm-demo/src/lib.rs:26-28` says the type is exposed "under a distinct … key derivation path". The implementation is `Aes256Gcm::new_from_slice(key)` with the caller's 32-byte key unchanged; there is no derivation. Reword to say it shares AES-256-GCM's key and parameters and differs only in type identity, so a reader does not assume the demo already proves a different-key-size cipher can be plugged in.

### What was checked
- Diff of all 9 files at the recorded base/head; the four contract tests in the crate are correct for what they assert (nonce/tag sizes, roundtrip, tampered tag, wrong AAD) and pass per the PR body; not re-run here because the crate has no consumer.
- Local guards on the PR head: `check_layer_dependencies.sh` pass, `check_doc_paths.sh` pass, `check_no_planning_docs.sh` pass (trivially, see P1), `cargo fmt --all --check` **fail**.
- Plan claims vs code: frame constants `FRAME_TYPE_V1/V2/V2_FINAL/END` (`0x00/0x01/0x02/0xFF`) in `crates/rio/src/encrypt_reader.rs` match the plan's description; no runtime path reads `RUSTFS_ENCRYPTION_CIPHER` today, as the PR body states.

### CI status (observed)
- Quick Checks: **fail** (rustfmt). Test and Lint: **fail** (dependency on Quick Checks). Cargo Deny, Dependency Review, Typos, Workflow Pin Report, CLA: pass. Upgrade/e2e lanes: pending at review time.

reatang (commented):
Thanks for the PR, and for the careful write-up — the walkthroughs clearly took real effort, and the core idea (read side dispatches purely on the on-disk frame type byte, never on configuration) is the right instinct for keeping old data readable. The contract tests in the demo crate are also a nice touch.

That said, I don't think this is ready to merge in its current shape. A few thoughts, grouped by area.

### Code structure

- **Documents under `reading-record/`.** The repo intentionally does not keep one-shot plans or code walkthroughs in the tree (see `AGENTS.md`, "Sources of Truth"); durable design notes go under `docs/architecture/` and cite symbols rather than line numbers, because line anchors go stale within days. Roughly 300 `file:line` references here would need re-verifying on every rebase — the plan header already records one such re-verification. I'd suggest moving the design into a GitHub issue or discussion, where it can be iterated on, and dropping the walkthroughs from the PR.
- **Unused dependency.** `crates/rio` gains `rustfs-aes256-gcm-demo`, but nothing in the workspace references the crate yet. It's better to introduce the dependency in the same change that uses it, so the wiring is exercised by a test.
- **Crate placement.** `vendor/` is conventionally the output of `cargo vendor`, so a first-party crate there (plus the `.gitignore` exception) will collide with that workflow. If the crate lands, `crates/<name>` with the workspace version and `publish = false` fits the existing layout better.
- **CI.** Quick Checks is currently failing at `cargo fmt --all --check`, and the branch conflicts with `main` in `Cargo.toml`. A `cargo fmt --all` plus a rebase will let the rest of the gates run.

### Design compatibility

- **Forward compatibility.** Nodes that predate the change fail closed on frame type `0x03`. That's the right failure mode, but it means the feature relies entirely on operator discipline during rolling upgrades. It would be worth writing down how a mixed-version cluster is expected to behave, and whether the write side should refuse to enable the new cipher until every node advertises read support.
- **Algorithm identity lives only inside the data frames.** Paths that move ciphertext without parsing it (replication, tiering, heal) won't know which cipher an object uses, so an object written with the new cipher and shipped to an older peer becomes unreadable there. Recording the cipher in object metadata alongside the existing SSE headers would let those paths reason about it. Related: the object still advertises `AES256` externally regardless of the actual cipher, which needs a decision rather than silently diverging.
- **What the demo proves.** Because `Aes256GcmDemo` reuses AES-256-GCM's key, nonce and tag sizes, it exercises the type-byte plumbing but not the harder question of whether a cipher with different parameters can be plugged in. Locking nonce/tag at the type level is good; I'd just be explicit in the docs that the extension point is "AES-GCM-shaped AEADs" rather than arbitrary algorithms.
- **Configuration scope.** A process-wide env var read once is simple, but algorithm choice usually ends up needing per-bucket or per-key control. Not a blocker for a first iteration, but worth noting in the design so the env var isn't treated as the long-term interface.

### Suggested path

1. Open an issue or discussion with the design and the motivating use case (which algorithms, and why), so we can settle the on-disk representation before code lands. Design review is much cheaper at that stage than in a PR.
2. Land the cipher crate together with its write/read wiring and a frame-level round-trip test, once the design is agreed.
3. In the meantime, if you'd like to keep contributing, issues tagged `good first issue` and small, self-contained fixes with a test are a great way to get familiar with the review conventions here — they tend to merge quickly and build trust for larger changes like this one.

Happy to help refine the design in the issue.

PR #7938, opened 2026-09-16
fix(heal): treat nil version id as unversioned in replacement readback

cxymds:
## Related Issues

N/A

## Summary of Changes

Automatic replacement never completed for an unversioned erasure set. Every rebuilt object was reported as:

```
Skipped heal for <bucket>/<object> because replacement target readback did not confirm the committed version
```

The pass then returns a transient skip. With `max_retries = 3` and no backoff, the three attempts burned the whole budget within seconds and left the generation in the terminal `unrecoverable` state, after which it made no further progress.

Field evidence (`1.0.0-preview.4`, unversioned, 30-minute window):

- State went `incomplete` -> `unrecoverable` in four seconds, then showed zero changes for 30 minutes (405 samples) and never reached `completed`.
- Each object's skip message appeared three times in the journal, exactly `max_retries`.
- On the physical side at the same time, the shard bytes of all 24/24 objects on the replacement disk were byte-identical to the baseline (`part.*`, `*.integrity.*`), `xl.meta` was rewritten as expected, and 405 physical checks reported zero errors.

In other words, the readback was rejecting objects that had in fact been rebuilt successfully. In the same run the versioned (DV) branch reached a `completed` receipt; only the unversioned branch failed.

Root cause: an unversioned object reaches the replacement path with the all-zero UUID as its version id (`"object":"object-003.bin","version_id":"Some(\"00000000-0000-0000-0000-000000000000\")"`). The caller forwards it verbatim (`version_id.unwrap_or("")`), so `replacement_targets_have_version` treats it as a concrete version:

```rust
if !version_id.is_empty() && file_info.version_id.as_ref().map(ToString::to_string).as_deref() != Some(version_id)
{
    return Ok(false);
}
```

An unversioned object stores no version id on disk to match; the disk layer already treats a nil version as "no version" (`crates/ecstore/src/disk/local.rs`):

```rust
if fi.size == 0 || fi.version_id.is_none_or(|v| v.is_nil()) {
```

The same nil UUID was also passed to `read_version` as a concrete version instead of the empty id the disk layer uses for unversioned reads.

The fix normalizes a nil or empty version id to the disk layer's "no version" spelling (empty string) at the top of `replacement_targets_have_version`, so both the read and the comparison operate on the null version. Versioned objects are unaffected because their ids are concrete.

```rust
fn is_null_version_id(version_id: &str) -> bool {
    version_id.is_empty() || Uuid::parse_str(version_id).is_ok_and(|parsed| parsed.is_nil())
}
```

A unit test, `null_version_ids_are_not_treated_as_concrete_versions`, covers the change.

## Verification

Tested commit: the PR head rebased onto the release merge, no local changes on top. The workspace MSRV is 1.98.1, so the commands below use that toolchain.

```
cargo +1.98.1 test -p rustfs-ecstore --lib null_version_ids_are_not_treated_as_concrete_versions
# crate compiles, test result: ok. 1 passed; 0 failed; 5546 filtered out
cargo fmt --all --check   # clean
```

The regression test is new with the fix, so no failing-before run exists; the failing-before behavior is the field evidence above, where unversioned objects were rejected while their shards matched the baseline. Not run: the end-to-end replacement scenario, which needs the E06 unversioned case rerun on a real fleet. Remaining risk: the change only affects the unversioned readback path, so versioned readback behavior is unchanged.

## Impact

Bug fix for automatic replacement on unversioned erasure sets: objects that have already been rebuilt are now confirmed by the readback instead of being reported as skipped, so the replacement generation can progress toward `completed` instead of ending in `unrecoverable`. No API, configuration, or on-disk format change.

## Additional Notes

Follow-ups not covered here:

1. Retry accounting: a single transient skip burns the three-attempt budget within seconds and marks the generation terminal `unrecoverable`, while the same pass logs "retry scheduled". The message contradicts the state (`from_state` states `unrecoverable` implies an exhausted retry budget, but the observed cause was transient), and there is no backoff.
2. Diagnosability: the per-object skip reason lives only in the system journal, is lost by bounded collection, and is absent from the durable replacement record.
3. End-to-end confirmation requires rerunning E06 unversioned on a real fleet and observing whether the generation reaches `completed`.

Rollback is a revert of the normalization at the top of `replacement_targets_have_version` and the `is_null_version_id` helper.


cxymds:

houseme (approved):

cxymds:
Superseded by #7933, which is already merged into `main` and fixes the same root cause: `filemeta::FileMeta::to_fileinfo_with_part_checksums` now treats a nil UUID selector as the null version slot (`matches_version`), and `replacement_targets_have_version` compares parsed UUIDs instead of version-id text.

The only delta left on this branch is the `is_null_version_id` normalization, and it is not equivalent. It rewrites the nil selector to an empty version id, and `to_fileinfo_with_part_checksums` uses `if !version_id.is_empty() && !matches_version(...)` to decide whether to filter by version at all. An empty id therefore skips that filter and selects the newest non-free version instead of the null slot, which is the opposite direction from #7933: a target whose null slot is absent but which holds a concrete version would be reported as confirmed. Merging it would loosen the guard #7933 added, so this PR is closed rather than merged.

The field evidence and the retry-accounting / diagnosability follow-ups here are already tracked by backlog#2559 (fix items 3 and 4), which #7933 references.

PR #7941, opened 2026-09-16
fix(ecstore): reuse metadata read guards during writes

jkossis:
## Related Issues

N/A. Found during three-site replication qualification of RustFS 1.0.0.

## Summary of Changes

A replica PUT can hold an Object Lock snapshot's bucket metadata read lock while initial sync queues a metadata writer. Quota admission then requests another read lock. Writer preference blocks that reader behind the writer, while the writer needs the original reader to finish. The join reports `initialSyncErrorMessage` after the five-second transaction-lock timeout. Multipart generation validation and post-write lifecycle expiry can enter the same cycle.

Reuse the existing guard after validating its store, bucket, generation, and lock-loss state. Quota reservations retain shared ownership through commit and cleanup; generation checks still read the authoritative disk identity. Post-write expiry uses the owning store and request options. Single and batch deletes share their externally held guard with their snapshots while retaining the outer guard lifetime. Runtime deadlines and retry behavior are unchanged.

Correct the simulated-peer fixtures so cloned pools and disk sets use their peer's context. The restore fixture gives both peers the same tier configuration, including its persisted backend prefix, and the same mock backend. An early restore failure now surfaces directly instead of appearing as a barrier timeout. Reserve nextest's full thread budget for the one real-server Connect performance test, whose short measurement includes bucket creation and cleanup. The existing test deadlines, assertions, and zero-retry policy remain intact. The imported dashboard report-upload step also receives its required two-minute CI deadline.

## Verification

- Runtime regression evidence on `015e7b9a66d24881ba8f68c7ae951da59c2bb0ed`, before the maintainer merged upstream: all six `metadata_read_lock_reuse_test` cases passed 100 consecutive parallel suite runs (600 executions), with `RUST_MIN_STACK=4194304`. Coverage includes queued-writer progress for PUT/multipart with lifecycle configured, exact versions/bodies, quota rejection, cancellation, foreign snapshot rejection, and prefix deletion. The forced-contention regressions failed before the fix. A native arm64 server built with Rust 1.98.1 and `--no-default-features` passed 100 fresh three-site joins with three pre-existing versions: 900 direct peer-state comparisons and 1,800 signed version/body reads with proxying disabled. No `initialSyncErrorMessage` or metadata transaction-lock timeout occurred. The published 1.0.0 image reproduced the timeout on attempt 20 under the same multi-version scenario; its one-version control passed 100 joins.
- Fixture and runner changes in `cd37e161` / `fd767d52`: the ECStore unit-test binary passed all 47 `decommission_lock_order_tests`, each in a separate process with no retries. The nine `connect_perf_object` tests passed once and then passed all five `cargo nextest run --profile ci --stress-count 5` iterations: 54 executions, no skips or retries. JUnit timestamps confirm the real-server case did not overlap the other eight tests with four nextest workers. These local binaries used native aarch64 Linux and Rust 1.98.1; Connect used `--no-default-features`. They were compiled from `302f245f` plus this fixture/configuration diff before committing it.
- `cargo fmt --all --check`, `git diff --check`, and `scripts/check_architecture_migration_rules.sh` passed. The earlier CI upload correction passed all 31 `scripts/test_functional_chain.py` tests and actionlint with shellcheck. Full workspace and feature-matrix execution remains a CI gate; it was not run locally.

The completed [CI run on `302f245f`](https://github.com/rustfs/rustfs/actions/runs/35125616589) resolved all eight earlier decommission failures. It exposed the restore fixture's missing peer tier and one Connect deadline failure: Workspace had 18,595 passes and two failures; rio-v2 had 11,058 passes and only the restore failure. Connect passed in rio-v2 and in the previous Workspace run. Comparing the exact CI merge commits found no production-code differences between those runs. The failing Workspace runner recorded substantial I/O pressure with 16 competing test workers; reserving runner capacity addresses that contention without extending the performance budget. Lifecycle, SFTP, Swift, distributed E2E, and all seven upgrade checks passed on `302f245f`. Fresh [CI on `fd767d52`](https://github.com/rustfs/rustfs/actions/runs/35134601265) is pending and must confirm the final diff.

## Impact

Prevents nested metadata-read acquisition from blocking replication and other writes behind a metadata writer. No S3, wire, disk-format, configuration, or runtime timeout change. One performance test runs without competing tests in the same nextest invocation. No production deployment is included. Reverting the runtime change restores the previous lock-acquisition behavior.

The qualified `015e7b9a` binary SHA-256 is `33636bb6fe9cd3bf848db8efefba776531ed314db08ffb099c5ff06782a13e93`; its local image ID is `sha256:d8d22bc8507facc3b5d6e877b57eea7b0a59488c2d02f6722a929a90ab9d1d81`. It was not published or deployed. All disposable test containers were removed. That runtime qualification does not substitute for CI on the final merged source.

## Additional Notes

Two sequential solo review passes covered the runtime change; the test/configuration follow-up also passed correctness and simplicity review:

- Correctness: authoritative reads and error propagation remain intact; peer fixtures retain matching ownership and tier identity. No findings.
- Simplicity: shared metadata helper, standard `Arc` ownership, and existing nextest capacity controls; no new dependency or lock manager.
- Test coverage: failing-before contention proof, repeated runtime tests and joins, all 47 decommission cases, and repeated Connect coverage including deadline/cancellation behavior.
- Security: invalid snapshot scope fails closed; Object Lock and lock-loss checks remain enforced.
- Concurrency/durability: guards remain owned through settlement and cancellation; no deadline increase or lock bypass.
- Compatibility: no persisted or transmitted representation changes.
- Performance: guard cloning replaces repeated lock acquisition, with a small `Arc` allocation for newly owned guards. No throughput claim is made.


overtrue (approved):

overtrue (approved):
Reviewed merge-base `3e7a6aa63` → head `015e7b9a6` (the recorded baseRefOid `58d1296a2` is not an ancestor of the head, so the merge-base was used). Risk tier: high (locking, multipart, lifecycle); two independent static passes plus local verification.

**Assessment.** The diagnosis holds: a replica PUT that overwrites an existing version builds an Object Lock snapshot in `prepare_put_object` and keeps the bucket metadata transaction read lock, then `reservation::begin` requests a second read lock that queues behind a waiting metadata writer, which in turn needs the first reader to finish. Sharing the snapshot's `Arc` guard after validating store id, bucket, incarnation, and lock-loss state removes the nested acquisition. Points verified:

- `metadata_transaction_guard_for` only shares when `(store_id, bucket, incarnation)` all match and neither fence is lost; a foreign or stale snapshot fails closed instead of silently taking a fresh lock. The new tests cover cross-store, cross-bucket, and wrong-incarnation rejection.
- `get_quota_config_and_incarnation_from_disk_in` and `get_authoritative_metadata` take only the bucket namespace read lock, so the fixed paths no longer perform any nested transaction-lock acquisition.
- Delete and batch-delete snapshots now carry the already-held metadata guard; the outer `_object_lock_metadata_guard` lifetime is unchanged and incarnation mismatches are still rejected with `BucketNotFound` at the store layer before `validate_bucket_incarnation` runs.
- Post-write expiry uses the request options: plain PUTs carry no snapshot and the internal one is dropped when `handle_put_object` returns, so the later acquisition is sequential, not nested; snowball extract and same-key copy carry their own snapshot and reuse it. Expiry actions are enqueued, not executed inline under the shared guard.
- Lock order, deadlines, and retry behavior are unchanged; `QuotaReservation` holding a shared `Arc` does not extend the guard's lifetime beyond the snapshot that already owned it.

**Local verification at head.**

- `RUST_MIN_STACK=4194304 cargo test -p rustfs-ecstore --features test-util --test metadata_read_lock_reuse_test`: 6 passed.
- RED check: with `crates/ecstore/src` reverted to the merge-base and the new tests kept, 4 tests fail with the expected 2s timeout (PUT, multipart, quota, foreign-snapshot rejection); the cancellation and prefix-delete tests pass on the baseline as guards for the new ownership semantics.
- `cargo clippy -p rustfs-ecstore --features test-util --tests`: clean.

**P3 (non-blocking).** The PR applies the shared-guard helper to copy, restore, and metadata-update paths, but the regression tests exercise only PUT, multipart, and prefix delete. Those three paths check out statically; a copy or restore case with a waiting writer would close the gap in a follow-up.

**Checks.** All required checks pass except `Workspace Test and Lint` and `Test and Lint (rio-v2)`, which were still pending at review time; merge once they are green.

jkossis:
@codex

PR #7943, opened 2026-09-16
Allow 127.0.0.0/8 as loopback addresses for binding

aredridel:
## Related Issues

N/A

## Summary of Changes

This is particularly useful for integration testing of services that use an S3 API, as it lets many different instances bind a port on different IPs.

It is relatively straightforward: on Windows, we retain the existing behavior; under unix, we add the entire 127.0.0.0/8 range to the test.

## Verification

Before (`main@2bf77c1c1936f71ee950eaf8603dc86f86d5faec`): 

```
:; target/debug/rustfs server --address 127.0.5.6:9090 .
{"timestamp":"2026-09-16T09:14:52.509812906-04:00","level":"ERROR","message":"Server runtime failed","event":"server_runtime_failed","component":"main","subsystem":"startup","error":"host in server address should be this server","target":"rustfs::main","filename":"rustfs/src/startup_entrypoint.rs","line_number":172,"threadName":"main","threadId":"ThreadId(1)"}
[FATAL] Server runtime failed: host in server address should be this server
```

After:

```
:; target/release/x86_64-unknown-linux-gnu/rustfs --address 127.0.5.8:9090 .
```

It's accepted.

and 

```
:; target/release/x86_64-unknown-linux-gnu/rustfs --address 192.0.0.0:9090 .
{"timestamp":"2026-09-16T11:12:25.566624216-04:00","level":"ERROR","message":"Server runtime failed","event":"server_runtime_failed","component":"main","subsystem":"startup","error":"host in server address should be this server","target":"rustfs::main","filename":"rustfs/src/startup_entrypoint.rs","line_number":172,"threadName":"main","threadId":"ThreadId(1)"}
[FATAL] Server runtime failed: host in server address should be this server
```

Non-local is rejected.

A `--allow-non-local` flag would be an okay alternative, since Linux does allow non-local binds, but that's a much more niche use-case.

## Impact

This strictly allows more, and doesn't reduce function for anyone. No impact expected.

## Additional Notes

N/A


aredridel:

overtrue (approved):
## Review: PR #7943 — Allow 127.0.0.0/8 as loopback addresses for binding

Reviewed base `e9f8beb69` → head `27ae62492` (merged locally onto `main` `cd1098113` for verification). Risk tier: low (single helper in `crates/utils/src/net.rs`, behavior strictly widens the local-host check; no metadata, tiering, auth, or RPC surface).

### Assessment

The change is sound. `is_local_ip` keeps the interface-configured set as the first check and only adds `IpAddr::is_loopback()` on non-Windows targets, so previously accepted addresses are unaffected and the port gate in `is_local_host` still applies. The distributed-layout guard at `crates/ecstore/src/layout/endpoints.rs:435` ("all local endpoints should not have different hostnames/ips") still rejects a cluster that lists several 127/8 addresses on one port as all-local, so the widening does not create a new silent double-ownership path there.

### Findings

**P3 — `cfg(not(windows))` is broader than the comment's claim.** `crates/utils/src/net.rs:233-239`. The comment says "Unix kernels route and bind the entire RFC 1122 loopback space", but that is Linux behavior. macOS (and other BSDs) only configure `127.0.0.1` on `lo0` unless an alias is added; on this Darwin host `bind(("127.0.5.9", 0))` fails with `EADDRNOTAVAIL`. Effect: on macOS `check_local_server_addr("127.0.5.9:9000")` now passes and the server fails later at listener bind with "Can't assign requested address" instead of the earlier "host in server address should be this server". Startup still fails, so this is not a correctness regression, and the new unit tests do not bind so they stay green on macOS. Suggested fix: gate on `#[cfg(target_os = "linux")]` (or `any(target_os = "linux", target_os = "android")`) and keep the `false` arm for everything else. BSD users who alias extra loopback addresses are already covered by the interface enumeration in `LOCAL_IPS`, so narrowing loses nothing.

No other findings.

### Verification

- `cargo test -p rustfs-utils --all-features --lib net::` on the merge with current `main`: 15 passed (includes the new `127.0.5.9` cases).
- `cargo clippy -p rustfs-utils --all-features --all-targets -- -D warnings`: clean.
- `cargo fmt -p rustfs-utils -- --check`: clean.

### CI

`Quick Checks` failed in `scripts/test_functional_chain.py::test_non_performance_suites_have_hard_and_step_deadlines` (fault-tolerance workflow's "Upload functional report to dashboard" step lacks `timeout-minutes`). This is pre-existing: the step was introduced by #7939 (`e9f8beb69`, this PR's base) and fixed on `main` by #7948 (`f7431ce74`); the script passes on current `main`. `Test and Lint` failed only as the downstream aggregate of that. A re-run on top of current `main` should be green. No PR-caused failures.

### Verdict

APPROVE (P3 is non-blocking; happy to see the cfg narrowed before or after merge).

PR #7944, opened 2026-09-16
fix(ilm): keep expiry pending gauge balanced and drain overwrite tails

### LINKED ISSUE #7921: test: tier overwrite cleanup recovery intermittently times out
## Failure

The `rio-v2` release CI failed `store::init::tests::tier_overwrite_put_and_self_copy_recover_persisted_cleanup_owners` after 31.4 seconds. `wait_for_tier_free_version_recovery` timed out at `crates/ecstore/src/store/init.rs:12608` with `tier free-version recovery should complete: Elapsed(())`.

[Failing job](https://github.com/rustfs/rustfs/actions/runs/35007430339/job/104513121581): commit `a708820cda9ab30127bb4a6d16446e12181808d4`, 11,040 tests passed and one failed.

The same test passed in 3.4 seconds in [the preceding PR's rio-v2 job](https://github.com/rustfs/rustfs/actions/runs/34999108873/job/104485273612), at commit `439d3d74fb14e4473fddfe030fee53cc354e3e98`. Both commits have tree `0b38de5eb7aab94cb7341f10006bcd7b883040b9`.

## Investigation

The fixture restarts a four-disk store, injects a failed remote cleanup, then drives durable free-version recovery until the owner disappears. The timeout does not yet identify which matrix case or cleanup condition stopped progressing. The original timeout has not been reproduced. The existing CI override already reserves the full test-thread budget for this test.

## Expected result

Identify why cleanup occasionally fails to converge, add a deterministic regression or useful failure diagnostics, and retain the final assertions for one remote deletion, durable-owner removal, and readable replacement bytes. This currently blocks 1.0.0 preview acceptance.

## Follow-up evidence

- The unchanged preview source passed 10 focused macOS repetitions.
- Diagnostic PR #7922 adds failure context without changing the 30-second deadline or assertions. Its head `78e85b0545dc822d0760da3488c54bdaea4e50c9` passed 100 macOS repetitions and 50 Linux ARM repetitions with `test-util,rio-v2,gcs` and a 4 MiB Rust stack.
- [The diagnostic PR rio-v2 job](https://github.com/rustfs/rustfs/actions/runs/35017948966/job/104548389832) passed all 11,041 tests; this case passed in 3.393 seconds. Workspace, upgrade, and distributed gates also passed.

One explicit rerun of the original failed CI job will run at the unchanged preview SHA `a708820cda9ab30127bb4a6d16446e12181808d4`. The original failure remains recorded here. A passing rerun will not establish a root cause or close this issue; no quarantine, automatic test retries, assertion changes, or timeout increases have been added.

The [tracked original-SHA rerun](https://github.com/rustfs/rustfs/actions/runs/35007430339/job/104589671410) passed all 11,041 rio-v2 tests; this case passed in 3.578 seconds. The full CI run is now green at the same preview SHA. The original intermittent timeout remains unexplained, and this issue stays open. Final release approval must explicitly account for this unresolved risk.

overtrue:
## Related Issues

Fixes #7921.

## Summary of Changes

Two independent defects made `tier_overwrite_put_and_self_copy_recover_persisted_cleanup_owners` time out intermittently. Both reproduced locally by running the test under CPU contention (12 concurrent instances); the failure rate before this change was roughly 1 in 8 runs.

**Expiry worker pending gauge drift (product fix).** `ExpiryStats` updated the pending and active gauges through a saturating helper that clamps at zero, while `send_expiry_task` and `enqueue_recovered_free_version_with_state` incremented the pending gauge only after `try_send` succeeded. When the worker dequeued before that increment landed, its decrement clamped the transient -1 to 0 and the later +1 left a permanent phantom pending task with nothing in the channel. Locally this surfaced as the fixture's `wait_for_expiry_workers_idle` timing out with `pending=1 active=0` and a live, idle worker. The gauges now use plain `fetch_add`/`fetch_sub` (readers already clamp negative snapshots), and both enqueue paths count before publishing and roll back on a full queue. The `queued` value exported to scanner lifecycle metrics was also affected by the same drift.

**Overwrite tail lost before the simulated restart (test fix).** The PUT commit path acknowledges at write quorum and drains the remaining rename fan-out in a background tail. The fixture "crashed" the store right after the overwrite, so on a loaded machine one disk still held the pre-overwrite live transitioned source while the other three held the replacement plus its cleanup owner. Exact cleanup then failed closed on every retry (`a live transitioned source still references the free-version remote tuple`), which is the intended minority-copy protection until heal repairs the disk, but this fixture never runs heal. This is the shape of the original CI failure: the recovery loop retried for 30 seconds while `enqueued` stayed at 1. The fixture now requests `WriteCompletion::TailDrained` for the seed, overwrite, and self-copy writes and asserts that every physical disk holds exactly one cleanup owner and no live source before restarting. The idle helper reports the gauges when it times out.

## Verification

Tested commit: the head of this branch, applied on top of `58d1296a2`.

- Reproduction before the fix: with the same binary features as the failing CI lane (`rio-v2,test-util,gcs`), 12 concurrent instances of the test failed 5 of 36 runs; with the pre-restart idle wait removed to match the timing of the original CI commit, per-disk `xl.meta` dumps at timeout showed disk0 still carrying the old live transitioned source while disks 1-3 carried the replacement and the free-version owner.
- After the fix: `cargo nextest run -p rustfs-ecstore --features rio-v2,test-util,gcs -E 'test(tier_overwrite_put_and_self_copy_recover_persisted_cleanup_owners)'` passed 96 of 96 runs under the same 12-way contention (plus 96 of 96 with only the gauge fix applied and the pre-restart wait in place).
- New unit tests: `expiry_pending_gauge_survives_dequeue_landing_before_enqueue_accounting` fails on the previous saturating gauge and passes now; `free_version_enqueue_rolls_back_pending_when_queue_full` pins the rollback on a full queue (the count-then-publish ordering itself is not observable single-threaded; the non-saturating gauge makes either ordering safe).
- Adversarial check that `TailDrained` alone closes the second defect: with the pre-restart idle wait temporarily removed again (the timing of the original CI commit), the final code passed 72 of 72 runs under the same contention, where the previous run of that timing without `TailDrained` failed 2 of 12; the new per-disk assertion never fired.
- `cargo nextest run -p rustfs-ecstore --features rio-v2,test-util,gcs -E 'test(bucket_lifecycle_ops::tests)'`: 203 passed.
- `cargo clippy -p rustfs-ecstore --all-targets --features rio-v2,test-util,gcs -- -D warnings`: clean.
- `make pre-commit`: passed.

Not run: the full workspace test suite and e2e lanes; the change is confined to the expiry gauge accounting and to this fixture, so the CI matrix covers the rest.

## Impact

No API or configuration change. The lifecycle expiry `queued` and `active` metrics no longer drift upward after enqueue/dequeue races. A store that crashes between the quorum acknowledgement and the tail drain of an overwrite still leaves a minority disk with the old live source; exact free-version cleanup keeps deferring for that object until heal converges the copy, which is the existing fail-closed behavior and is unchanged here.

## Additional Notes

The pre-restart idle wait added by #7922 masked the second defect most of the time because it gave the background tail time to land; it is kept, and `TailDrained` makes the requirement explicit instead of timing dependent.


overtrue:
The Quick Checks failure here is inherited from main, not from this diff: the merge ref picks up #7939, whose new fault-tolerance dashboard upload step has no step deadline, and `scripts/test_functional_chain.py` fails on main since e9f8beb69 for the same reason. #7948 fixes that workflow. This branch will be refreshed from main once it lands so the required checks can rerun.

houseme (approved):

Issue #7962, opened 2026-09-17
A slow lock endpoint produces hours of sustained lock-RPC timeouts (~1,300/s per client) with no backoff, failed releases, and lost held locks


stevapple:
**Describe the bug**

#7363 reported that lock-RPC timeouts against one slow lock endpoint spiralled into an `RST_STREAM` / `GOAWAY too_many_resets` / re-dial loop. #7473 fixed that loop: on 1.0.0 we see zero `GoAway` and zero `Evicting cached remote lock connection` lines. But the underlying condition — one endpoint answering lock RPCs slower than the 3 s deadline — now persists for hours instead of about one, at a much higher rate, and nothing in the client slows down in response:

- Every peer retries against the slow endpoint at full speed. Median **1,307 `Remote lock RPC timed out` lines per second** on one client (max 2,524/s), all with `addr = https://node3…:9000/mnt/rustfs11`, `timeout_ms = 3000`; 3.5–5.0 M per hour per node, on all three peers, continuously since 09:01:36 (+08:00) — three and a half hours and still running when this was written (a re-measurement at 12:33 showed ~775/s on another client, 99.9 % of it against the same endpoint) — after a first window 04:52:33–~06:00. On rc.5 the same storm was 5–10 k per hour.
- The endpoint is slow, not gone — and that is what keeps it in rotation. [`eviction_verdict`](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/ecstore/src/cluster/rpc/remote_locker.rs#L138-L159) suppresses eviction while `last_success` is younger than two deadlines, and detached RPCs that complete late [refresh `last_success`](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/ecstore/src/cluster/rpc/remote_locker.rs#L403-L462). Zero evictions across ~11 M timeouts means the verdict was `PeerRecentlyServed` every time: the endpoint keeps completing *some* RPCs, mostly after their callers gave up. That is correct as a channel policy, but there is no *request* policy behind it: no backoff, no per-endpoint circuit breaker, no "skip this locker, quorum is reachable without it".
[...]

PR #7982, opened 2026-09-17
fix(scanner): resume and persist checkpoints across leader handoff

### LINKED ISSUE #7964: Scanner: let a cycle's checkpoint survive a leadership change, and save it periodically, so multi-day cycles can complete

On 1.0.0 a scanner cycle's per-bucket progress is persisted as a checkpoint inside the bucket's `.usage-cache.bin`, but that checkpoint is only usable by the leadership term that wrote it. Every leadership claim increments the epoch ([`persisted_epoch.checked_add(1)`](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/scanner/leadership.rs#L360)), and [`prepare_bucket_checkpoint`](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/data_usage_define.rs#L913-L921) treats a cache as reusable only if `self.info.leader_epoch == leader_epoch`; otherwise it does [`*self = Self::default()`](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/data_usage_define.rs#L948), keeping only `pending_heals` and `size_reconciliation`. This is deliberate — the fixture test [asserts exactly this](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/scanner_folder/tests/checkpoint_fixture.rs#L385-L393) (a cache saved at epoch 7 opened at epoch 8 → `Reset`, empty) — so as designed the checkpoint serves pause/resume *within* one leadership term. A failover, or any restart of the leader, restarts every in-progress bucket from the beginning of its namespace, with the cycle number unchanged.

Two more properties mean there is usually nothing to resume from anyway:

1. There is no periodic checkpoint save during a bucket scan. The only three save sites in the bucket loop ([complete](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/scanner_io/io_cache.rs#L1225), [partial](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/scanner_io/io_cache.rs#L1287), [final](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/scanner_io/io_cache.rs#L1363)) all sit in the result handling after the bucket scan returns (complete, cancelled or budget-stopped). With one bucket of several hundred million objects, the only checkpoint that can exist is the one written at the moment the scan is interrupted.
2. That save is skipped when the bucket cache lock was lost ([`if cache_guard.is_lock_lost() { false }`](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/scanner_io/io_cache.rs#L1283)), and on leader-lock loss the cycle gets [30 s to wind down and is then dropped](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/scanner/cycle_state.rs#L4222-L4226) ([`SCANNER_LOCK_LOSS_SHUTDOWN_TIMEOUT`](https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/crates/scanner/src/scanner.rs#L266)). Losing leadership is precisely when the locks are lost, so the one save opportunity is the one that is forfeited.

The consequence on a large namespace: a cycle that needs ~10 days can only complete if no node restarts and leadership is never lost for ten days in a row — and leadership is easy to lose, since one lock endpoint answering slower than the lock-RPC deadline for about 30 s takes `leader.lock` away from a healthy leader (#7962). Every rolling upgrade also resets it by construction. We have now lost five consecutive cycles (three on rc.5, two on 1.0.0), and no data-usage snapshot has been published since 2026-07-24 — so usage-dependent features (quota, usage reporting, anything gated on a completed cycle) have been dark for that long.

1. **A checkpoint resumable by the next leader of the *same cycle*.** The epoch is the right fence against a stale leader *writing*; it need not invalidate what a previous, legitimately fenced-off leader already persisted. If the concern is a half-written frontier, the sealed coverage frontier (`seal_scan_frontier`) already describes what was covered, and a new leader could resume from the last sealed point rather than from zero.
2. **Periodic checkpoint saves during a long bucket scan** (time- or object-count-based), not only when the scan returns — otherwise the only save opportunity is the moment of failure.
3. **Do not forfeit the save outright on cache-lock loss.** The save is already CAS-protected by revision, so a fenced writer cannot clobber a newer leader's state; a best-effort final save (or relying on the last periodic one from 2) would preserve hours or days of progress.

**Describe alternatives you've considered**

- *Fixing only the lock layer (#7962).* Making held locks tolerate one slow locker would remove our most frequent trigger, and we want it regardless — but it does not help restarts, deployments, or an OOM-killed leader, all of which change the epoch. A rolling upgrade would still discard the cycle every time.
- *Using the scanner runtime budget to force saves.* A budget stop does write a checkpoint, so a short budget would bound the loss within a term — but the checkpoint still dies with the epoch, so it helps only against interruptions that do not change leadership, which is none of the ones we have had.
- *Shrinking the unit of work.* If buckets were small, restarting "every in-progress bucket" would be cheap. Ours is dominated by one JuiceFS bucket of several hundred million objects, and the bucket layout is dictated by the client; splitting it is not an option we control.
- *Pinning leadership / avoiding restarts.* We already serialize restarts and avoid them where we can; five lost cycles in seven weeks is the result with that discipline in place.

**Additional context**

What the loss looks like in practice — 4-node, 3-set, EC 12+4 cluster, no node restarted, times +08:00 on 2026-09-17:

1. Cycle 9846 running on node2 since 09-16 20:14, checkpoint walking `juicefs-bucket/jfsvol/chunks/…`, at `chunks/138/…` by 04:52.
2. 04:52:33 lock RPCs from all peers to one endpoint start timing out at 3000 ms (#7962).
3. 04:53:03 node2: `lock refresh lost quorum` on `.rustfs.sys/leader.lock`; 04:53:09 same on `.rustfs.sys/buckets/juicefs-bucket/.usage-cache.bin/.scanner-cycle.lock.pool-0.set-2`.
4. 04:53:33 node2: `Scanner leader lock lost during a scanner cycle` — exactly the 30 s shutdown timeout after step 3.
5. 04:53:42 node4 claims leadership, same cycle 9846, and walks the bucket from the start. Admin scanner status on node4: `scan_checkpoint_used = 0`; node2's `scan_checkpoint.resume_after` is empty. Seven hours later node4 is at `chunks/136/…` — i.e. it re-did the ground node2 had already covered.

A deterministic way to see it: start a cycle on a bucket large enough to take minutes, restart the leader (or block one of its lock endpoints for longer than the lock TTL), and compare the new leader's first listed prefix with the old leader's last checkpoint.

Environment: RustFS `1.0.0` (`d47f54bfb2f39f48bd1adda334bd27e151fe85b8`); 4 nodes × 12 NVMe, 3 erasure sets, EC 12+4, TLS internode; ~500 M objects dominated by one JuiceFS bucket; scanner runtime budget disabled (`0`), so a bucket scan never stops for budget and never saves mid-scan. `crates/scanner` is unchanged between the 1.0.0 tag and current `main`.

Two asides. The new leader also logged `Scanner cache is corrupt and will be rebuilt` three times for `cache_name=".usage-cache.bin"` (the per-set root aggregate); that is not the cause of the loss — the root aggregate is rebuilt each cycle — but we do not know why it fails to decode. And for history: on rc.5 any leadership change discarded the cycle outright, which is what 1.0.0's checkpointing looked like it would address; from the operator's side nothing changed — same cycle number, progress back to zero.

Filtered logs of 04:45–05:00 from all four nodes are available on request, and we are glad to test a build: this cluster loses leadership often enough to validate a resume path within days.

houseme:
## Related Issues

- Fixes rustfs/rustfs#7964
- Tracks rustfs/backlog#2585
- Related to rustfs/rustfs#7962

## Summary of Changes

- Allow a bucket checkpoint to cross a scanner leader handoff only when the cycle, bucket identity, source, tier generation, cache format, and sealed frontier/raw enumeration proof all match.
- Decouple the new coverage digest domain from the process-local leader epoch and migrate legacy epoch-bound receipts after validating them under the old epoch.
- Add bounded periodic checkpoint snapshots for local and remote namespace workers (1024 objects or 60 seconds, with a 5-second minimum interval and a capacity-one queue).
- Centralize checkpoint persistence fencing and CAS-revision refresh so local and remote workers share the same publication/leader validation and main/backup revision contract.
- Keep fail-closed reset behavior for missing/corrupt proof, cycle or identity changes, and stale writers.

## Verification

- `cargo fmt --all --check`, `git diff --check`, and `cargo clippy -p rustfs-scanner --all-targets -- -D warnings` pass on commit `be803820b`.
- Focused scanner tests pass: `checkpoint_fixture_` (17), periodic checkpoint threshold, legacy digest migration, scanner cache lock tests (2), and the existing leader-fence regression.
- A full `cargo test -p rustfs-scanner --lib` run reached the existing recovery-control stack overflow in `scanner::tests::full_rescan_reset_preserves_unknown_phase_and_retries_completed_cleanup`; this failure is outside the changed checkpoint paths, so the full crate is not claimed green.
- Docker validation: a main-based four-node image with 3,000 objects reproduced deadline fencing and leader-lock quorum failure after node interruption. The rebuilt fix image started a fresh four-node cluster with all four `/health` probes returning HTTP 200. A long soak, mixed-version upgrade, and deterministic stale-writer CAS interleaving were not run.

## Impact

Internal scanner cache recovery and persistence behavior changes. Legacy checkpoint receipts remain readable and are migrated only after validation. No S3 API, protobuf, or user configuration schema changes are introduced. Periodic saves add bounded cache serialization and metadata writes during long scans.

## Additional Notes

The change was independently reviewed for correctness/test coverage and concurrency/durability/compatibility/performance. The local and remote save paths were refactored around one shared fenced persistence helper to avoid divergent patch-on-patch behavior. The patch is intentionally limited to scanner checkpoint durability and leader handoff; lock-RPC storm remediation remains in PR #7970.


houseme:
Follow-up for the distributed observability failure:

- The bare local `cargo test` invocation was invalid because the repository requires the `scripts/e2e_binary.py` receipt wrapper.
- Reproducing with the wrapper passed repeatedly; the failure was a recovery-window race in the test. It waited for fresh peer probes and `/health/ready`, but not for all 16 remote disks to return to `ok`/`online`.
- Commit `6bafe9cdd` now waits for both fresh observations and full disk recovery before issuing the recovered PUT.

Validation: the exact wrapped test passed after the change.

houseme:
The reported `test_security_workflow.py` failure was from the pre-merge-main workflow shape. The PR branch now contains the latest `main` merge (`defc62880`), where `rustfs-performance-test.yml` uses the shared `rustfs-shared-functional-tests-v2` concurrency group. Re-running `python3 scripts/test_security_workflow.py` on the current head passes all 30 tests, so no additional workflow patch is needed.

houseme:
The security-workflow failure was caused by the automatic main synchronization reintroducing the performance workflow's private `rustfs-performance-suite` concurrency group. The shared-lock test correctly caught that merge result.

Commit `e76d47480` restores `rustfs-shared-functional-tests-v2` for the performance workflow. `python3 scripts/test_security_workflow.py` now passes all 30 tests locally; the fix has been pushed and CI is restarting on the corrected head.

houseme:
Correction after reviewing merged PR #7984: performance intentionally uses its own `rustfs-performance-suite` concurrency group because it runs on a separate pf-testing fleet. The correct fix is therefore in commit `1c33b561e`: the security workflow test now expects the dedicated performance group while retaining `rustfs-shared-functional-tests-v2` for the functional suites. `python3 scripts/test_security_workflow.py` passes all 30 tests.

houseme:
Follow-up for the latest distributed observability failure:

- The disk/peer recovery gate was already satisfied, but logs showed remote lock RPC timeouts, lease-release cleanup, and a transient S3 `503 ServiceUnavailable` during lock convergence.
- Commit `151bdfe18` keeps the recovery gate and adds a bounded retry around the recovered PUT. It records exactly one successful PUT while allowing transient 5xx attempts during lock convergence; the two intentional sub-quorum failures remain part of the denominator.
- The exact `scripts/e2e_binary.py` wrapped test passed after this change.

Issue #7958, opened 2026-09-17
false EOF in paginated listings: gather_results infers completion from post-filter count, not producer stop reason


HateTM:
### Summary

Follow-up to #7049 (merged 2026-09-03), which fixed one symptom of a broader
pagination bug (`scan_dir` emitting a later-sorting sibling entry after a
recursive subdirectory scan hit its limit).

During review of that PR, a deeper, still-open issue was flagged: the
disk-side producer (`LocalDisk::scan_dir`) never signals *why* it stopped
sending entries — whether it exhausted its own page/per-disk limit, or
genuinely reached the end of the tree. `gather_results`
(`crates/ecstore/src/store/list_objects.rs`) currently infers completion
purely from whether `entries.len() >= opts.limit` was hit on its own side;
when the input channel simply closes without that happening, it unconditionally
reports `GatherResultsState::InputClosed` with `err = Some(Unexpected)`,
which every call site in the file reads via `disk_has_more = err.is_none()`
as "no more data, not truncated".

### Impact

The per-disk scan limit (`per_disk_limit = opts.limit + 4 + opts.limit / 16`,
~6.25% headroom, see `SetDisks::list_path`) is sized for the *final*
candidate count, not the *raw* scanned-entry count. When a large fraction of
raw entries gets filtered out downstream (delete markers, non-matching
prefix/marker skips, directory entries — one production report measured
~62% filtered), the per-disk scan can exhaust its limit deep inside a
subtree, close its channel, and `gather_results` will report a confident,
non-truncated EOF — even though most of the real object tree was never
scanned. This causes silent, permanent data loss in paginated
`ListObjectsV2` (and structurally the same code path is shared by other
list-family APIs).

Reported production case: recursive listing over ~1.17M objects with
`MaxKeys` pagination continued to drop entries after #7049 was deployed.

### Minimal reproduction

Attached: `repro.diff` — a self-contained unit test added to
`crates/ecstore/src/store/list_objects.rs`'s existing test module
(`list_path_gather_results_reports_false_eof_when_producer_stops_at_its_own_limit_after_heavy_filtering`).

It feeds `gather_results` 20 raw entries that are entirely filtered out
(delete markers, `incl_deleted: false`) and then closes the channel —
standing in for a per-disk `scan_dir` call that hit its own limit deep
inside a much larger subtree it never finished walking. `gather_results`
returns `GatherResultsState::InputClosed` with `err = Some(Unexpected)`:
[...]

PR #7994, opened 2026-09-18
fix(ecstore): preserve bounded listing completion reason

houseme:
## Related Issues

- rustfs/rustfs#7958
- rustfs/backlog#2605

## Summary of Changes

Bounded disk listing producers previously closed their metacache stream without identifying whether they had naturally exhausted the tree or stopped at their own scan budget. After downstream marker, prefix, directory, and metadata filtering, `gather_results` could interpret that close as a definitive EOF and return `IsTruncated=false` while keys remained.

This change keeps the completion state in-process: the local walker reports `LimitReached` through a shared `AtomicBool` carried by `WalkDirOptions` and `ListPathRawOptions`. The state is observed once at producer completion and carried through ECStore/Sets/SetDisks to `gather_results`. No per-entry synchronization or on-wire metacache format change is added. Remote/legacy peers retain the existing behavior until an explicit capability is added.

## Verification

Tested commit: `30bcaf782bb77e8c84ca72baab7961f6512a8745`.

- `cargo check -p rustfs-ecstore -p rustfs-filemeta` passed.
- `cargo fmt --all --check` and `git diff --check` passed.
- Focused bounded-producer/gather regression passed.
- All three previously SIGTERM-terminated local scan tests passed individually: base-dir metadata, nested object directories, and explicit directory-marker deduplication.
- `disk::local::test::scan_dir` passed 3/3, and version pagination exact/one-over-limit coverage passed.

The first CI attempt exposed a trailer-framing issue in the earlier implementation; that protocol experiment has been removed from this final patch. The Docker check confirmed ordinary `MaxKeys=1` and delimiter pagination with a prebuilt RustFS image. A main-source image build was not completed within the available build window, and standard S3 operations could not reliably preserve pending version-purge metadata, so the full LocalDisk-to-HTTP false-EOF reproduction remains unverified.

## Impact

ListObjectsV2 and the shared local listing layers now preserve truncation when a bounded local producer stops early. No quorum, cancellation, or client-visible wire request shape is intentionally changed. The shared state adds one optional `Arc<AtomicBool>` per bounded listing and one release store at producer completion. Mixed-version remote peers that do not expose the in-process state remain an explicit rollout boundary.

## Additional Notes

The final patch is intentionally narrower than the first protocol-trailer implementation: it does not modify `filemeta` stream framing and does not add per-entry overhead. The backlog issue remains open for merged-branch Docker/RemoteDisk/quorum E2E coverage.

---


houseme:
The first CI run at `e9e65abe7` exposed a real framing bug in the new completion trailer: the writer used MessagePack `write_u8`, while the reader consumed the trailer as raw bytes. The upgrade logs showed `invalid metacache completion value 204`, which caused `.rustfs.sys` listing quorum failures and readiness timeouts.

This is fixed in the force-pushed head `5864da85b`: the trailer now uses an explicit raw magic plus raw reason byte, with tests for normal, legacy, invalid, and truncated tails. Focused `filemeta` and `ecstore` checks pass; CI has been restarted for the corrected head.

houseme:
Final refinement after the CI diagnosis: instead of shipping a new metacache trailer, the patch now carries the bounded local-walk completion state entirely in-process through `WalkDirOptions`/`ListPathRawOptions`. This removes the trailer framing risk and avoids per-entry or on-wire overhead. The three SIGTERM scan tests now pass individually; the remaining mixed-version remote capability boundary is explicitly documented.

houseme:
The three `scan_dir` SIGTERM tests were also traced to the discarded trailer experiment: they use legacy `MetacacheWriter::close()` and then call `reader.read_all()` while the duplex writer remains alive, so the new reader waited for a completion byte and the workspace nextest job eventually hit its global timeout. The final in-process flag design removes that wait; all three tests pass individually on `30bcaf782`.

cxymds (approved):

PR #8029, opened 2026-09-20
fix(ecstore): allow server-side copy at normal read quorum

### LINKED ISSUE #8025: Server-side copy fails with 503 when one node of a four-node cluster is unavailable, while PUT/GET/LIST/multipart keep working
## Summary

In a four-node cluster (four drives each, single erasure set of 16), taking one node down makes **every server-side copy fail**, while all other object operations continue to work normally.

This breaks `docker/distribution`-based registries (Harbor, and anything else using the `s3` storage driver): the driver uploads a blob to a temporary path and then *moves* it to its content-addressed location. A move is a copy plus a delete, so no image can be accepted while a node is down.

Affected: **1.0.0** and **1.0.0-rc.6** — this is not a regression in 1.0.0.

## Behaviour with one of four nodes down

| Operation | Result |
|---|---|
| `PUT` (1, 2, 8 MB), also into deep key prefixes | 200 |
| `GET` / `LIST` / `DELETE` | 200 / 200 / 204 |
| Multipart upload (create, upload part, complete) | 200 |
| 40 consecutive `PUT`s | 40/40 succeeded |
| `/health/ready` | **200 throughout, 180/180 samples over 8 minutes** |
| **`PUT` with `x-amz-copy-source` (CopyObject)** | **503 ServiceUnavailable** |
| **`UploadPartCopy`** | **InternalError** |

With all four nodes up, both copy variants return 200.

## What does not help

- **Retrying.** 20 consecutive copies, all 503. The error message says "ServiceUnavailable. Please retry." — but it is not transient.
- **A different node.** Same result when stopping a different node.
- **The multipart copy path.** `UploadPartCopy` fails with `InternalError`.
- **Restarting the client.** Not a stale-connection issue.

## Minimal reproduction

```yaml
# compose.yaml
x-node: &node
  image: rustfs/rustfs:${RUSTFS_TAG:-1.0.0}
  environment:
    RUSTFS_VOLUMES: "http://node{1...4}:9000/data{1...4}"
    RUSTFS_ACCESS_KEY: testkey
    RUSTFS_SECRET_KEY: testsecret12345
    RUSTFS_ADDRESS: ":9000"
  networks: [lab]
services:
  node1: { <<: *node, ports: ["19001:9000"], tmpfs: [/data1:size=96m, /data2:size=96m, /data3:size=96m, /data4:size=96m] }
  node2: { <<: *node, tmpfs: [/data1:size=96m, /data2:size=96m, /data3:size=96m, /data4:size=96m] }
  node3: { <<: *node, tmpfs: [/data1:size=96m, /data2:size=96m, /data3:size=96m, /data4:size=96m] }
  node4: { <<: *node, tmpfs: [/data1:size=96m, /data2:size=96m, /data3:size=96m, /data4:size=96m] }
networks: { lab: {} }
```

```sh
S=http://127.0.0.1:19001
sig() { curl -s --aws-sigv4 "aws:amz:us-east-1:s3" --user "testkey:testsecret12345" "$@"; }

docker compose up -d && sleep 20
sig -X PUT "$S/probe"
head -c 1000000 /dev/urandom > /tmp/l.bin
sig -X PUT --data-binary @/tmp/l.bin "$S/probe/src"
sig -o /dev/null -w '%{http_code}\n' -X PUT -H "x-amz-copy-source: /probe/src" "$S/probe/dst"   # 200

docker compose stop node4 && sleep 15
sig -o /dev/null -w '%{http_code}\n' "$S/probe/src"                                            # 200
sig -o /dev/null -w '%{http_code}\n' -X PUT --data-binary @/tmp/l.bin "$S/probe/src2"           # 200
sig -o /dev/null -w '%{http_code}\n' -X PUT -H "x-amz-copy-source: /probe/src" "$S/probe/dst2" # 503
```

Note: `tmpfs` is required rather than Docker volumes, because RustFS refuses to start when several drives share a physical device:

```
[FATAL] local erasure endpoints must use distinct physical disks;
detected shared devices [vda => /data1, /data2, /data3, /data4]
```

## Expected

Server-side copy should behave like the other write paths and tolerate the loss of one node, or at minimum return an error that distinguishes "temporarily degraded, retry" from "not possible in this state".

## Additional note on observability

`/health/ready` returned 200 for the entire duration. From the outside the cluster looks healthy while no registry can accept an image. A readiness signal that reflects the degraded copy path would make this failure mode visible.

## Environment

- RustFS 1.0.0 and 1.0.0-rc.6, official container images
- 4 nodes, 4 drives each, default erasure configuration
- Reproduced both in containers and on bare VMs with XFS drives

overtrue:
## Related Issues

Fixes #8025.

## Summary of Changes

With 12 data shards and 4 parity shards, taking one of four nodes offline leaves enough shards for ordinary reads but causes CopyObject and UploadPartCopy to fail. Copy sources use demand-bound scheduling for backpressure, which incorrectly also enabled the opt-in GET requirement for a surplus verification shard.

Apply that extra requirement only to the default GET policy. Copy sources retain the normal read quorum, demand-bound scheduling, integrity-proof validation, and consistency checks for any available surplus shards. Add four regression tests covering both codecs, full and ranged multi-stripe reads, shard loss after the first stripe, insufficient quorum, and inconsistent parity with valid bitrot framing.

## Verification

Tested source: base `22243e791d37ca66ed5aa6d4c69b3e02bc384013` plus exactly the patch now committed as `7da3aeb756f01621fffe3d7d51763ae11a9dea06`; the tested source file and committed file have identical SHA-256 hashes.

1. Before the fix, both new exact-quorum success tests failed with `insufficient source shards to verify reconstructed data`. After the fix, `cargo +1.98.1 nextest run -p rustfs-ecstore --lib -E 'test(erasure::coding::) | test(copy_source_)'` passed all 276 selected tests without retries, including the strict GET gate and copy-policy coverage.
2. Docker Compose reproduction on Linux ARM64: four nodes with four ephemeral drives each, 12+4 erasure layout, and 1/2/8 MiB objects created before stopping node 4. Official `rustfs/rustfs:1.0.0` (`sha256:8cc9801755448b71a786705ce76692c77e14936cccd87cf2fc31842e58f4d1ff`) reproduced CopyObject 503 and UploadPartCopy 500 for every size. The patched image `rustfs:issue8025-fixed` (`sha256:b5851c74c946fa192c73992c98b34e23545f944fefcf861a7e27f69a8713e4c8`) returned 200 for both operations at every size, completed multipart copies, and returned byte-identical objects. Healthy-cluster copies and PUT/GET/LIST/DELETE/ordinary multipart controls passed; readiness remained 200. Across both images and both cluster states, 176 requests and 48 byte comparisons matched their expected outcomes.
3. `cargo +1.98.1 fmt --all --check` and `git diff --check` passed.

The patched Docker image was built from the tested source with Rust 1.98.1 using `cargo +1.98.1 build --locked -p rustfs --bin rustfs --no-default-features` in the debug profile. Full workspace CI, a release-profile image, and optional default-feature integrations were not run locally. An initial in-process `cargo test` run encountered process-global environment interference; the repository-standard nextest run above passed all selected tests in isolation.

## Impact

Restores server-side copy availability at the existing read quorum during a node outage. No public API, on-disk format, configuration, or deployment changes. The opt-in strict GET policy still requires its surplus verification source, and copies below the data-shard quorum still fail.

## Additional Notes

Two independent reviews of the final diff found no issues in the following lenses:

- Correctness: exact quorum succeeds; insufficient quorum and inconsistent surplus parity fail before output.
- Simplicity: one shared policy predicate controls all three emission paths.
- Test coverage: both codecs, ranges, partial tails, deferred recovery, and negative integrity cases are covered.
- Concurrency and durability: backpressure, cancellation, and terminal error propagation remain unchanged.
- Compatibility: existing GET policy behavior and storage formats are preserved.
- Performance: one policy lookup per decode invocation; no new allocations, locks, or I/O.

Rollback: revert this commit to restore the previous copy restriction.


Miladddddddddd (dismissed):

Miladddddddddd (dismissed):

Miladddddddddd:
در تاریخ یکشنبه ۲۰ سپتامبر ۲۰۲۶،‏ ۰۸:۰۷ Chris ***@***.***>
نوشت:

PR #8032, opened 2026-09-20
fix(heal): preserve remote corruption and reject incomplete deep scans

cxymds:
## Related Issues

Related: rustfs/backlog#2595 (C04 truncate-half / truncate-near-tail)

## Summary of Changes

A remote node correctly rejected a truncated shard as `FileCorrupt`, but the internal HTTP response dropped that error type. The coordinator received an untyped HTTP 500, converted it to `CHECK_PART_UNKNOWN`, skipped the payload, and could report `verified_healthy` without rebuilding it. Local-disk truncation tests did not exercise this transport boundary; the bitrot reader already requires complete reads or returns an error.

- Preserve `FileCorrupt` through the existing internal HTTP disk-error header, byte/chunk readers, cloned errors, and terminal read errors so remote corruption schedules payload repair.
- For protected deep scans, only explicit file missing/version missing/corruption errors become repairable part states. Other payload-open failures retain their original error and abort verification.
- For legacy deep scans, propagate nonrepairable verifier errors and reject unknown, unavailable, unrecognized, or incomplete part results instead of treating unverified parts as healthy.
- Cover EC12+4 half/tail/stripe-boundary truncation, quorum and dry-run behavior, six remote error cases, and a real filesystem path failure through the complete heal call. Aborted-verification tests assert that the remaining shards and metadata are unchanged.

## Verification

Validated source was committed as `e3410f8369524d2958f50f7c871df36738a7ed93` without further production changes.

- 58 distinct ECStore regressions passed across the 19-test deep-scan run, 51-test heal-module run, and final filesystem-error regression, with overlap removed. Commands included `cargo +1.98.1 test --locked -p rustfs-ecstore --lib deep_ -- --nocapture`, the ECStore test executable filtered to `set_disk::ops::heal::heal_result_report_tests::`, and `cargo +1.98.1 test --locked -p rustfs-ecstore --lib deep_heal_legacy_verification_error_cannot_certify_healthy -- --nocapture`. The final fixture requires both the disk verifier and complete heal call to return exactly `DiskError::InvalidPath` and preserve the unaffected bytes.
- The unchanged wire/server/error-classification changes previously passed 53 HTTP-reader tests, 39 disk-error tests, and the server test `read_file_error_response_preserves_typed_disk_errors`. These used `cargo +1.98.1 test --locked -p rustfs-rio --lib http_reader::tests::`, the ECStore test executable filtered to `disk::error::tests::`, and `cargo +1.98.1 test --locked -p rustfs --lib --no-default-features read_file_error_response_preserves_typed_disk_errors`.
- `cargo +1.98.1 fmt --all --check` and `git diff --check` passed. Local tests used disabled incremental/debug information and explicit loopback proxy bypass where needed.

Pre-fix C04 field evidence from candidate `pr8032-eabc55` showed both truncated shards retaining their damaged hashes while the coordinator reported healthy. This final patch has not yet been deployed to nodes 004–007 for physical reconstruction verification. Full local workspace CI was not run; the updated PR starts fresh CI. The initial filesystem fixture expectation was corrected after the real path guard returned `InvalidPath` before per-part verification; UNKNOWN responses are covered separately by the response matrix.

## Impact

Normal scans retain their existing presence checks. Deep scans now fail explicitly when verification is incomplete, which can expose storage or transport failures previously hidden behind a successful result. Explicit missing/corrupt payloads remain eligible for normal reconstruction.

The existing internal disk-error header gains the `file-corrupt` token. No S3 API, configuration, or on-disk format changes are introduced. A new coordinator fails closed on an old server's untyped error; old coordinators retain the original behavior until upgraded. Rollback is a normal binary rollback because persisted formats are unchanged, but it restores the prior false-health risk.

## Additional Notes

Two independent adversarial reviewers covered the final changes:

- Correctness: repairable payload damage is distinguished from incomplete verification; no remaining supported finding.
- Compatibility: the new token is restricted to the existing read-file HTTP error contract; mixed-version limits are recorded above.
- Security: unknown transport/status inputs cannot certify health or be fabricated into proven corruption.
- Concurrency and durability: aborted verification preserves payload and metadata bytes; existing heal race and quorum regressions passed.
- Simplicity: existing transport and error-conversion paths are reused without a new configuration or service abstraction.
- Test coverage: remote error matrices and full-call legacy failure coverage close gaps masked by local file-open checks; final four-node reconstruction remains an explicit verification gap.

D04 harness receipt handling remains outside this production fix.


houseme (approved):

houseme (approved):

overtrue:
The fix is now committed directly on this PR branch as e3410f836.

The C04 field evidence identifies why the local truncation tests were insufficient: the remote node returned `FileCorrupt`, but the HTTP response lost its type. The coordinator then treated the resulting UNKNOWN part state as healthy and skipped reconstruction. This is an error-propagation and verification-completeness problem, rather than a missing short-read check in the bitrot reader.

The update preserves remote corruption through the HTTP/error wrappers, aborts protected scans on nonrepairable storage failures, and rejects unknown or incomplete legacy verification results. Explicit missing/corrupt shards still schedule repair.

58 distinct storage regressions passed, including the remote failure matrix and a complete legacy heal failure that preserves the original `InvalidPath` error and unchanged shard/metadata bytes. The wire/server/error-classification tests also passed before the final storage-only additions. Two independent adversarial reviews found no remaining supported defects; the PR description now records the per-lens results and compatibility limits.

The remaining acceptance gap is physical reconstruction on nodes 004–007 using this exact new build. That deployment has not been performed, so the earlier four-node results must not be treated as validation of this commit.

overtrue (approved):
Reviewed base 42dd76d3b256fb50cb7f92dddca7ed1458d1dc8b and head 613fe47804d336afa567fa8c39dd383ce2408f7a.

No findings. High-risk review covered correctness, concurrency and durability, compatibility, security, test coverage, simplicity, and performance. Checked typed remote corruption propagation, rejection of incomplete deep scans, mixed-version failure behavior, and abort-before-write paths against callers and existing tests.

Validation: diff whitespace check passed. Rust tests were not rerun locally. Current CI has no failures; Workspace Test and Lint is still running. Live multi-node reconstruction remains unverified in this review.

PR #8034, opened 2026-09-20
ci(docker): sync the Docker Hub overview from README.md on release

### LINKED ISSUE #7995: Docker Hub overview is a stale copy of README and links to a 404
**Describe the bug**

The overview shown on https://hub.docker.com/r/rustfs/rustfs is an old snapshot of `README.md`. It still links to `https://docs.rustfs.com/introduction.html`, which now returns 404.

The current `README.md` no longer contains that URL — it points to `https://docs.rustfs.com/en/installation` instead.

**To Reproduce**

1. Open https://hub.docker.com/r/rustfs/rustfs
2. Follow the documentation link in the overview
3. It lands on `https://docs.rustfs.com/introduction.html` → 404

Or from the API:

```console
$ curl -s https://hub.docker.com/v2/repositories/rustfs/rustfs/ \
    | jq -r .full_description | grep -o 'https://docs\.rustfs\.com[^ )]*' | sort -u
https://docs.rustfs.com
https://docs.rustfs.com/
https://docs.rustfs.com/introduction.html

$ curl -s -o /dev/null -w '%{http_code}\n' -L https://docs.rustfs.com/introduction.html
404
```

**Expected behavior**

The Docker Hub overview reflects the current `README.md`, and its documentation links resolve.

**Additional context**

The two have drifted a long way apart:

| | size |
|---|---|
| Docker Hub `full_description` | 8,434 chars |
| `README.md` on `main` | 22,879 chars |

Docker Hub's overview is not updated by `docker push` — it is a separate `full_description` field, written through the web UI or `PATCH /v2/repositories/<namespace>/<repo>/`. I could not find anything in this repository that writes it: `.github/workflows/docker.yml` uses `docker/login-action` and `docker/build-push-action`, neither of which touches the description, and searching the repo for `dockerhub-description`, `full_description` and `hub.docker.com/v2` returns nothing. So it appears to be maintained by hand, which would explain the drift.

If keeping it in sync automatically sounds useful, a step like this in `docker.yml` would do it, and it needs no new secrets — `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` are already used by the Docker Hub login step:

```yaml
- name: Sync Docker Hub description
  uses: peter-evans/dockerhub-description@v5.0.0
  with:
    username: ${{ env.DOCKERHUB_USERNAME }}
    password: ${{ secrets.DOCKERHUB_TOKEN }}
    repository: rustfs/rustfs
    readme-filepath: ./README.md
```

One caveat if you go that way: Docker Hub caps `full_description` at 25,000 characters, and `README.md` is at 22,879 today, so there is not much headroom left.

Happy to open a PR for this if it would be welcome.

Marukome0743:
## Related Issues

Fixes #7995

## Summary of Changes

Docker Hub's overview is a separate `full_description` field that `docker push` never touches, so it had drifted into an 8 KB snapshot of an old `README.md` that still linked to `https://docs.rustfs.com/introduction.html` (now 404). Nothing in this repository wrote that field, so it was maintained by hand.

`docker.yml` gains a `sync-dockerhub-description` job that runs after `build-docker`, only when images were actually pushed (`should_push == 'true'`), and republishes `README.md` from the same commit the images were built from (`source_ref`). It is a separate job rather than a step in `build-docker` so it runs once, not once per musl/glibc matrix entry.

- Uses `peter-evans/dockerhub-description`, pinned to the v5.0.0 commit SHA like the other actions in this workflow.
- Reuses the existing `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` credentials; no new secrets.
- `enable-url-completion: true`: `README.md` has ~12 relative links (`docs/...`, `CONTRIBUTING.md`) that would be broken on Docker Hub; the action rewrites them to `github.com/rustfs/rustfs/blob/<ref>/...` URLs.
- Docker Hub caps `full_description` at 25,000 bytes and the action truncates to fit with only a warning (`truncateToBytes` in its source), which would publish a README cut off mid-sentence. `README.md` is at 22,879 bytes today. A follow-up step reads the published overview back from the public API and fails the job if it landed within one UTF-8 character of the cap. The check is done after publishing because URL completion adds a few hundred bytes, so a pre-publish `wc -c` would not be exact.

`docs/testing/ci-gates.md` lists the new job in the release-validation table, per that file's change checklist.

## Verification

- `actionlint` 1.7.12 (with shellcheck 0.11.0) on `.github/workflows/docker.yml` at 54c9c052d46d: no findings for the new job. The only report is the pre-existing unknown-runner-label warning for `dind-sm-standard-2` in `build-docker`.
- Checked the action's source at the pinned SHA: an empty `short-description` is not sent (the existing Docker Hub short description is left untouched), and the size limit is enforced by truncation, not failure — which is why the post-publish check exists.
- `curl -s https://hub.docker.com/v2/repositories/rustfs/rustfs/ | jq -r '.full_description | length'` returns 8434 today, confirming the public endpoint the check step reads is available unauthenticated.

Not run: the job itself. It only executes on a pushed release build with the repository's `DOCKERHUB_TOKEN`, so it cannot be exercised from a fork. Remaining risk: the token needs read/write scope on Docker Hub for the `PATCH`; if it is push-only, the job fails at the sync step without affecting image publication (it is a separate job downstream of `build-docker`).

## Impact

- Docker Hub overview for `rustfs/rustfs` is rewritten on every pushed release from `README.md`; manual edits on hub.docker.com will be overwritten.
- Requires `DOCKERHUB_TOKEN` to have read/write scope (see above).
- `README.md` growing past ~25,000 bytes (after link expansion) will fail this job, not the image build.

## Additional Notes

N/A


Marukome0743:

majinghe (approved):
@Marukome0743 Thanks for your contribution.

PR #8038, opened 2026-09-20
fix(decommission): use owned data for source capacity

### LINKED ISSUE #8037: bug(decommission): local-path pools overestimate source usage from filesystem-wide used space
## Summary

When a RustFS pool uses a **local-path / local-static directory backed by a filesystem that also contains unrelated data**, starting pool decommission can fail the capacity precheck even though the RustFS pool itself contains very little data.

The current decommission capacity model appears to use the **entire backing filesystem's used space** as the source pool's physical usage, instead of the amount of data actually owned by RustFS under that pool's storage path.

This can dramatically overestimate migration requirements and block decommission.

## Environment

- RustFS image: `rustfs/rustfs:1.0.0`
- Kubernetes / Helm multi-pool deployment
- Storage class: `EC:1`
- Pool 0:
  - 4 drives
  - Longhorn-backed PVCs
- Pool 1:
  - 2 drives
  - local-static / local-path storage
  - 350 GiB per drive
- Decommission target: pool 1

## Observed behavior

The RustFS Console reports only about **5 GB used** in pool 1.

However:

```bash
rc admin decommission start rustfs 1 --by-id
```

fails with HTTP 500, and the server logs:

```text
decommission_capacity_blocked: Storage reached its minimum free drive threshold.:
failed to start decommission:
insufficient reserved physical target capacity:
operation 1806aa6e-bf16-4e1d-ba4b-e184cc751ff1 generation 1
requires 131062071296 bytes,
but 118171238400 bytes are available
after source/target parity and temporary-copy accounting
```

So a pool with roughly 5 GB of actual RustFS data results in a decommission reservation requirement of about 122 GiB.

Expanding pool 0 and restarting all RustFS pods does not change the fact that the source requirement is grossly overestimated.

## Suspected root cause

In `crates/ecstore/src/core/pools.rs`, decommission capacity information is derived from disk-level filesystem statistics.

The current implementation of `decommission_physical_pool_capacity()` computes source physical usage approximately as:

```rust
let max_used = set_disks
    .iter()
    .map(|disk| {
        (disk.used_space as usize)
            .max(
                (disk.total_space as usize)
                    .saturating_sub(disk.available_space as usize)
            )
    })
    .max()
    .unwrap_or_default();

physical_used =
    physical_used.saturating_add(max_used.saturating_mul(width));
```

That `physical_used` is then used to build the decommission reservation:

```rust
let source_data_equivalent_bytes =
    capacity_source_data_equivalent(source.physical_used, source.layout)?;

let predicted_physical_bytes =
    capacity_target_physical_bytes(source_data_equivalent_bytes, target_layout)?;

let temporary_physical_bytes =
    predicted_physical_bytes
        .saturating_mul(DECOMMISSION_CAPACITY_TEMPORARY_COPIES);

peak_physical_bytes =
    predicted_physical_bytes.saturating_add(temporary_physical_bytes);
```

For a local-path deployment, `disk.used_space` / `total_space - available_space` describes the **whole filesystem**, not the RustFS data directory.

For example:

```text
filesystem
├── unrelated data        ~40 GiB
└── RustFS pool directory  ~5 GiB
```

RustFS can therefore treat ~45 GiB filesystem usage as ~45 GiB RustFS-owned usage on that drive.

With erasure-layout conversion and the temporary-copy reservation, this becomes much larger again.

## Why this is problematic

This behavior is reasonable for a dedicated raw disk or dedicated PVC filesystem where all used bytes belong to RustFS.

It is incorrect for local-path/local-static deployments where the RustFS directory shares a filesystem with other directories or workloads.

In that case:

- Console/object usage may be small.
- Actual RustFS directory usage may be small.
- Filesystem-wide used space may be much larger.
- Decommission can be rejected even though enough capacity exists for the actual RustFS objects.

## Expected behavior

The **source migration demand** should be estimated from RustFS-owned data, not from filesystem-wide used space.

The **target free capacity** should still use actual filesystem free space, because that reflects whether target writes can succeed.

Conceptually:

```text
source required bytes
    = RustFS-owned physical/logical usage
      -> source EC conversion
      -> target EC conversion
      -> temporary-copy safety factor

target available bytes
    = actual filesystem free bytes
```

## Reproduction

1. Create a local filesystem with significant non-RustFS data.
2. Use a directory on that filesystem as a local-static/local-path RustFS pool.
3. Write only a small amount of data to that RustFS pool.
4. Add/retain another active pool as the migration target.
5. Run:

```bash
rc admin decommission start <alias> <pool-id> --by-id
```

6. Observe that required decommission capacity is derived from filesystem-wide used space rather than RustFS-owned usage.

A useful confirmation is:

```bash
df -B1 <rustfs-data-path>
du -sb <rustfs-data-path>
```

When `df` used bytes are much larger than `du` for the RustFS directory, the decommission precheck can be inflated accordingly.

## Additional note

The same `decommission_physical_pool_capacity()` / `source.physical_used` based logic appears to still exist on the current `main` branch, so this does not appear limited to the 1.0.0 release.

A fix should preserve the existing safety checks for target physical capacity while avoiding attribution of unrelated filesystem usage to the source pool.

cxymds:
## Related Issues

Fixes #8037.

## Summary of Changes

Decommission capacity admission previously used filesystem-wide `used_space` as the source pool's migration demand. On local-path or local-static deployments, unrelated files on a shared filesystem were therefore attributed to RustFS and could block decommission.

The source capacity model now prefers a complete per-set RustFS data-usage snapshot and converts the logical object bytes to the source erasure layout's physical footprint. Filesystem free space remains the source of target write-capacity checks. If an owned-data snapshot is unavailable or incomplete, the existing conservative disk-statistics fallback is retained.

## Verification

- `cargo test -p rustfs-ecstore decommission_physical_capacity --lib` — 2 tests passed, including a regression where filesystem usage is 360 bytes while RustFS-owned physical usage is 12 bytes.
- `cargo check -p rustfs-ecstore --lib` — passed.

The test binary emitted the existing macOS linker warning about `__eh_frame` section size; it did not affect test results. Full workspace tests were not run.

## Impact

Decommission source reservations no longer over-count unrelated data on shared filesystems when a complete RustFS usage snapshot is available. Target capacity safety checks and the conservative fallback behavior are unchanged. No public API or storage format changes.

## Additional Notes

The change is limited to the decommission capacity model and its unit coverage.


crates/ecstore/src/core/pools.rs
```diff
@@ -925,6 +925,29 @@ fn capacity_target_physical_bytes(data_bytes: usize, layout: DecommissionErasure
...
 }
 
+async fn decommission_owned_physical_usage(sets: &Sets, layout: DecommissionErasureLayout) -> HashMap<i32, usize> {
+    let mut usage = HashMap::with_capacity(sets.disk_set.len());
+    for (set_index, set) in sets.disk_set.iter().enumerate() {
+        let Ok(cache) = load_data_usage_cache(set.as_ref(), DATA_USAGE_CACHE_NAME).await else {
+            continue;
+        };
+        let data_usage = cache.dui(DATA_USAGE_ROOT, &[]);
+        if !data_usage.is_complete_bucket_usage_snapshot() {
```
overtrue, line 935: [P1] Reconcile writes made after the completed scan. is_complete_bucket_usage_snapshot() checks completion and the presence of last_update, but does not establish that the cache includes subsequent writes. For example, complete a scan of an empty source pool, upload an object, then start decommission before the next scan: this path reports zero source usage, so reserve_decommission_start_target_capacity() creates no target allocations and select_decommission_capacity_target_pool() rejects the object with "no target allocation". Recovery preserves an existing active reservation rather than increasing its estimate. Please reconcile post-scan usage or safely grow the reservation against current target capacity, with a regression test covering this sequence.

crates/ecstore/src/core/pools.rs
```diff
@@ -925,6 +925,29 @@ fn capacity_target_physical_bytes(data_bytes: usize, layout: DecommissionErasure
...
+        }
+        let Ok(logical_bytes) = usize::try_from(data_usage.objects_total_size) else {
+            continue;
+        };
+        let Ok(set_index) = i32::try_from(set_index) else {
+            continue;
+        };
+        // The data-usage snapshot excludes unrelated files on a shared mount.
+        // Convert logical object bytes to the source set's physical EC footprint.
+        usage.insert(set_index, capacity_mul_div_ceil(logical_bytes, layout.width(), layout.data));
```
overtrue, line 946: [P2] Include the per-object minimum and rounding in the reservation. Even a fresh, complete snapshot can contain zero-byte objects or delete markers while objects_total_size is zero. This conversion then reserves zero bytes and creates no target allocations, but target selection and mutation admission charge capacity_target_physical_bytes(expected_data_bytes.max(1), layout), so those entries cannot migrate. Aggregate rounding also undercounts nonzero objects: with 3 data + 1 parity on both sides, three one-byte objects reserve 4 physical bytes here but consume 6 under the per-object accounting. Please make the source estimate cover the same minimum and rounding rules as migration admission, and test both cases through reservation and target selection.

overtrue (changes requested):
Two capacity underestimates can block decommission even when the target has enough free space. Both need to be addressed before this replaces the filesystem-based source estimate. I traced reservation creation, target selection, and recovery, and checked the arithmetic with three standalone probes; I did not run an ECStore integration test.

crates/ecstore/src/core/pools.rs
overtrue: Fixed in 112fa37fb. An underestimated reservation now grows under the pool metadata lock after checking current target capacity and competing reservations, and is persisted before use. Regression coverage includes the empty snapshot followed by a nonempty object and cross-node reload.

crates/ecstore/src/core/pools.rs
overtrue: Fixed in 112fa37fb. Growth uses the same per-object minimum and EC rounding as admission, including existing-target writes. Tests cover zero-byte entries and three one-byte objects with a 3+1 layout, and verify consumption after each object.

overtrue (approved):
Both findings are addressed in 112fa37fb. Capacity growth is checked against current free space and competing reservations, persisted before use, and covered by stale-snapshot, zero-byte, rounding and cross-node reload regressions. The 28 scoped nextest tests pass.

overtrue (changes requested):
Reviewed base 42dd76d3b256fb50cb7f92dddca7ed1458d1dc8b and head 1e956fa991e9ca32c1d22e41f27c957928a60dfd.

[P1] Do not make an asynchronous usage snapshot a fixed migration capacity ceiling.

In crates/ecstore/src/core/pools.rs:934-946, any complete scanner snapshot is accepted as source usage without establishing that it covers writes made before decommission starts. A complete empty snapshot followed by an upload before the next scan produces a zero-capacity reservation. Subsequent target admission rejects migration even when the target has ample free space. Existing renewal only refreshes the lease, recovery keeps active reservations, and this reviewed commit does not contain the reservation growth described in the PR body.

Persist an increased reservation against live target capacity and competing reservations before admitting underestimated objects, or use a reliably conservative source bound. Add regressions for uploads after a complete snapshot and for zero-byte objects/delete markers, whose migration admission still budgets at least one byte per entry.

High-risk review covered correctness, concurrency and durability, compatibility, test coverage, simplicity, and performance; no additional findings. The new test only injects a usage map into the capacity calculator and does not exercise stale scanner data through actual migration. Validation: diff whitespace check passed and current CI passed. Rust tests were not rerun locally.

cxymds:
Addressed in `075991273`.

- A complete scanner snapshot is now treated as an estimate: before each admitted migration mutation, the reservation grows under the pool metadata lock when the per-object target requirement exceeds the remaining allocation.
- Growth is checked against the live target capacity plus all active reservations, persisted before the object write, and rolled back on a failed capacity check.
- Empty or zero-byte snapshots can now create a target allocation on demand; target selection falls back to an eligible active target pool when the original reservation has no target entries.
- Added a regression covering an empty snapshot followed by a one-byte object with a 3+1 target layout.

Validation: `cargo fmt --all --check`, `git diff --check`, `cargo check -p rustfs-ecstore --features test-util`, and 24 `decommission_capacity` unit tests passed.

houseme (approved):

cxymds:
CI failures were reproduced and fixed in `38ff48c7c4c2602c565a28dc888e6ecf70ad7fd7`. Reservation growth is now limited to stale empty-source estimates, preserves pending-intent conflicts and target-gate switching, and keeps source metadata validation consistent. The four failing lock-order tests pass locally; `cargo check -p rustfs-ecstore --features test-util` and the 24 capacity tests pass (one existing serialization test exceeded the local 60s nextest slow-test notice and was stopped after the targeted tests passed).

cxymds:
CI failure root cause was two clippy::collapsible_if errors in the reservation growth helper (pools.rs:1928 and 1946), not a runtime test failure. Fixed in `304f8a025`; rustfmt and diff check pass. PR 8044 is related to the same capacity-gate area but only adds bounded gate retries; its diff has no overlap conflict with this reservation change. CI has been restarted.

overtrue (approved):

